PR #106, opened 2020-04-30
Clarify Relationship between RuntimeType and ComInterface

rylev:


crates/winmd/src/types/delegate.rs
```diff
@@ -65,10 +65,10 @@ impl Delegate {
             unsafe impl<#constraints> ::winrt::RuntimeType for #name {
                 type Abi = ::winrt::RawPtr;
                 fn abi(&self) -> Self::Abi {
-                    self.ptr.get()
+                     <::winrt::IUnknown as ::winrt::ComInterface>::as_vtable(&self.ptr) as Self::Abi
```
kennykerr, line 68: This seems a lot more complicated... 😉

rylev: To be fair this is just because we want to ensure no ambiguity in code gen. If it were regular code it would just be `self.ptr.as_vtable()`

kennykerr: Oh right, got it.

crates/winmd/src/types/class.rs
```diff
@@ -123,15 +123,15 @@ impl Class {
                 #type_name
                 unsafe impl ::winrt::ComInterface for #name {
                     type VTable = #abi_name;
-                    const GUID: ::winrt::Guid = ::winrt::Guid::from_values(#guid);
+                    const IID: ::winrt::Guid = ::winrt::Guid::from_values(#guid);
                 }
                 unsafe impl ::winrt::RuntimeType for #name {
-                    type Abi = *const *const <Self as ::winrt::ComInterface>::VTable;
+                    type Abi = ::winrt::RawComPtr<Self>;
```
kennykerr, line 129: Nice 👍

src/com_interface.rs
```diff
@@ -6,24 +6,35 @@ use crate::*;
...
+/// Additionally, because ComInterfaces are just pointers to vtables,
+/// it must be safe to zero initialize the interface.
 pub unsafe trait ComInterface: Sized {
-    const GUID: Guid;
+    const IID: Guid;
     type VTable;
 
-    fn as_vtable(&self) -> *const *const Self::VTable {
+    #[inline(always)]
+    fn as_raw(&self) -> RawComPtr<Self> {
```
kennykerr, line 21: I prefer this name - the name vtable is too confusing as it typically refers to the vtable layout rather than a particular vfptr.

PR #113, opened 2020-05-01
Model WinRT enums as structs

### LINKED ISSUE #65: Need to model WinRT enums as structs
While it would be preferable use Rust enums, their variants are required to be unique and WinRT's enumerators are not guaranteed to be unique. Using structs would also allow flag/unsigned WinRT enums to provide bitmask operators. Something like this:

```Rust
#[derive(PartialEq, Eq)]
struct AsyncStatus {
    value: i32,
}

impl AsyncStatus {
    pub const Canceled: AsyncStatus = AsyncStatus { value: 2 };
    pub const Completed: AsyncStatus = AsyncStatus { value: 1 };
    pub const Error: AsyncStatus = AsyncStatus { value: 3 };
    pub const Started: AsyncStatus = AsyncStatus { value: 0 };
}
```

kennykerr:
Technically fixes #65 although I'll leave that issue open until I have finished the support for bitmask operators.

### DIFF
--- a/tests/enum.rs
+++ b/tests/enum.rs
@@ -6,15 +6,17 @@ winrt::import!(
         "windows.application_model.appointments"
 );
 
+use winrt::RuntimeType;
+
 #[test]
 fn signed_enum() {
     use windows::foundation::AsyncStatus;
 
     assert!(AsyncStatus::default() == AsyncStatus::Canceled);
-    assert!(AsyncStatus::Canceled as i32 == 2);
-    assert!(AsyncStatus::Completed as i32 == 1);
-    assert!(AsyncStatus::Error as i32 == 3);
-    assert!(AsyncStatus::Started as i32 == 0);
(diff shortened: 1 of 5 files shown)


tests/enum.rs
```diff
@@ -6,32 +6,34 @@ winrt::import!(
...
 #[test]
 fn signed_enum() {
     use windows::foundation::AsyncStatus;
 
-    assert!(AsyncStatus::default() == AsyncStatus::Canceled);
-    assert!(AsyncStatus::Canceled as i32 == 2);
-    assert!(AsyncStatus::Completed as i32 == 1);
-    assert!(AsyncStatus::Error as i32 == 3);
-    assert!(AsyncStatus::Started as i32 == 0);
+    assert!(AsyncStatus::default().abi() == 0);
```
rylev, line 15: I think it would make sense to implement PartialEq for these for the type that represents the enum

PR #116, opened 2020-05-04
Add initial support for bitwise operations on WinRT enums

### LINKED ISSUE #115: Enable WinRT enums use as bitflags
This is a follow up to enum support added in #113. The following should work, but does not currently:

```rust
let weekend = AppointmentDaysOfWeek::Sunday | AppointmentDaysOfWeek::Saturday;
assert!(weekend.abi() == 0x41);
```

rylev:
Fixes #115. 

This only adds support for `&` and `|`. We need to decide if every enum should support `&=`, `|=`, `^`, and `^=`.

### DIFF
--- a/crates/winmd/src/types/enum.rs
+++ b/crates/winmd/src/types/enum.rs
@@ -82,6 +82,20 @@ impl Enum {
                     &mut self.value
                 }
             }
+            impl ::std::ops::BitOr for #name {
+                type Output = Self;
+
+                fn bitor(self, rhs: Self) -> Self {
+                    Self { value: self.value | rhs.value }
+                }
+            }
+            impl ::std::ops::BitAnd for #name {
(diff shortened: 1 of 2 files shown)


jcotton42:
Is there any reason you're doing this by hand and not with the `bitflags` crate?

crates/winmd/src/types/enum.rs
```diff
@@ -82,6 +82,20 @@ impl Enum {
                     &mut self.value
                 }
             }
+            impl ::std::ops::BitOr for #name {
```
kennykerr, line 85: This should only be included for enums who's representation is unsigned.

rylev:
@jcotton42 I didn't want to add a dependency until it became clear we would actually gain something. I'd like to see how much code it would be to implement this on our own.

Issue #76, opened 2020-04-27
Publish to crates.io


rylev:
Before we publish, we need to decide who is owner of the crate.

Boddlnagg:
As one of the owners of https://crates.io/crates/winrt, I think we can discuss whether you want to take over that crate name. From my perspective it makes no sense to develop two Rust language projections for WinRT in parallel (and also I don't have time to maintain https://github.com/contextfree/winrt-rust in a meaningful sense any longer).

I'm really happy that there is now an alternative that is supported by Microsoft directly 👍

kennykerr:
Thanks Patrick, that would be ideal.

dns2utf8:
I would prefere if you could continue with the semver schema and release this project as `0.7.0`.

Boddlnagg:
I agree that it would be necessary to release the first version as `0.7.0` to be semver-compatible. And `7` looks almost like `1`, doesn't it? 😉

contextfree:
Yeah, having this project take over the winrt crate name would definitely be fine.

PR #117, opened 2020-05-04
Lift version to 0.7.0 to enable a smoth transition on crates.io

dns2utf8:
Signed-off-by: Stefan Schindler <dns2utf8@estada.ch>

References #76


kennykerr:
Let's revisit this once we've decided what we are going to do about https://github.com/microsoft/winrt-rs/issues/76

PR #123, opened 2020-05-04
WinRT classes and interfaces are convertible to winrt::Object

kennykerr:
`winrt::Object` is the equivalent of `System.Object` in C# and `IInspectable` in C++. It is logically the base class for all WinRT interfaces and classes. Any interface or class can be represented as a `winrt::Object` in a polymorphic sense.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -11,7 +11,7 @@ pub struct Object {
 
 impl Object {
     pub fn type_name(&self) -> Result<HString> {
-        let this = self.ptr.get();
+        let this = self.ptr.as_raw();
         if this.is_null() {
             panic!("The `this` pointer was null when calling method");
         }
@@ -25,7 +25,7 @@ impl Object {
 
 unsafe impl ComInterface for Object {
     type VTable = abi_IInspectable;
-    const GUID: Guid = Guid::from_values(
+    const IID: Guid = Guid::from_values(
         0xAF86_E2E0,
         0xB12D,
         0x4C6A,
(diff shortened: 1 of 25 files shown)


src/object.rs
```diff
@@ -46,10 +46,8 @@ unsafe impl RuntimeType for Object {
 }
 
 #[repr(C)]
-pub struct abi_IInspectable {
```
rylev, line 49: I know users don’t interact with this often, but why the name change? It’s nice to know they this is IInspectable so you can read about it in docs

src/unknown.rs
```diff
@@ -25,7 +25,8 @@ unsafe impl ComInterface for IUnknown {
 
 #[repr(C)]
 pub struct abi_IUnknown {
-    pub(crate) query: extern "system" fn(RawComPtr<IUnknown>, &Guid, *mut RawPtr) -> ErrorCode,
-    pub(crate) addref: extern "system" fn(RawComPtr<IUnknown>) -> u32,
-    pub(crate) release: extern "system" fn(RawComPtr<IUnknown>) -> u32,
+    pub(crate) unknown_query:
```
rylev, line 28: I’d also prefer for these to have their canonical names (but transformed to camel case)

src/object.rs
```diff
@@ -46,10 +46,8 @@ unsafe impl RuntimeType for Object {
...
+pub struct abi_Object {
     __base: [usize; 4],
-    type_name: extern "system" fn(
-        *const *const object::abi_IInspectable,
-        *mut <HString as RuntimeType>::Abi,
-    ) -> ErrorCode,
+    inspectable_type_name:
```
rylev, line 51: Same thing here about naming

src/object.rs
kennykerr: Yeah... it seemed weird that Object's ABI representation is called something else entirely. In C# the `IInspectable` is entirely hidden. In C++ it's just called `IInspectable` throughout.

src/unknown.rs
kennykerr: Happy to do so, I just need to prepend unknown/inspectable in order to avoid collisions as I started adding support for implementations.

src/object.rs
rylev: I personally wouldn’t be against calling it IInspectable everywhere. I don’t see what you gain by not using the name, but you lose the ability for people to easily search for more info on it if you make it something else

PR #125, opened 2020-05-05
Fix build status in Readme

### LINKED ISSUE #121: Build status is inaccurate
The build status badge just provides the results of the last CI build rather than the last build of the master branch. So if a PR build happens to fail then the readme gives the impression that the project doesn't currently build.

rylev:
Fixes #121

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,10 @@
 name: Build and Test
 
-on: [pull_request]
+on: 
+  pull_request:
+  push:
+    branches:
+      - master
 
 env:
   RUSTFLAGS: -Dwarnings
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -1,6 +1,10 @@
 name: Build and Test
 
-on: [pull_request]
+on: 
+  pull_request:
+  push:
+    branches:
+      - master
```
kennykerr, line 7: Will this still support building PR branches?

rylev: Yea the syntax here means build on every PR, and build on pushes to master (and pushes happen on merges)

PR #127, opened 2020-05-07
Add default target for docs.rs

GuillaumeGomez:
So we'll land on windows documentation when browsing it on docs.rs. Question though: Do you want to only build for windows targets? If so, I can just list the windows targets (there are 2 on docs.rs) and that will be it.


kennykerr:
@GuillaumeGomez yes I think building only for Windows is a safe bet for now.

GuillaumeGomez:
I added the specific windows targets (I did it in another commit but I can squash if you want?).

kennykerr:
Thanks. No need to squash - I'll do that when I merge.

Issue #93, opened 2020-04-28
Import macro must support 3rd party dependencies


kennykerr:
Dependencies include:

* Specific Windows SDK (such as [this package](https://www.nuget.org/packages/Microsoft.Windows.SDK.Contracts))
* Specific WinRT component (such as [Win2D](https://www.nuget.org/packages/Win2D.uwp))

Except for the SDK, WinRT components will include both metadata (winmd files) and executable binaries (DLL files) and most WinRT components are already distributed via nuget so perhaps the `winrt::import` macro needs to know how to unpack them directly.

rylev:
Talking with @kennykerr about this, this is our plan for action:
* The `import!` macro will not do any network requests.

For nuget:
* We'll add a `nuget_dependencies` key to the Cargo.toml [manifest section)(https://doc.rust-lang.org/cargo/reference/manifest.html#the-metadata-table). 
* Create a tool `cargo-nuget` which can read that section and store it in a well known place. 
* The `import!` macro can then read nuget dependencies from that well known place.

Jack-Works:
Is it possible to import type for `Microsoft.Update.Session`?

chuckries:
@rylev @kennykerr it looks like the impl of this has changed since #118 was merged and this doesn't appear to be doc'd anywhere. Is there currently support for generating bindings for 3rd party winmd's? Trying to pull in Win2D.

robmikh:
You'll need to place the winmd files you wish to generate bindings for in the `.winmd/winmd` folder. You'll also have to place the binaries in the `.winmd/[arch]` folder (e.g. `.winmd/x64`). In addition, you'll need to place the appropriate binaries from the [Microsoft.VCRTForwarders.140 nuget package]() in that folder as well. From there, just include the desired namespaces/types in your `windows::build!` macro.

Checkout Kenny's [Win2D sample](https://github.com/kennykerr/win2d-rs) as a guide.

kennykerr:
I've added Win2D to the samples repo: https://github.com/kennykerr/samples-rs

Note that the `windows` crate first attempts to load metadata from the `.windows/winmd` folder, that Robert mentioned. Only if no winmd files are found will it fall back to the [default metadata](https://github.com/microsoft/windows-rs/tree/master/crates/winmd/default).

PR #132, opened 2020-05-11
Return `T` instead of `winrt::Result<T>` for `noexcept` methods

reflectronic:
Methods with the `noexcept` attribute should not return errors. This PR allows methods marked as such to return just `T` instead of `winrt::Result<T>`.

Unfortunately, I don't know of a good way to test this end-to-end, since there are no APIs in the Windows SDK that make use of `noexcept` at the moment. Currently, this PR builds its own testing WinMD from IL, which is used to test that the metadata parsing logic works correctly. However, nothing is testing whether the actually generated code is correct (or even compiles). I can take a look into testing this 'manually' with a one-off library in a couple of days.

Additionally, I don't know a good way to invoke `ilasm` in the build.rs (I just pasted in the full path when testing), so I would appreciate a good way to do that. The tests pass otherwise.

I do not know too much about Rust, and this is my first time doing anything serious with it. Please go easy on me 😄

### DIFF
--- a/crates/winmd/build.rs
+++ b/crates/winmd/build.rs
@@ -1,0 +1,15 @@
+use std::process::Command;
+use std::env;
+
+fn main() {
+    let out = env::var("OUT_DIR").unwrap();
+
+    Command::new(r"C:\Windows\Microsoft.NET\Framework\v4.0.30319\ilasm.exe")
+        .args(&["/DLL"])
+        .arg(format!("/output={}/RustWinRT.Tests.winmd", out).as_str())
+        .arg("./tests/RustWinRT.Tests.il")
+        .status()
+        .unwrap();
+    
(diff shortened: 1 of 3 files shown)


kennykerr:
Hi John, thanks for jumping in, but this is not a feature we can add just yet. As you noted, there aren’t any OS APIs that use this attribute at the moment, so there are a few things that need to happen first because only having metadata isn’t enough. We need #93 to enable us to import https://github.com/microsoft/testwinrt. We also need that component to publish a nuget package. Then we need to make sure all those tests pass. Then we need to add some `noexcept` tests to that test project. And only then can we add noexcept support to Rust/WinRT.

kennykerr:
I created an issue for this: https://github.com/microsoft/winrt-rs/issues/133

Issue #134, opened 2020-05-13
Document any support channels, even if unofficial


ndarilek:
Hey folks, new to working with WinRT. I thought I'd attempt to use this library to add speech synthesis support to my [tts crate](https://gitlab.com/ndarilek/tts-rs).

First, are there any support channels? I recognize this is a public preview--I'm just having some struggles mapping API concepts to this binding. I hope I'm just missing something obvious.

Assuming there aren't any, here's what I'm struggling with. This is my quick-and-dirty adaptation of the speech synthesis examples from the docs:

```
#[cfg(windows)]

winrt::import!(
    dependencies
        "os"
    modules
        "windows.foundation"
        "windows.media.playback"
        "windows.media.speechsynthesis"
);

use log::{info, trace};
use windows::media::playback::MediaPlayer;
use windows::media::speech_synthesis::SpeechSynthesizer;
...

    fn speak(&self, text: &str, interrupt: bool) -> std::result::Result<(), Error> {
        trace!("speak({}, {})", text, interrupt);
        let stream = self
            .synth
            .synthesize_text_to_stream_async(text)
            .expect("Failed to create stream")
            .get_results()
            .expect("Failed to get results");
        self.player.set_uri_source(stream.get_absolute_uri());
        self.player.play();
        Ok(())
    }
...
```

Here are some of the errors I get:

```
error[E0369]: binary operation `==` cannot be applied to type `backends::winrt::windows::foundation::IReference<u64>`
  --> src\backends\winrt.rs:3:1
   |
3  | / winrt::import!(
4  | |     dependencies
5  | |         "os"
6  | |     modules
...  |
9  | |         "windows.media.speechsynthesis"
10 | | );
   | |__^
   |
   = note: an implementation of `std::cmp::PartialEq` might be missing for `backends::winrt::windows::foundation::IReference<u64>`
   = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0369]: binary operation `!=` cannot be applied to type `backends::winrt::windows::foundation::IReference<u64>`
  --> src\backends\winrt.rs:3:1
   |
3  | / winrt::import!(
4  | |     dependencies
5  | |         "os"
6  | |     modules
...  |
9  | |         "windows.media.speechsynthesis"
10 | | );
   | |__^
   |
   = note: an implementation of `std::cmp::PartialEq` might be missing for `backends::winrt::windows::foundation::IReference<u64>`
[...]

Issue #131, opened 2020-05-10
`windows::foundation::IReference<u64>` errors from `import!`


DusterTheFirst:
with an import of
```rs
import!(
    dependencies
        "os"
    modules
        "windows.media.speech_synthesis"
);
```

I am given errors that i cannot understand how to fix

```
error[E0369]: binary operation `==` cannot be applied to type `windows::foundation::IReference<u64>`
 --> src/synth.rs:3:1
  |
3 | / import!(
4 | |     dependencies
5 | |         "os"
6 | |     modules
7 | |         "windows.media.speech_synthesis"
8 | | );
  | |__^
  |
  = note: an implementation of `std::cmp::PartialEq` might be missing for `windows::foundation::IReference<u64>`

error[E0369]: binary operation `!=` cannot be applied to type `windows::foundation::IReference<u64>`
 --> src/synth.rs:3:1
  |
3 | / import!(
4 | |     dependencies
5 | |         "os"
6 | |     modules
7 | |         "windows.media.speech_synthesis"
8 | | );
  | |__^
  |
  = note: an implementation of `std::cmp::PartialEq` might be missing for `windows::foundation::IReference<u64>`

error[E0277]: `windows::foundation::IReference<u64>` doesn't implement `std::fmt::Debug`
 --> src/synth.rs:3:1
  |
3 | / import!(
4 | |     dependencies
5 | |         "os"
6 | |     modules
7 | |         "windows.media.speech_synthesis"
8 | | );
  | |__^ `windows::foundation::IReference<u64>` cannot be formatted using `{:?}`
  |
  = help: the trait `std::fmt::Debug` is not implemented for `windows::foundation::IReference<u64>`
  = note: add `#[derive(Debug)]` or manually implement `std::fmt::Debug`
  = note: required because of the requirements on the impl of `std::fmt::Debug` for `&windows::foundation::IReference<u64>`
  = note: required for the cast to the object type `dyn std::fmt::Debug`
```

kennykerr:
Hi Zachary, 

Yes, that is a known issue and the same root cause as https://github.com/microsoft/winrt-rs/issues/104

I hope to get that working soon.

K

DusterTheFirst:
@kennykerr thanks for the quick response. I didn't see that issue earlier due to the weird error but am glad that it has already been found!

kennykerr:
[...]

PR #137, opened 2020-05-14
Implement Debug and PartialEq for WinRT types

### LINKED ISSUE #104: Support for procedural Xaml
While you can import `windows.ui.xaml`, importing `windows.ui.xaml.controls` fails for a few different reasons. I've resolved a few of them, but at least one more remains (structs with `IReference<T>` members). Coming soon.

* WinRT types (identified by `RuntimeType`) must implement the `Debug` trait. This will allow `IReference<T>` to implement the `Debug` trait that will then allow WinRT structs that already implement this trait to compile.
* WinRT types must also implement the `PartialEq` trait. This again supports structs with `IReference<T>` fields. For class/interface types this should query for `IUnknown` and compare those pointer values for COM identity.

kennykerr:
Fixes #104 #131 #134

### DIFF
--- a/crates/winmd/src/types/delegate.rs
+++ b/crates/winmd/src/types/delegate.rs
@@ -60,7 +60,7 @@ impl Delegate {
 
         quote! {
             #[repr(transparent)]
-            #[derive(Default)]
+            #[derive(Default, PartialEq)]
             pub struct #definition where #constraints {
                 ptr: ::winrt::ComPtr<#name>,
                 #phantoms
--- a/crates/winmd/src/types/delegate.rs
+++ b/crates/winmd/src/types/delegate.rs
@@ -105,6 +105,15 @@ impl Delegate {
                     self.ptr.set_abi()
                 }
             }
+            impl<#constraints> ::std::fmt::Debug for #name {
+                fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::fmt::Result {
+                    write!(
+                        f,
+                        "{:?}",
+                        <Self as ::winrt::RuntimeType>::abi(self)
+                    )
+                }
+            }
             #[repr(C)]
             struct #impl_definition where #constraints {
                 vtable: *const #abi_definition,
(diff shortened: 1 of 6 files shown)


crates/winmd/src/types/delegate.rs
```diff
@@ -105,6 +105,15 @@ impl Delegate {
                     self.ptr.set_abi()
                 }
             }
+            impl<#constraints> ::std::fmt::Debug for #name {
+                fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::fmt::Result {
+                    write!(
+                        f,
+                        "{:?}",
```
rylev, line 112: This doesn't seem too helpful. Shouldn't we also print the name of the type as well?

crates/winmd/src/types/delegate.rs
```diff
@@ -60,7 +60,7 @@ impl Delegate {
 
         quote! {
             #[repr(transparent)]
-            #[derive(Default)]
+            #[derive(Default, PartialEq)]
```
rylev, line 63: Derives do the conservative thing and place a matching constraint on all generics for the derived trait. So for example:
```rust
#[derive(PartialEq)]
struct Foo<T> {
  f: std::marker::PhantomData<T>
}
// Foo<T> is only PartialEq when T: PartialEq
```
Is this what we want? Or do we just want to differ to the `self.ptr == other.ptr`? If so, we need to implement `PartialEq` ourselves.

kennykerr: Ah right. Yes, we just want to compare the pointers.

crates/winmd/src/types/delegate.rs
kennykerr: The was primarily for completeness to unblock the issues with importing WinRT modules that used WinRT structs that have interface fields. WinRT structs mainly have primitive value fields and its very useful to have the Debug trait for those, but there are a handful that also have interface fields. I created https://github.com/microsoft/winrt-rs/issues/138 to address making these more helpful.

crates/winmd/src/types/delegate.rs
kennykerr: https://github.com/microsoft/winrt-rs/issues/140

PR #150, opened 2020-05-19
Acquire extended information when an error occurs

roblabla:
Call GetRestrictedErrorInfo() to acquire full error description when an error occurs. This change also adds an FFI wrapper around the BSTR type, and manual bindings to the IRestrictedErrorInfo COM object.

This makes printing the errors a lot more useful, giving detailed information. Here's an example of a detailed error I acquired while working on a project: 

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: ErrorCode(-2147418113), info: Some(IRestrictedErrorInfo { description: "Catastrophic failure\r\n", error_code: ErrorCode(-2147418113), restricted_description: "WindowsXamlManager and DesktopWindowXamlSource are supported for apps targeting Windows version 10.0.18226.0 and later.  Please check either the application manifest or package manifest and ensure the MaxTestedVersion property is updated.", capability_sid: "" }) }', src\main.rs:31:26
```


kennykerr:
Thanks Robin! That's the gist of it for consumption but there are a few issues I need to address before we can turn this on - some things I haven't fully documented and some things I need to clean up in the way Rust/WinRT takes dependencies. I'll prioritize this as I know getting good error information is invaluable.

PR #151, opened 2020-05-19
A better debug

### LINKED ISSUE #138: Specialize Debug trait for well-known foundation types
Types like `IStringable` and `IReference<T>` should have specialized `Debug` traits generated that produce more helpful results. For `IStringable` it should call `to_string()` and for `IReference<T>` it should delegate to `T`'s `Debug` trait implementation.

### LINKED ISSUE #149: Implement std::fmt::Display trait for all classes/interfaces that implement IStringable

rylev:
Fixes #138

### DIFF
--- a/crates/winmd/src/types/debug.rs
+++ b/crates/winmd/src/types/debug.rs
@@ -1,0 +1,60 @@
+use crate::types::RequiredInterface;
+use crate::types::TypeName;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+pub fn debug_tokens(name: &TypeName, interfaces: &Vec<RequiredInterface>) -> TokenStream {
+    let constraints = name.constraints();
+    let clean_name = name.clean_name();
+    let name = name.to_tokens(&name.namespace);
+
+    let default_impl = quote! { ::std::format!("{}({:?})", #clean_name, <Self as ::winrt::RuntimeType>::abi(self)) };
+
+    let implements_istringable = interfaces.iter().any(|interface| {
+        interface.name.name == "IStringable" && interface.name.namespace == "Windows.Foundation"
+    });
+    let is_istringable = &interfaces.get(0).map(|i| i.name.name.as_str()) == &Some("IStringable")
+        && &interfaces[0].name.namespace == "Windows.Foundation";
+
+    let implementation = if implements_istringable && !is_istringable {
+        quote! {
+            "{:?}",
+            {
+                let s: crate::windows::foundation::IStringable = self.into();
+                s
+            }
+        }
+    } else if is_istringable {
+        quote! {
+            "{}",
+            match self.to_string(){
+                Ok(s) => s.to_string(),
+                Err(_) => #default_impl
+            }
+        }
+    } else {
+        quote! {
+            "{}",
+            {
+                // TODO: assume IStringable
+                // let s: ::winrt::Result<IStringable> = <self as ::winrt::TryInto>::try_into();
+                // match s {
+                //     ::std::result::Result::Ok(s) => format!("{}", s),
+                //     ::std::result::Result::Err(_) => #default_impl
+                // }
+                #default_impl
+            }
+
+        }
+    };
+    quote! {
+        impl<#constraints> ::std::fmt::Debug for #name {
+            fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> ::std::fmt::Result {
+                write!(
+                    f,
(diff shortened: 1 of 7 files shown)


tests/debug.rs
```diff
@@ -0,0 +1,22 @@
...
+        os
+    modules
+        "windows.data.xml.dom"
+        "windows.foundation.collections"
+);
+
+use windows::data::xml::dom::XmlDocument;
+use windows::foundation::Uri;
+
+// Simple test to validate that default constructors are projected as static `new` methods.
```
kennykerr, line 12: Copy-pasted commented...

crates/winmd/src/types/delegate.rs
```diff
@@ -57,6 +58,7 @@ impl Delegate {
             .params
             .iter()
             .map(|param| param.to_invoke_arg_tokens());
+        let debug = debug::debug_tokens(&self.name, &Vec::new());
```
kennykerr, line 61: Delegates never implement IStringable.

crates/winmd/src/types/type_name.rs
```diff
@@ -351,6 +351,13 @@ impl TypeName {
 
         TokenStream::from_iter(generics)
     }
+
+    pub fn clean_name(&self) -> &str {
+        if self.name.starts_with("#r") {
```
kennykerr, line 356: Not sure why this is here. This will never be true.

crates/winmd/src/types/debug.rs
```diff
@@ -0,0 +1,60 @@
+use crate::types::RequiredInterface;
+use crate::types::TypeName;
+use proc_macro2::TokenStream;
+use quote::quote;
+
+pub fn debug_tokens(name: &TypeName, interfaces: &Vec<RequiredInterface>) -> TokenStream {
+    let constraints = name.constraints();
```
kennykerr, line 7: Since the caller of debug_tokens already has the constraints, perhaps we can just pass it in to avoid recomputing.

crates/winmd/src/types/debug.rs
```diff
@@ -0,0 +1,60 @@
...
+    let clean_name = name.clean_name();
+    let name = name.to_tokens(&name.namespace);
```
kennykerr, line 9: Same for the name.

crates/winmd/src/types/debug.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let default_impl = quote! { ::std::format!("{}({:?})", #clean_name, <Self as ::winrt::RuntimeType>::abi(self)) };
+
+    let implements_istringable = interfaces.iter().any(|interface| {
+        interface.name.name == "IStringable" && interface.name.namespace == "Windows.Foundation"
+    });
+    let is_istringable = &interfaces.get(0).map(|i| i.name.name.as_str()) == &Some("IStringable")
```
kennykerr, line 16: Why not just compare against the name?

crates/winmd/src/types/delegate.rs
kennykerr: But its probably harmless and not bad to keep the Debug code gen in one spot.

crates/winmd/src/types/type_name.rs
rylev: https://github.com/microsoft/winrt-rs/blob/master/crates/winmd/src/lib.rs#L25

crates/winmd/src/types/debug.rs
rylev: Not sure what you mean?

crates/winmd/src/types/type_name.rs
kennykerr: But that's the token representation - this is the original string name.

crates/winmd/src/types/debug.rs
rylev: Definitely more tedius to do this, but if we think it will make a perf difference I can do it.

kennykerr: Yes, since there's potentially tens of thousands of interfaces. Welcome suggestions for making this less tedious. We can also profile - not sure how expensive it is to compute the token stream.

crates/winmd/src/types/debug.rs
kennykerr:
```rust
let is_istringable = name.name == "IStringable" && name.namespace == "W.F";
```

PR #154, opened 2020-05-20
Add basic build crate for build.rs support

rylev:
This is based on @fasterthanlime's work. I'd like to get their approval first. I can also add @fasterthanlime as an author. 

In a build.rs script for your project you can run:

**EDIT: THIS HAS CHANGED. SEE BELOW**
```rust
fn main() {
    winrt_build::build(&["windows.foundation"], "./src/bindings.rs").unwrap()
}
```

And then use them in:

```rust
mod bindings;
use bindings::windows;
```

### DIFF
--- a/crates/winrt-build/Cargo.toml
+++ b/crates/winrt-build/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "winrt-build"
+version = "0.1.0"
+authors = ["Ryan Levick <ryan.levick@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+winmd = { path = "../winmd" }
--- a/crates/winrt-build/Cargo.toml
+++ b/crates/winrt-build/Cargo.toml
@@ -1,10 +1,8 @@
 [package]
 name = "winrt-build"
 version = "0.1.0"
-authors = ["Ryan Levick <ryan.levick@gmail.com>"]
+authors = ["Microsoft"]
 edition = "2018"
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 winmd = { path = "../winmd" }
--- a/crates/winrt-build/src/lib.rs
+++ b/crates/winrt-build/src/lib.rs
@@ -17,6 +17,33 @@ pub enum Dependencies {
     OsAndNuget(Vec<String>),
 }
 
+impl Dependencies {
+    fn winmds(&self) -> Vec<winmd::WinmdFile> {
+        let mut result = std::collections::BTreeSet::new();
+        let deps = match self {
+            Dependencies::Os => vec![winmd::dependencies::system_metadata_root()],
+            Dependencies::Nuget(deps) => deps
+                .iter()
+                .map(|d| winmd::dependencies::nuget_root().join(d))
+                .collect(),
+            Dependencies::OsAndNuget(deps) => {
+                let mut deps: Vec<PathBuf> = deps
+                    .iter()
+                    .map(|d| winmd::dependencies::nuget_root().join(d))
+                    .collect();
+
+                deps.push(winmd::dependencies::system_metadata_root());
+                deps
+            }
+        };
+        for dep in deps {
+            let dep = winmd::dependencies::nuget_root().join(dep);
+            winmd::dependencies::expand_paths(dep, &mut result, true)
+        }
+        result.iter().map(winmd::WinmdFile::new).collect()
+    }
+}
+
 impl Default for Builder {
     fn default() -> Self {
         Self {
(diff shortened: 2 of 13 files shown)


fasterthanlime:
Looks good to me!

My two cents: generating files in build scripts outside of the `OUT_DIR` environment variable is generally frowned upon, see https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

So maybe the example should encourage using `OUT_DIR` instead, with a stub `src/bindings.rs` files that has an `include!()` directive?

Mentioning [rerun-if-changed](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) might be a good idea too, otherwise they'd pay the same penalty as when using the proc macro version.

You may also want to mention that it helps to put the bindings in a separate crate, for recompilation speeds. Do verify that this isn't superstition though, I'm fuzzy on the details. Maybe we can get some Cargo folks to look at this?

roblabla:
I can definitely say that putting bindings in a separate crate is a huge boon on recompilation speed. In https://github.com/roblabla/ubuntu-installer, if I move the bindings (which takes a whooping 75MiB...) in the main crate, it takes ~10 minutes to build the main crate on my machine, even with incremental compilation. By splitting the bindings in their own crate, I got compile time way down to a second or two after the bindings crate is compiled.

crates/winrt-build/src/lib.rs
```diff
@@ -0,0 +1,51 @@
+use std::{fs, io::Write, path::Path, process::Stdio};
+
+pub fn build<P: AsRef<Path>>(
+    namespaces: &[&str],
+    output: P,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let output = output.as_ref();
+    let tr = winmd::TypeReader::from_os();
```
kennykerr, line 8: The dependencies should be configurable as it is with the `import` macro.

kennykerr:
I think this is great. Thanks for the inspiration @fasterthanlime!

About the "whooping 75MiB" of generated code, https://github.com/microsoft/winrt-rs/issues/119 should help - Xaml is huge. 😉

crates/winrt-build/src/lib.rs
watislaf: ца

rylev:
Just making a note that before we merge this we should plan to address the comments in #159

rylev:
This is ready for re-review.

Now the build.rs file is:
```rust
use winrt::Builder;

fn main() {
    Builder::default()
        .insert_namespaces(&["windows.foundation"])
        .insert_nuget(&["Microsoft.Windows.SDK.Contracts"])
        .build()
        .unwrap();
}
```

And usage is:

```rust
#![allow(overflowing_literals)]

include!(concat!(env!("OUT_DIR"), "/bindings.rs"));

use windows::foundation::PropertyValue;

fn main() {
    let _value = PropertyValue::create_boolean(true).unwrap();
}
```

crates/winrt-build/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "winrt-build"
+version = "0.1.0"
+authors = ["Ryan Levick <ryan.levick@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
kennykerr, line 7: For consistency with the other toml files, should we set the author to "Microsoft" and remove this comment?

crates/winrt-build/src/lib.rs
```diff
@@ -0,0 +1,122 @@
...
+    pub fn insert_nuget<T>(mut self, nuget_deps: T) -> Self
+    where
+        T: IntoIterator,
+        T::Item: std::string::ToString,
+    {
+        let new_deps = nuget_deps.into_iter().map(|i| i.to_string());
+        match &mut self.dependencies {
+            Dependencies::Os => self.dependencies = Dependencies::Nuget(new_deps.collect()),
+            Dependencies::Nuget(deps) => deps.extend(new_deps),
+            Dependencies::OsAndNuget(deps) => deps.extend(new_deps),
```
kennykerr, line 51: It's not obvious how we can use this to get "OsAndNuget".

kennykerr: Does it know that the SDK nuget package supersedes the "OS" metadata?

rylev: You can't as of now. I can remove it or we can block this PR until I implement a way to do that.

kennykerr: May need to wait since the nuget support doesn't yet support the sdk nuget.

rylev:
This fixes #159

crates/winrt-build/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+            let child = cmd.spawn()?;
+            let mut stdin = child.stdin.unwrap();
+            let stdout = child.stdout.unwrap();
+
+            let t = std::thread::spawn(move || {
+                let mut s = stdout;
+                std::io::copy(&mut s, &mut f).unwrap();
+            });
+
+            // workaround for, well, overflowing literals
```
kennykerr, line 117: Old comment?

crates/winrt-build/src/lib.rs
```diff
@@ -0,0 +1,137 @@
...
+            // writeln!(&mut stdin, "#![allow(overflowing_literals)]")?;
+
+            // Only rerun if the
```
kennykerr, line 120: If the what?

PR #156, opened 2020-05-20
Fix stack overflow in From<HString> for String impl

roblabla:
The current impl does an infinite recursion into itself, eventually leading to a stack overflow. I suppose the implementation was supposed to call the `impl From<&HString> for String` defined above.

### DIFF
--- a/src/hstring.rs
+++ b/src/hstring.rs
@@ -188,7 +188,7 @@ impl<'a> From<&'a HString> for String {
 
 impl From<HString> for String {
     fn from(hstring: HString) -> Self {
-        hstring.into()
+        String::from(&hstring)
     }
 }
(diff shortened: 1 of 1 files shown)


src/hstring.rs
```diff
@@ -317,4 +317,9 @@ mod tests {
         let h = HString::from("");
         assert!(format!("{}", h) == "");
     }
+
+    #[test]
+    fn hstring_to_string() {
+        String::from(HString::from("test"));
```
kennykerr, line 323: Good catch! Please add an assert for the result. Perhaps something like this:

```rust
let h = HString::from("test");
let s = String::from(h);
assert!(s == "test");
```

PR #161, opened 2020-05-27
Provide richer error information via Result<T>

### LINKED ISSUE #89: Provide richer error information via Result<T>
`winrt::Error` needs to encapsulate an `IRestrictedErrorInfo` in the same way that C++/WinRT's `winrt::hresult_error` exception type carries the error information from the call site for safe propagation and chaining and to provide richer error information for debugging.

kennykerr:
Fixes #89 

This provides the basics of WinRT error origination - the underlying API isn't super easy to use, but this takes care of some basic fallback scenarios for older API conventions. There's more to be done, but this should help with error information in the short term.

Inspired by @roblabla, I added a `Debug` trait implementation for `Error` to provide better error information for failing callers. For example, given the following buggy program:

```rust
winrt::import!(
    dependencies
        os
    modules
        "windows.foundation"
);

fn main() -> winrt::Result<()> {
    windows::foundation::Uri::create_uri("Invalid")?;

    Ok(())
}
```

The output will be as follows:

```
> target\debug\scratch.exe
Error: Error { code: 0x80070057, message: "Invalid is not a valid absolute URI." }
```


src/error.rs
```diff
@@ -1,39 +1,124 @@
...
 }
 
-/// A WinRT related error
-#[derive(Debug)]
-pub struct Error {
-    code: ErrorCode,
-    // TODO: add `info: IErrorInfo`
+impl std::convert::From<ErrorCode> for Error {
+    fn from(code: ErrorCode) -> Self {
+        unsafe {
```
rylev, line 38: I would really love for us to be in the habit of only marking the code that actually requires `unsafe` as such. Doing it like this makes it much harder to know what code needs to have closer inspection and what doesn't

src/error.rs
```diff
@@ -1,39 +1,124 @@
...
+            let mut info = IErrorInfo::default();
+            GetErrorInfo(0, info.set_abi() as _);
+            let restricted = info.query::<IRestrictedErrorInfo>();
+
+            if !restricted.is_null() {
+                let info2 = info.query::<ILanguageExceptionErrorInfo2>();
```
rylev, line 44: nit: I think just shadowing the `info` binding is better than using a name like `info2`

src/error.rs
```diff
@@ -1,39 +1,124 @@
...
+
+                return Self {
+                    code,
+                    info: restricted,
+                };
+            }
+
+            let mut message = String::new();
+
+            if !info.is_null() {
```
rylev, line 58: All the null checking is very annoying especially since the compiler won't enforce it. More reason to have us be able to model WInRT types as not nullable.

src/error.rs
```diff
@@ -1,39 +1,124 @@
...
 
 impl Error {
+    pub fn new(code: ErrorCode, message: &str) -> Self {
+        unsafe {
+            let message: HString = message.into();
+            RoOriginateError(code, message.abi() as _);
+            let mut info = IErrorInfo::default();
+            GetErrorInfo(0, info.set_abi() as _);
+            let info = info.query::<IRestrictedErrorInfo>();
+            return Self { code, info };
```
rylev, line 75: `return` isn't needed here

src/error.rs
```diff
@@ -1,39 +1,124 @@
...
 
-/// The ErrorCode (a.k.a HRESULT) of an error
-#[repr(transparent)]
-#[derive(Copy, Clone, Debug, PartialEq)]
-pub struct ErrorCode(pub HRESULT);
+        let mut message = HeapString::new();
+
+        unsafe {
+            let size = FormatMessageW(
+                0x00001300, // FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS
```
rylev, line 95: Can we use constants instead of the comments?

src/error.rs
```diff
@@ -71,3 +156,218 @@ impl ErrorCode {
...
+            ptr: std::ptr::null_mut(),
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.ptr.is_null() || self.len() == 0
+    }
+
+    pub fn len(&self) -> usize {
+        unsafe { SysStringLen(self.ptr) as usize }
```
rylev, line 177: I'm assuming that `SysStringLen` can support null pointers? It would be also be good for us to get in the habit of noting why code marked as unsafe is in fact safe.

src/error.rs
```diff
@@ -71,3 +156,218 @@ impl ErrorCode {
 
     pub(crate) const NOT_INITIALIZED: ErrorCode = ErrorCode(0x8004_01F0);
 }
+
+#[repr(transparent)]
+struct BString {
+    ptr: RawPtr,
```
rylev, line 162: Any reason to not model this as `*mut u16` instead?

src/error.rs
```diff
@@ -71,3 +156,218 @@ impl ErrorCode {
...
+                from.ptr as *const u16,
+                from.len(),
+            ))
+        }
+    }
+}
+
+#[repr(transparent)]
+struct HeapString {
+    ptr: RawPtr,
```
rylev, line 218: Same as above. Why not `*mut u16`?

src/error.rs
tim-weis: [SysStringLen](https://docs.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-sysstringlen) can indeed be safely called with a null pointer (*"If `bstr` is null the return value is zero."*).

src/error.rs
kennykerr: I used it to clearly identify that it is referring to a `ILanguageExceptionErrorInfo2` (note the `Info2` suffix) as apposed to one of the regular info objects. These interfaces don't have great names. Perhaps I'll call this variable `capture` to make it clearer what this is about.

rylev: OH... sorry. Yes weird name... But that makes the variable name better considering...

src/error.rs
rylev: @kennykerr did you miss this one?

PR #162, opened 2020-05-27
Moving to use non-null pointers

rylev:
This change moves: `RawComPtr` from `*mut *mut VTable` to `Option<NonNull<NonNull<Vtable>>`. This more correctly models reality that the ComPtrs can be null but they point to non-null pointers to vtables

### DIFF
--- a/crates/winmd/src/types/delegate.rs
+++ b/crates/winmd/src/types/delegate.rs
@@ -142,7 +142,8 @@ impl Delegate {
                     };
                     unsafe {
                         let mut result: #name = std::mem::zeroed();
-                        *<#name as ::winrt::RuntimeType>::set_abi(&mut result) = ::std::boxed::Box::into_raw(::std::boxed::Box::new(value)) as *const *const #abi_definition;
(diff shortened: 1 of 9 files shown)


crates/winmd/src/types/delegate.rs
```diff
@@ -152,7 +153,7 @@ impl Delegate {
                     interface: *mut ::winrt::RawPtr,
                 ) -> ::winrt::ErrorCode {
                     unsafe {
-                        let this = this as *const Self as *mut Self;
+                        let this: *mut Self = this.map(|s| s.cast::<Self>()).unwrap().as_ptr();
```
kennykerr, line 156: Do we know how much more expensive this pattern is at run time?

rylev: It's the same except it adds a null check and a possible branch to a panic handler (due to the unwrap). Of course, if the ptr is null in the original code, who knows what will actually happen.

https://godbolt.org/z/6zNWvN

rylev:
I'll need to update the latest changes from your error handling PR before we can merge this.

Issue #142, opened 2020-05-16
Generate standalone libs so that projects can link without the Windows SDK


Boddlnagg:
In a prototype, I migrated from winrt-rust to winrt-rs in my MIDI library [midir](https://github.com/Boddlnagg/midir), see https://github.com/Boddlnagg/midir/pull/56.

In the CI for that project, I also test building with the windows-gnu toolchain, and apparently it doesn't work with winrt-rs. I'm getting a linker error:

```
error: linking with `i686-w64-mingw32-gcc` failed: exit code: 1
  |
[...]

PR #166, opened 2020-05-27
Fix GNU toolchain builds

Alovchin91:
This PR fixes the GNU toolchain (-pc-windows-gnu) builds for the apps that use winrt-rs.

Note that it is still not possible to build all the tests in winrt-rs itself with the GNU toolchain, but with MSVC toolchain all the tests pass.

It is probably a temporary workaround, though I see that the TODO to dynamically load the APIs from OneCore has been removed.

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -22,7 +22,7 @@ extern "system" {
     ) -> u32;
 }
 
-#[link(name = "onecore")]
+#[link(name = "windowsapp")]
 extern "system" {
     pub fn CoIncrementMTAUsage(cookie: *mut RawPtr) -> ErrorCode;
 
--- a/tests/clipboard.rs
+++ b/tests/clipboard.rs
@@ -22,7 +22,7 @@
 //     Ok(())
 // }
 
-// #[link(name = "onecore")]
+// #[link(name = "windowsapp")]
 // extern "system" {
 //     pub fn CoInitializeEx(reserved: usize, apartment: u32) -> ErrorCode;
 // }


src/runtime.rs
```diff
@@ -22,7 +22,7 @@ extern "system" {
     ) -> u32;
 }
 
-#[link(name = "onecore")]
+#[link(name = "windowsapp")]
```
kennykerr, line 25: Do you know why the GNU toolchain accepts `windowsapp` and not `onecore`? Both are just umbrella libs. And yes, I'm planning to avoid the problem with these libs with #142. This change is ok, but in general I'm unconfortable making a change that we don't verify in the build. Is there a way to validate both MSVC and GNU?

Alovchin91: The reason is because there is `libwindowsapp.a` in the MinGW-w64 toolchain but no `libonecore.a` - as simple as that :)

I can play with the tests to see if I can compile them for GNU, but the reason is quite similar - there is no `libcoremessaging.a` in the GNU toolchain.

Alovchin91: The approach in #142 is definitely a much better one.

kennykerr: We can probably simplify the tests to avoid the additional link dependencies. The concern I have with #142 is that I don't know whether that will work with something other than the MSVC lib tool.

Alovchin91: The only usage of `coremessaging` is for `CreateDispatcherQueueController`. This is in `composition.rs` tests. I'm not sure how this can be simplified tbh, but we could probably load this function dynamically instead?

Alovchin91: The approach in #142 should work for MinGW-w64 just fine from my understanding. What you need is a *.def file and the `dlltool` that will generate a *.a file based on it. To automate *.def file generation you could also use `gendef` first. Example (from [here](https://dev.my-gate.net/2018/06/02/generate-a-def-file-from-a-dll/)):

```
gendef.exe csfml-graphics-2.dll
dlltool.exe -d csfml-graphics-2.def -D csfml-graphics-2.dll -l libcsfml-graphics.a
```

These tools are part of MinGW-w64 toolchain and should be available on all platforms.

kennykerr: Let's take this change for now and deal with the tests separately. It would be nice to have a simple LoadLibrary/GetProcAddress helper.

PR #168, opened 2020-05-27
Avoid the need for allowing overflowing_literals

### LINKED ISSUE #164: Make ErrorCode unsigned to avoid overflowing_literals error

kennykerr:
Fixes #164 

HRESULT is traditionally defined as a signed 32-bit integer but we don't need to do so in Rust so long as we handle Severity bit checks (bit 31) properly.

Also fixed a few code tidiness issues.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,12 +1,10 @@
-#![allow(overflowing_literals)]
-
 use crate::runtime::*;
 use crate::*;
 
 /// The ErrorCode (a.k.a HRESULT) of an error
 #[repr(transparent)]
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub struct ErrorCode(pub i32);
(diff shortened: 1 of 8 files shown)


src/error.rs
```diff
@@ -1,12 +1,10 @@
-#![allow(overflowing_literals)]
-
 use crate::runtime::*;
 use crate::*;
 
 /// The ErrorCode (a.k.a HRESULT) of an error
 #[repr(transparent)]
 #[derive(Copy, Clone, Debug, PartialEq)]
-pub struct ErrorCode(pub i32);
+pub struct ErrorCode(pub u32);
```
rylev, line 7: Is it useful to have a From/Into conversion for Error and i32?

kennykerr: I don't expect that to be common with WinRT APIs.

PR #170, opened 2020-05-28
NonNullRawComPtr

rylev:
Building off of #162, this provides an explicit type for `RawComPtr`s that are known to be non-null. This mostly happens at ABI function calling boundary. The code now forces that nulls have been checked and that the  `this` pointer is not null. 

The name `NonNullRawComPtr` is pretty terrible, but it is descriptive. Suggestions on improvements are more than welcome.

### DIFF
--- a/crates/winmd/src/types/class.rs
+++ b/crates/winmd/src/types/class.rs
@@ -125,7 +125,7 @@ impl Class {
             quote! {
                 #[repr(transparent)]
                 #[derive(Default, Clone, PartialEq)]
-                pub struct #name { ptr: ::winrt::ComPtr<#name> }
+                pub struct #name { ptr: ::winrt::ComPtr<#default_name> }
                 impl #name {
                     #new
                     #methods
@@ -139,12 +139,12 @@ impl Class {
                     }
                 }
                 unsafe impl ::winrt::RuntimeType for #name {
-                    type Abi = ::winrt::RawComPtr<Self>;
+                    type Abi = ::winrt::RawComPtr<#default_name>;
                     fn signature() -> String {
                         #signature.to_owned()
                     }
                     fn abi(&self) -> Self::Abi {
-                        <::winrt::ComPtr<Self> as ::winrt::ComInterface>::as_raw(&self.ptr)
+                        self.ptr.abi()
                     }
                     fn set_abi(&mut self) -> *mut Self::Abi {
                         self.ptr.set_abi()
--- a/crates/winmd/src/types/delegate.rs
+++ b/crates/winmd/src/types/delegate.rs
@@ -176,7 +176,7 @@ impl Delegate {
                 }
                 extern "system" fn unknown_release(this: ::winrt::NonNullRawComPtr<::winrt::IUnknown>) -> u32 {
                     unsafe {
-                        let this: *mut Self = this.as_raw() as _;                        let remaining = (*this).count.release();
+                        let this: *mut Self = this.as_raw() as _;
                         let remaining = (*this).count.release();
 
                         if remaining == 0 {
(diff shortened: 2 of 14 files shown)


crates/winmd/src/types/class.rs
```diff
@@ -125,7 +125,7 @@ impl Class {
             quote! {
                 #[repr(transparent)]
                 #[derive(Default, Clone, PartialEq)]
-                pub struct #name { ptr: ::winrt::ComPtr<#name> }
+                pub struct #name { ptr: ::winrt::ComPtr<#default_name> }
```
rylev, line 128: This might be slightly controversial. Now classes are wrappers around ComPtrs to their default interfaces. This is of course functionaliy equivalent to what was there before, but means that the ABI type is different. For example, `windows::foundation::Uri`'s ABI type is now `RawComPtr<windows::foundation::IUriRuntimeClass>`. Again this is functionally equivalent, and it could be argued more "correct", but different than before.

crates/winmd/src/types/delegate.rs
```diff
@@ -100,7 +100,7 @@ impl Delegate {
                     #signature
                 }
                 fn abi(&self) -> Self::Abi {
-                    <::winrt::ComPtr<Self> as ::winrt::ComInterface>::as_raw(&self.ptr)
+                    self.ptr.abi()
```
rylev, line 103: This seems cleaner. We're now delegating to the inner pointer's ABI

src/agile_object.rs
```diff
@@ -22,7 +22,7 @@ unsafe impl ComInterface for IAgileObject {
 #[repr(C)]
 pub struct abi_IAgileObject {
     pub unknown_query_interface:
-        extern "system" fn(RawComPtr<IUnknown>, &Guid, *mut RawPtr) -> ErrorCode,
-    pub unknown_add_ref: extern "system" fn(RawComPtr<IUnknown>) -> u32,
-    pub unknown_release: extern "system" fn(RawComPtr<IUnknown>) -> u32,
+        unsafe extern "system" fn(RawComPtr<IUnknown>, &Guid, *mut RawPtr) -> ErrorCode,
```
rylev, line 25: This is a preview of a change to come. All abi methods should be marked unsafe since it's up to us to make sure they're used properly.

crates/winmd/src/types/class.rs
kennykerr: No objections, I would argue this is more correct.

crates/winmd/src/types/delegate.rs
```diff
@@ -168,15 +168,15 @@ impl Delegate {
...
+                        let this: *mut Self = this.as_raw() as _;
                         (*this).count.add_ref()
                     }
                 }
-                extern "system" fn unknown_release(this: ::winrt::RawComPtr<::winrt::IUnknown>) -> u32 {
+                extern "system" fn unknown_release(this: ::winrt::NonNullRawComPtr<::winrt::IUnknown>) -> u32 {
                     unsafe {
-                        let this: *mut Self = this.map(|s| s.cast::<Self>()).unwrap().as_ptr();
+                        let this: *mut Self = this.as_raw() as _;                        let remaining = (*this).count.release();
                         let remaining = (*this).count.release();
```
kennykerr, line 180: There's a double release here.

kennykerr:
>                 pub unknown_query_interface: extern "system" fn(::winrt::RawComPtr<::winrt::IUnknown>, &::winrt::Guid, *mut ::winrt::RawPtr) -> ::winrt::ErrorCode,

I assume the interface ABI should also use NonNullRawComPtr

---
Refers to: crates/winmd/src/types/interface.rs:100 in 28d05b6. [](commit_id = 28d05b6765be5e1b5a97c1170e9754d9d2bef60f, deletion_comment = False)

PR #172, opened 2020-05-28
Documentation

kennykerr:
First stab at documentation and some code cleanup.

### DIFF
--- a/src/com/interface.rs
+++ b/src/com/interface.rs
@@ -1,58 +1,56 @@
 use crate::*;
 
-/// A COM interface
+/// A COM interface.
 ///
 /// # Safety
 /// Implementors of this trait must be transparent wrappers
-/// around pointers to the associated VTable for the interface.
+/// around pointers to the associated vtable for the interface.
 /// In other words, implementators must be exactly equivalent to
-/// *mut *mut Self::VTable.
-///
-/// VTables must be COM compliant VTables where the first three function
-/// pointers are the IUknown methods.
-///
-/// Additionally, because ComInterfaces are just pointers to vtables,
-/// it must be safe to zero initialize the interface.
+/// `*mut *mut Self::VTable`.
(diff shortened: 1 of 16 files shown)


src/com/interface.rs
```diff
@@ -1,58 +1,56 @@
...
-///
-/// VTables must be COM compliant VTables where the first three function
-/// pointers are the IUknown methods.
-///
-/// Additionally, because ComInterfaces are just pointers to vtables,
-/// it must be safe to zero initialize the interface.
+/// `*mut *mut Self::VTable`.
 pub unsafe trait ComInterface: Sized {
+    /// `VTable` must be a COM compliant vtable where the first three function
+    /// pointers are the `IUnknown` methods. And because ComInterfaces are just
```
rylev, line 12: This last sentence is about ComInterfaces as a whole not about vtables. I think we should move it back up

src/com/ref_count.rs
```diff
@@ -1,28 +1,37 @@
-use std::sync::atomic::{self, AtomicU32, Ordering};
+use std::sync::atomic::{self, AtomicI32, Ordering};
 
+/// A thread-safe reference count for use with COM implementations.
 #[repr(transparent)]
 pub struct RefCount {
-    value: AtomicU32,
+    value: AtomicI32,
```
rylev, line 6: Why is this better as an i32? Isn’t it a bug if the count goes below 0?

src/com/interface.rs
kennykerr: Will do.

src/com/ref_count.rs
kennykerr: Yes, it is a bug if the count goes below zero. Using `i32` lets us detect this and panic if the object is used after being released. That's the panic below when we release and realize we're already at zero. Hard to test for though...

PR #174, opened 2020-05-29
Allow partial import of WinRT namespaces

### LINKED ISSUE #119: Import macro should support importing types to improve compilation time
Today, the import macro lets you specify WinRT namespaces (called modules). Two things specifically contribute to compilation time:

* Every type from every listed module is imported
* Every type from every parent module is automatically included

This is convenient but affects compilation time for larger modules. If you only need to use two controls from the `windows.ui.xaml.controls` module you would still end up with every other type in the controls module as well as every type in the `windows.ui.xaml` module and every type in the `windows.ui` module. 

We should use the Rust path syntax and allow more fine grained imports. Instead of `windows.ui` we should be able to say `windows::ui::*` and also optionally identify a subset of types e.g. `windows::ui::xaml::controls::{Button, CalendarView}`

rylev:
Fixes #119


tests/enum.rs
```diff
@@ -2,8 +2,8 @@ winrt::import!(
     dependencies
         os
     modules
-        "windows.foundation"
-        "windows.application_model.appointments"
+        windows::foundation::*
+        windows::application_model::appointments::AppointmentDaysOfWeek
```
kennykerr, line 6: Perhaps `modules` should be renamed `types`...

kennykerr:
I think the build is failing because of the code in the readme.

PR #175, opened 2020-05-29
Add cargo nuget

### LINKED ISSUE #173: Move cargo nuget into this repo
[`cargo nuget`](https://github.com/rylev/cargo-nuget) can be used to install NuGet packages that work in the context of WinRT-rs. We should move that crate into this repository.

rylev:
Fixes #173


crates/cargo-nuget/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+use std::path::{Path, PathBuf};
+
+fn main() {
+    let Opt::Nuget { subcommand } = Opt::from_args();
+    match subcommand {
+        Subcommand::Install(i) => i.perform().unwrap(),
+    }
+}
+
+/// A utility for interacting with nuget packages
```
kennykerr, line 17: This comment seems out of place.

crates/cargo-nuget/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+            for dll in dep.dlls() {
+                dll.write(&dep_directory).unwrap();
+            }
+        }
+
+        Ok(())
+    }
+}
+
+fn workspace_root() -> PathBuf {
```
kennykerr, line 74: Maybe just remove this function for now?

crates/cargo-nuget/src/main.rs
rylev: I believe `StructOpt` uses this to display helpful info the user.

crates/cargo-nuget/src/main.rs
rylev: I'd rather just implement it :-)

Issue #176, opened 2020-05-29
Calling ABI methods not always unsafe


kennykerr:
The following generated `unsafe` now sometimes causes an "unnecessary unsafe block" warning. Previously all of these calls to the ABI were unsafe but now thanks to recent safety improvements only those involving certain parameters are unsafe. 

https://github.com/microsoft/winrt-rs/blob/dfb144b85a253067e59cf4bc543b156606c2dc85/crates/winmd/src/types/method.rs#L250

rylev:
I would argue that all ABI calls should be marked as unsafe. It is not safe to call outside of Rust because Rust cannot guarantee its safety properties are upheld when control flows to another language. I think the right solution is actually to mark all vtable fields as unsafe.

kennykerr:
I mainly want to avoid the warnings, so that works for me. 👍

PR #177, opened 2020-05-29
Support reg-free activation

### LINKED ISSUE #158: Support reg-free activation
C++/WinRT offers registration-free activation so that WinRT components need not be registered in order to support activation from any desktop/console app. Rust/WinRT needs the same support to allow activation of 3rd party components outside of store apps.

kennykerr:
Fixes #158 and #176 

This update adds support for registration-free activation consistent with [C++/WinRT](https://github.com/microsoft/cppwinrt). 

Also added some documentation improvements and minor fixes.


src/factory.rs
```diff
@@ -1,26 +1,108 @@
...
-// https://github.com/microsoft/cppwinrt/blob/master/strings/base_activation.h
+/// Attempts to load the factory interface for the given WinRT class.
+///
+/// Note that factory caching still needs to be implemented.
 pub fn factory<C: RuntimeName, I: ComInterface>() -> Result<I> {
-    let mut ptr = std::ptr::null_mut();
+    let mut factory = std::ptr::null_mut();
+    let name = HString::from(C::NAME);
+
     unsafe {
```
kennykerr, line 11: I have a single unsafe block because this function is riddled with "unsafe" calls. It didn't seem to add value to add the unsafe block around each call.

src/factory.rs
```diff
@@ -1,26 +1,108 @@
...
+                continue;
+            }
+
+            let library_call: DllGetActivationFactory = std::mem::transmute(library_call);
+            let mut default_factory: IActivationFactory = std::mem::zeroed();
+
+            // Now call DllGetActivationFactory to request the given class.
+            if library_call(name.abi(), default_factory.set_abi()).is_err() {
+                continue;
+            }
```
rylev, line 71: Might be paranoia, but should we `debug_assert` that default_factory is not null?

src/factory.rs
```diff
@@ -1,26 +1,108 @@
...
+            // Attempt to load the DLL.
+            let library: Library = LoadLibraryW(path.as_ptr()).into();
+
+            if library.handle.is_null() {
+                continue;
+            }
+
+            // If the DLL was found then get the export used to retrieve the factory.
+            let library_call =
+                GetProcAddress(library.handle, b"DllGetActivationFactory\0".as_ptr());
```
rylev, line 59: nit: maybe `b"DllGetActivationFactory\0"` should be a constant?

src/factory.rs
rylev: That's fine. At some point in the future it would be nice to have safe  versions of the win32 APIs...

src/factory.rs
```diff
@@ -1,26 +1,108 @@
...
+        if !self.handle.is_null() {
+            unsafe {
+                FreeLibrary(self.handle);
+            }
+        }
+    }
+}
+
+impl From<RawPtr> for Library {
+    fn from(handle: RawPtr) -> Library {
```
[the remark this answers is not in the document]
rylev: Wonder if `LoadLibraryW` should return a `HMODULE` type and then this could be `From<HMODULE>`

src/factory.rs
kennykerr: Definitely.

PR #191, opened 2020-06-03
Move signatures to being static

rylev:
This makes `RuntimeType::signature` calls return a static reference. For the one place where we need to allocate, we cache into a global.

This is only temporary as eventually we'll do all of this at compile time, but at least now the type returned is the correct one.

### DIFF
--- a/crates/winmd/src/types/type_name.rs
+++ b/crates/winmd/src/types/type_name.rs
@@ -104,7 +104,7 @@ impl TypeName {
 
     pub fn to_signature_tokens(&self, signature: &str) -> TokenStream {
         if self.generics.is_empty() {
-            return quote! { #signature.to_owned() };
+            return quote! { #signature };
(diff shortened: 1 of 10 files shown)


crates/winmd/src/types/type_name.rs
```diff
@@ -123,7 +123,15 @@ impl TypeName {
         };
 
         quote! {
-            #format
+            static mut VAL: &'static str = "";
```
kennykerr, line 126: Are you sure this works? I believe the static storage is shared by all specializations of the generic. So `IVector<i32>` will have the same signature as `IVector<f32>`. This is unlike in C++ where each specialization gets its own static storage.

rylev: Ach sorry. You're right. I was wondering why it was so simple this time around.

PR #195, opened 2020-06-03
Add "build" macro for build script code gen

### LINKED ISSUE #183: Harmonize import and code generation options
Right now we have the `import` macro for generating code within a module and the `Builder` struct for generating code within a build script. The `import macro is pretty flexible and expressive:

```rust
winrt::import!(
    dependencies
        os
        nuget: Microsoft.AI.MachineLearning
    types
        microsoft::ai::machine_learning::*
        windows::foundation::{IStringable, IClosable}
);
```

The `Builder` struct relies on procedural code, which is more difficult to write and doesn't provide full fidelity with the `import` macro.

```rust
Builder::default()
    .insert_nuget(&["Microsoft.AI.MachineLearning"])
    .insert_namespaces(&["microsoft::ai::machine_learning", "windows.foundation"])
    .build()
    .unwrap();
```

Note that it doesn't support Rust paths to limit code generation to specific types. I suggest we replace the `Builder` struct with a `build_import` macro that provides the same or similar syntax to the existing `import` macro but generates code to a file.

kennykerr:
Fixes #183 

This replaces the `Builder` struct and provides feature parity with the `import` macro.


crates/macros/src/lib.rs
```diff
@@ -54,6 +53,52 @@ use std::path::PathBuf;
...
+    to_tokens(stream)
+}
+
+#[proc_macro]
+pub fn build(stream: TokenStream) -> TokenStream {
+    let tokens = to_tokens(stream);
+
+    // TODO: OUT_DIR is not available from the proc_macro..
+    // let path = PathBuf::from(env::var("OUT_DIR").expect("No `OUT_DIR` env variable set"));
+    let path = PathBuf::from("C:\\git\\rust");
```
kennykerr, line 65: The `OUT_DIR` environment variable isn't yet available when the proc_macro is executed. What to do...

crates/winmd/src/types/delegate.rs
```diff
@@ -189,10 +189,8 @@ impl Delegate {
                     }
                 }
                 #invoke_sig {
-                    unsafe {
-                        let this: *mut Self = this.as_raw() as _;
-                        ((*this).invoke)(#(#invoke_args,)*).into()
-                    }
+                    let this: *mut Self = this.as_raw() as _;
+                    ((*this).invoke)(#(#invoke_args,)*).into()
```
kennykerr, line 193: The beauty of the `build` macro and then including the source, beyond debuggability and code completion, is that we can pick up compiler warnings that are otherwise silent when only a `TokenStream` is used (as with the `import` macro). In this case, the `unsafe` was redundant.

examples/nuget-build/Cargo.toml
```diff
@@ -14,4 +14,4 @@ winrt = { path = "../.." }
 "Microsoft.AI.MachineLearning" = "1.3.0"
 "Microsoft.Windows.SDK.Contracts" = "10.0.19041.1"
 
-[workspace]
\ No newline at end of file
+[workspace]
```
kennykerr, line 17: Having the examples in their own workspace means that commands like `cargo test --all` and `cargo fmt --all` ignore the examples. This is undesirable, but adding them to the root workspace causes problems for `cargo winrt`.

examples/nuget-build/src/main.rs
```diff
@@ -0,0 +1,16 @@
+// TODO: OUT_DIR is not available from the proc_macro..
+// include!(concat!(env!("OUT_DIR"), "/winrt.rs"));
+include!("C:\\git\\rust\\winrt.rs");
```
kennykerr, line 3: Same here. We need to figure out how to use a cargo-provided build directory.

PR #208, opened 2020-06-06
Publish api docs to GitHub Pages

Michael-F-Bryan:
Usually when you are looking into a library the first thing to do is check out the API docs. The `winrt` crate isn't published on *crates.io* so you can't look at *docs.rs*, it'd be nice if we published API docs to GitHub Pages instead of making users clone and build the repo locally.

This adds a new GitHub Action that runs `cargo doc` and pushes the generated docs to GitHub Pages.

It's originally based on [the *Generate API Docs* job](https://github.com/Michael-F-Bryan/thumbnails/blob/2fede6d1b51f6a8ae686ece52dc51cdfbcae1f2d/.github/workflows/api-docs.yml) for my `thumbnails` project (which in turn came from [my Rust project template](https://github.com/Michael-F-Bryan/github-template/blob/20bd3930401879d8832b29a60ffc2618edcee74b/.github/workflows/main.yml#L37-L69)).

To wire this up, somewhere with owner rights will need to [create a Personal Access Token](https://github.com/settings/tokens/new ) with the `public_repo` scope then [create a `GH_PAGES_ACCESS_TOKEN` secret](https://github.com/microsoft/winrt-rs/settings/secrets) with the token.

### DIFF
--- a/.github/workflows/api-docs.yml
+++ b/.github/workflows/api-docs.yml
@@ -1,0 +1,33 @@
+name: Publish API Docs
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  api-docs:
+    name: Publish API Docs to GitHub Pages
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: doc
+          args: --all --verbose
+        env:
+          RUSTDOCFLAGS: "--cfg=docsrs"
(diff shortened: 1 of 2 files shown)


tim-weis:
Rust/WinRT was published to [crates.io](https://crates.io/crates/winrt/0.7.0) two days ago. The documentation is [here](https://docs.rs/winrt/0.7.0/winrt/).

rylev:
As @tim-weis notes, the crate is published and the documentation is on docs.rs. Let us know if there's some other issue with the docs though.

Michael-F-Bryan:
Thanks @rylev. I started playing around with `winrt-rs` a couple weeks ago and at the time there was no way of reading docs other than cloning the repo and running `cargo doc --open`.

If the project has been published to crates.io then that works for me.

PR #216, opened 2020-06-09
Don't use raw identifiers unless necessary

rylev:
This makes looking at docs much easier since there's not a bunch of `r#` everywhere. I also cleaned up checking `this` pointers for null.


crates/gen/src/types/method.rs
```diff
@@ -230,29 +230,21 @@ impl Method {
...
-                    match self.ptr.get_abi() {
-                        None => panic!("The `this` pointer was null when calling method"),
-                        Some(this) => {
-                            unsafe {
-                                let mut __ok: #return_type = ::std::mem::zeroed();
-                                (this.vtable().#method_name)(this, #args #return_arg)
-                                    .and_then(|| __ok )
-                            }
-                        }
+                    let this = self.ptr.get_abi().expect("The `this` pointer was null when calling method");
```
Josh015, line 233: Can't this be condensed down to "self.get_abi()" now?

crates/gen/src/types/method.rs
```diff
@@ -230,29 +230,21 @@ impl Method {
...
                 pub fn #method_name<#constraints>(&self, #params) -> ::winrt::Result<()> {
                     use ::winrt::AbiTransferable;
-                    match self.ptr.get_abi() {
-                        None => panic!("The `this` pointer was null when calling method"),
-                        Some(this) => {
-                            unsafe {
-                                (this.vtable().#method_name)(this, #args).ok()
-                            }
-                        }
+                    let this = self.ptr.get_abi().expect("The `this` pointer was null when calling method");
```
Josh015, line 245: Can't the "self.ptr.get_abi()" be condensed down to "self.get_abi()" now?

crates/gen/src/types/method.rs
rylev: Yes it can. I can push that real quick.

Josh015:
@rylev Don't forget to add ".expect()" to "activation_factory.rs".

PR #226, opened 2020-06-16
Improved tokenization and caching

kennykerr:
This is quite a big change to pay down some technical debt around how calling namespaces are calculated during tokenization. Rather than repeatedly generating tokens for type names, we calculate the calling namespace at the point at which the type name is created and then cache as appropriate. 

There is probably more opportunity to cache and clean up the code, but this overhaul is pretty substantial and I'd rather leave it here and address the remaining improvements separately.

Even though this work wasn't done to improve performance specifically, this does improve tokenization performance by about 20%. 

I also reintroduced a basic test for xaml as it ended up uncovering a few tokenization bugs that weren't covered by other tests. Once formal testing is introduced (https://github.com/microsoft/winrt-rs/issues/213), we'll be able to remove this again as it is quite slow (the Rust compiler is quite slow at parsing all of generated code).

### DIFF
--- a/crates/gen/src/types/interface.rs
+++ b/crates/gen/src/types/interface.rs
@@ -20,7 +20,7 @@ impl Interface {
 
         // Ensures that the default interface is first in line.
         let mut default_interface =
-            RequiredInterface::from_type_def(reader, name.def, &name.namespace); // TODO: RequiredInterface::from_type_name
+            RequiredInterface::from_type_def(reader, name.def, &name.namespace);
         default_interface.kind = InterfaceKind::Default;
         interfaces.push(default_interface);
 
--- a/crates/gen/src/types/interface.rs
+++ b/crates/gen/src/types/interface.rs
@@ -24,7 +24,7 @@ impl Interface {
         default_interface.kind = InterfaceKind::Default;
         interfaces.push(default_interface);
 
-        RequiredInterface::append_required(reader, &name, &mut interfaces);
+        RequiredInterface::append_required(reader, &name, &name.namespace, &mut interfaces);
         let signature = name.base_interface_signature(reader);
 
         Self {
--- a/crates/gen/src/types/interface.rs
+++ b/crates/gen/src/types/interface.rs
@@ -19,7 +19,7 @@ impl Interface {
         let mut interfaces = Vec::new();
 
         // Ensures that the default interface is first in line.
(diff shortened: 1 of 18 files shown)


crates/gen/src/types/interface.rs
```diff
@@ -135,101 +138,3 @@ impl Interface {
         }
     }
 }
-
-#[cfg(test)]
```
kennykerr, line 139: Removed internal tests as they required extra scaffolding that isn't needed otherwise and the external tests now more than cover this code.

rylev: Hmm whole I agree that this code will be exercised in integration tests, having unit tests makes debugging issues easier because it's usually quicker to spot the issue. I think these tests are pretty helpful.

crates/gen/src/types/type_name.rs
```diff
@@ -425,183 +431,21 @@ fn format_abi_ident(name: &str) -> proc_macro2::Ident {
     quote::format_ident!("abi_{}", name)
 }
 
-#[cfg(test)]
```
rylev, line 428: Same thought here. These tests have helped me in the past.

crates/gen/src/types/type_name.rs
```diff
@@ -354,12 +369,7 @@ impl TypeName {
     /// definition tokens and regular tokens. Definition tokens are those that require a
     /// colon spearater between the name and the generics (e.g., `Vector::<OtherType>`) vs.  
     /// `Vector<OtherType>`
-    fn generate_tokens<F>(
-        &self,
-        namespace: Option<&TokenStream>,
-        calling_namespace: &str,
-        format: F,
-    ) -> TokenStream
+    fn generate_tokens<F>(&self, namespace: Option<&TokenStream>, format: F) -> TokenStream
```
rylev, line 372: Why two generate_tokens functions?

kennykerr: No reason. Just ran out of steam refactoring this monster PR. Will clean this up soon.

crates/gen/src/types/type_name.rs
kennykerr: Fair enough. Restored all tests.

crates/gen/src/types/type_name.rs
kennykerr: Done.

PR #228, opened 2020-06-16
Dynamically load Windows 10 APIs.

simonbuchan:
This allows dependent crates to be able to run down-level.

The user experience is APIs return an `Error` where `code()` returns
`ErrorCode(HRESULT_FROM_WIN32(ERROR_MOD_NOT_FOUND))`
("The specified module could not be found.").

Tested on Windows 7 with the `Windows.System.UserProfile.GlobalizationPreferences.Languages` API.

Should probably also include a test in the `.gitlab/workflows/build.yml`, but I don't really
understand what would be needed in there. An exhaustive test of all APIs is probably overkill?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ readme = "README.md"
 [dependencies]
 winrt_macros = { path = "crates/macros",  version = "0.7" }
 sha1 = "0.6.0"
+wchar = "0.6.0"
 
 [dev-dependencies]
 doc-comment = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ readme = "README.md"
 [dependencies]
 winrt_macros = { path = "crates/macros",  version = "0.7" }
 sha1 = "0.6.0"
-wchar = "0.6.0"
 
 [dev-dependencies]
 doc-comment = "0.3"
--- a/src/error.rs
+++ b/src/error.rs
@@ -22,8 +22,10 @@ impl Error {
         let message: HString = message.into();
 
         // RoOriginateError creates the error object and associates it with the thread.
+        // Need to ignore the result, as that is the delay-load error, which would mean
+        // that there's no WinRT to tell about the error.
         unsafe {
-            RoOriginateError(code, message.get_abi() as _);
+            let _ = RoOriginateError(code, message.get_abi() as _);
         }
 
         let mut info = IErrorInfo::default();
@@ -137,8 +139,10 @@ impl<T> std::convert::From<Result<T>> for ErrorCode {
             if let Some(info) = error.info.get_abi() {
                 // Set the error information on the thread if the result is `Err`
                 // so that the caller can pick it up.
+                // Need to ignore the result, as that is the delay-load error, which would mean
+                // that there's no WinRT to tell about the error.
                 unsafe {
-                    SetRestrictedErrorInfo(info.as_raw() as _);
+                    let _ = SetRestrictedErrorInfo(info.as_raw() as _);
                 }
             }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -131,6 +131,19 @@ impl ErrorCode {
 
     /// Indicates that COM has not been initialized.
     pub(crate) const NOT_INITIALIZED: ErrorCode = ErrorCode(0x8004_01F0);
+
+    /// Creates a failure code from GetLastError()
+    #[inline]
+    pub(crate) fn last_win32_error() -> Self {
+        Self::from_win32(unsafe { GetLastError() })
+    }
+
+    /// Creates a failure code with the provided win32 error code.
+    #[inline]
+    pub(crate) fn from_win32(error: u32) -> Self {
+        // equivalent to MAKE_WIN32_HRESULT(error)
+        Self(0x8007_0000 | error & 0xFFFF)
+    }
 }
 
 impl<T> std::convert::From<Result<T>> for ErrorCode {
--- a/src/factory.rs
+++ b/src/factory.rs
@@ -9,19 +9,27 @@ pub fn factory<C: RuntimeName, I: ComInterface>() -> Result<I> {
 
     unsafe {
         // First attempt to get the activation factory via the OS.
-        let mut code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory)?;
+        let code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory);
+
+        // Treat any delay-load errors like standard errors, so that the heuristics
+        // below can still load registration-free libraries on Windows versions below 10.
+        let mut code = code.unwrap_or_else(|code| code);
 
         // If this fails because combase hasn't been loaded yet then load combase
         // automatically so that it "just works" for apartment-agnostic code.
         if code == ErrorCode::NOT_INITIALIZED {
             let mut _cookie = std::ptr::null_mut();
-            CoIncrementMTAUsage(&mut _cookie)?;
+
+            // Won't get any delay-load errors here if we got NOT_INITIALIZED, so quiet the
+            // warning from the #[must_use] on the returned Result<>.
+            let _ = CoIncrementMTAUsage(&mut _cookie);
 
             // Now try a second time to get the activation factory via the OS.
-            code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory)?;
+            code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory)
+                .unwrap_or_else(|code| code);
         }
 
-        // If this succeeded then retun the resulting factory interface.
+        // If this succeeded then return the resulting factory interface.
         if code.is_ok() {
             return Ok(std::mem::transmute_copy(&factory));
         }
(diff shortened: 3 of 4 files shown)


kennykerr:
Hi Simon, this looks great. I've been meaning to do the very same thing to support Windows 7 as C++/WinRT does today. Just going to leave a few comments.

Cargo.toml
```diff
@@ -12,6 +12,7 @@ readme = "README.md"
 [dependencies]
 winrt_macros = { path = "crates/macros",  version = "0.7" }
 sha1 = "0.6.0"
+wchar = "0.6.0"
```
kennykerr, line 15: I'd like to avoid dependencies as much as possible. If we can void the `wchar` dependency that would be great.

src/factory.rs
```diff
@@ -9,16 +9,16 @@ pub fn factory<C: RuntimeName, I: ComInterface>() -> Result<I> {
 
     unsafe {
         // First attempt to get the activation factory via the OS.
-        let mut code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory);
+        let mut code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory)?;
```
kennykerr, line 12: Short circuiting here won't allow reg-free activation of WinRT components on Windows 7. https://github.com/microsoft/winrt-rs/pull/177

src/factory.rs
```diff
@@ -9,16 +9,16 @@ pub fn factory<C: RuntimeName, I: ComInterface>() -> Result<I> {
...
         // If this fails because combase hasn't been loaded yet then load combase
         // automatically so that it "just works" for apartment-agnostic code.
         if code == ErrorCode::NOT_INITIALIZED {
             let mut _cookie = std::ptr::null_mut();
-            CoIncrementMTAUsage(&mut _cookie);
+            CoIncrementMTAUsage(&mut _cookie)?;
 
             // Now try a second time to get the activation factory via the OS.
-            code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory);
+            code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory)?;
```
kennykerr, line 21: Same here.

src/factory.rs
```diff
@@ -9,16 +9,16 @@ pub fn factory<C: RuntimeName, I: ComInterface>() -> Result<I> {
...
         // First attempt to get the activation factory via the OS.
-        let mut code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory);
+        let mut code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory)?;
 
         // If this fails because combase hasn't been loaded yet then load combase
         // automatically so that it "just works" for apartment-agnostic code.
         if code == ErrorCode::NOT_INITIALIZED {
             let mut _cookie = std::ptr::null_mut();
-            CoIncrementMTAUsage(&mut _cookie);
+            CoIncrementMTAUsage(&mut _cookie)?;
```
kennykerr, line 18: This shouldn't fail either since it would force a Windows 7 project to create a COM apartment beforehand.

Cargo.toml
simonbuchan: I generally want to prefer avoiding a dep for a small usage, but in this case the alternative do kind of suck too.
Which would you prefer: copying the `wch_c!` macro into your own macro crate, manually expanding the strings to u16 slices or dynamically converting the str literals to utf-16?

src/factory.rs
simonbuchan: Huh, didn't even consider that you could do that!
Looks like it would be fine to just squish the delay-load error into the regular one?

src/factory.rs
simonbuchan: Not completely sure I understand what needs to happen here and in on my phone ATM: this function isn't present on win7 at least, so to do that I would need to fall back to CoInitializeEx?

kennykerr: Oh, nevermind about this one. For reg-free on Windows 7 the previous call to RoGetActivationFactory would not fail with NOT_INITIALIZED but with the GetProcAddress error code, so this should be fine.

Cargo.toml
kennykerr: Whatever is simplest. I'd rather not copy the macro, but just inline the conversion since its only needed in one place.

src/factory.rs
kennykerr: Yes, if it fails to load or fails to activate we can treat it the same regardless.

src/runtime.rs
```diff
@@ -26,18 +28,55 @@ extern "system" {
     ) -> u32;
 }
 
-#[link(name = "windowsapp")]
-extern "system" {
-    pub fn CoIncrementMTAUsage(cookie: *mut RawPtr) -> ErrorCode;
+unsafe fn load_proc(library: &[u16], sym: &str) -> Result<RawPtr, ErrorCode> {
+    let library = LoadLibraryW(library.as_ptr());
+    if library.is_null() {
+        return Err(ErrorCode(0x8007_0000 | GetLastError()));
```
kennykerr, line 34: Maybe wrap this in a `from_win32` function. Note that we tend to mask the result of `GetLastError` with `0x0000FFFF` when forming an `HRESULT`.

Cargo.toml
simonbuchan: Just to be clear, the point of wchar is that the conversion is static, e.g. it emits `&[65u16, ...]`, so do you mean change it to be dynamically converting?

kennykerr: Yes, that would be ideal. If you can do that without too much effort that's great. Otherwise, since it is behind a `call_once` I don't think its the end of the world if the conversion is dynamic. None of the functions that we dynamically load are in hot code paths.

kennykerr:
> Should probably also include a test in the .gitlab/workflows/build.yml, but I don't really
understand what would be needed in there. An exhaustive test of all APIs is probably overkill?

We need to add a build that runs on Windows 7 to validate this PR. Note that this build won't be able to run `cargo test` until I have a dedicated test component since many of the existing tests assume Windows 10 - that's what I'm working on now. https://github.com/microsoft/winrt-rs/issues/213

We'll still need some tests to run that are specific to Windows 10, so we'll need a way to distinguish between `cargo test` for Windows 10 and `cargo test` for Windows 7.

simonbuchan:
Note that I also switched from using `LoadLibraryW` to `LoadLibraryExW` with `LOAD_LIBRARY_SEARCH_SYSTEM32` as [that's what chromium does](https://chromium.googlesource.com/chromium/src/+/53e40463423e4c016ec5c243d2e785e4840d58f4), it seems to be more likely to be what the user wants.

src/error.rs
```diff
@@ -22,8 +22,10 @@ impl Error {
         let message: HString = message.into();
 
         // RoOriginateError creates the error object and associates it with the thread.
+        // Need to ignore the result, as that is the delay-load error, which would mean
+        // that there's no WinRT to tell about the error.
```
kennykerr, line 26: Technically we need to create the `IRestrictedErrorInfo` that `RoOriginateError` would have created ourselves as a fallback, but Rust/WinRT doesn't yet support that so this is sufficient for now.

src/runtime.rs
```diff
@@ -26,18 +30,64 @@ extern "system" {
     ) -> u32;
 }
 
-#[link(name = "windowsapp")]
-extern "system" {
-    pub fn CoIncrementMTAUsage(cookie: *mut RawPtr) -> ErrorCode;
+unsafe fn load_proc(library_name: &str, sym: &str) -> Result<RawPtr, ErrorCode> {
```
rylev, line 33: Can we document why this is unsafe?

src/factory.rs
```diff
@@ -9,19 +9,27 @@ pub fn factory<C: RuntimeName, I: ComInterface>() -> Result<I> {
 
     unsafe {
         // First attempt to get the activation factory via the OS.
-        let mut code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory);
+        let code = RoGetActivationFactory(name.get_abi(), &I::iid(), &mut factory);
+
+        // Treat any delay-load errors like standard errors, so that the heuristics
+        // below can still load registration-free libraries on Windows versions below 10.
+        let mut code = code.unwrap_or_else(|code| code);
```
rylev, line 16: nit: it's a bit strange that the functions return `Result<ErrorCode, ErrorCode>` is there any reason to not just return ErrorCode?

kennykerr: This is as a result of the delay loading macro wrapping the result of the function. Ideally, it would fold the resulting error code if the result of the function happens to be an `ErrorCode` to begin with.

simonbuchan: Not all of the functions return ErrorCode, so the macro can't assume it.
This could use something like:

```rust
type DemandLoadResult<T> = Result<T, DemandLoadError>;
struct DemandLoadError(ErrorCode);
```

which I'd definitely do for a crate, but doesn't seem worth it for a couple of internal uses?

src/runtime.rs
simonbuchan: Hmm, it isn't actually unsafe according to rust rules, AFAICT, it's just unsafe contents being called by unsafe code and I was lazy. I can fix it.

src/factory.rs
kennykerr: Agreed. Perhaps a future enhancement.

src/runtime.rs
rylev: Thanks! FYI: Unsafe functions in the future will require unsafe parts to be marked as unsafe. This will be a warning soon, and a hard error in a future edition of Rust.

simonbuchan:
Thanks! Sorry for the late update, dumb work stuff 😉

kennykerr:
Simon, because the PR is created from a fork I'll need you to make sure the branch is up to date before I can merge this PR.

simonbuchan:
🤞

simonbuchan:
I don't have the Win 7 machine I was testing on available to confirm that the other branch didn't break anything downlevel, but it all looks clean.

Issue #213, opened 2020-06-08
Set up test component for comprehensive testing


kennykerr:
We should add support for testing this component:

https://github.com/microsoft/testwinrt

The trouble is that this component doesn't yet publish a nuget package.

PR #230, opened 2020-06-19
Complete array support

### LINKED ISSUE #194: Support for arrays
Most of the support for WinRT arrays is already in place, but a bit is missing and a lot of testing is still required to validate it all.

WinRT doesn't have an array type as such. Arrays only occur as parameters to WinRT APIs. That is why `winrt::Array<T>` does not implement the `RuntimeType` trait but is constrained to only allow arrays of `RuntimeType`.

There are three different array parameter patterns. 

1. An **input** array parameter where the caller provides a pointer to the array as well as the number of elements in the array that the callee may **read** from. Storage is provided by the caller (stack or heap).

2. An **output** array parameter where the caller provides a pointer to the array as well as the number of elements in the array that the callee may **write** to. Storage is provided by the caller (stack or heap).

3. An **output** array parameter where the callee returns a pointer allocated using `CoTaskMemAlloc` as well as the number of elements in the array and the caller assumes ownership of the storage and must ultimately call `CoTaskMemFree` to free the storage, but only after dropping the individual elements in the array. 

The `winrt::Array<T>` type needs to be completed to properly handle creating (allocating) and dropping arrays for (3) above. (1) and (2) should be handled directly by Rust slices (and is mostly already working I believe).

There are a few more subtleties to arrays particularly around the ABI handling, but that's the gist of it.

Comprehensive tests are needed for all three patterns.

kennykerr:
Fixes #194

Still needs more comprehensive testing #213, but this covers basic array support.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -1,12 +1,12 @@
 use crate::*;
 
 /// A WinRT array
-pub struct Array<T> {
+pub struct Array<T: RuntimeType> {
     data: *mut T,
     len: u32,
 }
 
-impl<T> Default for Array<T> {
+impl<T: RuntimeType> Default for Array<T> {
     fn default() -> Self {
         Array {
             data: std::ptr::null_mut(),
@@ -16,19 +16,31 @@ impl<T> Default for Array<T> {
 }
 
 impl<T: RuntimeType> Array<T> {
+    /// Creates an empty array.
     pub fn new() -> Self {
         Self::default()
     }
 
-    pub fn clear(&mut self) {
-        // TODO: drop members, CoTastkMemFree, zero members
+    /// Returns `true` if the array is empty.
+    pub fn is_empty(&self) -> bool {
+        self.len == 0
     }
 
-    pub fn as_slice(&self) -> &[T] {
-        if self.data.is_null() {
-            return &[];
+    /// Returns the length of the array.
+    pub fn len(&self) -> usize {
+        self.len as usize
+    }
+
+    /// Clears the contents of the array.
+    pub fn clear(&mut self) {
+        if self.is_empty() {
+            return;
         }
-        unsafe { std::slice::from_raw_parts(self.data, self.len as usize) }
+
+        unsafe {
+            std::ptr::drop_in_place(&mut self[..]);
+            CoTaskMemFree(self.data as _);
+        }
     }
 
     pub unsafe fn set_abi_len(&mut self) -> *mut u32 {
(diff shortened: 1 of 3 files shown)


src/array.rs
```diff
@@ -41,8 +53,30 @@ impl<T: RuntimeType> Array<T> {
...
 
-impl<T> Drop for Array<T> {
+impl<T: RuntimeType> std::ops::Deref for Array<T> {
+    type Target = [T];
+
+    fn deref(&self) -> &[T] {
+        if self.data.is_null() {
+            return &[];
+        }
+
```
rylev, line 63: Just for some sanity checking might be useful to have a `debug_assert!` here that `self.len > 0`.

src/array.rs
```diff
@@ -41,8 +53,30 @@ impl<T: RuntimeType> Array<T> {
...
+        unsafe { std::slice::from_raw_parts_mut(self.data, self.len as usize) }
+    }
+}
+
+impl<T: RuntimeType> Drop for Array<T> {
     fn drop(&mut self) {
-        // TODO: CoTaskMemFree
+        self.clear();
     }
 }
```
rylev, line 82: Can we add some unit tests for Array?

src/array.rs
```diff
@@ -16,19 +16,31 @@ impl<T> Default for Array<T> {
...
+
+    /// Clears the contents of the array.
+    pub fn clear(&mut self) {
+        if self.is_empty() {
+            return;
+        }
+
+        unsafe {
+            std::ptr::drop_in_place(&mut self[..]);
+            CoTaskMemFree(self.data as _);
```
rylev, line 42: We need to null the pointer out, otherwise the following is UB:

```rust
my_array.clear();
let slice: &[T] = &*my_array;
```

src/array.rs
```diff
@@ -16,19 +16,31 @@ impl<T> Default for Array<T> {
...
+    }
+
+    /// Clears the contents of the array.
+    pub fn clear(&mut self) {
+        if self.is_empty() {
+            return;
+        }
+
+        unsafe {
+            std::ptr::drop_in_place(&mut self[..]);
```
rylev, line 41: I think this is not fully correct. If we call `clear` and one of the values in the array panics during its drop, we'll unwind the stack and call `clear` again in the drop impl of Array leading us right back here to the where the array elements drop impls will be called again (double free). 

I think we have to first swap the pointers and then drop. If one of the elements panics, we may leak memory, but we won't double free.

src/array.rs
kennykerr: Switched to `is_empty` for consistency.

src/array.rs
kennykerr: I have some basic tests that cover the various issues here in the tests/boxing.rs - its difficult to cover all the edge cases without #213 - that's up next.

src/array.rs
kennykerr: Fixed.

src/array.rs
kennykerr: Fixed.

src/array.rs
```diff
@@ -16,19 +16,37 @@ impl<T> Default for Array<T> {
...
+    pub fn clear(&mut self) {
+        if self.is_empty() {
+            return;
+        }
+
+        let mut data = std::ptr::null_mut();
+        let mut len = 0;
+
+        std::mem::swap(&mut data, &mut self.data);
+        std::mem::swap(&mut len, &mut self.len);
```
rylev, line 44: This isn’t really necessary. `self.len = 0` is also fine

Issue #193, opened 2020-06-03
Send and Sync Support


rylev:
Currently all WinRT classes, interfaces, etc. are `!Send` and `!Sync`. Rust cannot automatically implement these marker traits are their underlying types are not `Send` and `Sync`. However, some WinRT types are indeed thread safe, namely those that implement `IAgileObject`. We should implement `Send` and `Sync` for these types.

zachlute:
I've been meaning to play around with this codebase and simultaneously learn about WinRT, and picked this issue at semi-random this weekend to see if I could do an implementation. The implementation itself actually seems pretty straightforward, but it's not quite clear how to test it. So, a couple of questions:

1. Is there a known type in the existing WinRT projection that implements `IAgileObject`? This is surprisingly difficult to determine, since the metadata is essentially opaque and all of the documentation is about user objects.
1. If not, is there any documentation about implementing custom types? I found [documentation for the C++/WinRT projection](https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/agile-objects), where deriving from `implements<MyType>` will automatically make something `IAgileObject` unless you opt out, but I couldn't find any examples of defining custom types in the current Rust codebase or in the Minesweeper example.

Additionally, while trying to learn about this, I came across the concept of tagging types with `MarshalingBehaviorAttribute(MarshalType.Agile)` in the metadata, which I believe is supposed to indicate that types like `Uri` can be safely marshaled between threads, but it's not quite clear to me whether that means those things should also be able to implement `Send` and `Sync` or if additional machinery is required.

kennykerr:
In response to your questions:

1. You can query any object using `IUnknown`'s `QueryInterface` for `IAgileObject` at run time to determine whether it is agile. The is the only foolproof way to make this determination. 

2. Rust/WinRT doesn't yet support implementing WinRT types. That support will arrive in a [future milestone](https://github.com/microsoft/winrt-rs/milestone/4). For now, you can use C++/WinRT to implement WinRT types that you can then call from Rust/WinRT.

You can certainly use the `MarshalingBehaviorAttribute` attribute to query the agility of a given WinRT type. Note that there may well be many objects that are agile that do not carry this attribute. For example, you might have a WinRT class that has the `MarshalingBehaviorAttribute` attribute and asserts its agility and this class has a method that returns an `IVector<u32>`. Unless documented otherwise, you'd still have to rely on `IAgileObject` to determine whether this returned value is agile.

[...]

PR #231, opened 2020-06-20
Make known-agile types generate Send/Sync implementations

zachlute:
If a type has a MarshalingBehaviorAttribute with MarshalingType.Aglile, we know it can be marshalled between threads.

This does not include support objects that implement IAgileObject, which we cannot know at compile time.

The change also includes a basic implementation of the ElementType enum in the interest of sanity, as well as a lot of TypeSig plumbing in service of getting underlying enum types.

Related to #193.


crates/gen/src/element_type.rs
```diff
@@ -0,0 +1,165 @@
+use crate::blob::Blob;
```
zachlute, line 1: It wasn't quite clear to me where the best place for the ElementType implementation was, so I settled on here, but please tell me if there's a better place!

crates/gen/src/element_type.rs
```diff
@@ -0,0 +1,165 @@
...
+            0x53 => ElementType::Field,
+            0x54 => ElementType::Property,
+            0x55 => ElementType::Enum,
+
+            unknown_type => panic!(format!("Unexpected ElementType: {:x}", unknown_type)),
+        }
+    }
+}
+
+impl std::fmt::Debug for ElementType {
```
zachlute, line 115: This implementation could be better. Enum needed this to have a Debug implementation so it could derive, but making this able to Derive took me down a rabbit hole of changing too many files so that's a future exercise.

crates/gen/src/tables/attribute.rs
```diff
@@ -33,42 +36,61 @@ impl Attribute {
...
 
-        values.read_u16();
-        sig.read_unsigned();
-        let count = sig.read_unsigned();
-        sig.read_unsigned();
+        let prolog = values.read_u16();
+        assert!(prolog == 0x0001, "CustomAttribute Prolog must be 0x0001"); // Required by spec.
 
-        let mut args: Vec<(String, AttributeArg)> = Vec::with_capacity(count as usize);
+        let _this_and_gen_param_count = sig.read_unsigned();
```
zachlute, line 42: I gave a lot of stuff in here names as I was attempting to understand what values were being read while going through the ECMA spec, and decided there was no reason to take them back out since it's more descriptive this way, but obviously let me know if you disagree.

crates/gen/src/element_type.rs
```diff
@@ -0,0 +1,165 @@
...
+    U2,
+    I4,
+    U4,
+    I8,
+    U8,
+    R4,
+    R8,
+    String,
+
+    Ptr(TypeSig),
```
kennykerr, line 22: Please remove everything that isn't specific to WinRT support. The ECMA-335 spec is enormous and the rabbit hole is long. 😉 

Let's just add what we need when we need it as we cannot possibly test all of these permutations and I don't want to have to support it.

crates/gen/src/types/enum.rs
```diff
@@ -33,14 +37,21 @@ impl Enum {
...
+
                 fields.push((name, value));
             }
         }
 
         Self {
-            name,
-            fields,
-            signature,
+            name: name,
```
kennykerr, line 51: `name: name` is redundant - why not just `name`?

tests/send_sync.rs
```diff
@@ -0,0 +1,22 @@
...
+use windows::foundation::Uri;
+
+// Simple test to validate that types with MarshalingType.Agile are marked Send and Sync
+// (if this compiles it worked)
+#[test]
+fn send_sync() -> winrt::Result<()> {
+    let url = Uri::create_uri("http://kennykerr.ca")?;
+
+    thread::spawn(move || {
+        println!("{:?}", url);
```
kennykerr, line 18: Tests should `assert` rather than `println`. Also, can the test wait to ensure the assert in the thread has executed?

crates/gen/src/types/enum.rs
zachlute: I honestly didn't know you could mix named and unnamed arguments like that, so I learned something today!

crates/gen/src/element_type.rs
zachlute: Good call. This also reduces the surface area of this change a lot, because none of the signature portion is necessary any longer, so I removed all of that until its necessary as well.

crates/gen/src/element_type.rs
kennykerr: `Debug` was important when I was bootstrapping the parser and had no tokenization. Now that it's up and running I don't know that we need to keep `Debug` implemented. I certainly don't want to manage a bunch of extra code if it can't be derived.

crates/gen/src/tables/attribute.rs
```diff
@@ -33,42 +36,62 @@ impl Attribute {
             AttributeType::MemberRef(method) => (reader.blob(method.0, 2), reader.blob(self.0, 2)),
         };
 
-        values.read_u16();
-        sig.read_unsigned();
-        let count = sig.read_unsigned();
-        sig.read_unsigned();
+        let prolog = values.read_u16();
+        assert!(prolog == 0x0001, "CustomAttribute Prolog must be 0x0001"); // Required by spec.
```
kennykerr, line 40: Let's `debug_assert` this for now. There's a bunch of validation work we may need to do, but I'd like to apply that consistently.

crates/gen/src/tables/attribute.rs
```diff
@@ -96,6 +119,22 @@ impl Attribute {
     }
 }
 
+fn read_enum(element_type: &ElementType, blob: &mut Blob) -> AttributeArg {
+    match element_type {
+        //ElementType::Boolean => AttributeArg::Bool(blob.read_bool()),
+        //ElementType::Char => AttributeArg::Char(blob.read_char()),
```
kennykerr, line 125: Remove commented out code.

crates/gen/src/element_type.rs
```diff
@@ -0,0 +1,73 @@
+use crate::blob::Blob;
+use crate::codes::{Decode, TypeDefOrRef};
+
+#[allow(dead_code)]
```
kennykerr, line 4: Still need this?

zachlute: Nope! Fixed.

crates/gen/src/types/class.rs
```diff
@@ -68,6 +69,12 @@ impl Class {
                         None => default_constructor = true,
                     }
                 }
+                ("Windows.Foundation.Metadata", "MarshalingBehaviorAttribute") => {
+                    let args = attribute.args(reader);
+                    if let AttributeArg::I16(marshaling_type) = args[0].1 {
```
rylev, line 74: nit: this can just be `if let AttributeArg::I16(2) = args[0].1`

crates/gen/src/types/enum.rs
```diff
@@ -41,6 +51,7 @@ impl Enum {
             name,
             fields,
             signature,
+            underlying_type: underlying_type.unwrap(),
```
rylev, line 54: Can we use expect instead of unwrap?

crates/gen/src/types/class.rs
```diff
@@ -121,6 +129,17 @@ impl Class {
             let async_get = async_get_tokens(&self.name, &self.interfaces);
             let debug = debug::debug_tokens(&self.name, &self.interfaces);
 
+            let send_sync = if self.is_agile {
+                let constraints = &self.name.constraints;
+                let name = &self.name.tokens;
+                quote! {
+                    unsafe impl<#constraints> ::std::marker::Send for #name {}
+                    unsafe impl<#constraints> ::std::marker::Sync for #name {}
```
rylev, line 137: I'm convinced that deriving Send is find, but I'm wondering how sure we can be about Sync. @kennykerr the semantics on `Sync` are not 100% fully documented but in short a type is `Sync` if `&T` can be safely sent between threads and there are never any unsynchronized reads and writes to those references.

zachlute: If we're unsure, it would probably still be useful to just implement `Send` while figuring out `Sync`, but I don't know enough about the semantics of the underlying types to have much input beyond that.

kennykerr: Yes, any type that is agile should be both `Send` and `Sync`. WinRT types should be implemented in a manner that is thread safe both in terms of their COM lifetime (thread-safe reference counting) and their internal state (thread-safe mutation). There are types that may violate this definition, but we should regard those as bugs.

crates/gen/src/element_type.rs
rylev: I personally don't see too much of a burden maintaining this particular debug implementation. It shouldn't change often (or ever?), and I too have used debug implementations for debugging and just for understanding what's happening.

Issue #233, opened 2020-06-23
Issues using WinUI with winrt-rs


joekgh:
I am attempting to use a WinUI control via Xaml Islands and winrt-rs.  I can get the windows::ui::xaml::controls to work however I ran into a number of issues with microsoft::ui::xaml::controls.

1. The dll from [Microsoft.UI.Xaml](https://www.nuget.org/packages/Microsoft.UI.Xaml/2.5.0-prerelease.200609001) does not get copied to `target/debug` automatically by `cargo winrt install`.  It is downloaded to `target/nuget/Microsoft.UI.Xaml/win10-x64/`.  Dlls from other NuGet packages do get copied as expected.  I manually copied it to move forward.
1. The Microsoft.UI.Xaml.dll depends on [Microsoft.VCRTForwarders.140](https://www.nuget.org/packages/Microsoft.VCRTForwarders.140/) dlls.  This results in the same error, `HRESULT 0x80040154` I received before I copied Microsoft.UI.Xaml.dll. I am not sure if `cargo winrt` can detect this situation and provide the user with additional feedback because it was very confusing at first.  I added the VCRTForwarders to `[package.metadata.winrt.dependencies]`.  The dlls were downloaded into `target/nuget/Microsfot.VCRTForwarders.140` but not were not copied to `target/debug`.  I manually copied them to move forward.
1. At this point my sample will compile and run, however, I get an `STATUS_ACCESS_VIOLATION` when I try to create a WinUI control, NumberBox in this case (see an example [here](https://github.com/joekgh/win_ui/blob/c2c3c4f3a7c06c6c86d831a321ef8c45642ee716/src/main.rs#L41)).

I could not create a concise example to share since all of the Xaml Islands setup needs to be done before I encountered the issue in step 3.  I have pushed a full repo to show the issue: https://github.com/joekgh/win_ui

Some other notes:

- The error message `cargo winrt` generates for a missing dll is the same as the error for a missing NuGet package

> Compiling bindings v0.1.0 (C:\Users\joe\workspace\win_ui\bindings)
error: could not read dependency: dependency path 'C:\Users\joe\workspace\win_ui\target\nuget\Microsoft.UI.Xaml' does not exist

- `cargo winrt` does not generate any output while it is running.  It is hard to know what happened (especially if something goes wrong).  Visual Studio NuGet Package Manager outputs quite a bit of status:

[...]

PR #234, opened 2020-06-24
Properly move over DLLs

rylev:
This PR fixes an issue reported in #233 where dlls were not being properly moved over from the `nuget` directory to the debug and release directories. The issue was that we were looking for targets in the form of `win-$ARCH` instead of `win10-$ARCH`. 

Additionally this improves logging by adding a `--verbose` flag which can help with simply debugging.

### DIFF
--- a/crates/cargo-winrt/Cargo.toml
+++ b/crates/cargo-winrt/Cargo.toml
@@ -19,3 +19,4 @@ zip = "0.5"
 console = "0.11"
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
+once_cell = { version = "1.4", default_features = false }
--- a/crates/cargo-winrt/Cargo.toml
+++ b/crates/cargo-winrt/Cargo.toml
@@ -18,4 +18,4 @@ anyhow = "1.0"
 zip = "0.5"
 console = "0.11"
 serde_json = "1.0"
-serde = { version = "1.0.111", features = ["derive"] }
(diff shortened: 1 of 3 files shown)


joekgh:
@rylev, I tested this branch and can confirm that it does sym link the Microsoft.UI.Xmal.dll into `target/debug`.  However for:
```
"Microsoft.VCRTForwarders.140" = "1.0.6"
```
I now get an empty `target/nuget/Microsoft.VCRTForwarders.140` directory.

> writing dll file "win10-x64\\release\\vcruntime140_app.dll" into C:\Users\joe\workspace\win_ui\target\nuget\Microsoft.VCRTForwarders.140
           not creating symlink for win10-x64\release\vcruntime140_app.dll because of differing architecture to host architecture

I'm not sure what the intended behavior is in this case, but the version in master would download VCRTForwarders and extract the files to `target/nuget/Microsoft.VCRTForwarders.140`

One other nit.  "Fetching" is repeated:

> cargo winrt install
        Fetching: fetching Microsoft.UI.Xaml
        Fetching: fetching Microsoft.VCRTForwarders.140

rylev:
@joekgh it seems that that nuget package differentiates between debug and release... I'll have to put in special logic for that.

rylev:
@joekgh can you give this branch another try? I think it should work now when dlls are inside release directories.

joekgh:
@rylev, it now works as expected.  Thanks.

PR #237, opened 2020-06-24
Initial support for test component

kennykerr:
This update adds support for the TestWinRT component:

https://github.com/microsoft/testwinrt

This WinRT component is written in C++/WinRT and provides more comprehensive test coverage than is possible by simply exercising the Windows API. 

This initial commit mainly provides tweaks to code generation to actually support parsing the test component's winmd file. Exhaustive test coverage of the test component's many test methods is coming up next!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: cargo winrt
+      run: cargo install --git https://github.com/microsoft/winrt-rs cargo-winrt
+
+    - name: winrt install
+      run: cargo winrt install
+
     - name: check
       run: cargo check --all --bins --examples
 
@@ -31,6 +37,12 @@ jobs:
 
     - name: install
       run: rustup install 1.40.0
+
+    - name: cargo winrt
+      run: cargo install --git https://github.com/microsoft/winrt-rs cargo-winrt
+
+    - name: winrt install
+      run: cargo winrt install
 
     - name: check
       run: cargo +1.40.0 check --all --bins --examples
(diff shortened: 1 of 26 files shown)


crates/gen/src/types/delegate.rs
```diff
@@ -58,6 +58,32 @@ impl Delegate {
             .map(|param| param.to_invoke_arg_tokens());
         let debug = debug::default_debug_tokens(&self.name);
 
+        let invoke_upcall = if let Some(return_type) = &self.method.return_type {
```
kennykerr, line 61: This is just roughed in to allow the resulting Rust code to compile - it doesn't actually work yet. That's up next and I'm sure will change.

crates/gen/src/types/method.rs
```diff
@@ -67,7 +67,7 @@ impl Method {
         let return_type = if blob.read_expected(0x01) {
             None
         } else {
-            let name = "__result".to_owned();
+            let name = "result__".to_owned();
```
kennykerr, line 70: Switched from leading to trailing underscores for internal generated names to allow unused variables to quickly be detected.

.github/workflows/build.yml
```diff
@@ -15,6 +15,12 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: cargo winrt
+      run: cargo install --git https://github.com/microsoft/winrt-rs cargo-winrt
```
rylev, line 19: Let's not forget to change this with we release a new version

src/abi_transferable.rs
```diff
@@ -23,6 +23,14 @@ pub unsafe trait AbiTransferable {
         // correctly implement this trait
         unsafe { std::mem::transmute_copy(&abi) }
     }
+
+    fn slice_from_abi<'a>(abi: *const Self::Abi, len: usize) -> &'a [Self] {
```
rylev, line 27: I think this function should be unsafe since it's up to the caller to ensure the pointer is valid for `'a`

kennykerr: Agreed.

PR #244, opened 2020-07-01
Add array/delegate/collection tests

kennykerr:
Delegates still need more work but this provides enough test coverage so I can do some broader refactoring to support implementations more cleanly.


src/array.rs
```diff
@@ -21,6 +21,21 @@ impl<T: RuntimeType> Array<T> {
         Self::default()
     }
 
+    pub fn with_len(len: usize) -> Self {
+        unsafe {
+            assert!(len < u32::MAX as usize);
```
kennykerr, line 26: Hmm, looks like `::MAX` is not available on Rust 1.40.0

src/array.rs
```diff
@@ -21,6 +21,21 @@ impl<T: RuntimeType> Array<T> {
         Self::default()
     }
 
+    pub fn with_len(len: usize) -> Self {
+        unsafe {
+            assert!(len < 0xFFFF_FFFF as usize);
```
rylev, line 26: You can use the constant instead: https://doc.rust-lang.org/stable/std/u32/constant.MAX.html this has been there since 1.0

src/array.rs
```diff
@@ -21,6 +21,21 @@ impl<T: RuntimeType> Array<T> {
         Self::default()
     }
 
+    pub fn with_len(len: usize) -> Self {
+        unsafe {
```
rylev, line 25: Can we move the unsafe block to only be around lines 27-33? The other lines are safe

src/array.rs
```diff
@@ -21,6 +21,21 @@ impl<T: RuntimeType> Array<T> {
...
+            assert!(len < 0xFFFF_FFFF as usize);
+            let data = CoTaskMemAlloc(len * std::mem::size_of::<T>()) as *mut T;
+
+            if data.is_null() {
+                panic!("Could not successfully allocate for Array");
+            }
+
+            std::ptr::write_bytes(data, 0, len);
```
rylev, line 33: In the spirt of better documentation around safety, can we document this is safe because `RuntimeTypes` are safe to zero-initialize?

src/array.rs
```diff
@@ -57,6 +72,12 @@ impl<T: RuntimeType> Array<T> {
         self.clear();
         &mut self.data as *mut _ as *mut _
     }
+
+    pub unsafe fn into_abi(self) -> (*mut T::Abi, u32) {
```
rylev, line 76: The std lib usually marks functions like this as safe since it's only unsafe to _use_ the raw pointer not to produce it. I'm fine with keeping a more conservative approach to this, but then we need to document its safety properties.

src/array.rs
```diff
@@ -21,6 +21,21 @@ impl<T: RuntimeType> Array<T> {
         Self::default()
     }
 
+    pub fn with_len(len: usize) -> Self {
```
rylev, line 24: docs please :-)

src/abi_transferable.rs
```diff
@@ -42,6 +42,12 @@ pub unsafe trait AbiTransferable: Sized {
     unsafe fn slice_from_mut_abi<'a>(abi: *mut Self::Abi, len: usize) -> &'a mut [Self] {
         std::slice::from_raw_parts_mut(std::mem::transmute_copy(&abi), len)
     }
+
+    fn into_abi(self) -> Self::Abi {
+        let abi = unsafe { std::mem::transmute_copy(&self) };
```
rylev, line 47: Can we document why this is safe?

src/abi_transferable.rs
```diff
@@ -42,6 +42,12 @@ pub unsafe trait AbiTransferable: Sized {
     unsafe fn slice_from_mut_abi<'a>(abi: *mut Self::Abi, len: usize) -> &'a mut [Self] {
         std::slice::from_raw_parts_mut(std::mem::transmute_copy(&abi), len)
     }
+
+    fn into_abi(self) -> Self::Abi {
```
rylev, line 46: Docs please :-)

crates/gen/src/types/delegate.rs
```diff
@@ -60,21 +61,31 @@ impl Delegate {
...
+
                 quote! {
                     match ((*this).invoke)(#(#invoke_args,)*) {
-                        ::std::result::Result::Ok(_) => {
+                        ::std::result::Result::Ok(ok__) => {
+                            *#return_name = <#return_kind as ::winrt::AbiTransferable>::into_abi(ok__);
                             ::winrt::ErrorCode(0)
                         }
-                        ::std::result::Result::Err(result__) => result__.into()
+                        ::std::result::Result::Err(err__) => err__.into()
```
rylev, line 88: I'm not convinced these local variables need the funny `__` on their names

kennykerr:
Closing this PR for now - needs more work to deal with some ABI edge cases.

src/abi_transferable.rs
rylev: Nit: We should mark safety documenation in an easily discoverable way: 
```rust
// SAFETY: 
```

PR #251, opened 2020-07-07
Support for async await

### LINKED ISSUE #80: Support for async
Need to integrate Rust's async-await support with WinRT's async support via its IAsyncXxx interfaces.

kennykerr:
You now have the option of using either the blocking `get()` function or the `await` operator within an `async` function to wait for an async API call to complete. For example, here's the RSS feed example I first used to bootstrap coroutine support in C++. 

Here's how you would write it in Rust with the `get()` function to cause a blocking wait for the results:

```rust
use winrt::*;
use windows::foundation::*;
use windows::web::syndication::*;

fn main() -> Result<()> {
    let uri = Uri::create_uri("https://kennykerr.ca/feed")?;
    let client = SyndicationClient::new()?;
    let feed = client.retrieve_feed_async(uri)?.get()?;

    for item in feed.items()? {
        println!("{}", item.title()?.text()?);
    }

    Ok(())
}
```

And here's how you can now write it using an `async` function and the `await` operator:

```rust
async fn main_async() -> Result<()> {
    let uri = Uri::create_uri("https://kennykerr.ca/feed")?;
    let client = SyndicationClient::new()?;
    let feed = client.retrieve_feed_async(uri)?.await?;

    for item in feed.items()? {
        println!("{}", item.title()?.text()?);
    }

    Ok(())
}

fn main() -> Result<()> {
    futures::executor::block_on(main_async())
}
```

Fixes #80


crates/gen/src/types/futures.rs
```diff
@@ -0,0 +1,109 @@
...
+        quote! {
+            pub fn get(&self) -> ::winrt::Result<#return_type> {
+                if self.status()? == #namespace AsyncStatus::Started {
+                    unsafe {
+                        let event = ::winrt::CreateEventW(::std::ptr::null_mut(), 1, 0, ::std::ptr::null_mut());
+                        self.set_completed(#namespace #handler::new(move |_sender, _args| {
+                            ::winrt::SetEvent(event);
+                            Ok(())
+                        }))?;
+                        ::winrt::WaitForSingleObject(event, 0xFFFFFFFF);
```
rylev, line 81: Do we also want to offer a `get_with_timeout` method?

crates/gen/src/types/futures.rs
```diff
@@ -0,0 +1,109 @@
...
+                }
+                self.get_results()
+            }
+        },
+        quote! {
+            impl<#constraints> ::std::future::Future for #name {
+                type Output = ::winrt::Result<#return_type>;
+
+                fn poll(self: ::std::pin::Pin<&mut Self>, context: &mut ::std::task::Context) -> ::std::task::Poll<Self::Output> {
+                    if self.status().expect("Failed to get async status") == #namespace AsyncStatus::Started {
```
rylev, line 93: This reminds me of a point I've been wanting to make. Because `winrt::Result` is linked to HRESULTs for errors, we usually result to panicking when an error occurs that doesn't map to an HRESULT. In this particular case, I think panicking is perhaps appropriate but we might want to think about if it's the right thing for `winrt::Result::Err` to always be an HRESULT.

crates/gen/src/types/futures.rs
```diff
@@ -0,0 +1,109 @@
...
+                        let waker = context.waker().clone();
+
+                        self.set_completed(#namespace #handler::new(move |_sender, _args| {
+                            waker.wake_by_ref();
+                            Ok(())
+                        })).expect("Failed to set async completed handler");
```
rylev, line 99: We use `?` on the error here above. We should probably be consistent.

crates/gen/src/types/futures.rs
```diff
@@ -0,0 +1,109 @@
...
+        quote! {
+            impl<#constraints> ::std::future::Future for #name {
+                type Output = ::winrt::Result<#return_type>;
+
+                fn poll(self: ::std::pin::Pin<&mut Self>, context: &mut ::std::task::Context) -> ::std::task::Poll<Self::Output> {
+                    if self.status().expect("Failed to get async status") == #namespace AsyncStatus::Started {
+                        let waker = context.waker().clone();
+
+                        self.set_completed(#namespace #handler::new(move |_sender, _args| {
+                            waker.wake_by_ref();
```
rylev, line 97: Is the status guaranteed to be switched away from `Started` when this callback is called?

rylev: Also, what happens if the user sets their own completed handler?

Cargo.toml
```diff
@@ -15,6 +15,7 @@ sha1 = "0.6.0"
 
 [dev-dependencies]
 doc-comment = "0.3"
+futures = "0.3"
```
Josh015, line 18: I might be overthinking this, but is there any value in making "async/await" a feature so the ```futures``` crate can be excluded if the user doesn't need it?

crates/gen/src/types/futures.rs
kennykerr: I just realized that the `Poll<T>` supports propagation - that's much cleaner.

crates/gen/src/types/futures.rs
kennykerr: Same as above.

crates/gen/src/types/futures.rs
kennykerr: Only one completed handler is allowed by the async state machine. A second handler will return an error.

PR #253, opened 2020-07-07
Add TimeSpan <> std::time::Duration interop

### LINKED ISSUE #90: Support for Windows.Foundation's TimeSpan
Similar to how C++/WinRT uses std::chrono and C#/WinRT uses .NET's TimeSpan, it would be nice if developers could use std::time::Duration to specify a WinRT TimeSpan.

kennykerr:
This update elevates `Windows.Foundation.TimeSpan` to a well-known type within the `winrt` crate rather than simply being generated in order to enable improved usability and interop with Rust's `std::time::Duration`. 

You can now pass a `Duration` to any input parameter expecting a `TimeSpan` and can use `into()` to convert between the types.

Fixes #90

### DIFF
--- a/src/time_span.rs
+++ b/src/time_span.rs
@@ -1,0 +1,49 @@
+use crate::*;
+
+/// Represents a time interval as a signed 64-bit integer value.
+///
+/// TimeSpan represents the WinRT [TimeSpan](https://docs.microsoft.com/en-us/uwp/api/Windows.Foundation.TimeSpan)
+/// struct and provides convertibility with `std::time::Duration`.
+#[repr(C)]
+#[derive(Clone, Default, Debug, PartialEq)]
+pub struct TimeSpan {
+    pub duration: i64,
+}
+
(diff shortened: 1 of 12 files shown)


src/time_span.rs
```diff
@@ -0,0 +1,49 @@
+use crate::*;
+
+/// Represents a time interval as a signed 64-bit integer value.
+///
+/// TimeSpan represents the WinRT [TimeSpan](https://docs.microsoft.com/en-us/uwp/api/Windows.Foundation.TimeSpan)
+/// struct and provides convertibility with `std::time::Duration`.
+#[repr(C)]
+#[derive(Clone, Default, Debug, PartialEq)]
```
rylev, line 8: Any reason for this not to be copy?

kennykerr: Good catch - I just copied the default we generate for WinRT structs. But for TimeSpan Copy makes sense.

Issue #247, opened 2020-07-02
cargo winrt consistency


kennykerr:
A few minor polish issues.

* Notice the indentation difference between `cargo install` and `cargo winrt install`.
* The latter injects a `:` which we should probably just remove now that we have colorization.
* Add a `Finished in 1m 03s` at the end both to tell you how long it took and as a quick visual that it completed and didn't silently fail.
* Add a `cargo winrt clean` option to remove nuget packages and ensure `cargo winrt install` fetches them again. Right now I have to run `cargo clean` just to make sure `cargo winrt install` actually updates the nuget packages but that blows away the cargo cache as well.

![image](https://user-images.githubusercontent.com/9845234/86400642-d7fec580-bc5d-11ea-9626-558c5b1da6d0.png)

kennykerr:
@rylev pointed out there's a "force" option - we should document that.

https://github.com/microsoft/winrt-rs/blob/8a5d2a9a56327a3ac77ee3594931217da23a5951/crates/cargo-winrt/src/main.rs#L70

zachlute:
With #258 merged, I believe everything mentioned in here is handled and this can be closed.

kennykerr:
Thanks Zach!

PR #254, opened 2020-07-08
Update cargo-winrt output to better match cargo's style

zachlute:
Add print_status and print_verbose_status methods with consistent justified formatting.

As specified in #247 .

cargo_winrt_output

### DIFF
--- a/crates/cargo-winrt/src/main.rs
+++ b/crates/cargo-winrt/src/main.rs
@@ -181,20 +181,27 @@ fn elapsed(duration: Duration) -> String {
     }
 }
 
-fn print_status<D>(status: &str, message: D)
-where
-    D: std::fmt::Display,
-{
-    println!("{:>12} {}", console::style(status).green().bold(), message);
+macro_rules! print_status {
+    ($status:expr, $message:expr) => ({
+        println!("{:>12} {}", console::style($status).green().bold(), $message);
+    });
+    ($status:expr, $message_fmt:expr, $($message_args:tt)*) => ({
+        println!("{:>12} {}", console::style($status).green().bold(), format!($message_fmt, $($message_args)*));
+    });
 }
 
-fn print_verbose_status<D>(status: &str, message: D)
-where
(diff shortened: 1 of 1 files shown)


crates/cargo-winrt/src/main.rs
```diff
@@ -215,10 +237,10 @@ impl Install {
...
                 .collect::<anyhow::Result<Vec<DependencyDescriptor>>>()?
         };
-        debug!(
-            "{} {} nuget dependencies",
-            console::style("Installing").green().bold(),
-            dependency_descriptors.len()
+
+        print_verbose_status(
+            "Installing",
+            format_args!("{} nuget dependencies", dependency_descriptors.len()),
```
zachlute, line 243: The `format_args!` are *probably* overkill, but the previous `debug!` macro for verbose output wouldn't evaluate the formatted string unless `-v` was actually specified, so I wanted to at least prove that this change could maintain that behavior if desired.

crates/cargo-winrt/src/main.rs
```diff
@@ -169,32 +170,53 @@ fn verbose() -> bool {
...
+    let secs = duration.as_secs();
+
+    if secs >= 60 {
+        format!("{}m {:02}s", secs / 60, secs % 60)
+    } else {
+        format!("{}.{:02}s", secs, duration.subsec_nanos() / 10_000_000)
+    }
+}
+
+fn print_status<D>(status: &str, message: D)
```
rylev, line 184: I think I would prefer these as macros so you're not required to do the funny `format_args` or `format` dance. Is there a reason to not use macros for this?

zachlute: I was mostly borrowing from the way Cargo does this output, while maintaining the semantics already used here, so it wasn't explicitly a decision to 'not use macros'. Happy to give it a shot with macros if you think that'd be more useful.

kennykerr:
Thanks @zachlute - works great!

PR #258, opened 2020-07-14
Improve help output of cargo-winrt

zachlute:
Updates help messaging to use cargo-standard syntax and output, including separate help documentation for each subcommand.

Added documentation for the previously undocumented -f and -v options, as mentioned in #247 .

While testing this, I also realized that the previous "Finished" message didn't say what finished, which makes it compose poorly with `cargo winrt build` for example, since you'll get two finished messages, so I updated that as well.

### DIFF
--- a/crates/cargo-winrt/src/main.rs
+++ b/crates/cargo-winrt/src/main.rs
@@ -270,7 +270,11 @@ impl Install {
         self.install_from_manifest(manifest)?;
 
         let time_elapsed = elapsed(start_time.elapsed());
-        print_status!("Finished", "in {}", time_elapsed);
+        print_status!(
+            "Finished",
+            "installing WinRT dependencies in {}",
+            time_elapsed
+        );
 
         Ok(())
     }
--- a/crates/cargo-winrt/src/main.rs
+++ b/crates/cargo-winrt/src/main.rs
@@ -45,7 +45,7 @@ fn run() -> Result<(), i32> {
         Subcommand::Install(i) => i.perform(),
         Subcommand::Run(r) => r.perform(),
         Subcommand::Build(b) => b.perform(),
-        Subcommand::Help => print_help(),
+        Subcommand::Help(h) => h.perform(),
     };
     if let Err(ref e) = result {
         cmd_err!("{}", e);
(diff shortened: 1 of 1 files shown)


crates/cargo-winrt/src/main.rs
```diff
@@ -67,14 +67,32 @@ fn parse_args() -> Result<Subcommand, ArgsError> {
...
 
-    let verbose = args.contains(["-v", "--verbose"]);
-    let force = args.contains(["-f", "--force"]);
-    args.finish()?;
-
     let subcommand = match subcommand.as_deref() {
-        Some("install") => Subcommand::Install(Install { verbose, force }),
-        Some("run") => Subcommand::Run(Run { verbose, force }),
-        Some("build") => Subcommand::Build(Build { verbose, force }),
+        Some("install") => {
```
kennykerr, line 71: This seems very repetitive. Is it worth cleaning up?

crates/cargo-winrt/src/main.rs
```diff
@@ -137,6 +158,22 @@ impl Build {
         install.perform()?;
         cargo::build()
     }
+
+    fn print_help() -> anyhow::Result<()> {
```
kennykerr, line 162: Same here.

crates/cargo-winrt/src/main.rs
zachlute: Do you mean just the duplicated argument parsing? My thought was that although they're currently all the same, they wouldn't  necessarily *always* be, so splitting it up by command seemed sensible. Originally they were always parsed, but since, for example, the `help` subcommand doesn't use those arguments, it needed to be rethought in some way.

It's also possible that, for example, we'd want to get crazier with the `run` and `build` subcommands that forward to Cargo and let you pass additional arguments that get forwarded on.

So that's kind of a lot of words to say, this could probably be cleaned up (for example, we could move parsing into constructors for each subcommand so at least it wasn't just repetitive here), but it wasn't clear to me what the most straightforward path for potential future evolution was here, and this seemed at least simplest.

crates/cargo-winrt/src/main.rs
zachlute: Well, if we were using `clap`, obviously this would all just be magically done for us. We could do the same kind of attribute/description magic to generate these messages, or at least separate it out into a function with a template string we could call with the list/descriptions and just have it do the output.

As above, I erred on the side of simplicity for this initial change, but a case could certainly be made that now it's TOO simple. :)

kennykerr: I like simple. 😉 And I'd rather not bring in additional dependencies so this is fine. We can always revisit if it gets more complicated.

PR #261, opened 2020-07-15
Fix named arguments in attribute parsing

robmikh:
Currently it seems that named arguments in the attribute parsing isn't working properly. I don't think it comes up in any of the Windows APIs, but I discovered it while messing with custom attributes.

It appears that the first byte is the identification, which must be either 0x53 or 0x54 according to the spec:

```
9. The following rules apply to the structure of NamedArg (§II.23.3):
    *  A NamedArg shall begin with the single byte FIELD (0x53) or PROPERTY (0x54) for identification [ERROR]
```
Page 217 (243 in the [pdf](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-335.pdf))

The next byte is the argument type, followed by the name.

Unfortunately I'm not sure what's the best way to test this as-is. I tested it in a different branch I'm working on that consumes winrt_gen with some custom idls/winmds.

### DIFF
--- a/tests/custom_attribute.rs
+++ b/tests/custom_attribute.rs
@@ -1,0 +1,86 @@
+// This is used to make sure our dependencies are picked up by cargo winrt
+import!(
+    dependencies
+        nuget: Microsoft.Windows.SDK.Contracts
+        nuget: KennyKerr.Windows.TestWinRT
+    types
+        test_component::*
+        windows::foundation::*
+);
+
+use std::path::{Path, PathBuf};
+use winrt::*;
+use winrt_gen::{load_winmd, AttributeArg, TypeReader};
+
+fn find_winmds<P: AsRef<Path>>(directory: P) -> Vec<PathBuf> {
+    let mut result = Vec::new();
+    for entry in std::fs::read_dir(directory).unwrap() {
+        let entry = entry.unwrap();
+        let path = entry.path();
+        if path.is_file() {
+            if let Some(ext) = path.extension() {
+                if ext == "winmd" {
+                    result.push(path.clone());
+                }
+            }
+        }
+    }
+    result
+}
+
+#[test]
+fn named_arguments() -> Result<()> {
+    // Get our WinMD files from the target directory
+    let files = {
+        let mut files = Vec::new();
+        let paths = [
+            "target/nuget/Microsoft.Windows.SDK.Contracts",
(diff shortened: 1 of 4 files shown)


kennykerr:
Yeah that's a problem with testing this capability. You could add metadata with a custom attribute to this IDL:

https://github.com/microsoft/TestWinRT/blob/master/TestComponent/TestComponent.idl

This is used by winrt-rs for testing. I'd love to get some tests for named arguments as that's been a buggy bit of code.

robmikh:
Ah, that's a good idea! I'll put together a PR to add a custom attribute with named arguments to TestWinRT and then update this PR with a test that consumes/tests everything.

robmikh:
I made the PR: https://github.com/microsoft/TestWinRT/pull/32

I was able to successfully run the test in this change with the new metadata. After pulling down the nuget package, I replaced the local metadata with my custom version.

tests/custom_attribute.rs
```diff
@@ -0,0 +1,86 @@
...
+                        ),
+                    }
+                }
+            }
+            _ => {}
+        }
+    }
+    assert_eq!(some_string, 1);
+    assert_eq!(some_int, 1);
+    assert_eq!(some_bool, 1);
```
robmikh, line 83: If there's a better way of making sure each attribute was seen only once, please let me know!

tests/custom_attribute.rs
```diff
@@ -0,0 +1,86 @@
+// This is used to make sure our dependencies are picked up by cargo winrt
+import!(
+    dependencies
+        nuget: Microsoft.Windows.SDK.Contracts
+        nuget: KennyKerr.Windows.TestWinRT
+    types
+        test_component::*
+        windows::foundation::*
+);
```
robmikh, line 9: I just noticed the `package.metadata.winrt.dependencies` section in the root `Cargo.toml` file. I'm guessing that is what is informing cargo winrt of the dependencies? If so, I'll delete this block.

kennykerr:
Awesome, now once 1.0.17 is published you can just update the toml on this end to pick up the latest version.

tests/custom_attribute.rs
robmikh: Yup, looks like this isn't needed.

robmikh:
Thanks for getting the package published, I was able to pull down the new version without modifying it! Everything seems to pass.

I also removed some code I didn't need.

Issue #260, opened 2020-07-15
cargo winrt install is working pretty slow in low bandwidth environment


uhuntu:
I tried "cargo winrt install" to install, but in my environment, it always stuck in
```
C:\Users\hunt\source\repos\winrt-rs>cargo winrt install
    Fetching KennyKerr.Windows.TestWinRT
    Fetching Microsoft.AI.MachineLearning
```

I tried adding verbose to see more information
```
C:\Users\hunt\source\repos\winrt-rs>cargo winrt install --verbose
   Resolving winrt_macros
  Installing 0 nuget dependencies
   Resolving winrt
  Installing 3 nuget dependencies
    Fetching KennyKerr.Windows.TestWinRT
  Requesting https://www.nuget.org/api/v2/package/KennyKerr.Windows.TestWinRT/1.0.16
  Requesting https://globalcdn.nuget.org/packages/kennykerr.windows.testwinrt.1.0.16.nupkg
   Retrieved data from https://globalcdn.nuget.org/packages/kennykerr.windows.testwinrt.1.0.16.nupkg
    Starting extraction of 'KennyKerr.Windows.TestWinRT'
   Searching zip file: _rels\.rels
   Searching zip file: KennyKerr.Windows.TestWinRT.nuspec
   Searching zip file: lib\uap10.0\TestComponent.winmd
       Found winmd file: "TestComponent.winmd"
   Searching zip file: runtimes\win10-x86\native\TestComponent.dll
       Found dll "TestComponent.dll" with arch "win10-x86" at path runtimes\win10-x86\native\TestComponent.dll
   Searching zip file: runtimes\win10-x64\native\TestComponent.dll
       Found dll "TestComponent.dll" with arch "win10-x64" at path runtimes\win10-x64\native\TestComponent.dll
   Searching zip file: [Content_Types].xml
   Searching zip file: package\services\metadata\core-properties\d677ecd8bcef4b7c825af36506771d04.psmdcp
   Searching zip file: .signature.p7s
    Fetching Microsoft.AI.MachineLearning
  Requesting https://www.nuget.org/api/v2/package/Microsoft.AI.MachineLearning/1.3.0
  Requesting https://globalcdn.nuget.org/packages/microsoft.ai.machinelearning.1.3.0.nupkg
```

I realized that machinelearning download always stuck at header request and almost never end.

I guess in good network environment, you might not ever observe it.

[...]

PR #262, opened 2020-07-16
Re-implement try_download to prevent twice download.

uhuntu:
As my mentioned in #260, there was a bug in cargo winrt install procedure, without setting nobody when doing a curl header request, it will cost much time that not just request header. The patch is for fixing it.

### DIFF
--- a/crates/cargo-winrt/src/main.rs
+++ b/crates/cargo-winrt/src/main.rs
@@ -5,7 +5,7 @@ mod manifest;
 use error::Error;
 use manifest::Manifest;
 
-use anyhow::{bail, Context};
+use anyhow::Context;
 use curl::easy::Easy;
 
 use std::ffi::OsString;
@@ -380,14 +380,14 @@ impl DependencyDescriptor {
         match self {
             DependencyDescriptor::NugetOrg { name, version } => {
                 let url = format!("https://www.nuget.org/api/v2/package/{}/{}", name, version);
-                let bytes = try_download(url, 5)?;
+                let bytes = try_download(url)?;
                 Ok(RawNuget::Zipped {
                     bytes,
                     name: name.clone(),
                 })
             }
             DependencyDescriptor::Url { url, name } => {
-                let bytes = try_download(url.as_str().to_owned(), 5)?;
+                let bytes = try_download(url.as_str().to_owned())?;
 
                 Ok(RawNuget::Zipped {
                     bytes,
@@ -438,70 +438,41 @@ impl DependencyDescriptor {
     }
 }
 
-fn try_download(url: String, recursion_amount: u8) -> anyhow::Result<Vec<u8>> {
-    if recursion_amount == 0 {
-        bail!(Error::DownloadError(
-            anyhow::anyhow!("Too many redirects").into(),
-        ));
-    }
+fn try_download(url: String) -> anyhow::Result<Vec<u8>> {
     let mut handle = Easy::new();
     handle
         .url(&url)
         .map_err(|e| Error::DownloadError(e.into()))?;
-    let status = &mut None;
+    // Instruct curl to follow redirections
+    handle
+        .follow_location(true)
+        .map_err(|e| Error::DownloadError(e.into()))?;
+    // Optionally set the max number of redirects
+    handle
+        .max_redirections(5)
+        .map_err(|e| Error::DownloadError(e.into()))?;
+
+    let mut bytes = Vec::new();
     {
         let mut transfer = handle.transfer();
         transfer
-            .header_function(|header| {
-                let header = std::str::from_utf8(header).unwrap();
-                if header.starts_with("HTTP/1.1 ") {
-                    let n = &header[9..12];
-                    *status = Some(n.parse::<u16>().expect("Should be number"));
-                }
-                true
+            .write_function(|d| {
+                bytes.extend_from_slice(d);
+                Ok(d.len())
             })
-            .unwrap();
+            .map_err(|e| Error::DownloadError(e.into()))?;
         print_verbose_status!("Requesting", &url);
-        transfer.perform()?;
+        transfer
+            .perform()
+            .map_err(|e| Error::DownloadError(e.into()))?;
     }
-    match status.expect("HTTP request did not have a status code") {
-        200u16 => {
-            print_verbose_status!("Retrieved", "data from {}", url);
-            let mut bytes = Vec::new();
-            {
-                let mut transfer = handle.transfer();
-                transfer
-                    .write_function(|d| {
-                        bytes.extend(d);
-                        Ok(d.len())
-                    })
-                    .map_err(|e| Error::DownloadError(e.into()))?;
-                transfer.perform()?;
-            }
-            Ok(bytes)
-        }
-        302 => {
-            let redirect_url = &mut None;
-            {
-                let mut transfer = handle.transfer();
-                transfer
-                    .header_function(|header| {
-                        let header = std::str::from_utf8(header).unwrap();
-                        if header.starts_with("Location: ") {
-                            *redirect_url = Some(header[10..header.len() - 2].to_owned())
-                        }
-                        true
-                    })
-                    .map_err(|e| Error::DownloadError(e.into()))?;
-                transfer.perform()?;
-            }
-            let redirect_url = redirect_url.take().unwrap();
 
-            try_download(redirect_url, recursion_amount - 1)
-        }
-        s => bail!(Error::DownloadError(
-            anyhow::anyhow!("Non-successful response: {} {}", url, s).into(),
-        )),
+    match handle
+        .response_code()
+        .map_err(|e| Error::DownloadError(e.into()))?
+    {
+        200 => Ok(bytes),
+        code => Err(Error::DownloadError(curl::Error::new(code as _).into()).into()),
     }
 }
 


crates/cargo-winrt/src/main.rs
```diff
@@ -332,7 +332,7 @@ impl DependencyDescriptor {
     }
 }
 
-fn try_download(url: String, recursion_amount: u8) -> anyhow::Result<Vec<u8>> {
+fn try_download(url: String, recursion_amount: u8, redirect: bool) -> anyhow::Result<Vec<u8>> {
```
tim-weis, line 335: The formal parameter name (`redirect`) doesn't quite match the argument's meaning. Something like `request_body` or `include_body` is maybe a better fit.

uhuntu: Changed to request_nobody.

tim-weis: I don't not think this is not an improvement ;) For one, it exposes naming details of the currently used library. If that library ever gets swapped out for another library with different naming (say `no_body()` rather than `nobody()`), the parameter name no longer matches.

What's more troublesome, as I hinted to above: Using a negative in a parameter name is an invitation for double and triple negations in code, with an unavoidable tendency to make it harder to comprehend. Plus, the name suggests that you are *"requesting no body"*, when really the semantics are to *"not request a body"*.

Though I have no say in this. I'm not affiliated with this repository, just a user of this library. This is really only a comment.

zachlute: As @tim-weis mentioned, I'm also not super keen on this negation in the parameter name. Maybe more than that for me, I have a real brain problem parsing `request_nobody` and even in that case would definitely prefer `request_no_body`, but maybe `nobody` is just a common term of art and I'm not steeped enough in HTTP culture to know that.

I would definitely prefer `include_body` in this case for both of those reasons, but I'm also just a humble drive-by commenter. :)

kennykerr: `include_body` seems the cleanest without imposing too much rework on a simple internal function. If this were a library I'd avoid a bool parameter altogether.

kennykerr:
@uhuntu how are you testing this to confirm before/after that the network usage is satisfactory? Also, remember to run `cargo fmt` on the code to allow the CI build to pass.

crates/cargo-winrt/src/main.rs
uhuntu: Thanks for comments, changed to include_body

uhuntu:
> @uhuntu how are you testing this to confirm before/after that the network usage is satisfactory? Also, remember to run `cargo fmt` on the code to allow the CI build to pass.

About the cargo fmt, it's pretty strange that in my machine it will not result like CI did. Anyway, I format it like CI suggest.

About the testing, you can see the ouput below after the fix:
```
C:\Users\hunt\source\repos\winrt-rs>cargo winrt install --verbose
   Resolving winrt_macros
  Installing 0 nuget dependencies
   Resolving winrt
  Installing 3 nuget dependencies
    Fetching KennyKerr.Windows.TestWinRT
  Requesting https://www.nuget.org/api/v2/package/KennyKerr.Windows.TestWinRT/1.0.17
  Requesting https://globalcdn.nuget.org/packages/kennykerr.windows.testwinrt.1.0.17.nupkg
   Retrieved data from https://globalcdn.nuget.org/packages/kennykerr.windows.testwinrt.1.0.17.nupkg
    Starting extraction of 'KennyKerr.Windows.TestWinRT'
   Searching zip file: _rels\.rels
   Searching zip file: KennyKerr.Windows.TestWinRT.nuspec
   Searching zip file: lib\uap10.0\TestComponent.winmd
       Found winmd file: "TestComponent.winmd"
   Searching zip file: runtimes\win10-x86\native\TestComponent.dll
       Found dll "TestComponent.dll" with arch "win10-x86" at path runtimes\win10-x86\native\TestComponent.dll
   Searching zip file: runtimes\win10-x64\native\TestComponent.dll
       Found dll "TestComponent.dll" with arch "win10-x64" at path runtimes\win10-x64\native\TestComponent.dll
   Searching zip file: [Content_Types].xml
   Searching zip file: package\services\metadata\core-properties\0e77aaa72e4f4184a4a9efbcbb14f1ca.psmdcp
   Searching zip file: .signature.p7s
    Fetching Microsoft.AI.MachineLearning
  Requesting https://www.nuget.org/api/v2/package/Microsoft.AI.MachineLearning/1.3.0
  Requesting https://globalcdn.nuget.org/packages/microsoft.ai.machinelearning.1.3.0.nupkg
   Retrieved data from https://globalcdn.nuget.org/packages/microsoft.ai.machinelearning.1.3.0.nupkg
```

Before the fix, you should never see "Retrieved data from https://globalcdn.nuget.org/packages/microsoft.ai.machinelearning.1.3.0.nupkg"

crates/cargo-winrt/src/main.rs
```diff
@@ -284,14 +284,14 @@ impl DependencyDescriptor {
         match self {
             DependencyDescriptor::NugetOrg { name, version } => {
                 let url = format!("https://www.nuget.org/api/v2/package/{}/{}", name, version);
-                let bytes = try_download(url, 5)?;
+                let bytes = try_download(url, 5, true)?;
```
rylev, line 287: I'm confused why nuget packages include the body by default and downloading from url does not. I would expect the opposite to be true... Can you explain?

uhuntu: If you make a debug here to print header:

```
diff --git a/crates/cargo-winrt/src/main.rs b/crates/cargo-winrt/src/main.rs
index 40b2fd0..103b592 100644
--- a/crates/cargo-winrt/src/main.rs
+++ b/crates/cargo-winrt/src/main.rs
@@ -349,6 +349,7 @@ fn try_download(url: String, recursion_amount: u8, include_body: bool) -> anyhow
         transfer
             .header_function(|header| {
                 let header = std::str::from_utf8(header).unwrap();
+                println!("header: {}", header);
                 if header.starts_with("HTTP/1.1 ") {
                     let n = &header[9..12];
                     *status = Some(n.parse::<u16>().expect("Should be number"));
```

You will see:

```
    Fetching Microsoft.AI.MachineLearning
	
https://www.nuget.org/api/v2/package/Microsoft.AI.MachineLearning/1.3.0
	
header: HTTP/1.1 302 Found
header: Cache-Control: private
header: Transfer-Encoding: chunked
header: Content-Type: text/html; charset=utf-8
header: Location: https://globalcdn.nuget.org/packages/microsoft.ai.machinelearning.1.3.0.nupkg
header: Request-Context: appId=cid-v1:338f6804-b1a9-4fe3-bba7-c93064e7ae7b
header: Access-Control-Expose-Headers: Request-Context
header: Content-Security-Policy: frame-ancestors 'none'
header: X-Frame-Options: deny
header: X-XSS-Protection: 1; mode=block
header: X-Content-Type-Options: nosniff
header: Strict-Transport-Security: max-age=31536000
header: Set-Cookie: ARRAffinity=e56c4413884e44fa2949035523b026d52af6ee3f783d256e3cbeb42fa95d6022;Path=/;HttpOnly;Domain=nuget-prod-v2gallery-appservice.trafficmanager.net
header: Date: Tue, 21 Jul 2020 01:18:30 GMT
header:

https://globalcdn.nuget.org/packages/microsoft.ai.machinelearning.1.3.0.nupkg

header: HTTP/1.1 200 OK
header: Accept-Ranges: bytes
header: Access-Control-Allow-Origin: *
header: Access-Control-Expose-Headers: x-ms-request-id,Server,x-ms-version,Content-Length,Date,Transfer-Encoding
header: Cache-Control: max-age=86400
header: Content-MD5: AAB1USeDetROmk8fD/XLqg==
header: Content-Type: application/octet-stream
header: Date: Tue, 21 Jul 2020 01:18:32 GMT
header: Etag: 0x8D7FB804A83621E
header: Expires: Wed, 22 Jul 2020 01:18:32 GMT
header: Last-Modified: Mon, 18 May 2020 23:07:52 GMT
header: Server: Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0
header: Strict-Transport-Security: max-age=31536000
header: X-Cache: HIT
header: X-CDN-Rewrite: Root path in dist
header: X-Content-Type-Options: nosniff
header: x-ms-blob-type: BlockBlob
header: x-ms-lease-status: unlocked
header: x-ms-meta-da7b2905_0f3c_4262_921c_b1593d1336f1_ESRP_RequestId: 8de9e9d0-befe-491f-849c-7e0fd05eba05
header: x-ms-meta-SHA512: gxrx74NSZh8TcDRLwYO9MqdTfn/5NfMVOMlAmdSMzyVNNFKogQWQVxsZxLlp2oIuDu0uByUQ0PNPjmFvk3ndLQ==
header: x-ms-request-id: 30fe239d-101e-0015-38fc-5ef2cc000000
header: x-ms-version: 2009-09-19
header: Content-Length: 172281918
header:
```

First time try_download from https://www.nuget.org/api/v2/package/Microsoft.AI.MachineLearning/1.3.0, it would include body. As it will cause a 302 redirect, otherwise it will cause 404

Second time it try to reach real url https://globalcdn.nuget.org/packages/microsoft.ai.machinelearning.1.3.0.nupkg, for now it should not include body, otherwise it will stick there to download both header and body.

kennykerr: I'm sorry but I still don't get what the problem is. The first request results in a 302 response telling the caller to use a different URL. Even if the body is downloaded, it should at most contain a few bytes of html describing the redirect but not the entire nuget package.

uhuntu: > I'm sorry but I still don't get what the problem is. The first request results in a 302 response telling the caller to use a different URL. Even if the body is downloaded, it should at most contain a few bytes of html describing the redirect but not the entire nuget package.

Hi Kenny, that's right, to get 302 response is include body that's fine, and we must include body to let 302 happen. My concern is 200 response after 302. To get 200 from the redirected URL and if it include body it will cause much time to download body when just want header.

I push a test repo here: https://github.com/uhuntu/curl-rust.git

You can just run a sample I wrote by: ```cargo run --example test_hunt```

You will measure that if I don't set ```handle.nobody(true).unwrap();```, the program should be seeing never end in low bandwidth environment, but actually I just want header. If you put ```handle.nobody(true).unwrap();``` back, the program will quit immediately after it get header.

```
use std::str;

fn main() {
    let mut handle = curl::easy::Easy::new();
    handle.url("https://globalcdn.nuget.org/packages/microsoft.ai.machinelearning.1.3.0.nupkg").unwrap();
    // handle.nobody(true).unwrap();
    handle.header_function(|header| {
        print!("header: {}", str::from_utf8(header).unwrap());
        true
    }).unwrap();
    handle.perform().unwrap();
}
```

tim-weis: I think I'm following now. The code logic in [`try_download`](https://github.com/microsoft/winrt-rs/blob/6ab99ef9df3840caadc1f0ed9c82ea78b208a8e6/crates/cargo-winrt/src/main.rs#L441) first requests the resource (including the body), evaluates the HTTP response code, and then either follows the redirect (302) or re-initiates the transfer (200), including the body. Either way, the body from the first request gets dropped. The net result is that the body gets downloaded twice.

That is indeed wasteful, although I'm skeptical that the proposed change is the right way to address this issue: For one, it imposes the assumption, that the first 302 response is never followed by another 302. While I don't know whether HTTP allows for that, it seems plausible that it would. Worse, though, the proposed change doesn't help in situations, where the *first* request already returns a 200 code.

Moving the `write_function` from the [`200u16` match arm](https://github.com/microsoft/winrt-rs/blob/6ab99ef9df3840caadc1f0ed9c82ea78b208a8e6/crates/cargo-winrt/src/main.rs#L468) into the initial request (or otherwise preserving the downloaded document) would address both of these issues.

uhuntu: I just re-submit the code as @tim-weis mentioned.

crates/cargo-winrt/src/error.rs
```diff
@@ -4,6 +4,7 @@ use thiserror::Error;
 type DynError = dyn StdError + Send + Sync + 'static;
 
 #[derive(Debug, Error)]
+#[allow(dead_code)]
```
tim-weis, line 7: This can be removed, once the `DownloadError` enum is used again.

crates/cargo-winrt/src/main.rs
```diff
@@ -438,70 +437,27 @@ impl DependencyDescriptor {
...
-            anyhow::anyhow!("Too many redirects").into(),
-        ));
-    }
+fn try_download(url: String) -> anyhow::Result<Vec<u8>> {
     let mut handle = Easy::new();
-    handle
-        .url(&url)
-        .map_err(|e| Error::DownloadError(e.into()))?;
-    let status = &mut None;
+    handle.url(&url)?;
```
tim-weis, line 442: Curl errors should be [mapped](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err) to `DownloadError`.

crates/cargo-winrt/src/main.rs
```diff
@@ -438,70 +437,27 @@ impl DependencyDescriptor {
...
-            .header_function(|header| {
-                let header = std::str::from_utf8(header).unwrap();
-                if header.starts_with("HTTP/1.1 ") {
-                    let n = &header[9..12];
-                    *status = Some(n.parse::<u16>().expect("Should be number"));
-                }
-                true
-            })
-            .unwrap();
-        print_verbose_status!("Requesting", &url);
```
tim-weis, line 464: This line should be kept to communicate progress.

uhuntu:
@tim-weis I re-submitted as you wished.

crates/cargo-winrt/src/main.rs
```diff
@@ -438,70 +438,30 @@ impl DependencyDescriptor {
...
-        ));
-    }
+fn try_download(url: String) -> anyhow::Result<Vec<u8>> {
     let mut handle = Easy::new();
     handle
         .url(&url)
         .map_err(|e| Error::DownloadError(e.into()))?;
-    let status = &mut None;
+    // Instruct curl to follow redirections
+    handle.follow_location(true)?;
```
tim-weis, line 447: That error would also need to get mapped to `DownloadError`.

crates/cargo-winrt/src/main.rs
```diff
@@ -438,70 +438,30 @@ impl DependencyDescriptor {
...
+    // Optionally set the max number of redirects
+    handle.max_redirections(5)?;
```
tim-weis, line 449: Same here

crates/cargo-winrt/src/main.rs
```diff
@@ -438,70 +438,30 @@ impl DependencyDescriptor {
...
-                    let n = &header[9..12];
-                    *status = Some(n.parse::<u16>().expect("Should be number"));
-                }
-                true
-            })
-            .unwrap();
+        transfer.write_function(|d| {
+            bytes.extend_from_slice(d);
+            Ok(d.len())
+        })?;
```
tim-weis, line 457: And here

crates/cargo-winrt/src/main.rs
```diff
@@ -438,70 +438,30 @@ impl DependencyDescriptor {
...
         print_verbose_status!("Requesting", &url);
         transfer.perform()?;
```
tim-weis, line 459: Probably here, too, though you should probably double-check with @rylev why the original code didn't. Maybe there's more to it.

crates/cargo-winrt/src/main.rs
```diff
@@ -438,70 +438,30 @@ impl DependencyDescriptor {
...
-            let redirect_url = redirect_url.take().unwrap();
 
-            try_download(redirect_url, recursion_amount - 1)
-        }
-        s => bail!(Error::DownloadError(
-            anyhow::anyhow!("Non-successful response: {} {}", url, s).into(),
-        )),
+    match handle.response_code()? {
+        200 => Ok(bytes),
+        code => Err(curl::Error::new(code as _).into()),
```
tim-weis, line 464: Also needs to get mapped.

crates/cargo-winrt/src/main.rs
rylev: We should map this to a DownloadError as well. Just add `.map_err(|e| Error::DownloadError(e.into()))` to all of these.

rylev:
@uhuntu this looks goood! Thanks for being patient. I think the only thing to do is what @tim-weis said - map the errors from cURL to `Error::DownloadError`

uhuntu:
@rylev @tim-weis Thanks for being patient too :) I re-submitted the code.

crates/cargo-winrt/src/main.rs
```diff
@@ -438,69 +438,39 @@ impl DependencyDescriptor {
...
+        transfer
+            .perform()
+            .map_err(|e| Error::DownloadError(e.into()))?;
+    }
 
-            try_download(redirect_url, recursion_amount - 1)
-        }
-        s => bail!(Error::DownloadError(
-            anyhow::anyhow!("Non-successful response: {} {}", url, s).into(),
+    match handle.response_code()? {
```
tim-weis, line 470: Sorry for keeping this busy, I missed this error return in my earlier review. That should also get mapped, I guess.

crates/cargo-winrt/src/main.rs
```diff
@@ -438,69 +438,39 @@ impl DependencyDescriptor {
...
+        200 => Ok(bytes),
+        code => bail!(Error::DownloadError(
```
tim-weis, line 472: As I understand [bail!](https://docs.rs/anyhow/1.0.32/anyhow/macro.bail.html), it's meant for early returns. This isn't, though, and `Err(...)` should be sufficient. With that change, the `anyhow::bail` import could be dropped, too.

uhuntu:
Re-submitted.

PR #263, opened 2020-07-16
Implement factory caching

### LINKED ISSUE #114: Add factory caching
C++/WinRT has a high-performance factory cache that dramatically improves the performance of most API calls. I need to get the same working for Rust/WinRT.

kennykerr:
The fixes #114 to add support for factory caching, a very important optimization that speeds up access to the static/constructor methods of a class. 

The `windows::ui::Colors` class provides a set of static methods that are useful for testing this. Here's a simple example:

```rust
use windows::ui::*;
use winrt::*;

fn main() -> Result<()> {
    {
        let start = std::time::Instant::now();

        for _ in 0..1_000_000 {
            let white = Colors::white()?;

            debug_assert!(
                white
                    == Color {
                        a: 255,
                        r: 255,
                        g: 255,
                        b: 255
                    }
            );
        }

        println!("single {:5}ms", start.elapsed().as_millis());
    }

    {
        let start = std::time::Instant::now();
        let mut threads = Vec::new();

        for _ in 0..16 {
            threads.push(std::thread::spawn(|| {
                for _ in 0..1_000_000 {
                    let white = Colors::white().unwrap();

                    debug_assert!(
                        white
                            == Color {
                                a: 255,
                                r: 255,
                                g: 255,
                                b: 255
                            }
                    );
                }
            }));
        }

        for thread in threads {
            thread.join().unwrap();
        }

        println!("multi  {:5}ms", start.elapsed().as_millis());
    }

    Ok(())
}
```

Here are the results without caching:

```
C:\git\rust\scratch>cargo run --release
    Finished release [optimized] target(s) in 0.22s
     Running `target\release\scratch.exe`
single   639ms
multi   7229ms
```

And here are the results with caching:

```
C:\git\rust\scratch>cargo run --release
    Finished release [optimized] target(s) in 0.12s
     Running `target\release\scratch.exe`
single    24ms
multi     21ms
```

### DIFF
--- a/crates/gen/src/types/class.rs
+++ b/crates/gen/src/types/class.rs
@@ -2,10 +2,10 @@ use super::object::to_object_tokens;
 use crate::tables::*;
 use crate::types::*;
 use crate::TypeReader;
+use crate::*;
 use proc_macro2::TokenStream;
 use quote::quote;
 use std::iter::FromIterator;
-use crate::*;
 
 /// A WinRT Class
 #[derive(Debug)]
@@ -120,6 +120,7 @@ impl Class {
         let name = &self.name.tokens;
         let type_name = self.type_name(&name);
         let methods = to_method_tokens(&self.interfaces);
+        let call_factory = self.to_call_factory_tokens();
 
         if self.interfaces[0].kind == InterfaceKind::Default {
             let conversions = TokenStream::from_iter(
@@ -131,7 +132,7 @@ impl Class {
             let new = if self.default_constructor {
                 quote! {
                     pub fn new() -> ::winrt::Result<Self> {
-                        ::winrt::factory::<Self, ::winrt::IActivationFactory>()?.activate_instance::<Self>()
+                        Self::IActivationFactory(|f| f.activate_instance::<Self>())
                     }
                 }
             } else {
@@ -147,7 +148,6 @@ impl Class {
             let abi_name = self.interfaces[0].name.to_abi_tokens();
             let (async_get, future) = get_async_tokens(&self.name, &self.interfaces);
             let debug = debug::debug_tokens(&self.name, &self.interfaces);
-            let call_factory = self.to_call_factory_tokens();
 
             let send_sync = if self.is_agile {
                 let constraints = &self.name.constraints;
@@ -202,7 +202,10 @@ impl Class {
         } else {
             quote! {
                 pub struct #name {}
-                impl #name { #methods }
+                impl #name {
+                    #methods
+                    #call_factory
+                }
                 #type_name
             }
         }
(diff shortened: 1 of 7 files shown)


src/factory.rs
```diff
@@ -1,37 +1,82 @@
 use crate::*;
+use std::marker::PhantomData;
+use std::sync::atomic::{AtomicPtr, Ordering};
+
+/// Attempts to load and cache the factory interface for the given WinRT class.
+pub struct FactoryCache<C: RuntimeName, I: ComInterface> {
+    pub shared: AtomicPtr<std::ffi::c_void>,
+    pub _c: PhantomData<C>,
+    pub _i: PhantomData<I>,
```
kennykerr, line 9: Unfortunately, all of these members must be public because Rust doesn't support const functions on generics and the callsite has to initialize these fields because there's now `new` method to wrap it all up.

crates/gen/src/types/class.rs
```diff
@@ -233,6 +239,52 @@ impl Class {
...
+    fn to_named_call_factory(&self, method_name: &str, interface: &TokenStream) -> TokenStream {
+        let self_name = &self.name.tokens;
+        let method_name = format_ident(method_name);
+
+        quote! {
+            #[allow(non_snake_case)]
+            fn #method_name<R, F: FnOnce(&#interface) -> ::winrt::Result<R>>(
+                callback: F,
+            ) -> ::winrt::Result<R> {
+                static mut SHARED: ::winrt::FactoryCache<#self_name, #interface> =
```
kennykerr, line 277: Unfortunately, Rust doesn't support type-specific statics so we need to generate a function for each class interface to house the static storage.

src/factory.rs
```diff
@@ -1,37 +1,82 @@
...
+}
+
+impl<C: RuntimeName, I: ComInterface + Default> FactoryCache<C, I> {
+    pub fn call<R, F: FnOnce(&I) -> Result<R>>(&mut self, callback: F) -> Result<R> {
```
rylev, line 13: I'm don't think I fully understand why a callback is necessary instead of just returning `&I`.

src/factory.rs
```diff
@@ -1,37 +1,82 @@
...
+    }
+}
 
 /// Attempts to load the factory interface for the given WinRT class.
-///
-/// Note that factory caching still needs to be implemented.
-pub fn factory<C: RuntimeName, I: ComInterface>() -> Result<I> {
-    let mut factory = std::ptr::null_mut();
+pub fn factory<C: RuntimeName, I: ComInterface + Default>() -> Result<I> {
+    let mut factory = I::default();
```
rylev, line 52: I like this change. It would be nice to eliminate as many uses of std::mem::transmute as possible since it's such a sharp tool.

src/factory.rs
kennykerr: Because the lifetime of `I` depends on whether the factory may be cached and that is only known at run time. If the factory is agile we cache it and a static reference is available. If it's not agile, we cannot cache it and merely use it before releasing.

PR #267, opened 2020-07-20
Add a const constructor for FactoryCache

rylev:


kennykerr:
Much better - I also pushed a tweak as the cache is always null to begin with.

PR #275, opened 2020-07-21
Add exhaustive required interface testing

kennykerr:
This is based on a stress test found here:

https://github.com/microsoft/TestWinRT/blob/479bfae3e3e290c37b05bea91e7f0a55233a986e/TestComponent/TestComponent.idl#L217-L256

This does a few things. It provides types with redundant required interfaces. This is unnecessary but allowed by MIDLRT and thus language projections must be able to deal with it. It also introduces required interfaces across the class hierarchy, further complicating the job for the language projection. All of these redundant interfaces must be folded.

The composable test in Rust now ensure not only that this generates valid Rust bindings but that all of the conversions are sound. Note that the RequiredInterface/RequiredInterfaces implementation in Rust is not very efficient. I've wanted to improve that for a while but lacked the tests to ensure I'd get it right. So here are the tests and one less reason to keep the slow Xaml tests around.

### DIFF
--- a/crates/gen/src/types/required_interface.rs
+++ b/crates/gen/src/types/required_interface.rs
@@ -117,6 +117,10 @@ impl RequiredInterface {
                 kind = InterfaceKind::NonDefault;
             }
 
+            if interfaces.iter().any(|i| i.name == append_name) {
+                continue;
+            }
+
             interfaces.push(RequiredInterface::from_type_name_and_kind(
                 reader,
                 append_name,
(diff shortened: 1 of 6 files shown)


crates/gen/src/types/required_interface.rs
```diff
@@ -117,6 +117,10 @@ impl RequiredInterface {
                 kind = InterfaceKind::NonDefault;
             }
 
+            if interfaces.iter().any(|i| i.name == append_name) {
```
kennykerr, line 120: Previously, duplicates were creeping in and causing invalid Rust to be generated with duplicate trait implementations for some structs (in some rarely used Xaml namespaces).

rylev: I wonder if storing the interfaces in a HashMap would be better. It uses a tiny bit more memory, but the lookup for duplicates will be faster.

PR #278, opened 2020-07-21
Support versions in nuget packages and overwrite on force

### LINKED ISSUE #274: Cargo winrt install doesn't update nuget packages
Even when using `-f` it simply won't update a nuget package from one version to the next. To repro:

Use `cargo init force` to create a new package

Add the following to the Cargo.toml file:

```
"KennyKerr.Windows.TestWinRT" = "1.0.19"
```

Note the version number. Now run the following to download the nuget package:

```
cargo winrt install
```

This successfully downloads version 1.0.19. Run the following and note the time stamp and file size:

```
dir /s testcomponent.winmd
```

Now update the version of the nuget package in the Cargo.toml file:

```
"KennyKerr.Windows.TestWinRT" = "1.0.20"
```

Then run the following to update the nuget package to the new version:

```
cargo winrt install
```

Check the time stamp and file size. Nothing changed. Try to force it:

```
cargo winrt install -f
```

Check the time stamp and file size. Nothing changed. Just nuke the whole thing and start again:

```
cargo clean
cargo winrt install
```

Check the time stamp and file size. After running `cargo clean`, the `cargo winrt` tool can successfully download the correct version.

This should not be necessary and is a big waste of time on a large project.

rylev:
Fixes #274

### DIFF
--- a/crates/cargo-winrt/src/main.rs
+++ b/crates/cargo-winrt/src/main.rs
@@ -299,7 +299,7 @@ impl Install {
 
         let deps = self.get_dependencies(dependency_descriptors)?;
         for dep in deps {
-            dep.save()?;
+            dep.save(self.force)?;
         }
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


crates/cargo-winrt/src/main.rs
```diff
@@ -417,14 +417,24 @@ impl DependencyDescriptor {
         }
     }
 
+    fn version(&self) -> &str {
+        match self {
+            DependencyDescriptor::NugetOrg { version, .. } => version,
+            DependencyDescriptor::Url { .. } => "unknown",
+            DependencyDescriptor::Local { .. } => "unknown",
```
Josh015, line 424: Wouldn't it be better to make these last two cases into a catch all?
```rust
_ => "unknown",
```

rylev: I imagine we'll add more varieties and I don't want to miss changing this code when we add them

Josh015: Understood.

PR #279, opened 2020-07-21
Const guid

### LINKED ISSUE #136: Calculate generic GUIDs at compile time
Following on #129, the next step it to calculate the generic GUIDs at compile time using Rust const functions to avoid the expensive calculation at run-time.

rylev:
Fixes #136 

Note that this will push the MSRV to Rust 1.46 which is currently in beta.


src/guid.rs
```diff
@@ -35,29 +35,37 @@ impl Guid {
     /// Creates a `Guid` for a "generic" WinRT type.
     ///
     /// Note this needs to be a const function as soon as [Rust supports it](https://github.com/microsoft/winrt-rs/issues/136).
```
Boddlnagg, line 37: This comment is now obsolete ;-)

rylev:
@kennykerr this is ready for review. It won't pass CI until I update const-sha1 with the changes necessary to get this to work, but I'd like to have your eyes on this before I push a new version. Once that's done, I'll push a new version of const-sha1 and we can merge this.

Cargo.toml
```diff
@@ -11,7 +11,7 @@ readme = "README.md"
 
 [dependencies]
 winrt_macros = { path = "crates/macros",  version = "0.7.1" }
-sha1 = "0.6.0"
+const-sha1 = { path = "../const-sha1" }
```
kennykerr, line 14: Where is this crate going to live?

crates/gen/src/types/class.rs
```diff
@@ -173,14 +173,12 @@ impl Class {
...
-                    }
+                    const IID: ::winrt::Guid = {
+                        <#default_name as ::winrt::ComInterface>::IID
+                    };
                 }
                 unsafe impl ::winrt::RuntimeType for #name {
-                    fn signature() -> String {
-                        #signature.to_owned()
-                    }
+                    const SIGNATURE: ::winrt::ConstBuffer = { ::winrt::ConstBuffer::from_slice(#signature) };
```
kennykerr, line 181: I think this can be:

```rust 
const SIGNATURE: ::winrt::ConstBuffer = ::winrt::ConstBuffer::from_slice(#signature);
```

Cargo.toml
rylev: It's on my personal GitHub and already live on crates.io: https://github.com/rylev/const-sha1 https://crates.io/search?q=const-sha1

crates/gen/src/types/class.rs
```diff
@@ -173,14 +173,12 @@ impl Class {
                 #type_name
                 unsafe impl ::winrt::ComInterface for #name {
                     type VTable = #abi_name;
-                    fn iid() -> ::winrt::Guid {
-                        <#default_name as ::winrt::ComInterface>::iid()
-                    }
+                    const IID: ::winrt::Guid = {
+                        <#default_name as ::winrt::ComInterface>::IID
+                    };
```
kennykerr, line 178: Same here about braces.

Cargo.toml
kennykerr: Got it - so we just need that update and then we can point to the crate.

crates/gen/src/types/delegate.rs
```diff
@@ -109,9 +109,9 @@ impl Delegate {
             }
             unsafe impl<#constraints> ::winrt::ComInterface for #name {
                 type VTable = #abi_definition;
-                fn iid() -> ::winrt::Guid {
+                const IID: ::winrt::Guid = {
                     #guid
-                }
+                };
```
kennykerr, line 114: And here.

Cargo.toml
rylev: Exactly. I just didn't want to push a new version if we decided in this PR that the API wasn't quite right

crates/gen/src/types/delegate.rs
```diff
@@ -130,9 +130,7 @@ impl Delegate {
                 #phantoms
             }
             unsafe impl<#constraints> ::winrt::RuntimeType for #name {
-                fn signature() -> String {
-                    #signature
-                }
+                const SIGNATURE: ::winrt::ConstBuffer = { #signature };
```
kennykerr, line 133: And here.

crates/gen/src/types/delegate.rs
```diff
@@ -192,9 +190,9 @@ impl Delegate {
                     unsafe {
                         let this: *mut Self = this.as_raw() as _;
 
-                        if iid == &<#name as ::winrt::ComInterface>::iid()
-                            || iid == &<::winrt::IUnknown as ::winrt::ComInterface>::iid()
-                            || iid == &<::winrt::IAgileObject as ::winrt::ComInterface>::iid()
+                        if iid == &<#name as ::winrt::ComInterface>::IID
+                            || iid == &<::winrt::IUnknown as ::winrt::ComInterface>::IID
+                            || iid == &<::winrt::IAgileObject as ::winrt::ComInterface>::IID
```
kennykerr, line 195: Lovely. 🎉

crates/gen/src/types/enum.rs
```diff
@@ -85,9 +85,7 @@ impl Enum {
                 #(#fields)*
             }
             unsafe impl ::winrt::RuntimeType for #name {
-                fn signature() -> String {
-                    #signature.to_owned()
-                }
+                const SIGNATURE: ::winrt::ConstBuffer = { ::winrt::ConstBuffer::from_slice(#signature) };
```
kennykerr, line 88: Here.

crates/gen/src/types/interface.rs
```diff
@@ -86,9 +86,7 @@ impl Interface {
             }
             unsafe impl<#constraints> ::winrt::ComInterface for #name {
                 type VTable = #abi_definition;
-                fn iid() -> ::winrt::Guid {
-                    #guid
-                }
+                const IID: ::winrt::Guid = { #guid };
```
kennykerr, line 89: And here.

crates/gen/src/types/interface.rs
```diff
@@ -105,9 +103,7 @@ impl Interface {
                 #phantoms
             }
             unsafe impl<#constraints> ::winrt::RuntimeType for #name {
-                fn signature() -> String {
-                    #signature
-                }
+                const SIGNATURE: ::winrt::ConstBuffer = { #signature };
```
kennykerr, line 106: And here.

crates/gen/src/types/struct.rs
```diff
@@ -56,9 +56,7 @@ impl Struct {
                 #(#fields),*
             }
             unsafe impl ::winrt::RuntimeType for #name {
-                fn signature() -> String {
-                    #signature.to_owned()
-                }
+                const SIGNATURE: ::winrt::ConstBuffer = { ::winrt::ConstBuffer::from_slice(#signature) };
```
kennykerr, line 59: I wish type inference worked on const/static.

```rust
const SIGNATURE = ::winrt::ConstBuffer::from_slice(#signature);
```

crates/gen/src/types/delegate.rs
rylev: The braces are necessary here because the signature is calculated through several lines of const code.

crates/gen/src/types/interface.rs
rylev: Same as above.

kennykerr:
Can't wait to do some run time comparative performance testing with this change.

Boddlnagg:
Note that this is a breaking change for consumers of `winrt-rs` who have their own impls of `winrt::ComInterface`. This should be reflected in the next release version.

rylev:
Yes, our next version will be 0.8 for multiple reasons 👍 .

kennykerr:
Here's the performance impact of const guids. Consider the following sample:

```rust
fn main() -> Result<()> {
    let uri = Uri::create_uri("http://kennykerr.ca")?;
    let query = uri.query_parsed()?;

    let start = std::time::Instant::now();

    for _ in 0..1_000_000 {
        let s = query.size()?;
        debug_assert!(s == 0);
    }

    println!("{:5}ms", start.elapsed().as_millis());
    Ok(())
}
```

The `size` method is on a required but non-default interface of the `WwwFormUrlDecoder` class returned by the `query_parsed` method. That means that every time `size` is called, `QueryInterface` must first be called to retrieve the appropriate vtable. This is just one of those details that a language projection hides from the language as a convenience, but such hidden costs need to be as minimal as possible.

Before this commit:

```
C:\git\rust\scratch>cargo run --release
    Finished release [optimized] target(s) in 0.09s
     Running `target\release\scratch.exe`
  829ms
```

After this commit:

```
C:\git\rust\scratch>cargo run --release
    Finished release [optimized] target(s) in 0.09s
     Running `target\release\scratch.exe`
   21ms
```

That's about 40 times faster and on par with C++/WinRT.

nu8:
@kennykerr looks like your snippet is missing this:

~~~rust
import!(
   dependencies
      os
   types
      windows::foundation::Uri
);
use windows::foundation::Uri;
use winrt::*;
~~~

otherwise its not valid Rust.

Issue #285, opened 2020-07-23
Typos in build/import macro produce ugly error message


kennykerr:
When these macros can't find the requested type/namespace the error message is not that pretty.

Here I mistyped "contacts" as "contants".

```rust
winrt::build!(
    dependencies
        os
    types
        windows::application_model::contants::IContactFieldFactory
);
```

```
error: 'windows.applicationmodel.contants' is not a known namespace
 --> C:\git\rust\scratch\target\debug\build\bindings-c0911e994fd7a983\out/winrt.rs:1:1
  |
1 | compile_error! { "\'windows.applicationmodel.contants\' is not a known namespace" }
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #288, opened 2020-07-27
Use squote for code generation

rylev:
This moves `winrt_gen` to using `squote`, a new crate I created which works just like `quote` except it simply stores things in a string instead of using the compiler backed TokenStream APIs. Since we're generating code from non-user input, this is a much more performant way to generate code. I'm hoping that this can be merged into [`quote` proper at some point](https://github.com/alexcrichton/proc-macro2/pull/239), but until then the squote crate is pretty minimal and easy to maintain. It's not yet on crates.io, but if we're happy with this direction, I can push it there.

A happy coincidence is now code generation is thread safe and so we can parallelize it. I used the [`rayon`](https://crates.io/crates/rayon) crate to simply make some of our iterators parallel, but perhaps we'll want to just do a simple thread pool instead so we don't take on an additional dependency. 

For my test crate which normally takes ~10 seconds to generate code (as its generating all of Windows.UI.Xaml), this change now generates code in 600ms. Without parallelization it generates code in about 2 seconds.


kennykerr:
That's awesome - I originally created an elaborate coroutine scheduler for C++/WinRT but finally settled on a very simple threadpool-based task parallelizer that was just as fast and much simpler as it just used standard library primitives.

https://github.com/microsoft/cppwinrt/blob/master/cppwinrt/task_group.h

rylev:
This is ready to go. For reference, on the stress test repo it takes 67 seconds to generate the token string in the `build!` macro using the current version of WinRT-rs. It now takes 8 seconds.

rylev:
This also fixes #285

simonbuchan:
I don't suppose you have memory comparison too? I would expect it to also be much better, of course.

rylev:
I don't have a direct memory comparison as memory consumption is not a first order concern (given memory is typically not a bottleneck for compilation) though at some point I will do an analysis just to make sure we're being reasonable. I could see memory consumption being slightly worse as the proc_macro bridge does cache in local storage, but we'll have to see.

Issue #79, opened 2020-04-27
Support for numerics


kennykerr:
The Windows.Foundation.Numerics types need special math operations as provided by C++/WinRT and C#/WinRT.

PR #290, opened 2020-07-28
Generate math operations for numeric types

robmikh:
This change takes the general idea from the [minesweeper-rs numerics helper](https://github.com/robmikh/minesweeper-rs/blob/master/bindings/src/numerics.rs) and generates those implementations within winrt-rs. Math operations are supported for both the struct and for a borrowed struct. This addresses #79.

While I'm still working on some of it, I have enough that I wanted to get some feedback.

Currently implemented:

- [x] Vector2/3/4 basic operations (Add, Sub, Div, Mul)
- [x] Matrix3x2/Matrix4x4 basic operations (Add, Sub, Mul)
- [ ] Vector2/3/4 helper functions (length, dot, etc)
- [ ] Matrix3x2/Matrix4x4 helper functions (identity, from scale, etc)
- [ ] Plane/Rational/Quaternion basic operations

This change also doesn't bother with anything like SIMD, although that would be interesting to investigate in the future.


kennykerr:
Sweet, here are the [tests for the OS numerics](https://microsoft.visualstudio.com/OS/_git/os.2020?path=%2Fonecoreuap%2Fwindows%2FDirectX%2Fmisc%2Fnumerics%2Ftests). May be helpful as a reference.

robmikh:
Ooh, thanks! I'll incorporate some of these as I bring online more of the special helper functions.

crates/gen/src/numerics.rs
```diff
@@ -0,0 +1,190 @@
+use crate::format_ident;
+use crate::types::*;
+use proc_macro2::TokenStream;
+use quote::quote;
+use std::iter::FromIterator;
+
+pub fn generate_struct_extensions(winrt_struct: &Struct) -> TokenStream {
```
kennykerr, line 7: So this seems nice and generic in terms of how it iterates the struct fields and so on, but I'm wondering whether this is overkill since the numerics can never change and are unlikely to ever grow. Perhaps we can simply write out the numeric operations by hand and simply include them all in a single quote! (per type). I suspect that will be easier to reason about.

crates/gen/src/types/struct.rs
```diff
@@ -69,6 +71,11 @@ impl Struct {
                     self as *mut Self::Abi
                 }
             }
+            #extensions
         }
     }
+
+    fn add_extensions(&self) -> TokenStream {
+        crate::numerics::generate_struct_extensions(self)
```
kennykerr, line 79: Maybe just call this function above rather than wrapping it.

robmikh: Sounds good to me, it can always be introduced later if we have more types of extensions.

crates/gen/src/numerics.rs
robmikh: Fair enough, I'll run this and paste the output in directly. I just didn't want to type all of it by hand 🙂

crates/gen/src/numerics.rs
```diff
@@ -0,0 +1,190 @@
...
+
+enum Operation {
+    Add,
+    Sub,
+    Div,
+    Mul,
+}
+
+impl Operation {
+    pub fn to_tokens_with_same(&self, winrt_struct: &Struct) -> TokenStream {
```
rylev, line 98: This is a confusing name for a method. The same what?

robmikh: In hindsight maybe it should have been named same_type or something. The method is supposed to generate all the operations against both a struct of the same type and a borrow of that type of struct.

Either way, this is being deleted in favor of just putting it all in one big quote macro.

robmikh:
I'm going to close this PR for now. I'm going to continue down the generation path until I have all the stuff I'd like implemented and then will open a new PR that puts all the output from that into the quote macro.

PR #291, opened 2020-07-28
Basic boxing support

kennykerr:
WinRT's type system takes after .NET with its split of reference types and value types. Reference types logically derive from `Object` and value types must be "boxed" in order to store their values in an `Object` polymorphically. This is required by Xaml but also comes into play in some other areas like nullable types. 

This update adds support for boxing basic primitive types using the `TryFrom` and `TryInto` traits. Full boxing support needs to also cover boxing user-defined types, but that's going to have to wait until I have full authoring support since that requires us to be able to implement `IReference<T>`. 

As an example, you can simply store any class/interface type as an `Object` simply by using the `Into` or `From` traits:

```rust
let uri = Uri::create_uri("http://kennykerr.ca/")?;
let o: Object = uri.into();
```

```rust
let uri = Uri::create_uri("http://kennykerr.ca/")?;
let o = Object::from(uri);
```

Now with this update you can also box (and unbox) basic primitives as well:

```rust
let o: Object = 123.try_into()?;
let v: i32 = o.try_into()?;
assert!(v == 123);
```

```rust
let o = Object::try_from(123)?;
let v = i32::try_from(o)?;
assert!(v == 123);
```

I also had to move `Object` from the `winrt` crate into the generated code in order to get around Rust's restrictions around where traits may be implemented relative to either the implementing type or the trait itself (and the fact that boxing requires support from types in the `windows::foundation` module). This rule is going to mean more and more types and even traits from the `winrt` crate are likely to move into generated code. `Object` is now included inside the `windows::foundation` module and is always included, along with its boxing dependencies.

I also removed the WinRT `TryInto` trait as that makes the two traits inconvenient to use together due to naming ambiguity. Instead, the `ComInterface` trait simply also provides a `try_query` method along with the existing `query` method.


crates/gen/src/types/futures.rs
```diff
@@ -50,7 +50,7 @@ fn to_async_tokens(
     name: &TypeName,
     self_name: &TypeName,
 ) -> (TokenStream, TokenStream) {
-    let namespace = to_namespace_tokens("Windows.Foundation", &self_name.namespace);
```
rylev, line 53: FWIW I like `namespace` better

src/com/interface.rs
```diff
@@ -41,6 +41,24 @@ pub unsafe trait ComInterface: Sized + crate::AbiTransferable {
...
+            unsafe { (ptr.vtable().unknown_query_interface)(ptr, &Into::iid(), &mut into).ok()? };
+
+            debug_assert!(
+                !into.is_null(),
+                "Null pointer found after successful QueryInterface call"
+            );
+
+            unsafe { Ok(std::mem::transmute_copy(&into)) }
+        } else {
+            Err(ErrorCode(0x80004002).into())
```
rylev, line 58: nit: Can we put the error number into a descriptive constant?

PR #293, opened 2020-07-29
Reduce code gen by deriving common implementations

kennykerr:


rylev:
Does this have a noticeable effect on compile times?

kennykerr:
> For the generic types like Class, we don't ever expect there to be a generic type that isn't Clone right?

No, all WinRT types will always be Clone.

> Does this have a noticeable effect on compile times?

No, it just removed a bunch of clutter in generated code. The compiler still has to expand the code out so it's a wash.

Issue #295, opened 2020-07-31
To use ScrollViewer.


uhuntu:
I tried to push a branch here: https://github.com/uhuntu/win_ui/tree/sv to demonstrate how is going I tried to use ScrollViewer.

In the file https://github.com/uhuntu/win_ui/blob/sv/src/main.rs, there is a code section:
```
    let sp = StackPanel::new()?;
    let tb = TextBox::new()?;
    let sv = ScrollViewer::new()?;

    sp.children()?.append(&tb)?;
    sp.update_layout()?;
    desktop_source.set_content(&sv)?;
    sv.set_content(&sp)?;
```

But the build result:
```
C:\Users\hunt\source\repos\win_ui>cargo build
   Compiling win_ui v0.1.0 (C:\Users\hunt\source\repos\win_ui)
error[E0277]: the trait bound `winrt::param::Param<'_, winrt::object::Object>: std::convert::From<&bindings::windows::ui::xaml::controls::StackPanel>` is not satisfied
  --> src\main.rs:52:8
   |
52 |     sv.set_content(&sp)?;
   |        ^^^^^^^^^^^ the trait `std::convert::From<&bindings::windows::ui::xaml::controls::StackPanel>` is not implemented for `winrt::param::Param<'_, winrt::object::Object>`
   |
   = help: the following implementations were found:
             <winrt::param::Param<'a, T> as std::convert::From<&'a T>>
             <winrt::param::Param<'a, T> as std::convert::From<T>>
             <winrt::param::Param<'a, winrt::hstring::HString> as std::convert::From<&'a str>>
             <winrt::param::Param<'a, winrt::hstring::HString> as std::convert::From<std::string::String>>
   = note: required because of the requirements on the impl of `std::convert::Into<winrt::param::Param<'_, winrt::object::Object>>` for `&bindings::windows::ui::xaml::controls::StackPanel`     

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: could not compile `win_ui`.

To learn more, run the command again with --verbose.
```

joekgh:
The signature for set_content is:
```
pub fn set_content<'a, T0__: Into<Param<'a, Object>>>(
    &self,
    value: T0__
) -> Result<()>
````
You need to convert the `StackPanel` to an `Object` first.

One example would be: `sv.set_content(&sp.into())?;`

kennykerr:
That's a bug. The following should work:

```rust
let sp = StackPanel::default();
[...]

PR #297, opened 2020-07-31
Make the foundation namespaces part of the winrt crate

### LINKED ISSUE #92: Support for boxing and unboxing values
This can probably take the form of `Into` and `TryInto` trait implementations around `winrt::Object`.

kennykerr:
The `Windows.Foundation` namespace contains types that are central to the design and operation of a language projection. So much so, that the C# language projection hides or renames most of those types. The C++ language projection doesn’t quite go that far but still ends up adding a ton of value-add for those types into the generated code. Simply generating this code is problematic in Rust because other libraries cannot then easily extend or build on these types in a reusable way and the `winrt` crate cannot make use of them directly without a dependency on code that doesn't yet exist.

In this update I’m pulling the central foundation namespaces directly into the `winrt` crate so that they will always be present and can thus be relied upon to service the projection. This enables things like boxing support as well as convertibility for `TimeSpan`, which previously required this type to receive special treatment.

The `winrt` crate now uses the build macro to generate the following namespaces and bakes them into the winrt crate:

```
Windows.Foundation
Windows.Foundation.Collections
Windows.Foundation.Numerics
```

The types in these namespaces are now available in the following paths:

```rust
winrt::foundation::*
winrt::foundation::collections::*
winrt::foundation::numerics::*
```

Note that these modules, while public, are hidden for documentation and are reexported into the generated bindings so developers can continue to rely on them in their usual placement in the type hierarchy.

The `TimeSpan` type is no longer hand-rolled and naturally appears in the `winrt` crate as part of code gen with the addition of `std::time::Duration` convertibility. A lot of special handling for `TimeSpan` could be removed thanks to the ingestion of the foundation namespaces.

Boxing support is now also provided for primitive types. WinRT's type system takes after .NET with its split of reference types and value types. Reference types logically derive from `Object` and value types must be "boxed" in order to store their values in an `Object` polymorphically. This is required by Xaml but also comes into play in some other areas like nullable types. 

This update adds support for boxing basic primitive types using the `TryFrom` and `TryInto` traits. Full boxing support needs to also cover boxing user-defined types, but that's going to have to wait until I have full authoring support since that requires us to be able to implement `IReference<T>`. 

As an example, you can simply store any class/interface type as an `Object` simply by using the `Into` or `From` traits:

```rust
let uri = Uri::create_uri("http://kennykerr.ca/")?;
let o: Object = uri.into();
```

```rust
let uri = Uri::create_uri("http://kennykerr.ca/")?;
let o = Object::from(uri);
```

Now with this update you can also box (and unbox) basic primitives as well:

```rust
let o: Object = 123.try_into()?;
let v: i32 = o.try_into()?;
assert!(v == 123);
```

```rust
let o = Object::try_from(123)?;
let v = i32::try_from(o)?;
assert!(v == 123);
```

I also removed the WinRT `TryInto` trait as that makes the two traits inconvenient to use together due to naming ambiguity. Instead, the `ComInterface` trait simply also provides a `try_query` method along with the existing `query` method.

Fixes #92 #295 #253


Boddlnagg:
I haven't looked into the diff to see if you already do this, but maybe it would make sense to reexport the Windows.Foundation namespace in generated SDK binding? So that a user can still request bindings generation for `windows::foundation` and will get reexports of the types from `winrt::*` instead?

build.rs
```diff
@@ -0,0 +1,10 @@
+winrt_macros::build!(
+    foundation
+    dependencies
+        nuget: Microsoft.Windows.SDK.Contracts
+    types
```
rylev, line 5: observation: the syntax we have for including everything in the dependencies is a bit awkward...

crates/gen/src/type_tree.rs
```diff
@@ -31,6 +31,16 @@ impl TypeTree {
         }
     }
 
+    pub fn remove(&mut self, namespace: &str) {
```
rylev, line 34: observation: I do wish our namespace handling was less "stringly-typed".

crates/gen/src/types/class.rs
```diff
@@ -197,9 +195,28 @@ impl Class {
                         <::winrt::ComPtr<#default_name> as ::winrt::AbiTransferable>::set_abi(&mut self.ptr)
                     }
                 }
+                impl ::std::convert::From<#name> for ::winrt::Object {
```
rylev, line 198: Is there a way to achieve this with a blanket impl? I wonder if having a marker trait for all WinRT classes and having a blanket impl `impl <T: Class> From<T> for Object` would be better for compile times.

crates/gen/src/types/delegate.rs
```diff
@@ -150,7 +150,7 @@ impl Delegate {
                     invoke: #impl_name::invoke,
                     #phantoms
                 };
-                pub fn new(invoke: F) -> #name {
+                pub fn make(invoke: F) -> #name {
```
rylev, line 153: What's the reason for the name change?

src/lib.rs
```diff
@@ -33,14 +35,15 @@
...
-//! domain: kennykerr.ca
-//! port: 80
-//! string: http://kennykerr.ca/
-//! ```
+
+mod bindings {
+    include!(concat!(env!("OUT_DIR"), "/winrt.rs"));
+}
+
+extern crate self as winrt;
```
rylev, line 43: Does this show up in the docs? Perhaps we want to do `#[doc(hidden}]` for this.

tests/com_interface.rs
```diff
@@ -1,18 +1,11 @@
...
-
 #[test]
 fn com_interface() -> winrt::Result<()> {
     use winrt::AbiTransferable;
     use winrt::ComInterface;
 
-    use windows::foundation::IStringable;
-    use windows::foundation::IUriRuntimeClass;
-    use windows::foundation::Uri;
+    use winrt::IStringable;
```
rylev, line 6: I'm wondering if having the foundation types in a `foundation` module would be better (e.g., `use winrt::foundation::IStringable`

src/com/interface.rs
```diff
@@ -41,6 +41,24 @@ pub unsafe trait ComInterface: Sized + crate::AbiTransferable {
         }
     }
 
+    fn try_query<Into: ComInterface>(&self) -> Result<Into> {
```
rylev, line 44: I'm confused how `query` above is actually safe now. What if you query for an interface that does not make sense?

tests/interface.rs
```diff
@@ -1,16 +1,8 @@
-winrt::import!(
-    dependencies
-        os
-    types
-        windows::foundation::{IStringable, Uri}
-);
-
 use winrt::ComInterface;
+use winrt::{IStringable, Uri};
```
rylev, line 2: I wonder if re-exporting the foundation namespace in the generated code would also make sense.

crates/gen/src/types/struct.rs
```diff
@@ -68,6 +70,35 @@ impl Struct {
                     self as *mut Self::Abi
                 }
             }
+            #extensions
+        }
+    }
+
+    fn extension_tokens(&self) -> TokenStream {
+        if self.name.name == "TimeSpan" && self.name.namespace == "Windows.Foundation" {
```
rylev, line 78: Why do this in generated code instead of in the WinRT crate proper? I would imagine that we want to put as much into "normal" code as we can and only generate what we have to.

rylev:
I'm also seeing a few warnings from the generated code:
```
warning: static is never used: `SHARED`
      --> C:\Users\ryanl\Code\stress-rs\target\debug\build\bindings-12d29370494acda0\out/winrt.rs:574672:21
       |
574672 | /                     static mut SHARED: ::winrt::FactoryCache<
574673 | |                         WiFiDirectDevice,
574674 | |                         IWiFiDirectDeviceStatics2,
574675 | |                     > = ::winrt::FactoryCache::new();
       | |_____________________________________________________^
```

And it seems to take ~3 minutes longer to compile the stress test repo on this branch rather than the main branch.

build.rs
kennykerr: Any suggestions? I was hoping not to have to add **_any_** syntax but I couldn't figure out how to sniff out the winrt crate build without confusing all of the tests.

crates/gen/src/types/class.rs
kennykerr: The trouble is that unlike in C++, Rust doesn't support negation, conjunction, or disjunction as generic constraints. Traits need to map to non-overlapping sets. I have been thinking about introducing non-overlapping marker traits for class/interface/delegate/struct so we can at least provide blanket implementations for subsets of WinRT types because so many of these are identical.

crates/gen/src/types/delegate.rs
kennykerr: Calling it new and *not* returning `Self` results in a clippy warning.

tests/com_interface.rs
kennykerr: Yeah, I'm not sure. I'd prefer to keep `winrt` module types to a minimum but this is more consistent with C# which accounts for the vast majority of existing docs/samples/apps so seems helpful to beginners.

tests/interface.rs
kennykerr: I thought having different names/aliases for all those types would be confusing and lead to a mix of code that used one name or the other.

tests/com_interface.rs
rylev: Yea that was my though. Having a 1 to 1 mapping is helpful for discoverability.

tests/interface.rs
rylev: I personally think mixing and matching would be unlikely and not that disruptive if it occurs. It's nice to be able to refer to the types in their "native" namespace.

kennykerr:
> warning: static is never used: `SHARED`

That's an existing bug that I'm aware of. It has to do with static/constructor overloads that don't quite map correctly to Rust (since it lacks overloads). I have a plan to fix it but that's unrelated to this change. #300

src/com/interface.rs
kennykerr: Yes, if query fails it returns a null pointer. The query function is primarily for the projection to use internally where it knows which queries make sense. So perhaps we need to rename things so that `try_query` has the "good name" and `query` is given a less obvious name.

src/lib.rs
kennykerr: Done.

crates/gen/src/type_tree.rs
kennykerr: Yep

kennykerr:
@Boddlnagg @rylev the foundation namespaces are now reexported into the generated code.

src/com/interface.rs
kennykerr: #302

tests/interface.rs
kennykerr: Done

robmikh:
Looks good to me! The only minor thing is that the pull request description is out of date. I tried to play around with it and discovered the `foundation` addition to the namespace.

kennykerr:
Thanks @robmikh - I've updated the PR description for historical record.

crates/gen/src/type_namespaces.rs
```diff
@@ -16,11 +16,20 @@ impl TypeNamespaces {
...
+            let tokens = tree.to_tokens().collect::<Vec<_>>();
 
-            quote! {
-                pub mod #name {
-                    #(#tree)*
+            if tree.foundation {
+                quote! {
+                    pub mod #name {
+                        #(#tokens)*
+                        pub use ::winrt::foundation::*;
```
rylev, line 25:
```rust 
let foundation  = if tree.foundation { quote! { pub use ::winrt::foundation::* } else { TokenStream::new() };
 quote! {
     pub mod #name {
         #(#tokens)*
         #foundation 
     }
 }


```

crates/gen/src/type_tree.rs
```diff
@@ -6,8 +6,9 @@ use squote::TokenStream;
 /// A namespaced tree of types
 #[derive(Default)]
 pub struct TypeTree {
-    types: Vec<Type>,
-    namespaces: TypeNamespaces,
+    pub types: Vec<Type>,
+    pub namespaces: TypeNamespaces,
+    pub foundation: bool,
```
rylev, line 11: nit: `include_foundation` might be a more descriptive name

crates/gen/src/types/namespace.rs
```diff
@@ -5,6 +5,22 @@ use squote::{quote, TokenStream};
 use std::iter::FromIterator;
 
 pub fn to_namespace_tokens(destination: &str, source: &str) -> TokenStream {
+    if destination == source {
+        return TokenStream::new();
+    }
+
+    // if destination == "Windows.Foundation" {
```
rylev, line 12: Is there a reason this are commented out?

crates/gen/src/type_namespaces.rs
rylev: Also I thought we wanted to export foundation as is (to mirror the same hierarchry). So this should be `pub use::winrt::foundation` without the `*`.

kennykerr: There's one foundation sub-namespace that's not pulled into winrt but I don't suppose there's much value in excluding it at this point.

PR #303, opened 2020-08-04
Include basic math operations for Numerics (Vector and Matrix)

robmikh:
This change takes the general idea from the [minesweeper-rs numerics helper](https://github.com/robmikh/minesweeper-rs/blob/master/bindings/src/numerics.rs) and includes those implementations within winrt-rs. Math operations are supported for both the struct and for a borrowed struct. This addresses oart of #79.

Plane, Rational, and Quaternion haven't been looked at yet. Additionally, only the basic operations plus some helpers for Vector* types are included this time around. I thought it was important to get these in since you can no longer define these math operations yourself like Minesweeper does.

The tests are a bit crude, I wanted to get more of the functionality before I started porting the OS versions of the numerics tests. Although I'm open to doing it now if we think that would be better.


robmikh:
Also we may want to squash this one, it contains history from when I was generating the implementations.

kennykerr:
Thanks again Rob. Note that since the numerics are now included in the winrt crate, you no longer need to use the code generator to inject extensions like this. Instead, you can simply create a numerics.rs file in the src folder and add the operations there. This is much like what I did for TimeSpan. It is still generated but the extensions for Duration support are simply provided in the src folder.

https://github.com/microsoft/winrt-rs/blob/master/src/time_span.rs

robmikh:
Ah yeah, sorry. I got mixed up since the definitions of the structs were still being generated. I'll get that feedback incorporated, thanks!

kennykerr:
> The tests are a bit crude, I wanted to get more of the functionality before I started porting the OS versions of the numerics tests. Although I'm open to doing it now if we think that would be better.

Happy to take this as a start, but yes porting the OS tests would be great. That will also help inform what's missing.

src/numerics.rs
```diff
@@ -0,0 +1,1270 @@
+use winrt::foundation::numerics::{Matrix3x2, Matrix4x4, Vector2, Vector3, Vector4};
+
+// Vector2
```
rylev, line 3: I know you moved from the proc_macro to this, but this seems like a perfect use for `macro_rules!`. I think that would make this much more maintainable.

kennykerr: Oops, I didn't see your comments until after I completed the PR.

kennykerr: Now that the code is not being generated through the `build`/`import` macros, it does make sense to compact the definitions with `macro_rules` where practical, but we should keep it as simple as possible i.e. it doesn't need to be driven by metadata.

PR #307, opened 2020-08-06
Document why code is safe in Array

rylev:
@timClicks gave a [great talk on safety in Rust](https://youtu.be/9M0NQI5Cp2c?t=1294). He used this crate as an example. Fortunately the code that Tim looked at seems to be safe. One piece of feedback was better documentation explaining why the code was indeed safe. Here's better docs for that particular code. We should make an effort to do the same across the entire project. 

There is however one change I made that would lead to unsafety although it would be unlikely to introduce - overflows in `Array::with_len`.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -24,15 +24,24 @@ impl<T: RuntimeType> Array<T> {
     /// Creates an array of the given length with default values.
     pub fn with_len(len: usize) -> Self {
         assert!(len < std::u32::MAX as usize);
+        let bytes_amount = len
+            .checked_mul(std::mem::size_of::<T>())
+            .expect("Attempted to allocate too large an Array");
 
         // WinRT arrays must be allocated with CoTaskMemAlloc.
-        let data = unsafe { CoTaskMemAlloc(len * std::mem::size_of::<T>()) as *mut T };
+        // SAFETY: the call to CoTaskMemAlloc is safe to perform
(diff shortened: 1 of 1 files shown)


src/array.rs
```diff
@@ -24,15 +24,24 @@ impl<T: RuntimeType> Array<T> {
     /// Creates an array of the given length with default values.
     pub fn with_len(len: usize) -> Self {
         assert!(len < std::u32::MAX as usize);
```
tim-weis, line 26: Unrelated to this pull request, but I need to ask: Why is this `<` rather than `<=`? It is my understanding, that `std::u32::MAX` is a valid argument to [CoTaskMemAlloc](https://docs.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-cotaskmemalloc).

Is there something I'm missing?

rylev: @kennykerr I believe @tim-weis has a point here. Can you confirm?

kennykerr: Technically yes. In practice arrays aren't used for large buffers because some language projections would have a real hard time with them (as they don't marshal them efficiently) and there are more efficient ways to share large buffers.

rylev: Alright we can make the change just to be "more correct", but it's not a priority. PRs welcome 😄

PR #310, opened 2020-08-10
Suggest `cargo winrt install` when missing dependencies

zachlute:
If you just do `cargo build` on a WinRT project (like `winrt-rs`, for example), you'll get error messages about missing nuget or dependency folders but no indication why.

Also fixes a minor slash-escaping issue in an error message.


zachlute:
Ask me how I know people can be confused by this! ;)

PR #312, opened 2020-08-12
Add 'version' subcommand

### LINKED ISSUE #296: "cargo winrt --version" no longer displays version information
Previously, I had *cargo-winrt* version 0.7.0 installed. Running `cargo winrt --version` on the command line produced the following output:

>     cargo-winrt 0.7.0

After installing the latest version (0.7.2) of *cargo-winrt* by executing `cargo install cargo-winrt` produces the following output instead:

>     error: missing subcommand

(followed by the generic help text, that does include the `-v` and `--version` options). Was the option `--version` removed on purpose, or this is a regression?

zachlute:
This makes --version and its ilk work again.

This change matches the `cargo` behavior of having -V/--version switches and also a `version` subcommand, and also happened to be the easiest way to implement it.

Fixes #296.

### DIFF
--- a/crates/cargo-winrt/src/main.rs
+++ b/crates/cargo-winrt/src/main.rs
@@ -13,6 +13,8 @@ use std::io::Read;
 use std::path::{Path, PathBuf};
 use std::time::{Duration, Instant};
 
+const VERSION: &'static str = env!("CARGO_PKG_VERSION");
+
 fn main() {
     if let Err(i) = run() {
         std::process::exit(i);
@@ -46,6 +48,7 @@ fn run() -> Result<(), i32> {
         Subcommand::Run(r) => r.perform(),
         Subcommand::Build(b) => b.perform(),
         Subcommand::Help(h) => h.perform(),
+        Subcommand::Version(v) => v.perform(),
     };
     if let Err(ref e) = result {
         cmd_err!("{}", e);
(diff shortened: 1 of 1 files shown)


crates/cargo-winrt/src/main.rs
```diff
@@ -368,6 +379,27 @@ USAGE:
     }
 }
 
+#[derive(Debug)]
+struct Version {}
```
rylev, line 383: perhaps `struct Version;` is the better syntax?

crates/cargo-winrt/src/main.rs
```diff
@@ -368,6 +379,27 @@ USAGE:
...
+
+impl Version {
+    fn perform(&self) -> anyhow::Result<()> {
+        println!("cargo-winrt {}", VERSION);
+        Ok(())
+    }
+
+    fn print_help() {
+        println!(
+            r#"
```
rylev, line 393: Just thinking out loud, I wonder if it would be a good idea to put all these help messages in text files and include them using `include!`...

crates/cargo-winrt/src/main.rs
zachlute: Sure!

crates/cargo-winrt/src/main.rs
zachlute: It's definitely ugly the way it is, but I also think there are advantages to keeping the help text with the command, in that you're more likely to see it and update it when updating the command. Though I suspect any conversation about command ergonomics is all a little bit pedantic, because it seems like the number of commands that would theoretically be added once things reach a steady state is pretty low?

crates/cargo-winrt/src/main.rs
```diff
@@ -86,6 +91,10 @@ fn parse_args() -> Result<Subcommand, ArgsError> {
             let subcommand = args.free()?.first().map(String::to_owned);
             Subcommand::Help(Help { subcommand })
         }
+        Some("version") => {
+            args.finish()?;
+            Subcommand::Version(Version {})
```
rylev, line 96: This should just be `Subcommand::Version(Version)`. Same with line 70

Issue #321, opened 2020-08-24
The crate no longer builds when used as a dependency due to missing WinRT dependencies


rylev:
Because of https://github.com/microsoft/winrt-rs/pull/297, WinRT-rs no longer builds when used as a dependency, because of missing WinRT dependencies. In order for the crate to build, the WinRT dependencies must be present. This requires the end user to specify the WinRT dependencies of WinRT-rs in its Cargo.toml file. This is obviously not ideal. 

There's no obvious way to fix this. We could have WinRT-rs install its dependencies inside of its build script. Rust's "best practices" discourage fetching resources from the internet during a build.rs script so we'll probably want to avoid that. We could, however, include the WinRT dependencies as submodules in WinRT code base and reference them as local dependencies. This would allow installation without accessing the internet.

kennykerr:
Once we have #320 we can simply copy the winmd files from the SDK nuget package into a subfolder in this repo. The build can then directly access them there instead of having to download them.

PR #325, opened 2020-08-28
Local winmds

### LINKED ISSUE #320: Allow specifying direct winmd files as dependencies
Currently when specifying a WinRT dependency, the dependency must be structured like a well-defined NuGet based WinRT dependency. We should also allow pointing an arbitrary folder that has winmd files. This will make packaging and testing.

rylev:
Fixes #320 and #321 

The following needs to be decided before we can merge this:

#### Who moves the winmd files into place?
Something needs to move the winmd files from the `winmds` folder to the well known place inside of the target folder. This usually falls on the shoulders of `cargo winrt`. We _could_ require that any user who wants to build WinRT-rs have cargo winrt installed but this would include all downsteam users of the crate even if the user is only transitively depending on WinRT. Another solution is add library support for `cargo-winrt` which would allow us to take it as a direct build dependency on WinRT-rs. Then inside of WinRT-rs's build.rs file, we can directly invoke cargo-winrt as a build library.

#### Cargo.toml syntax
We need to decide how users declare in their Cargo.toml that a local dependency is not laid out like a NuGet dependency but is rather just a flat collection of winmds. Although I did not take this approach in the code, I would like to see the following:

```toml
"Microsoft.Windows.SDK.Contracts" = { path = "./winmds/Contracts", blob = true }
```
This is the same syntax as a local NuGet dependency just with the added `blob = true` to indicate that the dependency is a blob rather than a well-formed NuGet dependency.


Cargo.toml
```diff
@@ -32,5 +32,5 @@ default-target = "x86_64-pc-windows-msvc"
 targets = ["x86_64-pc-windows-msvc", "i686-pc-windows-msvc"]
 
 [package.metadata.winrt.dependencies]
-"Microsoft.Windows.SDK.Contracts" = "10.0.19041.1"
-"KennyKerr.Windows.TestWinRT" = "1.0.21"
+"Microsoft.Windows.SDK.Contracts" = { blob = "./winmds/Contracts" }
+"KennyKerr.Windows.TestWinRT" = { blob = "./winmds/TestComponent" }
```
kennykerr, line 36: We should only have to do this for the `Windows.Foundation.*.winmd` files in the SDK nuget package. Those are the only ones that the `winrt-rs` crate needs a dependency on. The tests can still rely on nuget.

kennykerr:
> Something needs to move the winmd files from the winmds folder to the well known place inside of the target folder. 

Technically, the winmd files should not have to move at all. Only the DLLs (if any) need to be be copied.

> Cargo.toml syntax

I'd like to emulate the Cargo dependency syntax. I think the the `{ path = "folder/sub-folder" } ` syntax should suffice to indicate that it is a folder of winmd files. If we need to support local nuget layout we could have a `nuget = true` option. I'd just like to infer as much as possible and it's not obvious what `blob` means.

rylev:
Ok this makes sense. We were relying on the winmd files being in a well-known place, but you are right that we can now use the information provided by the Cargo.toml file to figure this out making moving the winmd files not necessary.

I'm not sure how I feel about opting into NuGet package form for local dependencies. I agree that this is likely the more common use case, but this behavior might be slightly surprising as for git and named dependencies we expect the dependency to be in NuGet form but for local, the user is required to opt-in.

kennykerr:
Going to take this on myself.

rylev:
Sorry for the delay here. Happy to help with code review and such!

Issue #81, opened 2020-04-27
Defining and Implementing Win32/COM/WinRT types in Rust


kennykerr:
Authoring support is part of the next major milestone - stay tuned!

kennykerr:
Rust presents both unique challenges and unique opportunities when it comes to creating a language projection for the Windows Runtime. Defining and implementing COM and WinRT types in Rust highlights how this language’s rich metaprogramming facilities allow developers rather ingenious flexibility to hook compilation for both importing and exporting type information. 

Today, Rust/WinRT lets you import WinRT types from metadata. The resulting imported types are projected as Rust types that are callable in an idiomatic way. This however only covers consumption of types. Now it's time to explore how Rust/WinRT is being extended to allow COM and WinRT types to be defined in Rust, for local consumption and optionally for export as well, and to allow Rust implementations of both imported or locally defined types. 

The problem can be broken down as follows:

* Declare WinRT and COM types in Rust 
* Implement WinRT and COM classes in Rust 
* Export the metadata representation of those types as a winmd file

Today, a Rust project imports types using either the `winrt::import` or `winrt::build` macros. Either way, a set of winmd files are gathered and the aggregate of all types are available for import into the Rust type system. This acts as a sort of cache of metadata that is at the Rust project’s disposal. This is important because the `import`/`build` macros don’t know ahead of time what types might need to be implemented by the project. Breaking down the problem like this is useful because you can quickly see how the solution comes together. Locally declared WinRT and COM types are immediately exported to a winmd file that is then added to the metadata cache that is then available to any COM and WinRT implementations within the Rust project. The winmd file can then be shared with other projects as needed.

To complement the winrt `import`/`build` macros, Rust/WinRT adds the `winrt::implements` attribute macro to implement any previously declared COM or WinRT classes and interfaces. This macro draws information from the metadata cache to fill in the scaffolding and boilerplate code necessary to implement the desired types in Rust. 

Here is an example of how the build macro might be used (and updated) to both import existing type information and declare new project-specific types.

```rust
winrt::build! {
    [import]
    nuget = Microsoft.Windows.SDK.Contracts
 
    [filter]
    windows::foundation::*
    windows::data::xml::dom::*
 
    [export]
    pub mod microsoft {
        pub mod windows {
            pub struct StructType { x: i32, y: i32 };
 
            pub interface IInterfaceType {
                fn method(&self) -> Result<StructType>;
            }
 
            pub class ClassType : IInterfaceType;
        }
    }
[...]

PR #329, opened 2020-09-22
Build macro refactoring

kennykerr:
A little cleanup to make room for the `implements` macro for #81.

The `ImportMacro`, originally used to describe the old `import` macro, has been renamed `BuildMacro` since it now only describes the `build` macro. All of the functionality of the build macro has been moved to the `build` module so that an `implements` macro can then be implemented in its own source file. The respective proc macros themselves cannot be defined in the sub modules as Rust doesn't yet support this.

### DIFF
--- a/crates/gen/src/type_reader.rs
+++ b/crates/gen/src/type_reader.rs
@@ -23,6 +23,25 @@ impl TypeReader {
         Self::new(crate::load_winmd::from_os())
     }
 
+    pub fn from_build() -> &'static Self {
+        use std::{mem::MaybeUninit, sync::Once};
+        static ONCE: Once = Once::new();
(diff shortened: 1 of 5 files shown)


crates/gen/src/type_reader.rs
```diff
@@ -23,6 +23,25 @@ impl TypeReader {
         Self::new(crate::load_winmd::from_os())
     }
 
+    pub fn from_build() -> &'static Self {
+        use std::{mem::MaybeUninit, sync::Once};
+        static ONCE: Once = Once::new();
+        static mut VALUE: MaybeUninit<TypeReader> = MaybeUninit::uninit();
```
rylev, line 29: I think a more common way to do this is with an Option type. This removes the last line of unsafe in favor of a checked unwrap. Slightly less efficient, but also safer. 🤷‍♀️

PR #333, opened 2020-10-09
Simpler and faster

kennykerr:
Time for some refactoring to simplify and get the complexity under control. This accounts for most of the churn here. 

I also optimized the code gen some by simplifying the limits/staging/tree pipeline and avoiding the token caching. Since this phase happens on a single thread, the caching actually makes it slower. Instead, tokens are generated in parallel without caching. This is about 30% faster.


.github/workflows/build.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         components: rustfmt
 
     - name: cargo winrt
-      run: cargo install --path ./crates/cargo-winrt
+      run: cargo install --path ./crates/cargo
```
rylev, line 28: FWIW I don't love calling this directory `cargo`. The crate is called `cargo-winrt`, and while it's marginally longer to have to type the additional 6 characters, it's much clearer what the directory contains for the uninitiated.

crates/gen/Cargo.toml
```diff
@@ -1,17 +1,16 @@
 [package]
-name = "winrt_gen"
+name = "winrt-gen"
```
rylev, line 2: I believe this means we'll have to create a new crate on Crates.io. Is that what we want?

crates/gen/macros/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
-name = "winrt_gen_macros"
+name = "winrt-gen-macros"
```
rylev, line 2: Same as above ^^

crates/gen/src/gen/interface.rs
```diff
@@ -50,34 +47,41 @@ impl Interface {
...
+    pub fn default_interface(&self) -> &gen::RequiredInterface {
+        self.interfaces
             .iter()
             .find(|i| i.kind == InterfaceKind::Default)
-            .unwrap();
+            .unwrap()
+    }
+
+    pub fn gen_vtable_initializer(&self) -> TokenStream {
+        panic!();
```
rylev, line 58: nit: "For work-in-progress" items: the `todo!()` macro is much more descriptive than using an empty `panic!`.

crates/gen/src/gen/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+mod namespace;
+mod param;
+mod required_interface;
+mod r#struct;
+mod r#type;
+mod type_guid;
+mod type_kind;
+mod type_name;
+
+pub use class::*;
```
rylev, line 18: I love the move to a `gen` submodule, but I also really wish we didn't rely so much on `*` glob imports. It's not very idiomatic to do `use submodule::*` and bring _everything_ into scope.

rylev (commented):
Looks good. Just a few comments.

.github/workflows/build.yml
kennykerr: Just trying to bring some consistency. The crates folder contains a number of crates that all include `winrt` in their names but this was the only one that spelled it out. Same of the other comments below. Some crates used `-` and others `_`.

crates/gen/src/gen/mod.rs
kennykerr: There's a convention here that considerably reduces the reliance on glob imports and makes the code more readable but I may have to write that down to explain.

rylev: Ah I think I saw this, where you're always prefixing with `gen`. I like that. I still see plenty of `use crate::*` that I'm not too thrilled about, but I like the strategy of using the `gen` prefix`

.github/workflows/build.yml
rylev: The directory names and the crate names don't _need_ to align, but it can help. I would personally prefer that the directories in `crates` were all named exactly like the crates themselves. It's unfortunate about the `-` vs `_`. We can rename crates to be consistent, but it's a bit of a shame we'll have to deprecate the old crates which use `_`. This isn't a huge deal though since they're just implementation details...

crates/gen/Cargo.toml
ctaggart: Any reason to switch to dashes? I settled on underscores for the package names, to match what is required in the `use` statements.

kennykerr: We had a mix of dashes and underscores - I just settled on dashes for consistency and because I wasn't sure whether the name of the `cargo-winrt` crate mattered. The others don't.

PR #334, opened 2020-10-09
Allow winrt-rs to build without first calling `cargo winrt install`

kennykerr:
Simpler fix for #321. Fixing #320 is less urgent. 

This change simply lets the winrt-rs crate build without calling `cargo winrt install` first.


rylev (commented):
I believe this fixes this issue only for local development. To fix for others who are using WinRT as a dependency, you'll have to explicitly include the WinMD files. https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

PR #344, opened 2020-10-20
Add TryFrom<HString> for String 

### LINKED ISSUE #227: From<HString> for String can panic
`String` in Rust requires that the underlying buffer be a valid UTF-8 string. `HSTRING` in Windows does not require that the underlying buffer be a valid UTF-16 string--it can be any sequence of 16-bit integers, including unmatched surrogates. Therefore, it's not always possible to translate an `HSTRING` into a `String` (but it is, of course, always valid to go the other direction).

However, we have:

```rust
impl<'a> From<&'a HString> for String {
    fn from(hstring: &HString) -> Self {
        String::from_utf16(hstring.as_wide()).unwrap()
    }
}
```

This will panic if the `HSTRING` contains non-UTF-16 characters.

This can happen, for example, if a user has a file with a non-UTF-16 name and calls one of the WinRT directory enumeration APIs.

It seems that the fix for this would be to implement `From<HString>` for `OsString` instead, as `OsString` can encode any Windows wide string, not just valid Unicode strings. The user can then try to convert it to a `String` or `&str` dynamically, falling back to some reasonable behavior (or panicking!) as appropriate.

This will make using `HString` more annoying for callers, but there's no real alternative I can see.

adumbidiot:
Closes #227.

I removed the implementation of `From<HString>` for `String` and replaced it with a `TryFrom<HString>`, modifying the existing tests accordingly. In order to add a new basic test for this new `TryFrom` impl, I created a constructor called `from_wide` that creates a `HString` from a slice of `u16`s. To reduce code duplication, I extracted most of the logic from `From<&str>` and put it in a new private constructor. I added the `to_string_lossy` method to `HString` to allow easier lossy conversions between `HString` and `String`, as well as a basic test for this functionality.

I refrained from adding a fallible `to_string` method as that would conflict with the `Display` trait that is implemented for `HString`. I felt the name `to_str` also wouldn't make sense here as this method would be forced to return a `String` as the internal data is not utf8. I was also unsure of the importance of such a method as the new `TryFrom<HString>` impl now exists, so I excluded it from this PR.

### DIFF
--- a/src/hstring.rs
+++ b/src/hstring.rs
@@ -47,6 +47,11 @@ impl HString {
         unsafe { std::slice::from_raw_parts((*header).data, (*header).len as usize) }
     }
 
+    /// Create a HString from a slice of 16 bit characters (wchars).
+    pub fn from_wide(value: &[u16]) -> HString {
+        HString::from_wide_iter(value.iter().copied(), value.len() as u32)
+    }
+
     /// Clear the contents of the string and free the memory if `self` holds the
     /// last reference to the string data.
     pub fn clear(&mut self) {
@@ -66,6 +71,29 @@ impl HString {
         }
 
         self.ptr = std::ptr::null_mut();
+    }
+}
+
+impl HString {
+    fn from_wide_iter<I: Iterator<Item = u16>>(iter: I, len: u32) -> HString {
+        if len == 0 {
+            return HString::new();
+        }
+
+        let mut ptr = Header::alloc(len);
+
+        // Place each utf-16 character into the buffer and
+        // increase len as we go along.
+        for (index, wide) in iter.enumerate() {
+            unsafe {
+                std::ptr::write((*ptr).data.add(index), wide);
+                (*ptr).len = index as u32 + 1;
+            }
+        }
+
+        // Write a 0 byte to the end of the buffer.
+        unsafe { std::ptr::write((*ptr).data.offset((*ptr).len as isize), 0) };
+        Self { ptr }
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/hstring.rs
```diff
@@ -67,6 +79,29 @@ impl HString {
     }
 }
 
+impl HString {
```
kennykerr, line 82: Why not just combine this in the previous `impl`?

rylev (changes requested):
Looking good, but there's a safety issue.

rylev:
Thanks for the contribution 🎉🦀!

src/hstring.rs
adumbidiot: I thought it was best to separate for readability since this method is a private constructor and others are all public. I'll combine the impl blocks.

src/hstring.rs
```diff
@@ -67,6 +79,29 @@ impl HString {
...
+    fn from_wide_iter<I: Iterator<Item = u16>>(iter: I, len: u32) -> HString {
```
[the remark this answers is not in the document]
adumbidiot: Oops, nice catch. I think I'll just make the function unsafe.

src/hstring.rs
```diff
@@ -65,6 +77,27 @@ impl HString {
 
         self.ptr = std::ptr::null_mut();
     }
+
+    /// # Safety
+    /// len must not be less than the number of items in the iterator.
+    unsafe fn from_wide_iter<I: Iterator<Item = u16>>(iter: I, len: u32) -> HString {
+        if len == 0 {
```
rylev, line 84: Since the safety invariant is easy to test for, I think we should have an `assert_debug!` that checks that the invariant is respected. This would only get triggered if we've made a mistake somewhere in winrt-rs (since this is a private function), but I think the small runtime hit on debug builds is worth it.

adumbidiot: Ok sounds good I'll put a `debug_assert!((index as u32) < len);` in the loop. 
Also, shouldn't `Header::allocate` allocate `len + 1` for the null terminator?

kennykerr: `Header::alloc` already accounts for the terminator.

rylev:
@adumbidiot I'm not sure why you made `from_wide_iter` safe as it's still not safe. If you want it to be safe you need to use `assert!` which will panic in both debug and release mode unlike `debug_assert!` which only panics in debug. Otherwise you can just leave it unsafe. If you do make it safe by panicking, please document the conditions when it panics.

adumbidiot:
Sorry I think i misunderstood your feedback. I'll make that function unsafe again.

kennykerr (approved):
Thanks!

rylev (approved):

PR #349, opened 2020-10-27
Copy dll files instead of using symlinks

### LINKED ISSUE #347: Avoid use of symlinks in cargo winrt
This requires the use of "Developer Mode" on Windows and should thus be avoided.

danielframpton:
Fixes #347


danielframpton:
This change simple uses a copy instead of symlink; if we we were also relying on the symlink to transparently update then we need to do more. But if that is the case we should probably also add some tests for that scenario?

rylev (approved):
This is probably an improvement over what we currently have.

kennykerr:
Thanks Daniel!

Issue #300, opened 2020-08-03
Some factory cache statics are unused


kennykerr:
```rust
warning: static is never used: `SHARED`
      --> C:\Users\ryanl\Code\stress-rs\target\debug\build\bindings-12d29370494acda0\out/winrt.rs:574672:21
       |
574672 | /                     static mut SHARED: ::winrt::FactoryCache<
574673 | |                         WiFiDirectDevice,
574674 | |                         IWiFiDirectDeviceStatics2,
574675 | |                     > = ::winrt::FactoryCache::new();
       | |_____________________________________________________^
```

It has to do with static/constructor overloads that don't quite map correctly to Rust (since it lacks overloads).

PR #362, opened 2020-11-24
Improved method collision management

### LINKED ISSUE #280: EmailAttachment's IEmailAttachmentFactory2 is not visible (Windows.ApplicationModel.Email)
This is because IEmailAttachmentFactory and IEmailAttachmentFactory2 both provide Create methods and the one is dropped as a dupe, but we should provide a simpler way to get at this constructor.

kennykerr:
Fixes #280 and #300 

Both of these issues stem from the fact that Rust lacks support for overloads that WinRT assumes languages can support. With this update, Rust/WinRT properly handles both cases where constructors and statics may be overloaded. There is another case, a bit more obscure, that should also be fixed but I need to add a bit more test coverage to make that possible to confirm. 

With all the name mangling, it becomes increasingly difficult to use the original interface method names as the name of a given method may differ depending on the context e.g. whether it is the immediate interface or inherited (required). I've addressed this by further relying on vtable ordinals/offset since they're stable for all method generation. This also simplifies code gen since all method generation is handled centrally. 

I am also making progress on improving the implementation support that I recently debuted (#81). There's still a long way to go but this update also smooths some wrinkles and centralizes the vtable generation to avoid generating them twice. Finally, I'm moving away from `syn::UseTree` for the `implement` macro's input since that won't support generic interfaces. I now have my own `ImplementTree` which doesn't quite support generics yet but will soon.


crates/gen/src/method.rs
```diff
@@ -135,116 +137,101 @@ impl Method {
...
+                        #vcall
+                    }
+                }
+            },
+            InterfaceKind::NonDefault | InterfaceKind::Overrides => {
+                let interface = interface.gen();
+                quote! {
+                    pub fn #method_name<#constraints>(&self, #params) -> ::winrt::Result<#return_type_tokens> {
+                        unsafe {
+                            let this = &::winrt::Interface::cast::<#interface>(self).unwrap();
```
rylev, line 213: nit: I believe this call is safe and so it doesn't need to be in the unsafe block.

crates/gen/src/method.rs
```diff
@@ -135,116 +137,101 @@ impl Method {
...
-                Self::#interface(|f| f.#method_name(#args))
-            }
-        }
-    }
+        let vtable_offset = Literal::u32_unsuffixed(self.vtable_offset);
 
-    pub fn gen_composable(&self, interface: &RequiredInterface) -> TokenStream {
-        let method_name = format_ident(&self.name);
-        let interface = format_ident(&interface.name.name);
+        let vcall = if let Some(return_type) = &self.return_type {
```
rylev, line 177: Same thing as before, this might be clearer as a match:

```rust
match &self.return_type {
    Some(return_type) if return_type.array => {}
    Some(return_type) => {}
    None => {}
}
```

crates/gen/src/method.rs
```diff
@@ -8,6 +8,7 @@ pub struct Method {
     pub params: Vec<Param>,
     pub return_type: Option<Param>,
     pub vtable_offset: u32,
+    pub overload: u32,
```
rylev, line 11: Out of curiousity, why did you choose `u32` instead of `u8` or `u16` for these? Presumably we'll never have more than 256 or certainly not 65536 overloads or virtual methods, no?

crates/gen/src/required_interface.rs
```diff
@@ -9,6 +9,26 @@ pub struct RequiredInterface {
...
 
+impl PartialEq for RequiredInterface {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name
+    }
+}
+
+impl Eq for RequiredInterface {}
+
+impl PartialOrd for RequiredInterface {
```
rylev, line 20: You should be able to derive all but `PartialEq` no?

crates/gen/src/required_interface.rs
```diff
@@ -188,39 +185,29 @@ pub fn add_dependencies(
...
-    for method in methods {
-        if names.contains(&method.name) {
-            method.name = format!("{}2", method.name);
-        } else {
-            names.insert(&method.name);
+    let mut count = BTreeMap::new();
+
+    for interface in interfaces {
+        for method in &mut interface.methods {
+            let count = count.entry(method.name.clone()).or_insert(0);
```
rylev, line 208: I believe you should be able to use `&method.name` instead of cloning here.

rylev (approved):
👍

crates/gen/src/required_interface.rs
kennykerr: The `derive` macro is too aggressive. Even though the custom `PartialEq` implementation is provided and deriving `Eq` should be simple, it attempts to validate that all fields are `Eq` when they don't need to be.

crates/gen/src/method.rs
kennykerr: I just tend to use `u32` for counters. It's not likely there would be anywhere near 256 overloads but there could well be a thousands offsets.

crates/gen/src/method.rs
kennykerr: I do like that but it doesn't let me share code gen for return types.

PR #363, opened 2020-11-24
Use full paths when implementing interfaces

### LINKED ISSUE #361: Issues attempting to implement IApplicationOverrides
Hello,

I am attempting to implement IApplicationOverrides in an attempt to write a UWP xaml app without requiring the use of xaml islands. I'm basing my attempt off of the [code here](https://github.com/mgnsm/CodeSamples/tree/master/MinimalWRLSampleApp), specifically how to manually generate an Application class, not the old WRL machinery. My attempt at replicating that setup is [here](https://gist.github.com/DJankauskas/a8ec5c771b21e11e838055fa8871aca2).

However, I'm running into these three errors:
```
error[E0433]: failed to resolve: use of undeclared crate or module `windows`
 --> src\main.rs:6:1
  |
6 | #[implement(windows::ui::xaml::IApplicationOverrides)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `windows`
  |
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: there are too many leading `super` keywords
 --> src\main.rs:6:1
  |
6 | #[implement(windows::ui::xaml::IApplicationOverrides)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ there are too many leading `super` keywords
  |
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0308]: mismatched types
 --> src\main.rs:6:1
  |
6 | #[implement(windows::ui::xaml::IApplicationOverrides)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `WindowCreatedEventArgs`, found reference
  |
  = note: expected struct `WindowCreatedEventArgs`
          found reference `&Option<WindowCreatedEventArgs>`
  = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to 3 previous errors; 1 warning emitted
```

Sorry for the code and error dump: if some other format or information would be helpful or would suit you better, I'd be happy to provide it.

The first error has to do with me moving winrt code generation to a `bindings` crate, which I can undo (although I would love suggestions on how to implement interfaces while retaining it), but the second and third errors are completely opaque to me, as they seem to revolve around proc macro output. Am I implementing something incorrectly, or is this a bug with the new `implement` feature?

Thanks again for your help!

kennykerr:
Fixes #361 

With this change, you can now implement interfaces that refer to types in other namespaces/modules. Here's an example of implementing Xaml's `IApplicationOverrides` and `IApplicationOverrides2`.

```rust
use bindings::*;
use winrt::Result;
use windows::application_model::activation;

fn main() -> Result<()> {
    let app = App{};
    let overrides: windows::ui::xaml::IApplicationOverrides = app.into();
    overrides.on_activated(None)?;

    Ok(())
}

#[winrt::implement(windows::ui::xaml::{IApplicationOverrides, IApplicationOverrides2})]
struct App {
}

impl App {
    fn on_activated(&self, _args: &Option<activation::IActivatedEventArgs>) -> Result<()> {
        println!("on_activated");
        Ok(())
    }

    fn on_launched(&self, _args: &Option<activation::LaunchActivatedEventArgs>) -> Result<()> {
        Ok(())
    }

    fn on_file_activated(&self, _args: &Option<activation::FileActivatedEventArgs>) -> Result<()> {
        Ok(())
    }

    fn on_search_activated(&self, _args: &Option<activation::SearchActivatedEventArgs>) -> Result<()> {
        Ok(())
    }

    fn on_share_target_activated(&self, _args: &Option<activation::ShareTargetActivatedEventArgs>) -> Result<()> {
        Ok(())
    }

    fn on_file_open_picker_activated(&self, _args: &Option<activation::FileOpenPickerActivatedEventArgs>) -> Result<()> {
        Ok(())
    }

    fn on_file_save_picker_activated(&self, _args: &Option<activation::FileSavePickerActivatedEventArgs>) -> Result<()> {
        Ok(())
    }

    fn on_cached_file_updater_activated(&self, _args: &Option<activation::CachedFileUpdaterActivatedEventArgs>) -> Result<()> {
        Ok(())
    }

    fn on_window_created(&self, _args: &Option<windows::ui::xaml::WindowCreatedEventArgs>) -> Result<()> {
        Ok(())
    }

    fn on_background_activated(&self, _args: &Option<activation::BackgroundActivatedEventArgs>) -> Result<()> {
        Ok(())
    }
}
```

A few things to note about this example.

1. The `implement` macro has no way of knowing the name of the bindings crate so you need to use a `use` declaration to bring its contents into scope e.g. `use bindings::*;`

2. Although you can freely use additional `use` declarations to simplify the code you need write, you must use full type paths in the `implement` macro because again, the `implement` macro has no way of knowing about these `use` declarations and can only see its input token stream.

### DIFF
--- a/crates/gen/src/method.rs
+++ b/crates/gen/src/method.rs
@@ -235,8 +235,11 @@ impl Method {
         }
     }
 
-    pub fn gen_upcall(&self, inner: TokenStream) -> TokenStream {
-        let invoke_args = self.params.iter().map(|param| param.gen_invoke_arg());
+    pub fn gen_upcall(&self, inner: TokenStream, relative: bool) -> TokenStream {
+        let invoke_args = self
+            .params
+            .iter()
+            .map(|param| param.gen_invoke_arg(relative));
 
         match &self.return_type {
             Some(return_type) if return_type.array => {
--- a/crates/gen/src/delegate.rs
+++ b/crates/gen/src/delegate.rs
@@ -46,7 +46,7 @@ impl Delegate {
             self.name.gen_signature(&format!("{{{:#?}}}", &self.guid))
         };
 
-        let invoke_upcall = self.method.gen_upcall(quote! { ((*this).invoke) });
+        let invoke_upcall = self.method.gen_upcall(quote! { ((*this).invoke) }, true);
 
         quote! {
             #[repr(transparent)]
--- a/crates/gen/src/param.rs
+++ b/crates/gen/src/param.rs
@@ -171,9 +171,14 @@ impl Param {
         }
     }
 
-    pub fn gen_invoke_arg(&self) -> TokenStream {
+    pub fn gen_invoke_arg(&self, relative: bool) -> TokenStream {
         let name = format_ident(&self.name);
-        let kind = self.kind.gen();
+
+        let kind = if relative {
+            self.kind.gen()
+        } else {
+            self.kind.gen_full()
+        };
 
         // TODO: This compiles but doesn't property handle delegates with array parameters.
         // https://github.com/microsoft/winrt-rs/issues/212
--- a/crates/gen/src/type_kind.rs
+++ b/crates/gen/src/type_kind.rs
@@ -240,6 +240,35 @@ impl TypeKind {
         }
     }
 
+    pub fn gen_full(&self) -> TokenStream {
+        match self {
+            Self::Bool => quote! { bool },
+            Self::Char => quote! { u16 },
+            Self::I8 => quote! { i8 },
+            Self::U8 => quote! { u8 },
+            Self::I16 => quote! { i16 },
+            Self::U16 => quote! { u16 },
+            Self::I32 => quote! { i32 },
+            Self::U32 => quote! { u32 },
+            Self::I64 => quote! { i64 },
+            Self::U64 => quote! { u64 },
+            Self::F32 => quote! { f32 },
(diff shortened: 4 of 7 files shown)


crates/gen/src/method.rs
```diff
@@ -235,8 +235,11 @@ impl Method {
         }
     }
 
-    pub fn gen_upcall(&self, inner: TokenStream) -> TokenStream {
-        let invoke_args = self.params.iter().map(|param| param.gen_invoke_arg());
+    pub fn gen_upcall(&self, inner: TokenStream, relative: bool) -> TokenStream {
```
rylev, line 238: Nit: you might want to consider changing the bool to an enum. Something like this:
```rust
enum CallPosition {
  Relative,
  Absolute
}
```
This can make the code easier to read.

kennykerr: Yep, good call. Was meaning to do something like this.

PR #370, opened 2020-12-01
Direct dependency management and removal of "cargo winrt" tool

### LINKED ISSUE #320: Allow specifying direct winmd files as dependencies

### LINKED ISSUE #351: Remove the need for the cargo-winrt tool
@rylev and I have been talking about the build and we think moving away from `cargo winrt` altogether would provide the best user experience. We'd still provide a tool for optionally downloading WinRT dependencies, but it would not be needed for the build. Projects can just store and manage their dependencies directly.

https://github.com/microsoft/winrt-rs/issues/346#issuecomment-717558621

### LINKED ISSUE #368: Failure to build a project using winrt-rs in a workspace
Example repo: https://github.com/Enigmatrix/winrt_workspace_test

Building the above repo gives an error when importing any type. e.g. in the case above,
image
Using commit https://github.com/microsoft/winrt-rs/commit/38b7bea9a17930e55145a2b577b956e8193521ad

kennykerr:
This update removes the `cargo winrt` prerequisite for fetching WinRT dependencies. You can now manage this directly with the `.windows` folder in the root of the workspace much like you would manage settings for Rust and Git and so on directly in your repos. Together, this puts you in complete control of your build without requiring network downloads and additional tools beyond `cargo` itself.

The `winrt::build` macro will look for metadata in the `.windows\winmd` folder. If none is found, it will fall back to using the local machine's metadata located here: `%windir%\System32\WinMetadata`, but only if no metadata is available in the `.windows\winmd` folder.

The `winrt::build` function will also copy any architecture-specific dependencies to the target folder. These are `.windows\x64`, `.windows\x86`, `.windows\arm`, and `.windows\arm64` respectively.

There is no longer any need for the `[package.metadata.winrt.dependencies]` section in your `Cargo.toml` file. This will now be ignored and you can safely remove it.

There is also no need for the `types` keyword in the build macro as only types can be specified. Simply remove the keyword and list the use paths you'd like to import.

Fixes #320
Fixes #351
Fixes #368


kennykerr:
The build is taking forever because of a degradation with GitHub Actions. https://www.githubstatus.com/

> We are investigating reports of degraded performance for GitHub Actions.
> Dec 1, 16:40 UTC

crates/macros/src/build_limits.rs
```diff
@@ -0,0 +1,191 @@
...
+use rayon::iter::ParallelIterator;
+use std::convert::{TryFrom, TryInto};
+use syn::spanned::Spanned;
+use winrt_gen::{NamespaceTypes, TypeLimit, TypeLimits, TypeTree};
+
+pub struct BuildLimits(pub std::collections::BTreeSet<TypesDeclaration>);
+
+impl BuildLimits {
+    pub fn to_tokens_string(self) -> Result<String, proc_macro2::TokenStream> {
+        let foundation = self.0.is_empty();
```
rylev, line 11: nit: typically predicate variables like this are in the form `is_foundation`

rylev: Also, is this the best way to detect this case? I can't shake the feeling that handling the foundation in the way we have been is a little hacky.

crates/macros/src/build_limits.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        for limit in self.0 {
+            let types = limit.types;
+            let syntax = limit.syntax;
+            if let Err(e) = limits.insert(types).map_err(|ns| {
+                syn::Error::new_spanned(syntax, format!("'{}' is not a known namespace", ns))
+            }) {
+                return Err(e.to_compile_error());
+            };
```
rylev, line 42:
```rust
limits.insert(types).map_err(|ns| {
                syn::Error::new_spanned(syntax, format!("'{}' is not a known namespace", ns)).to_compile_error()
  })?;
```

crates/macros/src/lib.rs
```diff
@@ -10,76 +12,105 @@ use syn::parse_macro_input;
...
+            path.push("winrt.rs");
+            let mut file = ::std::fs::File::create(&path).expect("Failed to create winrt.rs");
+            file.write_all(#tokens.as_bytes()).expect("Could not write generated code to output file");
+
+            let mut cmd = ::std::process::Command::new("rustfmt");
+            cmd.arg(&path);
+            let _ = cmd.output();
+
+            fn copy(source: &::std::path::PathBuf, destination: &mut ::std::path::PathBuf) {
+                if let ::std::result::Result::Ok(files) = ::std::fs::read_dir(source) {
```
rylev, line 70: Do we really want to ignore file system errors?

examples/simple/readme.md
```diff
@@ -0,0 +1,13 @@
+# Simple
+
+This is an example of the simplest possible configuration. It's great if only have a small number
+of Windows APIs that you'd like to use. For anything beyond that, you'll want to create a bindings
```
rylev, line 4: Should we link to a readme for a bindings example in case folks are unfamiliar with the bindings pattern?

crates/gen/src/windows.rs
```diff
@@ -0,0 +1,26 @@
+/// Returns the build's `.windows` directory in the root of the workspace as a `PathBuf`.
+pub fn build_windows_dir() -> std::path::PathBuf {
+    let mut path = workspace_dir();
+    path.push(".windows");
+    path
+}
+
+fn workspace_dir() -> std::path::PathBuf {
+    // This is all rather complicated and expensive because Cargo lacks an environment variable for the workspace dir.
```
rylev, line 9: Might want to consider using a `lazy_static!` or just `Once` like you have been to cache this.

rylev (commented):
This looks great though I wonder if we've thrown out the baby with the bathwater a little bit. Having a simple utility that can read some sort of manifest information and populate the `.windows` folder on your behalf sounds useful still. This utility would be completely optional but could still help folks manage their WinRT dependencies.

kennykerr:
@rylev agreed such a utility would be helpful but given the complexity of msbuild and nuget I think it would better served as a distinct project with a far deeper knowledge of msbuild and nuget. There are just too many ways the existing `cargo winrt` tool can fail.

crates/macros/src/build_limits.rs
kennykerr: Agreed, I was really looking for a way to identify the package being compiled and only do this if `winrt` is being compiled but I couldn't find a way to identify that.

PR #374, opened 2020-12-03
Ensure completed handler doesn't capture waiter

### LINKED ISSUE #373: Incorrect usage of winrt::Waiter in generated impls for IAsync{Action,Operation}{WithProgress,} interfaces
Looking at the [macro for generating `.get()` for the IAsync\{Action,Operation\}\{WithProgress,\} interfaces](https://github.com/microsoft/winrt-rs/blob/a3eae9e1454de33e6ea55389bd0eb59fd587ee96/crates/gen/src/futures.rs#L44), we can see that we use the [winrt::Waker](https://github.com/microsoft/winrt-rs/blob/a3eae9e1454de33e6ea55389bd0eb59fd587ee96/src/runtime/waiter.rs) struct to wait for completion to occur before getting the result.  However, instead of waiting for the [unnamed Event](https://github.com/microsoft/winrt-rs/blob/a3eae9e1454de33e6ea55389bd0eb59fd587ee96/src/runtime/waiter.rs#L9) created by `winrt::Waiter` via the [impl of the Drop trait](https://github.com/microsoft/winrt-rs/blob/a3eae9e1454de33e6ea55389bd0eb59fd587ee96/src/runtime/waiter.rs#L25), the waiter actually moves it to the completion closure, making the waiting to occur in the completion handler instead of the thread that created the waiter itself, which leads to a call to `.get_results()` before completion occurs!
``` rust
pub fn get(&self) -> ::winrt::Result<#return_type> {
    if self.status()? == ::winrt::foundation::AsyncStatus::Started {
        let waiter = ::winrt::Waiter::new();
        self.set_completed(::winrt::foundation:: #handler::new(move |_sender, _args| { // <--- waiter moved here!
             waiter.signal();
             Ok(())
             // ... and waiter gets Drop-ed here, where there is no point in the WaitForSingleObject!
        }))?;
        // If the below code was uncommented to show the intended code, it would error out
        // due to the `move` above.  The `move` is required to satisfy lifetime constraints pertaining
        // to the use of `waiter` in the completion handler before this scope is exited. Rust doesn't know
        // that we are waiting for the handler to fire first, so it assumes that it is possible to exit the
        // scope first and leave a dangling reference!
       
       // drop(waiter) < --- will error out
    }
    self.get_results()
}
```

As a workaround, I made this function and called it instead using the generated `.get()` method (temporarily):
``` rust
fn wait_complete<T: ::winrt::RuntimeType, R: ::winrt::RuntimeType>(op: windows::foundation::IAsyncOperationWithProgress<T, R>) -> ::winrt::Result<T> {
    if op.status()? == ::winrt::foundation::AsyncStatus::Started {
        let (sender,recver) = std::sync::mpsc::sync_channel(1);
        op.set_completed(::winrt::foundation::AsyncOperationWithProgressCompletedHandler::new(move |_sender, _args| {
            sender.send(());
            Ok(())
        }))?;
        recver.recv().unwrap();
    }
    op.get_results()
}
```

The solution would be similar to the use of channels as above, by splitting `Waiter` into 2: a `Waiter` and `Signaller`. Example code is as below, based on the existing Waiter code (please note that this is untested and I am running on adrenaline 🐵):
``` rust
use crate::*;

/// A simple blocking waiter used by the generated bindings and should not generally be used directly.
pub struct Waiter(RawPtr);
pub struct Signaller(RawPtr);

impl Waiter {
    pub fn new() -> (Waiter, Signaller) {
        unsafe {
            let ev = CreateEventW(
                std::ptr::null_mut(),
                1,
                0,
                std::ptr::null_mut(),
            );
            (Waiter(ev), Signaller(ev))
        }
    }
}

impl Signaller {
    pub fn signal(&self) {
        unsafe {
            SetEvent(self.0);
        }
    }
}

impl Drop for Waiter {
    fn drop(&mut self) {
        unsafe {
            WaitForSingleObject(self.0, 0xFFFFFFFF);
            CloseHandle(self.0);
        }
    }
}

#[link(name = "kernel32")]
extern "system" {
    fn CreateEventW(security: RawPtr, manual: i32, state: i32, name: RawPtr) -> RawPtr;
    fn SetEvent(handle: RawPtr) -> i32;
    fn WaitForSingleObject(handle: RawPtr, milliseconds: u32) -> u32;
    fn CloseHandle(handle: RawPtr) -> i32;
}
```

Code generated from `gen_async_kind` will have to be modified as such:
``` rust
pub fn get(&self) -> ::winrt::Result<#return_type> {
    if self.status()? == ::winrt::foundation::AsyncStatus::Started {
        let (waiter, signaller) = ::winrt::Waiter::new();
        self.set_completed(::winrt::foundation:: #handler::new(move |_sender, _args| {
            signaller.signal();
            Ok(())
        }))?;
        // waiter gets dropped here, which means it waits for singaller to `.signal()`
    }
    self.get_results()
}
```

kennykerr:
Fixes #373

### DIFF
--- a/src/runtime/waiter.rs
+++ b/src/runtime/waiter.rs
@@ -14,7 +14,12 @@ impl Waiter {
 }
 
 impl WaiterSignaler {
+    /// Signals the `Waiter`. This is unsafe because the lifetime of `WaiterSignaler` is not tied
+    /// to the lifetime of the `Waiter`. This is not possible in this case because the `Waiter`
+    /// is used to signal a WinRT async completion and the compiler doesn't know that the lifetime
+    /// of the delegate is bounded by the calling function.
     pub unsafe fn signal(&self) {
+        // https://github.com/microsoft/winrt-rs/pull/374#discussion_r535313344
         SetEvent(self.0);
     }
 }
--- a/src/runtime/waiter.rs
+++ b/src/runtime/waiter.rs
@@ -1,24 +1,21 @@
 use crate::*;
 
-/// A simple blocking waiter used by the generated bindings and should not generally be used directly.
+/// A simple blocking waiter used by the generated bindings and should not be used directly.
 pub struct Waiter(RawPtr);
+pub struct WaiterSignaler(RawPtr);
 
 impl Waiter {
-    pub fn new() -> Self {
+    pub fn new() -> (Waiter, WaiterSignaler) {
         unsafe {
-            Self(CreateEventW(
-                std::ptr::null_mut(),
-                1,
-                0,
-                std::ptr::null_mut(),
-            ))
+            let handle = CreateEventW(std::ptr::null_mut(), 1, 0, std::ptr::null_mut());
+            (Waiter(handle), WaiterSignaler(handle))
         }
     }
+}
 
-    pub fn signal(&self) {
-        unsafe {
-            SetEvent(self.0);
-        }
+impl WaiterSignaler {
+    pub unsafe fn signal(&self) {
+        SetEvent(self.0);
     }
 }
 
--- a/crates/gen/src/futures.rs
+++ b/crates/gen/src/futures.rs
@@ -66,9 +66,10 @@ fn gen_async_kind(
         quote! {
             pub fn get(&self) -> ::winrt::Result<#return_type> {
                 if self.status()? == ::winrt::foundation::AsyncStatus::Started {
-                    let waiter = ::winrt::Waiter::new();
+                    let (waiter, signaler) = ::winrt::Waiter::new();
                     self.set_completed(::winrt::foundation:: #handler::new(move |_sender, _args| {
-                        waiter.signal();
+                        // Safe because the waiter will only be dropped after being signaled.
+                        unsafe { signaler.signal(); }
                         Ok(())
                     }))?;
                 }


src/runtime/waiter.rs
```diff
@@ -1,24 +1,21 @@
 use crate::*;
 
-/// A simple blocking waiter used by the generated bindings and should not generally be used directly.
+/// A simple blocking waiter used by the generated bindings and should not be used directly.
 pub struct Waiter(RawPtr);
+pub struct WaiterSignaler(RawPtr);
```
rylev, line 5: This can made much safer if you tie the lifetime of `WaiterSignaler` was tied to the lifetime of `Waiter`. You can do this by having a `WaitSignaler<'a>` instead which you get by calling `signaler(&'a self)` on `Waiter`

rylev (changes requested):
It's easy to make this completely safe. I think we should avoid having an unsafe method here.

src/runtime/waiter.rs
kennykerr: nice!

kennykerr: Doesn't quite work:

```
error[E0597]: `waiter` does not live long enough
   --> D:\git\winrt-rs\target\debug\build\winrt-4b6f2188f7967158\out\winrt.rs:298:36
    |
298 |                     let signaler = waiter.signaler();
    |                                    ^^^^^^-----------
    |                                    |
    |                                    borrowed value does not live long enough
    |                                    argument requires that `waiter` is borrowed for `'static`
...
306 |                 }
    |                 - `waiter` dropped here while still borrowed
```

kennykerr: The compiler doesn't know that the completed handler's lifetime is bounded by the enclosing function.

rylev: Ah yes that's true. 😦 I guess we have to go back. We should just document the unsafe method then

kennykerr: Added a comment.

kennykerr: For reference, here's the implementation that would have been safe.

```rust
pub struct Waiter(RawPtr);
pub struct WaiterSignaler<'a>(RawPtr, std::marker::PhantomData<&'a Waiter>);

impl Waiter {
    pub fn new() -> Waiter {
        unsafe {
            Self(CreateEventW(
                std::ptr::null_mut(),
                1,
                0,
                std::ptr::null_mut(),
            ))
        }
    }

    pub fn signaler<'a>(&'a self) -> WaiterSignaler<'a> {
        WaiterSignaler(self.0, std::marker::PhantomData)
    }
}

impl<'a> WaiterSignaler<'a> {
    pub fn signal(&self) {
        unsafe {
            SetEvent(self.0);
        }
    }
}

impl Drop for Waiter {
    fn drop(&mut self) {
        unsafe {
            WaitForSingleObject(self.0, 0xFFFFFFFF);
            CloseHandle(self.0);
        }
    }
}
```

rylev (approved):

PR #378, opened 2020-12-06
copy files from .windows to root target directoty

milyin:
This is a fix for issue https://github.com/microsoft/winrt-rs/issues/377
The problem is that CARGO_MANIFEST_DIR variable contains the path to Cargo.toml of subcrate, so script tries to copy files from ",windows" directory to "win2d-rs\bindings\target" instead of "win2d-rs\target"

Function workspace_dir() made public to avoid multiple call to it as it's quite expensive

### DIFF
--- a/crates/gen/src/windows.rs
+++ b/crates/gen/src/windows.rs
@@ -1,7 +1,7 @@
 use lazy_static::lazy_static;
 
 /// Returns the build's `.windows` directory in the root of the workspace as a `PathBuf`.
-pub fn build_windows_dir() -> std::path::PathBuf {
+pub fn workspace_windows_dir() -> std::path::PathBuf {
     let mut path = WORKSPACE_DIR.clone();
     path.push(".windows");
     path
--- a/crates/gen/src/windows.rs
+++ b/crates/gen/src/windows.rs
@@ -8,7 +8,7 @@ pub fn build_windows_dir() -> std::path::PathBuf {
 }
 
 /// Returns the build's target directory in the passed workspace directory as a `PathBuf`.
-pub fn build_target_dir() -> std::path::PathBuf {
+pub fn workspace_target_dir() -> std::path::PathBuf {
     let mut path = WORKSPACE_DIR.clone();
     path.push("target");
     path
--- a/crates/gen/src/windows.rs
+++ b/crates/gen/src/windows.rs
@@ -1,11 +1,25 @@
 /// Returns the build's `.windows` directory in the root of the workspace as a `PathBuf`.
 pub fn build_windows_dir() -> std::path::PathBuf {
-    let mut path = workspace_dir();
+    build_windows_dir_at_workspace(&workspace_dir())
+}
+
+/// Returns the build's `.windows` directory in the passed workspace directory as a `PathBuf`.
+pub fn build_windows_dir_at_workspace(workspace_dir: &std::path::PathBuf) -> std::path::PathBuf {
+    let mut path = workspace_dir.clone();
     path.push(".windows");
     path
 }
 
-fn workspace_dir() -> std::path::PathBuf {
+/// Returns the build's target directory in the passed workspace directory as a `PathBuf`.
+pub fn build_target_dir_at_at_workspace(workspace_dir: &std::path::PathBuf) -> std::path::PathBuf {
+    let mut path = workspace_dir.clone();
+    path.push("target");
+    path
+}
+
+/// Returns the build's workspace directory. Executes 'cargo' command to get it because Cargo
+/// lacks an environment variable for the workspace dir.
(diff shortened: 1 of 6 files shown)


crates/gen/src/windows.rs
```diff
@@ -1,11 +1,25 @@
...
+/// Returns the build's `.windows` directory in the passed workspace directory as a `PathBuf`.
+pub fn build_windows_dir_at_workspace(workspace_dir: &std::path::PathBuf) -> std::path::PathBuf {
+    let mut path = workspace_dir.clone();
     path.push(".windows");
     path
 }
 
-fn workspace_dir() -> std::path::PathBuf {
+/// Returns the build's target directory in the passed workspace directory as a `PathBuf`.
+pub fn build_target_dir_at_workspace(workspace_dir: &std::path::PathBuf) -> std::path::PathBuf {
```
rylev, line 14: `build_target_dir_at_workspace` => `workspace_target_dir`

The current wording is pretty confusing.

crates/gen/src/windows.rs
```diff
@@ -1,11 +1,25 @@
 /// Returns the build's `.windows` directory in the root of the workspace as a `PathBuf`.
 pub fn build_windows_dir() -> std::path::PathBuf {
-    let mut path = workspace_dir();
+    build_windows_dir_at_workspace(&workspace_dir())
+}
+
+/// Returns the build's `.windows` directory in the passed workspace directory as a `PathBuf`.
+pub fn build_windows_dir_at_workspace(workspace_dir: &std::path::PathBuf) -> std::path::PathBuf {
```
rylev, line 7: `build_windows_dir_at_workspace` => `workspace_windows_dir`

crates/gen/src/windows.rs
```diff
@@ -1,11 +1,25 @@
...
+/// Returns the build's target directory in the passed workspace directory as a `PathBuf`.
+pub fn build_target_dir_at_workspace(workspace_dir: &std::path::PathBuf) -> std::path::PathBuf {
+    let mut path = workspace_dir.clone();
+    path.push("target");
+    path
+}
+
+/// Returns the build's workspace directory. Executes 'cargo' command to get it because Cargo
+/// lacks an environment variable for the workspace dir.
+pub fn workspace_dir() -> std::path::PathBuf {
```
rylev, line 22: Note: we should probably cache this call in a static since this is expensive. This would also remove the need to pass the workspace directory into the other functions which would simplify the calling code.

crates/macros/src/lib.rs
```diff
@@ -42,10 +42,15 @@ pub fn build(stream: TokenStream) -> TokenStream {
         Err(t) => return t.into(),
     };
 
-    let mut source = ::winrt_gen::build_windows_dir();
+    let workspace_dir = ::winrt_gen::workspace_dir();
+
+    let mut source = ::winrt_gen::build_windows_dir_at_workspace(&workspace_dir);
```
rylev, line 47: These uses of `winrt_gen` don't need to have the leading `::`.

rylev (changes requested):
Thanks! Looks good, but I think I'd rather use caching of the workspace_dir than passing it to each of the functions.

kennykerr:
Ah yes, the architecture binaries were meant to use the workspace calculation rather than the `CARGO_MANIFEST_DIR` environment variable as well. Thanks for catching that.

crates/gen/src/windows.rs
kennykerr: `workspace_windows_dir` sounds good.

crates/gen/src/windows.rs
kennykerr: `workspace_target_dir` sounds good.

crates/gen/src/windows.rs
milyin: done - renamed it in last update

crates/gen/src/windows.rs
milyin: done - renamed it in last update

crates/gen/src/windows.rs
milyin: ok, placed workspace_dir() result to lazy_static variable

crates/macros/src/lib.rs
milyin: done - removed ::

kennykerr:
Thanks @milyin - really appreciate it. I'm just in the middle of a large refactor to simplify the internal parser and it somewhat overlaps with this code so I've rolled in your fix into that branch. I've opted not to take the caching suggested by Ryan because I'm trying to reduce dependencies. I'll replace it with a hand-rolled cache when I get a moment. The good news is that this is fixed with your help. Thanks for pointing out the bug and suggesting the fix!

PR #379, opened 2020-12-08
Winmd parsing improvements

### LINKED ISSUE #377: new build scheme without "cargo winrt" command doesn't work when winrt bindings is placed to subcrate
I tried to modify project https://github.com/kennykerr/win2d-rs to move winrt dependency to subcrate 'bindings' in order to accelerate build time, like it's implemented in https://github.com/robmikh/minesweeper-rs project.

The test project is https://github.com/milyin/win2d-rs

I found that it doesn't work because in this configuration .dll files from ".windows" directory are not copied to "target" directory:

```
   Compiling bindings v0.1.0 (C:\TEMP\win2d-rs\bindings)
   Compiling win2d-rs v0.1.0 (C:\TEMP\win2d-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 0.90s
     Running `target\debug\win2d-rs.exe`
thread 'main' panicked at 'Failed to create CanvasDevice: Error { code: 0x80040154, message: "Class not registered" }', src\main.rs:4:39
error: process didn't exit successfully: `target\debug\win2d-rs.exe` (exit code: 101)
```

kennykerr:
Improvements to allow a broader set of types to be supported. Notably, the winmd parser is again local to the repo and uses a dedicated static type reader to simplify the library and not require a reader argument to be passed to each method. This greatly simplifies the parsing code. This also means that we can use `&str` in many places instead of having to allocate `String` values.

Also fixes #377


Cargo.toml
```diff
@@ -13,7 +13,7 @@ readme = "README.md"
 winrt-macros = { path = "crates/macros",  version = "0.8.0" }
 winrt-gen = { path = "crates/gen",  version = "0.8.0" }
 const-sha1 = "0.2"
-winmd = "0.6.0"
+winrt-winmd = { path = "crates/winmd",  version = "0.8.0" }
```
rylev, line 16: Why the change of name?

crates/gen/src/class32.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub fn from_type_name(name: TypeName) -> Self {
+        Self { name }
+    }
+
+    pub fn gen(&self) -> TokenStream {
+        quote! {}
+    }
+
+    pub fn dependencies(&self) -> Vec<winmd::TypeDef> {
+        Vec::new()
```
rylev, line 19: Is this just for consistency or will these eventually return something meaningful?

rylev (commented):
This is great! ❤ I did have some questions though.

crates/gen/src/class32.rs
kennykerr: Yes, just placeholder as all the types need a `dependencies` function and will soon return its dependencies.

Cargo.toml
kennykerr: I ran into some trouble with the name when I was porting but you're right the simpler name should suffice - switching back to `winmd`.

PR #382, opened 2020-12-08
Rerun if changed

milyin:
Use absolute path to .windows directory since relative path doesn't work for subcrates


kennykerr:
Thanks! I'm just not seeing an improvement when I try this change:

```
D:\git\minesweeper-rs>cargo run
   Compiling winrt-macros v0.8.0 (D:\git\winrt-rs\crates\macros)
   Compiling winrt v0.8.0 (D:\git\winrt-rs)
   Compiling bindings v0.1.0 (D:\git\minesweeper-rs\bindings)
   Compiling minesweeper-rs v0.1.0 (D:\git\minesweeper-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 21.47s
     Running `target\debug\minesweeper-rs.exe`

D:\git\minesweeper-rs>cargo run
   Compiling bindings v0.1.0 (D:\git\minesweeper-rs\bindings)
   Compiling minesweeper-rs v0.1.0 (D:\git\minesweeper-rs)
    Finished dev [unoptimized + debuginfo] target(s) in 15.21s
     Running `target\debug\minesweeper-rs.exe`
```

kennykerr:
If I comment out `cargo:rerun-if-changed` it doesn't make it any faster for me either. 🤔

kennykerr:
I take that back. That does indeed remove the slowdown.

rylev:
From the [docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed):

> If the path points to a directory, it does not automatically traverse the directory for changes. Only the mtime change of the directory itself is considered (which corresponds to some types of changes within the directory, depending on platform). To request a re-run on any changes within an entire directory, print a line for the directory and separate lines for everything inside it, recursively.

I’m not sure what the behavior on windows would be, but this seems to not be what we want.

kennykerr:
Yes, this doesn't really help us. And generating a separate line for each entry recursively doesn't make sense in this case since a developer may add/remove items and that wouldn't necessarily trigger a build. So, I suggest we simply remove the `cargo:rerun-if-changed`.

milyin:
For me the absolute path approach works very well. Any change inside .windows directory causes rebuild because "Last Write Time" attribute of directory is changed. If no changes - no rebuild.
Though if you decide to just remove the "rerun-if-changed" attribute, it's ok for me too.

rylev:
I've found the issue: minesweeper-rs doesn't have a .windows file. It relies on windows metadata stored at the path in `windir`. This means that for minesweeper-rs, cargo will look at a non-existent `.windows` file and rerun if that file changed. I imagine cargo sees that the file isn't there, and assumes it was deleted. Thus it reruns the build script every time. In case of minesweeper-rs, removing the "rerun-if-changed" line no longer makes the build script dependent on anything and so it will only rerun on a clean build (or if the build script itself changes. 

I believe the change is to only emit the "rerun-if-changed" directive if the path to ".windows" actually exists which can be done with `if ::std::path::Path::new(#workspace_windows_dir).exists()`. Also, we can probably do the same check for the `source` directory at the end of the build script as we only want to copy files in the case where we're not using the `windir` env variable. 

@kennykerr this would have been easier to debug if we didn't simply swallow errors in the build script. Perhaps we should also at least `eprintln!` those errors? They will only show up if you build with the verbose flag.

skreborn:
> @kennykerr this would have been easier to debug if we didn't simply swallow errors in the build script. Perhaps we should also at least `eprintln!` those errors? They will only show up if you build with the verbose flag.

I'm behind you on this one - it would be nice not to just swallow errors, especially if the build script gives no indication that something went wrong and how to get around to seeing the specific cause.

milyin:
> I believe the change is to only emit the "rerun-if-changed" directive if the path to ".windows" actually exists which can be done with `if ::std::path::Path::new(#workspace_windows_dir).exists()`
Updated PR - now it sets rerun-if-changed to absolute path of ".windows/target"directory and does it only if ".windows/target"directory exists

rylev:
@milyin this means we won't rerun the build script if the winmd files change (only if the dlls do) which doesn't seem right.

milyin:
> 
> 
> @milyin this means we won't rerun the build script if the winmd files change (only if the dlls do) which doesn't seem right.

@rylev Thank you, forgot about winmd folder. Fixing

rylev:
This looks good to me!

kennykerr:
Thanks all - glad we got to the bottom of this! Agreed about error reporting - we should get that updated.

kennykerr (approved):

kennykerr:
Today I started noticing long rebuilds again. Anyone else notice this?

skreborn:
> Today I started noticing long rebuilds again. Anyone else notice this?

@kennykerr I've been having them for a while, actually. Every time I save a file in the project, cargo-check runs and it practically freezes the IDE completely.

kennykerr:
`git clean -fdx` seems to have done it for me. A lot has changed in `windows-rs` lately and I think the tooling just collects a lot of cruft that doesn't quite get cleaned up.

kennykerr:
I will say that incremental builds within `windows-rs` - as in while working on this project - are shot. The tooling just can't handle the complexity/scale. But as a consumer of `windows-rs` it appears to work ok.

PR #383, opened 2020-12-11
Additional type system support

### LINKED ISSUE #353: Unroll derived traits as the derive macro is very slow
Most of the traits are now generated by winrt-rs (as part of #81) - just a few remain for WinRT structs.

kennykerr:
Also fixes #353


crates/gen/src/delegate32.rs
```diff
@@ -4,23 +4,45 @@ use squote::{quote, TokenStream};
...
+    pub fn dependencies(&self) -> Vec<winmd::TypeDef> {
+        self.method.dependencies()
     }
 
     pub fn gen(&self) -> TokenStream {
-        let definition = self.name.gen_definition();
+        let name = self.name.gen();
+
+        // TODO: here we're using gen_abi() because as an ABI function pointer it can't use the owning
+        // types. Perhaps we should introduce a NonOwning<T> concept for this case.
```
rylev, line 30: Is it possible to use `&T` instead? A `NonOwning<T>` would still need to be sure that it doesn't last longer than the `T` does which is exactly what `&T` tracks. If that's too difficult a `NonOwning<T>` would be fine, but it's use would always be unsafe since it's up to the programmer to track the lifetime.

crates/gen/src/delegate32.rs
```diff
@@ -4,23 +4,45 @@ use squote::{quote, TokenStream};
...
+        let params = self.method.params.iter().map(|param| param.gen_abi());
+
+        let return_type = if let Some(return_type) = &self.method.return_type {
```
rylev, line 33: Thought: If we implement `Default` for `TokenStream`, this could be changed to:

```rust
let return_type = self.method.return_type.map(|rt| {
  let rt = rt.kind.gen_abi();
  quote! {
    -> #rt 
  }
}).unwrap_or_default();
```

Not sure which you like better.

crates/gen/src/delegate32.rs
```diff
@@ -4,23 +4,45 @@ use squote::{quote, TokenStream};
...
+            let return_type = return_type.kind.gen_abi();
+
+            quote! {
+                -> #return_type
```
rylev, line 37: Nit: I think this could is always easier to understand if you don't include extras like `->` in the token stream. The variable is called `return_type` which makes me think it's only including the type itself not the `->`. This particular use is easy to tell because the code is rather small, but I've noticed in many places where the puzzle pieces fit together in surprising ways making the code hard to follow.

rylev (commented):
Great stuff!

crates/gen/src/delegate32.rs
kennykerr: This is one case where I don't have an abstraction over what the developer would write. These delegates are just function pointers and unless I come up with an abstraction, we need to honor the ABI directly. `T&` won't work if `T` is `NonNull` or some type that is not blittable. The ABI rules get in the way. Hopefully I can find a way to abstract this ABI the same as others and then we can just use the regular `Param<T>` than handles all of this, but in this case there's no "object" to hang it off.

PR #385, opened 2020-12-29
Use escaped method name & full abi signature for implement

Perlmint:
Escape method names in generated by deriving implement code.
Fix "too many super" error occurred by using relative ABI in implement code.


kennykerr:
Thanks for the contribution! Please create an issue that briefly describes the problem you're trying to solve along with a minimal repro to validate the problem.

Perlmint:
> Thanks for the contribution! Please create an issue that briefly describes the problem you're trying to solve along with a minimal repro to validate the problem.

I wrote some detailed error in #386.

kennykerr (approved):
Thanks - looks great!

PR #398, opened 2021-01-12
Improved BSTR support

kennykerr:
The `BString` class now provides the same capabilities as `HString`.

### DIFF
--- a/src/strings/bstring.rs
+++ b/src/strings/bstring.rs
@@ -1,21 +1,54 @@
 use crate::*;
+use std::convert::TryFrom;
+use std::result::Result as StdResult;
 
-#[repr(C)]
+/// A Basic string, sometimes called a [BSTR](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/automat/bstr),
+/// is a string data type used by certain COM interfaces and interop functions.
+#[repr(transparent)]
 pub struct BString(RawPtr);
 
 impl BString {
-    pub fn new() -> BString {
+    /// Create an empty `BString`.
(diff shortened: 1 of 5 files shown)


src/strings/bstring.rs
```diff
@@ -1,21 +1,54 @@
...
+
+    /// Get the contents of this `BString` as a `String` lossily.
+    pub fn to_string_lossy(&self) -> String {
+        String::from_utf16_lossy(self.as_wide())
+    }
+
+    /// Frees the memory occupied by the string.
     pub fn clear(&mut self) {
         unsafe {
             SysFreeString(self.0);
```
rylev, line 54: Nit: It might be "more correct" to store the pointer into a local, set the `self.0` to null and then call `SysFreeString` on the local. 

`SysFreeString` can't kick off a stack unwind can it?

src/strings/bstring.rs
```diff
@@ -34,39 +67,208 @@ unsafe impl Abi for BString {
...
     }
 }
 
-impl From<BString> for String {
-    fn from(from: BString) -> Self {
-        String::from(&from)
+impl From<&str> for BString {
+    fn from(value: &str) -> Self {
+        let value: Vec<u16> = value.encode_utf16().collect();
+        Self::from_wide(&value)
```
rylev, line 107: You might want to consider implementing a `from_wide_iter` that can move values from an iterator to a BString. Currently, you're allocating a temporary vector that will just be dropped at the end of this function which is a bit of a waste.

PR #448, opened 2021-01-25
Additional COM support

### LINKED ISSUE #430: Support COM interfaces inheritance
It appears that COM interfaces that inherit from other interfaces don't properly express it in their vtable layout. Take `ID3D11Texture2D` for example:

```rust
#[repr(C)]
pub struct ID3D11Texture2D_abi(
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        iid: &::windows::Guid,
        interface: *mut ::windows::RawPtr,
    ) -> ::windows::ErrorCode,
    pub unsafe extern "system" fn(this: ::windows::RawPtr) -> u32,
    pub unsafe extern "system" fn(this: ::windows::RawPtr) -> u32,
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        p_desc: *mut D3D11_TEXTURE2D_DESC,
    ),
);
```

In order to call `GetDesc` correctly, I had to define a custom vtable:
```rust
#[repr(C)]
pub struct ID3D11Texture2DCustom_abi(
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        iid: &::windows::Guid,
        interface: *mut ::windows::RawPtr,
    ) -> ::windows::ErrorCode,
    pub unsafe extern "system" fn(this: ::windows::RawPtr) -> u32,
    pub unsafe extern "system" fn(this: ::windows::RawPtr) -> u32,

    // ID3D11DeviceChild
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        pp_device: *mut ::std::option::Option<ID3D11Device>,
    ),
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        guid: *const ::windows::Guid,
        p_data_size: *mut u32,
        p_data: *mut ::std::ffi::c_void,
    ) -> ::windows::ErrorCode,
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        guid: *const ::windows::Guid,
        data_size: u32,
        p_data: *const ::std::ffi::c_void,
    ) -> ::windows::ErrorCode,
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        guid: *const ::windows::Guid,
        p_data: ::std::option::Option<::windows::IUnknown>,
    ) -> ::windows::ErrorCode,

    // ID3D11Resource
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        p_resource_dimension: *mut bindings::windows::win32::direct3d11::D3D11_RESOURCE_DIMENSION,
    ),
    pub unsafe extern "system" fn(this: ::windows::RawPtr, eviction_priority: u32),
    pub unsafe extern "system" fn(this: ::windows::RawPtr) -> u32,

    // ID3D11Texture2D
    pub  unsafe extern "system" fn(
        this: ::windows::RawPtr,
        p_desc: *mut D3D11_TEXTURE2D_DESC,
    ),
);
```

kennykerr:
Fixes #430 

This update adds support for COM interface inheritance. Primarily, this means that you can call methods on COM interfaces that inherit from other COM interfaces as their vtables will properly be stacked to ensure the correct offsets. 

Convertibility is also provided so you can use `From`/`Into` to move up the hierarchy and the `Param` specializations are provided so you can automatically bind arguments to parameters polymorphically.

Constants for COM classes (CLSID values) are now also generated, so the build macro can generate those values for use with functions like `CoCreateInstance`. 

`LARGE_INTEGER` and `ULARGE_INTEGER` are now mapped to `i64` and `u64` respectively for convenience. 

Finally, the `build` macro now expects the use paths to be comma separated. This makes it a lot easier to copy and paste the use paths between the build macro and the source file where those paths are actually used.


crates/gen/src/com_interface.rs
```diff
@@ -28,21 +30,43 @@ impl ComInterface {
     pub fn from_type_name(name: TypeName) -> Self {
         let mut count = BTreeMap::new();
 
-        let methods = name
-            .def
-            .methods()
-            .map(|method| {
+        let mut bases = Vec::new();
```
rylev, line 33: We probably want to use a `VecDeque` if you're going to be inserting at the beginning.

crates/gen/src/com_interface.rs
```diff
@@ -28,21 +30,43 @@ impl ComInterface {
...
+        }
+
+        let mut methods = Vec::new();
+
+        for def in bases
+            .iter()
+            .map(|name| name.def)
+            .chain(std::iter::once(name.def))
+        {
+            for method in def.methods() {
```
rylev, line 54: Nit: you can get rid of the inner `for` loop by adding `flat_map(|def| def.methods())` after the call to `chain`.

crates/gen/src/com_interface.rs
```diff
@@ -28,21 +30,43 @@ impl ComInterface {
...
                 let count = count.entry(method.name()).or_insert(0);
                 *count += 1;
 
-                Method {
+                methods.push(Method {
```
rylev, line 58: You can get rid of this by just making the for loop a `map` and then using `collect` at the end. This will could be slightly  more efficient as it might pre-allocate enough space on the heap all at once.

crates/gen/src/com_interface.rs
```diff
@@ -61,24 +85,16 @@ impl ComInterface {
...
-
+            let constraints = gen_constraint(method);
+            let params = gen_params(method);
+            let args = gen_abi_args(method);
             let name = method.gen_name();
             let vtable_offset = Literal::u32_unsuffixed((vtable_offset + 3) as u32);
 
             quote! {
-                pub fn #name(&self, #(#params),*) #return_type {
+                pub fn #name<#constraints>(&self, #params) #return_type {
```
rylev, line 95: nit: I know it was like this before, but it would be nicer if `return_type` really was just the type and not also the `->`. This might make it easier to understand and compose.

crates/gen/src/com_interface.rs
```diff
@@ -138,14 +215,81 @@ impl ComInterface {
...
+fn gen_constraint(method: &Method) -> TokenStream {
+    let mut tokens = Vec::new();
+
+    for (position, param) in method.signature.params.iter().enumerate() {
+        if !param.is_input || param.is_array {
+            continue;
+        }
+
+        match &param.kind {
+            TypeKind::IUnknown | TypeKind::Interface(_) => {
```
rylev, line 241: nit: this might be easier to read as an `if let`

crates/gen/src/com_interface.rs
```diff
@@ -138,14 +215,81 @@ impl ComInterface {
...
+        };
+    }
+
+    TokenStream::from_iter(tokens)
+}
+
+fn gen_abi_args(method: &Method) -> TokenStream {
+    let mut tokens = Vec::new();
+
+    for param in method.signature.params.iter() {
```
rylev, line 281: If you use `map` instead of collecting into a vector you can just pass the iterator to `TokenStream::from_iter` and avoid allocating a vector.

crates/gen/src/com_interface.rs
```diff
@@ -138,14 +215,81 @@ impl ComInterface {
...
+                tokens.push(quote! { #name: #type_tokens, });
+            }
+            _ => {
+                let type_tokens = param.gen_field();
+                tokens.push(quote! { #name: #type_tokens, });
+            }
+        };
+    }
+
+    TokenStream::from_iter(tokens)
```
rylev, line 275: Same point here about avoiding allocating a vector

rylev (commented):
Looking great!

crates/gen/src/com_interface.rs
kennykerr: Replaced with `push` and `rev` iterator.

crates/gen/src/com_interface.rs
kennykerr: Problem is I get this:

```
error[E0515]: cannot return value referencing function parameter `def`
  --> crates\gen\src\com_interface.rs:53:28
   |
53 |             .flat_map(|def|def.methods())
   |                            ---^^^^^^^^^^
   |                            |
   |                            returns a value referencing data owned by the current function
   |                            `def` is borrowed here
```

crates/gen/src/com_interface.rs
kennykerr: See above.

crates/gen/src/com_interface.rs
kennykerr: Good idea, but I'll leave that as homework - this pattern appears in a few places and I'd like to tackle them all if possible.

crates/gen/src/com_interface.rs
kennykerr: This is going to need more match patterns as I finish up the type system support.

crates/gen/src/com_interface.rs
kennykerr: Nice, done.

PR #450, opened 2021-01-26
Improve macro hygiene

ajeetdsouza:
Currently, the macros defined in `macros.rs` capture `std`, and their behaviour may vary if `std` is shadowed somewhere in the code. This PR changes this to use an unambiguous path (`::std`) instead.

[Additional reading](https://web.archive.org/web/20201125151957/https://gist.github.com/Koxiaet/8c05ebd4e0e9347eb05f265dfb7252e1)


src/macros.rs
```diff
@@ -14,11 +14,11 @@ macro_rules! demand_load {
     } )* ) => {
         $($(
             #[allow(non_snake_case)]
-            unsafe fn $sym( $( $param: $pty ),* ) -> std::result::Result<$rt, ErrorCode> {
-                use std::{sync::Once, mem::MaybeUninit};
+            unsafe fn $sym( $( $param: $pty ),* ) -> ::std::result::Result<$rt, ErrorCode> {
+                use ::std::{sync::Once, mem::MaybeUninit};
```
rylev, line 18: We should probably just get rid of these use statements.

src/macros.rs
```diff
@@ -14,11 +14,11 @@ macro_rules! demand_load {
     } )* ) => {
         $($(
             #[allow(non_snake_case)]
-            unsafe fn $sym( $( $param: $pty ),* ) -> std::result::Result<$rt, ErrorCode> {
-                use std::{sync::Once, mem::MaybeUninit};
+            unsafe fn $sym( $( $param: $pty ),* ) -> ::std::result::Result<$rt, ErrorCode> {
```
rylev, line 17: Perhaps we should refer to `ErrorCode` by its full path?

src/macros.rs
```diff
@@ -30,7 +30,7 @@ macro_rules! demand_load {
                 // transmute() doesn't work on generic types, as you can't constrain to a
                 // function pointer, so it must be done here outside load_proc().
                 type FnPtr = extern "system" fn ( $( $param: $pty ),* ) -> $rt;
-                let f = std::mem::transmute::<RawPtr, FnPtr>(VALUE.assume_init()?);
+                let f = ::std::mem::transmute::<RawPtr, FnPtr>(VALUE.assume_init()?);
```
rylev, line 33: Same with `RawPtr`?

rylev (dismissed):
Looks good! We could probably go even further by referring to more things by their full path.

ajeetdsouza:
@rylev thanks for the quick review! There are more macros in the crate, but those are internal and it's not as dangerous to leave them as-is, since we have full control over their usage. All the same, let me know if you'd like me to change those as well.

kennykerr:
Looks good. It's pretty hard to know for sure without tooling to validate that we haven't missed any.

kennykerr (approved):
Thanks!

PR #461, opened 2021-01-28
Remove unused or barely used dependencies

rylev:
This removes the following explicit dependencies:
* `serde`
* `serde_json`
* `rayon`
* `sha1`

Additionally several other transitive dependencies were removed. These dependencies are largely underutilized and so they're not really worth the compiler performance hit.

This makes a full build (i.e., `cargo clean; cargo build`) of the "clock" example repo build in 34s vs 43s on master. 

We may eventually want to add back the functionality we're losing from removing `rayon` (parallel code gen), but for smaller builds like the "clock" example building `rayon` itself more than eliminates any gains from doing code gen in parallel.


crates/gen/src/type_namespaces.rs
```diff
@@ -1,13 +1,12 @@
...
 use std::collections::*;
 
 #[derive(Default)]
 pub struct TypeNamespaces(pub BTreeMap<&'static str, crate::type_tree::TypeTree>);
 
 impl TypeNamespaces {
-    pub fn gen<'a>(&'a self) -> impl ParallelIterator<Item = TokenStream> + 'a {
-        self.0.par_iter().map(|(name, tree)| {
+    pub fn gen<'a>(&'a self) -> impl Iterator<Item = TokenStream> + 'a {
+        self.0.iter().map(|(name, tree)| {
```
kennykerr, line 9: Does this not have a big impact on build time for projects that are importing a few large namespaces?

kennykerr: But we could certainly parallelize those without these expensive dependencies.

rylev: I imagine this will have negative impacts on _very_ large dependency graphs like building the entire Windows API. For my subset that I use for testing (which is about half of the total win32 API), the changes compile a bit faster (6min19sec vs 6min26sec). I think we should merge this, and then separately do the work of seeing if adding a custom parallel code gen can improve performance in the extreme cases.

kennykerr: Sounds good.

kennykerr (approved):

crates/gen/src/type_namespaces.rs
kennykerr: This removes quite a few dependencies. Very nice speedup. 👍🏼

PR #469, opened 2021-01-29
Numerics

kennykerr:
Just some cleanup to make them easier to read and maintain.

### DIFF
--- a/src/runtime/numerics.rs
+++ b/src/runtime/numerics.rs
@@ -682,6 +682,7 @@ impl Matrix3x2 {
         }
         matrix
     }
+
     fn impl_add(&self, rhs: &Self) -> Self {
         Self {
             m11: self.m11 + rhs.m11,
--- a/src/runtime/numerics.rs
+++ b/src/runtime/numerics.rs
@@ -192,7 +192,6 @@ impl ::std::ops::Mul<f32> for &Vector2 {
     }
 }
 
-
 // Vector3
 impl Vector3 {
     pub fn zero() -> Self {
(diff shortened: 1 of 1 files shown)


robmikh (approved):
The refactor looks great!

src/runtime/numerics.rs
```diff
@@ -1,1053 +1,819 @@
...
+        (self - value).length()
+    }
+    pub fn distance_squared(&self, value: &Self) -> f32 {
+        (self - value).length_squared()
+    }
+    pub fn normalize(&self) -> Self {
+        self / self.length()
+    }
+
+    fn impl_add(&self, rhs: &Self) -> Self {
```
rylev, line 36: What's the purpose of these `impl_` methods?

src/runtime/numerics.rs
```diff
@@ -1,1053 +1,819 @@
...
-            y: self.y * rhs,
-        }
+        self.impl_mul_f32(rhs)
     }
 }
-impl Vector2 {
+
+// Vector3
+impl Vector3 {
     pub fn zero() -> Self {
```
rylev, line 197: These might be better suited as constants.

kennykerr: Yes, good idea.

src/runtime/numerics.rs
kennykerr: They're the common implementations of the various `add` methods in the traits that follow. It can't simply be called `add` because that confuses the compiler into thinking it might be recursive.

PR #473, opened 2021-01-29
Add support for error conversion

### LINKED ISSUE #424: std::error::Error trait not implemented for windows::Error
The following test will not compile, since windows::Error does not implement the std::error::Error trait:

```
#[cfg(test)]
mod tests {
    use super::*;

    fn system_call() -> std::result::Result<(), std::ffi::NulError> {
        Ok(())
    }

    fn windows_call() -> windows::Result<()> {
        Err(Error::new(ErrorCode(777), "real bad..."))
    }

    fn can_fail() -> std::result::Result<(), Box<dyn std::error::Error>> {
        let win_error = windows_call()?;
        let std_error = system_call()?;
        Ok(())
    }

    #[test]
    fn can_mix_window_errors_with_std_errors() {
        assert!(can_fail().is_err());
    }
}
```

By having windows::Error implement the std::error::Error trait, the code above will compile which allows you to mix different kinds of errors in the same function. Something like this would work:
```
impl std::error::Error for Error {}

impl Display for Error {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        f.write_str(&self.message())
    }
}
```

Is there another way to make working with different kinds of errors's easier when working with the `windows` crate?

kennykerr:
Fixes #424

### DIFF
--- a/crates/tests/tests/error.rs
+++ b/crates/tests/tests/error.rs
@@ -32,15 +32,23 @@ fn bad_uri() {
 
 #[test]
 fn convertible() {
+    fn windows_error() -> windows::Result<()> {
+        Err(windows::Error::new(
+            windows::ErrorCode::E_NOINTERFACE,
+            "test message",
+        ))
+    }
+
     fn convertible_error() -> std::result::Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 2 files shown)


rylev (approved):
Awesome!

PR #476, opened 2021-01-30
Dual licensed under MIT or Apache

### LINKED ISSUE #426: Lack of Apache-2.0 license option prevents this from being used by the standard library
The current crate license is MIT only, with no option for Apache-2.0

However, for a crate to become a dependency of the standard library it has to have `Apache-2.0 OR MIT` as the license (plus possibly other license options, if any).

So, this crate can't become an eventual part of the standard library if the Apache-2.0 license isn't added as an option.

(If that's _not_ a goal of this crate that's fine, but if so that should probably be affirmed somewhere, because if that's the case then `winapi` will need to continue to be used by the standard library)

kennykerr:
Fixes #426


Lokathor:
The Cargo.toml didn't get an update in the `license` field.

kennykerr:
https://github.com/microsoft/windows-rs/pull/477

Issue #423, opened 2021-01-22
Add support for Win32 struct fields of array type


thisKai:
In the [windows-rs docs](https://microsoft.github.io/windows-docs-rs/doc/bindings/windows/win32/windows_and_messaging/struct.NCCALCSIZE_PARAMS.html) the `rgrc` field of `NCCALCSIZE_PARAMS` is a `bool`, but the [microsoft docs](https://docs.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-nccalcsize_params) and the [winapi crate docs](https://docs.rs/winapi/0.3.9/winapi/um/winuser/struct.NCCALCSIZE_PARAMS.html) say it should be a `[RECT; 3]`

kennykerr:
Yes, I still need to add support for struct fields of arrays.

pravic:
Why use a wrong type (bool) instead of `todo!()` or something like that? People will struggle to realize what's wrong with their code.

DaLynX:
I'm having the same issue with `bindings::windows::win32::security::TOKEN_PRIVILEGES`.

kennykerr:
Ryan added a tweak to make it easier to identify such APIs. I'll get this fixed asap.

kennykerr:
Sorry for the delay. This is now fixed as of https://github.com/microsoft/windows-rs/pull/608.

PR #481, opened 2021-02-01
Win32 array and pointer type generation for struct members

nathanvoglsam:
This solves #423

After encountering this issue myself and seeing the outstanding issue referring to it I decided to hack on the code to see if I could get it implemented. The link in the code to where to start in the referenced issue was very useful and I now have the functionality implemented.

I'd like some input on what to do with the outstanding "TODO" items I've left

```rust
// Recursively parse the type
let kind = Self::read_from_blob(blob, generics, calling_namespace);

// TODO: Don't know how to handle multi dimensional arrays yet
let rank = blob.read_unsigned();
if rank != 1 {
    panic!("Unsupported array rank: {}", rank);
}

// TODO: Don't know how to handle unsized arrays yet
let num_upper_bounds = blob.read_unsigned();
if num_upper_bounds != 1 {
    panic!("Unsupported array upper bound count: {}", num_upper_bounds);
}

let upper_bound = blob.read_unsigned();

// TODO: Don't know how to handle unsized arrays yet
let num_lower_bounds = blob.read_unsigned();
if num_lower_bounds != 1 {
    panic!("Unsupported array lower bound count: {}", num_lower_bounds);
}

// It's not possible for rust to represent a non 0 lower bound in any sane ABI
// compatible manner without generating a boat load of support code
let low_bound = blob.read_signed();
if low_bound != 0 {
    panic!("Unsupported array lower bound: {}", low_bound);
}
 ```

Going over the ECMA 335 spec for the file format, there is potential for a type that can't be sanely expressed by rust to be encoded in the metadata and I was hoping on getting input on how to resolve that.

Currently I'm just panicking on anything that isn't a single dimensional array, panicking when the upper or lower bound is not specified, and panicking when the lower bound is non zero. I wouldn't expect anything from the win32 API surface area to trigger one of those panics.

### DIFF
--- a/crates/gen/src/type.rs
+++ b/crates/gen/src/type.rs
@@ -46,6 +46,7 @@ pub enum TypeKind {
     Struct(TypeName),
     Delegate(TypeName),
     Generic(&'static str),
+    Array(usize, Box<TypeKind>),
 }
 
 impl TypeKind {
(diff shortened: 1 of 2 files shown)


kennykerr:
Thanks for sharing! This is a pretty complex area that you're attempting to work on. Unless you're an expert in both ECMA-335 and WinRT internals, I'd suggest starting with something a little less ambitious. The code in question is pretty sensitive to change and will require a lot of review and testing and is not something I can easily walk someone through at this point. 

Regarding the bounds, ECMA-335 can express many things that neither WinRT nor Win32 require so it's not unusual to omit support for certain features that the metadata format might accommodate.

nathanvoglsam:
Understandable.

I assumed there was info I was missing as I'm definitely coming in as an outsider to the fine details of WinRT and the winmd format. I made the changes just to see if I could get it working. It's working well for D3D12 and DXGI, which are now much more usable than they were before as many structs had array members and the placeholder `bool` types were big ABI breaks.

I would certainly not call the changes rigorously tested though and would definitely not merge them as is.

My hope from the PR was just to get the ball rolling on the feature as it's a big hole for Win32 stuff. The existing winapi-rs crate is missing a lot of stuff from DX12 and I saw better use of my time improving this crate than just hand writing bindings in winapi-rs. d3d12.rs hasn't been touched in 2 years and there's a lot of holes to fill.

Worst case is I use my fork until the feature is implemented by someone with more knowledge than I.

kennykerr:
Thanks, it's a great start. I'll try to get to this as soon as possible, but I'm glad you were able to get unblocked!

PR #483, opened 2021-02-01
Improve BOOL

rylev:


kennykerr (approved):
Yes, that seems more natural - thanks!

src/result/bool32.rs
```diff
@@ -39,18 +44,54 @@ unsafe impl Abi for BOOL {
     type Abi = Self;
 }
 
+impl From<BOOL> for bool {
+    fn from(value: BOOL) -> Self {
+        value == TRUE
```
lespea, line 49: I'm far from a C expert, but shouldn't this be `value != FALSE` or `value > FALSE`?  I thought any non-zero result was true?  Or is this a "windows bool" thing?

kennykerr: Well spotted - I thought I'd checked every comparison. Indeed, we should always compare against zero/false.

rylev: This makes me think we should not derive `PartialEq` and instead normalize when doing equality checks.

kennykerr: Fixes here: https://github.com/microsoft/windows-rs/pull/492

PR #488, opened 2021-02-01
Fix the name of the 'expect' method on BOOL

rylev:
Whoops!


kennykerr:
I'm fixing a few more bugs in BOOL.

PR #491, opened 2021-02-01
Add spellchecker tutorial

rylev:
This pull request introduces a tutorial on how to use `windows-rs`. Included are the following:
* [x] Example code base
* [x] Written tutorial explaining the code


examples/spellchecker/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
kennykerr, line 1: If we're adding samples here, perhaps I should just move all the samples from here:

https://github.com/kennykerr/samples-rs

examples/spellchecker/bindings/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    windows::build!(
+        windows::win32::intl::{ISpellChecker, SpellCheckerFactory, ISpellCheckerFactory},
+        windows::win32::com::*
```
kennykerr, line 4: This generates a lot of code that is probably unused. Perhaps limit it to just the COM declarations you need.

examples/spellchecker/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+use win32::intl;
+use windows::*;
+
+fn main() {
+    let input = std::env::args()
+        .nth(1)
+        .expect("Expected one command line argument for text to be spell-corrected");
+    unsafe {
+        // Initialize COM apartment
+        com::CoInitializeEx(
```
kennykerr, line 12: I may provide a one-line helper for this directly in the windows crate.

examples/spellchecker/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+
+    // Create ISpellCheckerFactory
+    let factory = create_factory();
+
+    // Make sure that the "en-US" locale is supported
+    let mut supported: BOOL = false.into();
+    let locale: Vec<u16> = "en-US\0".encode_utf16().collect();
+    // https://github.com/microsoft/windows-rs/issues/489
+    factory
+        .IsSupported(locale.as_ptr(), &mut supported as *mut _ as *mut i32)
```
kennykerr, line 27: Just FYI - I plan to map trailing out params as `Result<T>` so that this would simply become:

```rust
let supported = factory.IsSupported(...).ok()?;
```

examples/spellchecker/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+        .unwrap();
+    supported.expects("en-US is supported");
+
+    // Create a ISpellChecker
+    let mut checker: Option<intl::ISpellChecker> = None;
+    factory
+        // https://github.com/microsoft/windows-rs/issues/486
+        .CreateSpellChecker(locale.as_ptr(), unsafe { checker.set_abi() as _ })
```
kennykerr, line 35:
```suggestion
        .CreateSpellChecker(locale.as_ptr(), &mut checker)
```

examples/spellchecker/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+    let mut supported: BOOL = false.into();
+    let locale: Vec<u16> = "en-US\0".encode_utf16().collect();
+    // https://github.com/microsoft/windows-rs/issues/489
+    factory
+        .IsSupported(locale.as_ptr(), &mut supported as *mut _ as *mut i32)
+        .unwrap();
+    supported.expects("en-US is supported");
+
+    // Create a ISpellChecker
+    let mut checker: Option<intl::ISpellChecker> = None;
```
kennykerr, line 32:
```suggestion
    let mut checker = None;
```

kennykerr: A few more examples of this below.

examples/spellchecker/src/main.rs
```diff
@@ -0,0 +1,148 @@
+use bindings::windows::win32;
+use win32::com;
+use win32::intl;
+use windows::*;
+
+fn main() {
```
kennykerr, line 6: Might be more readable if this returned `Result<()>` and you use `?`.

examples/spellchecker/src/main.rs
kennykerr: `initialize_mta()`

examples/spellchecker/src/main.rs
```diff
@@ -0,0 +1,148 @@
...
+            None,
+            // issue: https://github.com/microsoft/windows-rs/issues/482
+            com::CLSCTX::CLSCTX_INPROC_SERVER.0 as u32,
+            &intl::ISpellCheckerFactory::IID,
+            factory.set_abi(),
+        )
+    }
+    .unwrap();
+    factory.unwrap()
+}
```
kennykerr, line 131:
```suggestion
let factory: ISpellCheckerFactory = create_instance(&SpellCheckerFactory)?;
```

examples/spellchecker/Cargo.toml
rylev: If you're comfortable with that, I'd love that!

examples/spellchecker/bindings/build.rs
rylev: Oh yes, this helped a lot!

examples/spellchecker/Cargo.toml
```diff
@@ -0,0 +1,9 @@
...
+name = "spellchecker"
+version = "0.1.0"
+authors = ["Microsoft"]
+edition = "2018"
+
+[dependencies]
+spellchecker_bindings = { path = "./bindings" }
```
kennykerr, line 8:
```suggestion
bindings = { package="spellchecker_bindings", path = "bindings" }
```

kennykerr: That way the code can still refer to the simpler `bindings`.

examples/spellchecker/bindings/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    windows::build!(
+        windows::win32::intl::{ISpellChecker, SpellCheckerFactory, ISpellCheckerFactory},
+        windows::win32::com::{CoInitializeEx, CoTaskMemFree, IEnumString, CoCreateInstance, COINIT, CLSCTX}
```
kennykerr, line 4:
```suggestion
        windows::win32::com::{CoTaskMemFree, IEnumString}
```

examples/spellchecker/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+    let input = std::env::args()
+        .nth(1)
+        .expect("Expected one command line argument for text to be spell-corrected");
+    windows::initialize_mta()?;
+
+    // Create ISpellCheckerFactory
+    let factory: intl::ISpellCheckerFactory = windows::create_instance(&intl::SpellCheckerFactory)?;
+
+    // Make sure that the "en-US" locale is supported
+    let mut supported: BOOL = false.into();
```
kennykerr, line 15:
```suggestion
    let mut supported = FALSE;
```

examples/spellchecker/src/main.rs
Nerixyz: Is there any issue tracking this? I think it would be great to do this as it would make developing around these interfaces a lot easier.

kennykerr: https://github.com/microsoft/windows-rs/issues/509

docs/getting-started.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+And we're done bootstraping the project. Now we'll move on to the code of the `spellchecker` crate which will use both the generated bindings from the `bindings` crate as well as helper functionality from the `windows` crate.
+
+## Initialization Code
+
+Inside of the main.rs file for the `spellchecker` crate, we'll start by adding the set up code. We'll initialize the COM runtime on the main thread. We'll make the `main` function return a `windows::Result` which is simply a standard `Result` type with the error hardcoded as a `windows::Error`.
+
+```rust 
+fn main() -> window::Result<()> {
```
jyn514, line 82:
```suggestion
fn main() -> windows::Result<()> {
```

kennykerr:
Now that we have https://github.com/microsoft/win32metadata/issues/207 this tutorial should become a lot simpler. I'll work on getting that into the windows crate and then we can revisit this tutorial.

rylev:
@kennykerr I think now is a good time to get this tutorial in. Things will continue to evolve, but having it in a branch isn't very helpful.

docs/getting-started.md
```diff
@@ -0,0 +1,209 @@
...
+
+We can generate our bindings inside of the `bindings` crate. To do this, we'll need to add the `windows` crate as both a regular dependency and build dependency of the `bindings` crate. In the Cargo.toml for the `bindings` crate add the following:
+
+```toml
+# ^^ The rest of the Cargo.toml remains the same
+[dependencies]
+windows = { git = "https://github.com/microsoft/windows-rs" }  # TODO: change this to the crates.io version when that gets released
+
+[build-dependencies]
+windows = { git = "https://github.com/microsoft/windows-rs" }  # TODO: change this to the crates.io version when that gets released
```
kennykerr, line 35: TODO

kennykerr (approved):

PR #493, opened 2021-02-01
Add must_use attribute to ErrorCode

### LINKED ISSUE #489: Consider marking all COM functions as `#[must_use]`.
It's probably always necessary to check the result of a COM method call. In Rust, we can mark COM methods as `#[must_use]` in order to ensure that users check the result from calling the COM method.

kennykerr:
Fixes #489


rylev (approved):
🎉

PR #496, opened 2021-02-01
Add basic COM helpers

### LINKED ISSUE #495: Add CoCreateInstance wrapper
This is such a common construction function that a wrapper would make a lot of common scenarios a lot simpler.

https://github.com/kennykerr/samples-rs/blob/40f7c8b0ed13ffe370c193365f104e26e862dfbd/clock/src/main.rs#L552-L563

This example looks as follows:

```rust
fn create_manager() -> Result<IUIAnimationManager> {
    unsafe {
        let mut manager: Option<IUIAnimationManager> = None;

        CoCreateInstance(
            &UIAnimationManager,
            None,
            CLSCTX::CLSCTX_INPROC_SERVER.0 as u32,
            &IUIAnimationManager::IID,
            manager.set_abi(),
        )
        .and_some(manager)
    }
}
```

This could be simplified to a safe one-liner:

```rust
let manager: IUIAnimationManager = create_instance(&UIAnimationManager)?;
```

kennykerr:
These helpers simplify calling many COM APIs. Fixes #495 and addresses many of the recent questions about using the `windows` crate to call various APIs.

### DIFF
--- a/src/runtime/com.rs
+++ b/src/runtime/com.rs
@@ -1,0 +1,44 @@
+use crate::*;
+
+/// Initializes COM for use by the calling thread for the multi-threaded apartment (MTA).
+pub fn initialize_mta() -> Result<()> {
+    unsafe {
+        CoInitializeEx(0, 0).ok() // COINIT_APARTMENTTHREADED
+    }
+}
+
+/// Initializes COM for use by the calling thread for a single-threaded apartment (STA).
+pub fn initialize_sta() -> Result<()> {
+    unsafe {
+        CoInitializeEx(0, 2).ok() // COINIT_MULTITHREADED
+    }
+}
+
+/// Creates a COM object with the given CLSID.
+pub fn create_instance<T: Interface>(clsid: &Guid) -> Result<T> {
+    let mut object = None;
+
+    unsafe {
+        CoCreateInstance(
+            clsid,
+            None,
+            23, // CLSCTX_ALL
+            &T::IID,
+            object.set_abi(),
+        )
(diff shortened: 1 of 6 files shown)


src/runtime/com.rs
```diff
@@ -0,0 +1,44 @@
...
+pub fn initialize_mta() -> Result<()> {
+    unsafe {
+        CoInitializeEx(0, 0).ok() // COINIT_APARTMENTTHREADED
+    }
+}
+
+/// Initializes COM for use by the calling thread for a single-threaded apartment (STA).
+pub fn initialize_sta() -> Result<()> {
+    unsafe {
+        CoInitializeEx(0, 2).ok() // COINIT_MULTITHREADED
```
rylev, line 13: Nit: can you use a constant instead of the raw number plus the comment in all these cases?

src/runtime/com.rs
```diff
@@ -0,0 +1,44 @@
+use crate::*;
+
+/// Initializes COM for use by the calling thread for the multi-threaded apartment (MTA).
+pub fn initialize_mta() -> Result<()> {
```
rylev, line 4: Aren’t the names of the functions backwards? This initializes for a single threaded apartment but is called `mta`

kennykerr: Yes, the comments are backwards. I'll switch to constants. 😉

not-holar:
To anyone finding this later and getting confused, these have been removed: https://github.com/microsoft/windows-rs/pull/998

PR #498, opened 2021-02-02
Make methods unsafe if any of there parameters are pointers

### LINKED ISSUE #485: Safety of API bindings
Currently COM methods are not marked as unsafe, but they should be especially if they take any kind of raw pointers as arguments. COM methods with no arguments could potentially be marked safe since presumably, it's part of the COM ABI that the "this" pointer is always valid if there are valid handles to the interface, but I'm unsure if this reasoning fully holds.

rylev:
Fixes #485.

This is better but this is still not correct in the general case (since we don't control all WinRT or COM interfaces that will ever pass through the code generator), and I'm not sure it's even correct in the case of Windows APIs. Take the following as an example of a COM interface method that would be marked safe by this tooling but is not safe. I've written it in Rust even though this may obviously be written in another language.

```rust
/// It is up to the caller to *never* pass `TRUE` as the `do_terrible_things` param,
unsafe fn my_method(&self, do_terrible_things: BOOL) -> HRESULT {
   if do_terrible_things.as_bool() {
      /// Literally do anything including dereferencing random memory
  }
}
```

This example is admittedly a bit contrived, but hopefully you get the point that pointer arguments are not the only sign of a whether a COM function is safe. It's unfortunate, but it's probably correct to mark any COM interface that has any parameters (beyond the this pointer) as unsafe.

### DIFF
--- a/crates/gen/src/class.rs
+++ b/crates/gen/src/class.rs
@@ -115,7 +115,7 @@ impl Class {
     pub fn gen(&self) -> TokenStream {
         let name = self.name.gen();
         let type_name = self.type_name(&name);
-        let methods = gen_method(&self.interfaces);
+        let methods = gen_methods(&self.interfaces);
         let call_factory = self.gen_call_factory();
 
         if let Some(default_interface) = self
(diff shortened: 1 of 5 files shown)


kennykerr:
Agreed this is not a good measure of safety.

PR #503, opened 2021-02-02
Use COM types and traits from the com crate

hadeutscher:
This PR unifies the Abi, Guid and Interface types/traits with those defined in the `com` crate. In addition to removing code redundancy, this allows us to use facilities provided by the `com` crate (e.g. producing a com class that extends an interface) with the types defined in this crate.

The matching PR in `com-rs`: https://github.com/microsoft/com-rs/pull/205
Until it's merged, I've pinned the com dependency to my git repo.

NOTE: this PR is not ready yet as I still have some tests to fix, I am opening it beforehand so that you can let me know if you want anything done differently.


kennykerr:
Thanks for sharing! However, we are not planning to have the `windows` crate take a dependency on the `com` crate. Rather, we expect the `windows` crate to subsume much of what the `com` crate does today. In particular, the ability to [author COM classes is tracked here](https://github.com/microsoft/windows-rs/issues/81) and I should have that working soon.

kennykerr:
You can now implement many COM interfaces using the Windows crate. Let me know if you experience any issues.

PR #506, opened 2021-02-03
Add unsupported type

rylev:
As noted in #423, using `BOOL` as a stand-in type for another unsupported type can be very confusing. This exchanges that out for a temporary type explicitly named `NOT_YET_SUPPORTED_TYPE` which is not constructible and can never be used. Hopefully this makes it a bit clearer when a particular API is not yet fully supported.

### DIFF
--- a/crates/gen/src/type.rs
+++ b/crates/gen/src/type.rs
@@ -46,6 +46,9 @@ pub enum TypeKind {
     Struct(TypeName),
     Delegate(TypeName),
     Generic(&'static str),
+    /// A type that hasn't been supported yet.
+    /// For example, mutlidimensional arrays are not yet supported
+    NotYetSupported,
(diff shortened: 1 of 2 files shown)


crates/gen/src/type.rs
```diff
@@ -46,6 +46,9 @@ pub enum TypeKind {
     Struct(TypeName),
     Delegate(TypeName),
     Generic(&'static str),
+    /// A type that hasn't been supported yet.
+    /// For example, mutlidimensional arrays are not yet supported
```
pravic, line 50: typo: multidimensional

pravic (approved):

kennykerr (dismissed):
Nice!

kennykerr (approved):

Issue #484, opened 2021-02-01
Support wide strings


rylev:
Wide strings are common enough in win32 programming that it'd be a helpful to provide utilities that make it easier to work with them. In particular providing a macro for constructing wide strings at compile time. There is a fairly popular crate that provides some of these utilities already: [widestring](https://crates.io/crates/widestring), but that crate uses winapi (for obvious reasons). It's not clear what the best path forward is.

kennykerr:
It's not just wide strings, but the particular ownership models prescribed by different Windows APIs. Some use HSTRING, some BSTR, some use the COM task allocator, and some are simply views or slices. We have enough metadata to distinguish those and then make them available directly as unique types. We can then use the `Param` trait to support conversion and string literal conversion as we do for HSTRING.

kennykerr:
We've started annotating the metadata with enough information to be able to distinguish between the different string types. Some work still remains in metadata.

For example, `Windows.Win32.Com.IUri` now uses the `Windows.Win32.Automation.BSTR` type. This should be mapped to the `windows::BString` type. This is typically only used for out/return params.

Other APIs might be attributed to indicate that they expect either a utf8 and utf16 slice (not allocator implied). Examples of this can be found on the `CreateEventA` and `CreateEventW` functions. You can see the trailing parameter is either `[NativeTypeInfo(UnmanagedType.LPStr, IsNullTerminated = true)]` or `[NativeTypeInfo(UnmanagedType.LPWStr, IsNullTerminated = true)]`. When these appear on a COM interface, we can wrap them with `Param<T>`, but not with function imports like this. In that case, we need to provide some convenience macro perhaps because the ABI signature must remain intact. 

`FormatMessageW`/`FormatMessageA` is another example where the resulting string must be freed/dropped with with `LocalFree` in most (but not all) cases. Tricky stuff. 

Also, remember that in most cases the strings must be null terminated.

Lokathor:
check out the `utf16_lit` crate, which lets you make utf16 literals as a macro_rules

kennykerr:
#1891 introduced string literal macros that address this issue.

PR #514, opened 2021-02-04
Add a costring type

rylev:
This is the beginnings of addressing #484.

There are certainly more features we could add to this type. I thought we'd start first with enough to get the spellchecker demo working with it, and a test.

### DIFF
--- a/src/strings/costring.rs
+++ b/src/strings/costring.rs
@@ -16,6 +16,7 @@ impl CoString {
         I: IntoIterator<Item = u16>,
     {
         let start = CoTaskMemAlloc(len * 2 + 2) as *mut u16;
+        assert!(!start.is_null(), "Could not allocate memory for CoString");
         let mut cursor = start;
         for (index, c) in iter.into_iter().enumerate() {
             debug_assert!(index < len);
--- a/src/strings/costring.rs
+++ b/src/strings/costring.rs
@@ -1,0 +1,135 @@
+/// A null-terminated wide-string that has been allocated with `CoTaskMemAlloc` or `CoTaskMemRealloc`
+/// and is freed with `CoTaskMemFree`.
+#[repr(transparent)]
+pub struct CoString(*mut u16);
+
+impl CoString {
+    /// Create a new `CoString`
+    pub fn new() -> Self {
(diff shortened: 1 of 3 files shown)


src/strings/costring.rs
```diff
@@ -0,0 +1,136 @@
...
+        for (index, c) in iter.into_iter().enumerate() {
+            debug_assert!(index < len);
+            cursor.write(c);
+            cursor = cursor.add(1);
+        }
+        cursor.write(0);
+        Self(start)
+    }
+
+    /// Get the string as 16-bit wide characters (wchars).
```
kennykerr, line 30: Perhaps just a comment explaining why this doesn't return a slice (because the size is not computed) as that would otherwise be very handy for Windows development. Fortunately HString provides this so there's no compelling need for CoString to do so.

kennykerr (approved):

src/strings/costring.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn from(s: String) -> Self {
+        s.as_str().into()
+    }
+}
+
+impl std::fmt::Display for CoString {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        use std::fmt::Write;
+        for c in std::char::decode_utf16(self.as_wide()) {
+            f.write_char(c.unwrap())?
```
lzybkr, line 94: @rylev - Why panic here on invalid utf16?
Maybe `unwrap_or(std::char::REPLACEMENT_CHARACTER)` would be appropriate?

src/strings/costring.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn drop(&mut self) {
+        // Pointer is allowed to be null
+        unsafe { CoTaskMemFree(self.0 as _) }
+    }
+}
+
+impl From<&CoString> for String {
+    fn from(cs: &CoString) -> Self {
+        std::char::decode_utf16(cs.as_wide())
+            .map(|r| r.unwrap())
```
lzybkr, line 61: @rylev - because of the potential to panic here, maybe the impl should be on `Result<String, ?>` so the caller can decide how to handle the error?

PR #526, opened 2021-02-08
Add WString type

### LINKED ISSUE #484: Support wide strings

rylev:
Fixes #484 (We should make new issues for any remaining string types).

This takes a dependency on https://github.com/rylev/const-utf16.

We can improve `WString` a lot, but this is enough to handle the Spellchecker example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ readme = "README.md"
 windows_macros = { path = "crates/macros",  version = "0.3.1" }
 gen = { package = "windows_gen", path = "crates/gen",  version = "0.3.1" }
 const-sha1 = "0.2"
-const-utf16 = { git = "https://github.com/rylev/const-utf16", branch = "main" }
+const-utf16 = "0.2"
 
 [build-dependencies]
 windows_macros = { path = "crates/macros",  version = "0.3.1" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ readme = "README.md"
 windows_macros = { path = "crates/macros",  version = "0.3.1" }
 gen = { package = "windows_gen", path = "crates/gen",  version = "0.3.1" }
 const-sha1 = "0.2"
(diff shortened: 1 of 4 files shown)


kennykerr:
I'd like to avoid having a few dozen different string types if possible. 😉 If there's a way we can reuse what we have I think that would be preferable. Having so many options doesn't improve usability. I have been using `HString` in places where a null terminated UTF16 string is needed. It is only when we have unique allocation requirements that a new string type may be necessary. 

There may be value in compile time string literals to avoid allocations but I'm not yet convinced that is something `windows` needs to provide.

rylev:
@kennykerr While I'm sympathetic to the desire to not have many types of strings, isn't this the reality of the Windows API? Doesn't an `HString` assume a certain allocation (i.e., that it is a pointer to a header which contains another pointer to the data and a reference count)? The `WString` is the simplest one of all, it simply assuming an owned buffer of `u16` that are terminated by null byte. On drop the global allocator will free that data. I'm not sure how `HString` could be used to model that.

kennykerr:
Yes `HString` can be used in more places because it guarantees a certain allocation model, but it is also a "an owned buffer of u16 that are terminated by null" so it is compatible with APIs that are expecting a `LPCWSTR` input parameter.

kennykerr:
How about we work on improving the metadata and parsing around these types and then loop back on whether we need additional string types?

kennykerr:
New metadata is coming that provide utf8 and utf16 slices (e.g. `PWSTR` and `PSTR`) and will help to clean up a lot of the string handling. Going to close this PR for now as that should address much of this.

PR #531, opened 2021-02-09
Update win32 metadata

kennykerr:
Now using `10.0.19041.5-preview.2` from https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/

### DIFF
--- a/crates/tests/tests/win32.rs
+++ b/crates/tests/tests/win32.rs
@@ -1,5 +1,3 @@
-use windows::Abi;
-
 use tests::{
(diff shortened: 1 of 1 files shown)


Josh015:
You might also want to update lines 122, 123, & 156 in "crates/tests/tests/win32.rs" to use the new TRUE, FALSE constants.

kennykerr:
Yes, I'm planning to refactor testing soon. It needs a more systematic approach, now that it's mostly up and running.

PR #554, opened 2021-02-19
Fix readme.md adding , into the end of the lines on build.rs example

pauloalves86:
The current example does not work giving the lack of commas at the end of lines.
This is a simple fix.


kennykerr (approved):
Thanks!

agausmann:
Last I knew, comma separators were not part of the `build!` macro syntax. Did that change recently?

kennykerr:
Yes, that changed in #448. The build macro now expects the use paths to be comma separated. This makes it a lot easier to copy and paste the paths between the build macro and the source file where those paths are actually used as the same syntax is used for the build macro and for `use` declarations (which requires commas).

PR #557, opened 2021-02-23
Make Guid Copy

fghzxm:


kennykerr:
Thanks for the contribution! Please start by creating an issue describing the problem or motivation for such a change along with a simple example illustrating the issue.

Issue #561, opened 2021-02-24
Consider adding from_thread method to Error


kennykerr:
https://github.com/microsoft/windows-rs/issues/559#issuecomment-785378835

kennykerr:
I don't remember why I was going to do this. I'll leave it until someone finds a good reason. 😉

Issue #559, opened 2021-02-24
Question: GetLastError and human readable message


wbuck:
When some `Win32` functions fail they will return `0` which indicates that we need to check `GetLastError` in order to attain the last error code. We can then subsequently call `FormatMessageW` if we want a human readable error message.

Is this process still the same using `windows-rs`? 
Are there any helper functions to assist? 

For example in `C++` we can use the `std::error_code` (with `std::system_category`) to convert the error code returned from `GetLastError` into a human readable `string`.

kennykerr:
WinRT methods all return `std::result::Result<T, windows::Error>` aka `windows::Result`. 

Some COM and Win32 functions return either `windows::ErrorCode` (an `HRESULT`) or `windows::BOOL`. Both of these provide `ok()` methods that turn the result into `windows::Result`. This is useful for short-circuiting with `?`. There are plenty of examples of this [in the examples](https://github.com/microsoft/windows-rs/blob/master/examples/clock/src/main.rs).

You can also directly create `windows::ErrorCode` based on the error code stored on the calling thread via `ErrorCode::from_thread`. This calls `GetLastError` under the hood. 

The `windows::ErrorCode` doesn't provide a way to retrieve a message. Perhaps it should.

`windows::Error` doesn't provide a `from_thread` method, but I plan to add one. It does however provide a `message()` method that will call `FormatMessage` under the hood.

For now, you can do this:

```rust
let e = Error::fast_error(ErrorCode::from_thread());

println!("{:?}", e);
println!("{} {}", e.code().0, e.message());
```

Output:

```
Error { code: 0x00000000, message: "The operation completed successfully." }
0 The operation completed successfully.
```

wbuck:
All of the functions I have used thus far return a `DWORD` (`u32`) which either indicates the count of something (ex. number of characters copied into the output buffer) or `0` if an error occurred. In the case of a `0` return I would typically call `GetLastError`.

The snippet you provided is perfect, thanks you.

kennykerr:
`ErrorCode` now provides a `message` method directly.

fenhl:
What is the status of this? `windows::ErrorCode` seems to have disappeared.

tim-weis:
[...]

Issue #570, opened 2021-02-27
How to use an ErrorCode from a returned Result<_, Error>?


zaddach:
I'm trying to figure out how to use the returned Error/ErrorCode when I need to check for a certain value. Let's say I'm using `ReadFile`, and I want to check if `GetLastError()` was `ERROR_IO_PENDING` when using overlapped input. How would I do this nicely?

My idea was:
```rust
const ERROR_IO_PENDING: ErrorCode = ErrorCode::from_win32(bindings::windows::win32::system_services::ERROR_IO_PENDING);

let mut overlapped = OVERLAPPED::default();
match ReadFile(_, _, _, _, _, &mut overlapped).ok() {
    Ok(_) => ...,
    Err(e) if e.code() == ERROR_IO_PENDING => ...,
    Err(e) => ...,
}
```
Two drawbacks:
- ErrorCode::from_win32 is crate-only pub, so I need to copy-paste the function into my own code.
- It would be nicer if `windows-rs` could export the `ErrorCode` constants somewhere.

Is this how the ErrorCode is intended to be used?

kennykerr:
A few thoughts.

* We're trying to improve the metadata for error codes. https://github.com/microsoft/win32metadata/issues/159
* Then we'll likely individual types for HRESULT, WIN32, and NTSTATUS that provide common const values for such error codes.
* OVERLAPPED doesn't work yet as it requires nested structs and unions. That should be working in a few days.

zaddach:
Hey, thanks for the pointer to the discussion on the win32metadata repo. Sorry for spamming you a bit with issues, I'm just trying to keep track of everything that doesn't feel ergonomic to me in the windows-rs crate usage right now. The OVERLAPPED thing was just an example, there are other functions like [InitializeProcThreadAttributeList](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-initializeprocthreadattributelist) that return an error on successful operation by API convention which the user should check for. I'll comment that on the other discussion and close the issue here.

kennykerr:
No worries, appreciate the feedback - keep it coming! 😉

kennykerr:
Unions are now supported. Here's an example of using OVERLAPPED: 

https://github.com/microsoft/windows-rs/tree/master/examples/overlapped

PR #579, opened 2021-03-09
Error handling and enum flags

### LINKED ISSUE #576: Consider making ErrorCode::from_win32 public
There are many Win32 APIs that return error codes as LSTATUS (such as the registry APIs) or DWORD values (such as WaitForSingleObject), and it would be nice to be able to easily convert these into an HRESULT -> ErrorCode.

Currently the only way to do this is to use ErrorCode::from_thread(), which uses the value from GetLastError(), or to write your own From<LSTATUS> that copies what from_win32() is doing. Functions such as RegQueryValue don't explicitly say that they will also call SetLastError() with the error code that they are returning, so the latter option seems like the only safe choice.

This also applies to dealing with NTSTATUS codes, which I'm sure nobody using windows-rs should have any business dealing with, but is nevertheless a type exposed through windows::win32::system_services::NTSTATUS.

kennykerr:
This update improves error handling support by adding a message method to the ErrorCode types so you don't have to create a complete Error object just to get a formatted message. 

It also adds bitwise operators for Win32 enums.

And more of the windows crate is dogfooding the generated bindings. 😉 

Fixes #576 #561 #559

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,6 +3,6 @@ fn main() {
         windows::foundation::{IReference, IStringable, PropertyValue},
         windows::win32::automation::{BSTR, GetErrorInfo, IErrorInfo, SetErrorInfo},
         windows::win32::winrt::{IRestrictedErrorInfo, ILanguageExceptionErrorInfo2},
-        windows::win32::debug::GetLastError,
+        windows::win32::debug::{GetLastError, FormatMessageW},
     );
 }
--- a/crates/gen/src/types/enum.rs
+++ b/crates/gen/src/types/enum.rs
@@ -46,10 +46,17 @@ impl Enum {
         let name = self.0.gen_name(gen);
         let underlying_type = self.underlying_type();
 
-        // TODO: add bitwise operators if its u32 *or* if its got the Win32 flags attribute
+        // WinRT enums don't have the flags attribute but are paritioned merely based
+        // on whether they are signed.
         let bitwise = match underlying_type {
-            ElementType::I32 => TokenStream::new(),
-            ElementType::U32 | ElementType::U16 => quote! {
+            ElementType::U32 => true,
+            _ => false,
+        };
+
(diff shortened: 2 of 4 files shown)


jaysonsantos:
Hi there @kennykerr, sorry to comment on a closed issue, and thanks for these implementations, they are making dealing with windows API way easier and pleasant to work!
I saw that win32metadata already has an issue on the topic as you shared here https://github.com/microsoft/windows-rs/issues/570
In the meantime, what is the suggestion to work with `ErrorCode`? For example [PasswordVault.retrieve](https://github.com/jaysonsantos/keyring-rs/blob/99fa4fb19b2bcaa615b83e091cb6cbf33db7ed73/src/windows.rs#L60-L62) can return `0x80070490` which is something like not found and to deal with that specific error I created a [constant](https://github.com/jaysonsantos/keyring-rs/blob/99fa4fb19b2bcaa615b83e091cb6cbf33db7ed73/src/windows.rs#L12) so I [can map the error](https://github.com/jaysonsantos/keyring-rs/blob/99fa4fb19b2bcaa615b83e091cb6cbf33db7ed73/src/windows.rs#L88-L93).
Would it be better to compare that error code by using `ErrorCode.from_win32(bindings::windows::win32::system_services::ERROR_NOT_FOUND)` or should it be something else?

kennykerr:
Hi Jayson, the [documentation](https://docs.microsoft.com/en-us/uwp/api/windows.security.credentials.passwordvault.retrieve?view=winrt-19041) doesn't indicate any specific error codes. I would thus avoid checking the result against any specific codes unless there's a compelling reason to do so. Instead it says, "If a match does not exist, an exception will throw." That just means an error will be returned if the value cannot be retrieved. The actual error code may still be useful for diagnostics but probably not control flow.

Ideally `is_err` will suffice because the error code itself is not presumed to be stable. So for the `retrieve` method I would just map any failure to `NoPasswordFound`. That's my best guess having not used this API before. 😄

PR #587, opened 2021-03-10
Introduce conservative dependency tracking 

### LINKED ISSUE #567: Consider non-greed dependency tracking option
The build macro chases all dependencies and includes the definitions for any types referred to by the imported types whether directly or indirectly. This can include a lot of code transitively. 

Consider an option that would elide any methods on COM and WinRT interfaces that refer to types not specifically mentioned in the build macro. The vtable slots would be preserved but the projected methods would be omitted. This has the potential of generating far less code and thus speeding up builds considerably.

kennykerr:
This update avoids chasing dependencies from interface methods in order to reduce the amount of code that is generated. If you include a class or interface in the build script, it will only generate the methods of those types if their signatures only refer to types included by the build script. This can have a huge impact on build time because one method, which you may not even use, may refer to another type, and its methods may refer to other types, leading to a cascade of definitions that need to be generated. 

As an extreme example, consider the following build script:

```rust
fn main() {
  windows::build!(
      windows::ui::xaml::Application
  );
}
```

Here I'm only asking for the Xaml Application class. Prior to this update it would result in a 179,897 line windows.rs weighing in at 8.7 MB and take about 51 seconds to compile. With this update, it generates only 453 lines of code and takes only 21 seconds to compile. And that 21 seconds is mostly dominated by building crate dependencies. 

Fixes #567

### DIFF
--- a/tests/wildcard/Cargo.toml
+++ b/tests/wildcard/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "test_wildcard"
+version = "0.4.0"
+authors = ["Microsoft"]
+edition = "2018"
+
+[dependencies]
+windows = { path = "../.." }
+gen = { package = "windows_gen", path = "../../crates/gen",  version = "0.4.0" }
+
+[build-dependencies]
+windows = { path = "../.." }
--- a/tests/wildcard/build.rs
+++ b/tests/wildcard/build.rs
@@ -1,0 +1,5 @@
+fn main() {
+    windows::build!(
+        windows::foundation::*,
+    );
+}
--- a/tests/wildcard/src/lib.rs
+++ b/tests/wildcard/src/lib.rs
@@ -1,0 +1,1 @@
+::windows::include_bindings!();
--- a/tests/wildcard/tests/wildcard.rs
+++ b/tests/wildcard/tests/wildcard.rs
@@ -1,0 +1,9 @@
+use test_wildcard::windows::foundation::Uri;
+
+#[test]
+fn test() -> windows::Result<()> {
+    let uri = Uri::create_uri("http://kennykerr.ca")?;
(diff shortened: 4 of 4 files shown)


rylev:
This is great but it does present a usability issue. Now when a user does not generate a type all methods that take that type as a parameter are also not generated. This can lead to a confusing situation where the user does not know why a method is not available for use. The compiler simply reports that that method does not exist. Perhaps this algorithm is too aggressive.  I'm not sure exactly how to improve this but two potential different ways of handling this are:
* If a user asks for type `T` all methods on type `T` are generated including any types used as parameters or return types. However, no methods are generated for type `T` itself. This prevents cascading, but it might still be confusing to the user that the type `T` is available but does not have any methods on it. 
* If a method takes a parameter for which the type is not generated, put a placeholder type like `TYPE_NOT_GENERATED`, `UNGENERATED<T>`, or something similar. This way, the method itself exists, but the user gets some clear indication that the type is not generated. This prevents cascading and gives the user an indication of the issue, but is a bit ugly and can be confusing for newcomers.

kennykerr:
I'd love to be able to tap into the compiler's error message generation to provide a more meaningful error. For example, when you type a method that doesn't exist but a similar one does then the compiler suggests the alternative. If we could hook the compiler to let us take a stab at the error message then we could do a quick lookup and determine that the method exists but is excluded and provide more actionable feedback to the developer.

rylev:
@kennykerr while that does sound like the ideal solution, getting something like that into Rust will be a very long process. I think we should discuss more actionable items for now before considering that ideal. I'm not personally convinced that the way it currently is in master is the best we can do given current constraints.

kennykerr:
Of course, happy to discuss ideas. That was just me dreaming. 😉 Feel free to open an issue.

Issue #645, opened 2021-03-25
Request for implementations of From/Into to convert between PWSTR/PCWSTR and String/&str


wravery:
Most COM and Win32 apps `#define UNICODE` and prefer the `W` APIs which take `PWSTR` or `PCWSTR`. Rust has built-in `String::encode_utf16` and `String::from_utf16` methods, it would be great if `PWSTR` and `PCWSTR` could implement `From`/`Into` traits to encapsulate that.

It's tricky since it needs extra storage, but maybe the tuple could be extended with an `, Option<Vec<u16>>` to hold the extra allocation, and the `.0` could point to that when this is used.

kennykerr:
This already works for input parameters. You can pass a string to any function expecting a `PSTR` or `PWSTR` and it will create the null-terminated string of the appropriate encoding. For example, the following just works today:

```rust
 SetWindowTextA(handle, "Hello A");
 SetWindowTextW(handle, "Hello W");
```

I'm still working on making it easier for `PWSTR` and `PSTR` output parameters. That is tracked by #595.

Issue #581, opened 2021-03-09
Create PWSTR and PSTR from literals


rylev:
Right now literals need to be copied into new buffers every time one is passed into a function that expects a `PWSTR`. Instead we should provide some type of macro convenience for creating `PSTR` and `PWSTR` from literals. These would encode the literals properly and add a trailing `\0` at the end all at compile time. Bonus points if we can make macro that works for either kind of string pointer.

lespea:
Could something like this work? https://github.com/Lokathor/utf16_lit

rylev:
Yes, I also have my own crate that does this https://github.com/rylev/const-utf16 which introduced const compilation a bit before `utf16_lit` did. I think either one would be fine to use.

kennykerr:
Do we need to use a macro? It would be nice if we could (also) write a const function to do this so you can write something like this:

```rust
let s = PSTR::from("hello");
```

It would somehow have to distinguish between a string literal and a regular string reference. I'm not sure if that's possible.

agausmann:
Any function that takes a `&str`, `&[u8]`, `&[u16]` etc with the goal of null-terminating it or transcoding it (ie. UTF-8 to UTF-16) would have to use another buffer to store the result, because the value can't be modified in-place. To get around that, the idea here is to make a new literal string that gets stored in the program's static storage at compile-time.

Functions cannot distinguish literals in their input (although you can get close with `&'static str`), and furthermore, they cannot create new literals (even if they're `const`). Literals can only be created by using literal syntax, so they have to be written in the code, either directly or by a macro.

For example, to convert a literal `b"abcd"` to one that is null-terminated, we can write a macro to rewrite the literal, syntactically, to `b"abcd\0"`. That's the only way to do so at compile-time.

1Dragoon:
What's the current idiomatic way of creating a PWSTR from a string literal? I used this but it doesn't appear to work when I give it to a function that needs a pwstr for an input:

```rust
let text = String::from("my pwstr literal").encode_utf16().collect::<Vec<u16>>();
let mypwstr = PWSTR(text.as_mut_ptr());
```

kennykerr:
The API is likely assuming a null terminator.

agausmann:
Yep, I'd recommend using a crate like `widestring`. I use `widestring::U16CString` which automatically converts strings to UTF-16 and  null-terminates them.

1Dragoon:
[...]

PR #647, opened 2021-03-26
Build a WebView2 sample with the WinRT APIs

### LINKED ISSUE #538: Add WebView2 sample

wravery:
Fixes #538


examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+        if request.len() == 3 {
+            if let Some(Value::String(str)) = request.get(0) {
+                if str == "Add" {
+                    if let Some(Value::Number(a)) = request.get(1) {
+                        if let Some(Value::Number(b)) = request.get(2) {
+                            let result = a.as_f64().unwrap_or(0f64) + b.as_f64().unwrap_or(0f64);
+                            let result = Number::from_f64(result);
+                            if let Some(result) = result {
+                                let result = Value::Number(result);
+                                bound.resolve(id, 0, result).unwrap_or(&bound);
```
rylev, line 51: Is the `unwrap_or` here what you want? This just unwraps the result and ignores it.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+
+        Ok(webview)
+    }
+
+    pub fn run(self) -> Result<()> {
+        let webview = self.webview.as_ref();
+        let url = self.url.lock().expect("lock url").clone();
+        let (tx, rx) = mpsc::channel();
+
+        if url.len() > 0 {
```
rylev, line 275:
```suggestion
        if !url.is_empty() {
```

This is a bit more idiomatic.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+    loop {
+        if let Ok(result) = rx.try_recv() {
+            return Ok(result);
+        }
+
+        unsafe {
+            match WindowsAndMessaging::GetMessageA(&mut msg, hwnd, 0, 0).0 {
+                -1 => {
+                    let error = format!("GetMessageW failed: {}", Win32::Debug::GetLastError());
+                    return Err(Error::new(ErrorCode::E_NOINTERFACE, &error));
```
rylev, line 603: Is `E_NOINTERFACE` the right error?

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+        WindowsAndMessaging::WM_DESTROY => {
+            webview.terminate().expect("window is gone");
+            LRESULT(0)
+        }
+
+        _ => unsafe { WindowsAndMessaging::DefWindowProcA(hwnd, msg, w_param, l_param) },
+    }
+}
+
+fn wait_with_pump<T>(rx: mpsc::Receiver<T>) -> Result<T> {
```
rylev, line 590: While, in general I'd like to see more comments so that those unfamiliar with WebView2 can use this example to learn more, I think this function in particular could use more explanatory prose.

rylev: Also, is this a common enough need for it to potentially make sense inside of the windows crate itself?

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+            webview.eval(&js).expect("eval return script");
+        })
+    }
+
+    fn set_window_webview(hwnd: HWND, webview: Box<WebView>) {
+        unsafe {
+            WindowsAndMessaging::SetWindowLongPtrA(
+                hwnd,
+                WINDOW_LONG_PTR_INDEX::GWLP_USERDATA,
+                Box::into_raw(webview) as _,
```
rylev, line 510: `webview` will be dropped at the end of the function. Is that what we want or does that cause a dangling pointer?

rylev (changes requested):
This is awesome! ♥ I just have some questions and comments.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+                id, method, result, id
+            );
+
+            webview.eval(&js).expect("eval return script");
+        })
+    }
+
+    fn set_window_webview(hwnd: HWND, webview: Box<WebView>) {
+        unsafe {
+            WindowsAndMessaging::SetWindowLongPtrA(
```
kennykerr, line 507: `SetWindowLongPtrA` and `GetWindowLongPtrA` are only exported from the 64-bit flavors of `user32.dll`. Blindly calling these functions will thus cause x86 builds to fail. You need to tuck them behind `cfg` attributes as illustrated here:

https://github.com/microsoft/windows-rs/blob/2c0f5445a277b7514500307f3e0753eb7833fc9a/examples/clock/src/main.rs#L726-L748

examples/webview2/bindings/build.rs
```diff
@@ -0,0 +1,15 @@
+fn main() {
+    windows::build!(
+        Microsoft::Web::WebView2::Core::*,
+        Windows::Foundation::*,
+        Windows::Win32::Com::{CoInitialize, CoUninitialize},
```
kennykerr, line 5:
```suggestion
```

These functions aren't needed and can simply be removed.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+    thread_id: u32,
+    token: EventRegistrationToken,
+    bindings: Arc<Mutex<HashMap<String, Box<dyn FnMut(u64, Vec<Value>)>>>>,
+    frame: Option<FrameWindow>,
+    parent: Arc<HWND>,
+    url: Arc<Mutex<HString>>,
+}
+
+unsafe impl Send for WebView {}
+unsafe impl Sync for WebView {}
```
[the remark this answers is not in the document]
wravery: Yup, I think they would auto-implement them now that everything is using Arc<>.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+    }
+}
+
+impl FrameWindow {
+    fn new() -> Self {
+        let hwnd = {
+            let class_name = CString::new("WebView").expect("lpszClassName");
+            let mut window_class = WNDCLASSA::default();
+            window_class.lpfnWndProc = Some(window_proc);
+            window_class.lpszClassName = PSTR(class_name.as_ptr() as *mut _);
```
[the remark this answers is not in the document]
wravery: They should, IIRC. I'll double check the docs, but I don't think we use static string literals in most of our Win32 code for the window name at least. They tend to be more dynamic.. I suppose `RegisterClassA` could have that requirement for the window class though.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+                            let result = Number::from_f64(result);
+                            if let Some(result) = result {
+                                let result = Value::Number(result);
+                                bound.resolve(id, 0, result).unwrap_or(&bound);
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
wravery: Much nicer! 👍🏼

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+
+            unsafe {
+                let result = WindowsAndMessaging::GetMessageA(&mut msg, h_wnd, 0, 0).0;
+
+                match result {
+                    -1 => {
+                        break {
+                            let error =
+                                format!("GetMessageW failed: {}", Win32::Debug::GetLastError());
+                            Err(Error::new(ErrorCode::E_NOINTERFACE, &error))
```
kennykerr, line 317:
```suggestion
                            Err(ErrorCode::from_thread().into())
```

examples/webview2/src/main.rs
wravery: Still new to Rust (this experiment was actually my first project not from the Rust Book). What I really want is the `?` operator, but I was running into trouble with mapping different Error types to the Result. This was me getting tired of handling that and deciding to just ignore the error (since it takes arbitrary input from a call on the JS console, I decided crashing with `unwrap` would be unfriendly).

One way to deal with that which I'm thinking about trying is to make my own Error enum which can wrap all of the Error types I expect, and implement From for those types. I found at least one blog which ends up suggesting that, did I just reinvent another wheel/idiom? 😆

examples/webview2/src/main.rs
wravery: Yup. For some reason rust-analyzer didn't include that in the suggestions at the time, maybe I had a parse error elsewhere.

examples/webview2/src/main.rs
wravery: Probably not, but this was another example of trying to merge Error types.

examples/webview2/src/main.rs
wravery: > Also, is this a common enough need for it to potentially make sense inside of the windows crate itself?

Maybe? It's not uncommon for Win32 programs to embed shorter-lived message loops while waiting for an async operation or a synchronization primitive (e.g. using [MsgWaitForMultipleObjects](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-msgwaitformultipleobjectsex)], so the UI remains responsive. The WebView2 threading model runs everything on the UI thread, including callbacks which it triggers with `PostMessage`, and I'm using this here because it's waiting for async operations in WebView2 to finish before I've started the main message loop in `run`.

The trouble with using an embedded message loop like this from within the main message loop is that you can easily trigger reentrancy bugs on the main thread (you're already in a call to `DispatchMessage` from the main thread, your embedded loop dispatches another message, and a message handler which triggered the embedded loop gets called again, changes state, embedded loop finishes, world explodes when the handler that called it resumes). So if this were included in the `windows`  crate, I'd probably want to at least mark it as unsafe and include warnings about not using it in the main loop.

If it's not there, the main loop would be a useful thing to abstract in the `windows` crate, though.

A generic window_proc would be interesting too. Most C/C++ window procs are implemented with `switch`/`case`, the `match` idiom in Rust is a lot less fragile. To get good static analysis, it might even make sense to implement it as a macro rather than a dispatch table mapping messages to `fn`, so the caller implements it with `match` arms inline. Or maybe it could just be another `fn` to implement the `match` expression.

examples/webview2/src/main.rs
wravery: I thought `Box::into_raw` consumed the `Box` and avoided calling `drop`, is that incorrect?

examples/webview2/src/main.rs
wravery: Ah, makes sense.

examples/webview2/src/main.rs
kennykerr: I'm working on adding the same level of async support that C++/WinRT provides to the Windows crate. At that point you would not need these pumping loops at all and can simply use a Rust async function. That already works to some degree in Rust except that Rust async functions are cold-start whereas we need hot-start for WinRT. So for now, I'm fine with this function but I'll hopefully get rid of it soon as its just a stopgap to get the sample working.

examples/webview2/src/main.rs
kennykerr: See https://github.com/microsoft/windows-rs/pull/647#discussion_r602316505

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+        Ok(())
+    }
+
+    pub fn set_title(&self, title: &str) -> Result<&Self> {
+        if let Some(frame) = self.frame.as_ref() {
+            if let Ok(title) = CString::new(title) {
+                unsafe {
+                    WindowsAndMessaging::SetWindowTextA(
+                        *frame.window,
+                        PSTR(title.as_ptr() as *mut _),
```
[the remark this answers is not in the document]
kennykerr: Yes.

examples/webview2/src/main.rs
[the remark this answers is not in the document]
kennykerr: Yes, this is safe.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+                    );
+                }
+            }
+        }
```
kennykerr, line 350:
```suggestion
            unsafe {
                WindowsAndMessaging::SetWindowTextA(*frame.window, title);
            }
```

The Windows crate provides null termination automatically for functions like `SetWindowTextA`.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+fn set_process_dpi_awareness() -> Result<()> {
+    let code = unsafe {
+        HiDpi::SetProcessDpiAwareness(PROCESS_DPI_AWARENESS::PROCESS_PER_MONITOR_DPI_AWARE)
+    };
+
+    if code.is_err() {
+        Err(Error::fast_error(code))
+    } else {
+        Ok(())
+    }
```
kennykerr, line 544:
```suggestion
    unsafe {
        HiDpi::SetProcessDpiAwareness(PROCESS_DPI_AWARENESS::PROCESS_PER_MONITOR_DPI_AWARE).ok()
    }
```

examples/webview2/src/main.rs
wravery: Double checked docs, I think this is safe.

examples/webview2/src/main.rs
wravery: Interesting, but not for `WNDCLASSA`, so I still need to use `CString` there it looks like. Even though it I have a copy already null terminated there for `WNDCLASSA`, I decided it's safer to let the `CreateWindowExA` call re-terminate the &str literal instead of sharing the `CString` pointer.

examples/webview2/src/main.rs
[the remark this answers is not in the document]
wravery: Looks fine, there's no mention in the docs about needing static string buffers.

examples/webview2/src/main.rs
kennykerr: Yes, the `WNDCLASSA` has a `PSTR` field so I can't as easily inject the conversion as I do for input parameters (https://github.com/microsoft/windows-rs/issues/645#issuecomment-808372812). There's also a desire to add a `const` function to generate a `PSTR` at compile time that would make that cleaner (#581).

examples/webview2/src/main.rs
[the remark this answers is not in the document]
kennykerr: But it can be simplified https://github.com/microsoft/windows-rs/pull/647/files#r602403168

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+    pub fn navigate(&self, url: &str) -> Result<&Self> {
+        let url: Vec<u16> = url.encode_utf16().collect();
+        let url = HString::from_wide(&url);
+        *self.url.lock().expect("lock url") = url;
+        Ok(self)
+    }
+
+    pub fn init(&self, js: &str) -> Result<&Self> {
+        let js: Vec<u16> = js.encode_utf16().collect();
+        let js = HString::from_wide(&js);
```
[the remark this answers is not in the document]
wravery: Nice, I can just pass that into the WinRT projection directly since it expects an `HString`.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+            let data =
+                WindowsAndMessaging::GetWindowLongPtrA(hwnd, WINDOW_LONG_PTR_INDEX::GWLP_USERDATA);
+
+            match data {
+                0 => None,
+                _ => {
+                    let webview_ptr = data as *mut WebView;
+                    let raw = Box::from_raw(webview_ptr);
+                    let webview = raw.clone();
+                    mem::forget(raw);
```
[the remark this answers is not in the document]
wravery: I want to leave the original `Box::into_raw` pointer in the `GWLP_USERDATA` slot on the `HWND` without taking ownership back.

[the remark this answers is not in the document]
wravery: ...but that does remind me, it's leaking that pointer when it destroys the window.

examples/webview2/src/main.rs
```diff
@@ -0,0 +1,622 @@
...
+        let bindings: Arc<Mutex<HashMap<String, Box<dyn FnMut(u64, Vec<Value>)>>>> =
+            Arc::new(Mutex::new(HashMap::new()));
+        let bindings_ref = bindings.clone();
+        let token = webview.WebMessageReceived(TypedEventHandler::<
+            CoreWebView2,
+            CoreWebView2WebMessageReceivedEventArgs,
+        >::new(move |_sender, args| {
+            if let Some(args) = args {
+                if let Ok(message) = String::from_utf16(args.WebMessageAsJson()?.as_wide()) {
+                    if let Ok(value) = serde_json::from_str::<InvokeMessage>(&message) {
```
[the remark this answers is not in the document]
wravery: I suppose I could use `Windows::Data::Json` for that, but I think `serde_json` is enough of a de-facto standard for Rust that it's more valuable to show the interop. I think I'll leave this one as is unless you feel strongly that it should use the Windows API.

examples/webview2/src/main.rs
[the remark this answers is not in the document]
wravery: I had this simplified before, but when I was figuring out that I still needed the `CString` for `WNDCLASSA` I must have undone the change while narrowing down the cause of the crash. Latest commit puts it back in the simpler form you recommended.

examples/webview2/src/main.rs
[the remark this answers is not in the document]
kennykerr: My view is to use whatever is simplest. The challenge with examples like this is to keep them being good examples of doing X - where X is singular - rather than becoming a tour de force of all the things. 😉 It's a fine line - just something to keep in mind.

kennykerr (approved):

rylev (approved):
Looking great!

wravery:
Oops, that close button is too easy to hit in the GitHub mobile app. 😆

PR #660, opened 2021-03-31
Fix documentation link

nshipilov:
Fixed a documentation link, had a bit of trouble finding it myself, so it should be helpful for others.


kennykerr (approved):
Thanks!

Issue #432, opened 2021-01-22
RFC: Provide pre-built version of this library


Jasper-Bekkers:
First off: thanks for a nice addition to the Rust ecosystem, it's really wonderful to see Microsoft actively try to support what is going on and provide support for the win32 apis.

However as a maintainer of several code-generation crates I've found that it's usually significantly nicer for the user of my crates to provide a pre-built version of the wrappers that I'm exposing. This has a few benefits

 - `build.rs` is a serial dependency and very slow for more complex projects since cargo has to stall building code further down the line, and it has to wait for code-gen crates like `syn` and `quote` to have been built before it can run a `build.rs`.
 - Right now the recommended workflow is for everybody else to create a pre-built version, however, this means making windows-rs types part of your public interface will lead to incompatibilities between crates, whereas before you could make sure that everything is on the same version through `cargo deny` for example.

To see the impact of building this crate, one can run `cargo +nightly build -Ztimings` to see the build dependency graph and this crate's impact on it (for example on the `minesweeper-rs` example that you've provided it looks like the vast majority of it's ~70s build time is spent in code-gen related things).

If a pre-generated version of this crate becomes available (or nicers: this crate switches into a pre-built library) please add feature toggles similar to how `winapi-rs` does things, this lets the `cargo` feature resolver make sure that the minimal set of features is built across the entire dependency tree.

rylev:
I'm personally in favor of that. The API surface is huge though so it's not exactly clear how best one would offer a pre-built version. So, some additional questions:
* Should it be a single crate or multiple crates? 
* If multiple, how should they be split?

nagisa:
I don't think it is terribly wrong to give an ability to pre-generate and include in crates just the bindings they need. As you mention it can lead to incompatibilities if these are exposed in the public interface of the crates, but in my experience the need to do so is pretty rare.

repi:
The winapi 0.3 approach is proven to work well, single huge crate but with features for all areas/modules that are disabled by default so each crate using it has to opt in to the feature they do want to use. This compiles nicely and the cargo additive feature support makes sure winapi is built once with all the features that all dependencies have enabled. 

In our codebase we have ~50 direct usages from our own and third party crates to `winapi` (0.3) and it builds really fast and is easy to use. Think almost none of those crates expose the winapi types as public symbols, but some may/could and it would still work well also.

thomcc:
[...]

PR #661, opened 2021-04-01
Simplify windows_gen dependencies

kennykerr:
Removes a few dependencies and folds in `squote`. This makes it easier to debug and tweak code gen and gets us a little closer to #432


rylev:
Hmmm... did you measure compile time impact? Having squote as a separate crate means it can be compiled in parallel which is no longer possible.

kennykerr:
Build time was unchanged, but that's a good suggestion - the windows_gen crate is now quite large - I can see if factoring out some portions of it might improve build time.

PR #663, opened 2021-04-01
Add default "macros" feature

kennykerr:
This update introduces the "macros" feature that is the default, so nothing changes by default. 😉 

It allows the `windows` crate to be used as a dependency without necessarily bringing in all of its dependencies necessary for the `build` and `implement` macros. This dramatically improves build time in such cases, going from 16 seconds down to 2 seconds.

It does mean that the `windows` crate can't simply rely on a build script to generate the bindings it needs. That's what the `windows_bindings` crate is for. Any time the bindings need to be updated, this tool can be rerun to regenerate the bindings. And the `crates` tool now uses `default-features = false` to speed up build time for the pre-generated crates. 

One more step toward #432.


crates/bindings/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+    // TODO: Rust doesn't seem to support something like a C++ static_assert so that we can assert
+    // that the generated file matches the crate it is being built with.
+
+    // file.write_all(
+    //     format!(
+    //         "assert!(env!(\"CARGO_PKG_VERSION\") == \"{}\");\n\n",
+    //         env!("CARGO_PKG_VERSION")
+    //     )
+    //     .as_bytes(),
+    // )?;
```
kennykerr, line 35: @rylev suggestions? 😉

rylev: There’s a crate for that: https://docs.rs/static_assertions/1.1.0/static_assertions/index.html

kennykerr: Ah, clever trick. Unfortunately, I still can't use it to compare strings.

PR #671, opened 2021-04-05
WIP: Add support for adding item attributes to generated types

udoprog:
This adds support for parsing attributes in combination with the tree being generated. The idea is that the parsed attributes added to each item associated with the types being generated. That way, it's possible to selective add compile-time constraints like feature flags.

This patch only adds support for parsing feature flags right now. Also the generated attributes only apply to functions, see below for more.

## Example

```rust
windows_macros::generate!(
    #[cfg(feature = "coreaudio"]
    Windows::Win32::CoreAudio::*,
    #[cfg(feature = "xaudio2")]
    Windows::Win32::XAudio2::*,
)
```

This would for example cause `XAudio2CreateWithVersionInfo` to be generated like this:

```rust
#[cfg(feature = "xaudio2")]
pub unsafe fn XAudio2CreateWithVersionInfo(/* ... */) -> ::windows::ErrorCode {
    #[link(name = "XAudio2_9")]
    extern "system" {
        pub fn XAudio2CreateWithVersionInfo(/*  */) -> ::windows::ErrorCode;
    }
    XAudio2CreateWithVersionInfo(/* ... */)
}
```

## Motivation

I want to be able use feature flags to determine which which linked libraries should be required for compiling my crate, and the above is enough to avoid a linker error in case `XAudio2_9` is not available and the feature is disabled.

The attributes are currently only added to functions, because these are the only items which include `#[link]` attributes. Filtering these allows tree shaking in rustc to do the rest of the work right now. But in principle these could be added to any item that is recursively included.

## WIP

This is currently in WIP to garner feedback.

What I want to do before it's merged is:
* Clean up `TypeConstraints`. It is currently used to recursively add `any(...)` dependencies for the complete intersection if features in use for any given item, but it's a bit messy.
* Add parsing for `all(...)` and `any(...)`.


kennykerr:
Thanks, I'll take a look later this week. Just to set expectations, I'm in the middle of adding features for namespace lightup - to reduce build time - and we also have another project to deal with linker issues so this PR is likely to collide with both of those. So I'm happy to see the experimentation and ideas flowing but may not be able to accept something like this at the moment.

udoprog:
That's all right. Looking forward to have a look at what you come up with. And feel free to use anything in here if you find it useful.

udoprog:
> Thanks, I'll take a look later this week. Just to set expectations, I'm in the middle of adding features for namespace lightup - to reduce build time - and we also have another project to deal with linker issues so this PR is likely to collide with both of those. So I'm happy to see the experimentation and ideas flowing but may not be able to accept something like this at the moment.

@kennykerr I just wanted to ask. Is this project public somewhere so I can follow it?

kennykerr:
For the DLL imports: https://github.com/rust-lang/rust/issues/58713

For the features: https://github.com/microsoft/windows-rs/issues/432

PR #682, opened 2021-04-08
Statically assert that the bindings were generated against the current version of the crate

rylev:
Fixes: https://github.com/microsoft/windows-rs/pull/663#discussion_r605987397

The error message this generates is not great, but still gets the point across. Only those building a local version of the windows crate should ever see this error, so the fact that the error isn't great is not a deal breaker.

The error:
```
...    |
3252 | |             ["The current version of the crate does not match the version the bindings were generated against: 0.7.0."][100];
     | |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ index out of bounds: the length is 1 but the index is 100  
```


kennykerr (approved):
Thanks!

PR #694, opened 2021-04-12
Simplify convertible parameters

### LINKED ISSUE #677: Consider improving generic param names
Most functions generated are generic over the parameters to facilitate taking multiple types of arguments. In the docs this currently looks like this:

```rust
pub unsafe fn SomeFunction<'a, T1__: IntoParam<'a, SomeType>>(some_param: T1__);
```

While some of this complexity is unavoidable, it might help to give generic params more descriptive name. What strikes the right balance between not being overly verbose and providing proper guidance is tough. The best I can think of is `Param1`, but that might be too verbose.

kennykerr:
Fixes #677

### DIFF
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -1460,8 +1460,8 @@ pub mod Windows {
                     .from_abi::<::windows::Object>(result__)
                 })
             }
-            pub fn CreateString<'a, T0__: ::windows::IntoParam<'a, ::windows::HString>>(
-                value: T0__,
+            pub fn CreateString<'a>(
+                value: impl ::windows::IntoParam<'a, ::windows::HString>,
             ) -> ::windows::Result<::windows::Object> {
                 Self::IPropertyValueStatics(|this| unsafe {
                     let mut result__: <::windows::Object as ::windows::Abi>::Abi =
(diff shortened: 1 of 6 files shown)


rylev (changes requested):
Can you update the docs/FAQ with the new syntax?

src/bindings.rs
```diff
@@ -1747,11 +1747,8 @@ pub mod Windows {
             clippy::all
         )]
         pub mod Automation {
-            pub unsafe fn SysAllocStringLen<
-                'a,
-                T0__: ::windows::IntoParam<'a, super::SystemServices::PWSTR>,
-            >(
-                strin: T0__,
+            pub unsafe fn SysAllocStringLen<'a>(
+                strin: impl ::windows::IntoParam<'a, super::SystemServices::PWSTR>,
```
rylev, line 1751: fyi: the param name should probably be `string` not `strin`

kennykerr: The name is derived from metadata. It is originally `strIn` and converted to lowercase by the windows crate.

kennykerr:
Docs are updated.

PR #700, opened 2021-04-14
Fixed formatting and crate version

tim-weis:
The hyperlink was missing a closing backtick.

The crate version needs to be 0.8, otherwise the generated APIs don't match the text.


rylev (approved):
Thanks!!!

Issue #709, opened 2021-04-17
Add NULL to IBindCtx


ArturKovacs:
I have an existing code that I'm converting to use `windows-rs`, but I hit an issue. In my code I use `IShellFolder::BindToObject` which can take a NULL pointer for the `IBindCtx` parameter. However it seems that it's not possible to set that parameter to NULL in `windows-rs`.

This also affects `SHCreateItemFromParsingName`.

kennykerr:
You should be able to pass _None_.

ArturKovacs:
You are correct, I did not realize this, thank you! 😊 As far as I'm concerned the issue is resolved.

ArturKovacs:
Something to consider is to add a sentence or two about this in the readme or the getting started document. If it's already there then I missed it, but a search for "null" doesn't yield any useful match in either of those documents.

rylev:
@kennykerr Is it possible for there to be the following impl on for all COM interfaces?

```rust
impl <'a> IntoParam<'a, ISomeInterface> for *mut ISomeInterfaceVTable {}
```

kennykerr:
All convertible parameters now support `None` - this seems like the most natural way to represent this in Rust. An FAQ entry seems appropriate.

Interfaces are not modeled as pointers in Rust so I don't think adding an `impl` ultimately helps all that much. The vtable type should also be private - its not at the moment but that's a technical limitation that I'd like to fix at some point.

PR #715, opened 2021-04-20
Document that nullable parameters are often modeled with Option

rylev:
Addresses the issue in #709

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -24,7 +24,7 @@ The `windows` crate does not allow the use of null pointers when safer alternati
 
 For example, in C/C++ [`IShellFolder::BindToObject`](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishellfolder-bindtoobject)'s third parameter is a pointer to an `IBindCtx` interface. The C and C++ docs state that passing `NULL` is appropriate when the parameter is not used.
 
-In Rust this parameter is [translated](https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/Shell/struct.IShellFolder.html#method.BindToObject) as `impl IntoParam<'a, IBindCtx>`. If you try to pass a null pointer (from `std::ptr::null` or `std::ptr::null_mut`), the code won't compile. Instead you can pass `None`, and the code will compile.
+In Rust this parameter is [translated](https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/Shell/struct.IShellFolder.html#method.BindToObject) as `impl IntoParam<'a, IBindCtx>`. If you try to pass a null pointer (from `std::ptr::null` or `std::ptr::null_mut`), the code won't compile. Instead, you can pass `None` and the code will compile.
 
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -18,7 +18,15 @@ It turns out that `IntoParam<'a, PWSTR>` is implemented for `&'a str` so we can
 
 If you get an error message saying that a specific type does not satisfy the trait bound, then you are passing a type which cannot be converted to the correct type. If you think this is in error (i.e., you think the type you have should be trivially convertible to the param type in question), don't hesitate to [file an issue](https://github.com/microsoft/windows-rs/issues).
 
-## I get a "method missing" error when the documentation says the method exists.
+## I get an error when trying to pass a NULL pointer to a function that should take null?
```
kennykerr, line 21:
```suggestion
## Why do I get an error when trying to pass a NULL pointer to a function that should take null?
```

docs/FAQ.md
```diff
@@ -18,7 +18,15 @@ It turns out that `IntoParam<'a, PWSTR>` is implemented for `&'a str` so we can
...
+
+Often times, the `windows` crate does not allow the use of null pointers when safer alternatives exist. In Rust, optional values are modeled using the `Option` type which the `windows` crate takes full advantage of for COM interfaces. So instead of passing a `NULL` pointer, you can pass `None` instead.
```
kennykerr, line 23:
```suggestion
The `windows` crate does not allow the use of null pointers when safer alternatives exist. In Rust, optional values are modeled using the `Option` type which the `windows` crate takes advantage of for COM interfaces. So instead of passing a `NULL` pointer, you can pass `None` instead.
```

docs/FAQ.md
```diff
@@ -18,7 +18,15 @@ It turns out that `IntoParam<'a, PWSTR>` is implemented for `&'a str` so we can
...
+
+For example, in C/C++ [`IShellFolder::BindToObject`](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishellfolder-bindtoobject)'s third parameter is a pointer to an `IBindCtx` interface. The C and C++ docs state that passing `NULL` is appropriate when the parameter is not used.
+
+In Rust this parameter is [translated](https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/Shell/struct.IShellFolder.html#method.BindToObject) as `impl IntoParam<'a, IBindCtx>`. If you try to pass a null pointer (from `std::ptr::null` or `std::ptr::null_mut`), the code won't compile. Instead you can pass `None`, and the code will compile.
```
kennykerr, line 27:
```suggestion
In Rust this parameter is [translated](https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/Shell/struct.IShellFolder.html#method.BindToObject) as `impl IntoParam<'a, IBindCtx>`. If you try to pass a null pointer (from `std::ptr::null` or `std::ptr::null_mut`), the code won't compile. Instead, you can pass `None` and the code will compile.
```

kennykerr (approved):

PR #718, opened 2021-04-21
Add weak reference (consumer) support

kennykerr:
On the way to #81, first up is support for weak references as this is required for Xaml/WinUI and is just generally what COM and WinRT objects should provide by default. This update adds support for creating a weak reference to an existing implementation. This is the easy part. Next up is implementing weak references as part of the `implement` macro. That's the hard part...

### DIFF
--- a/tests/weak/tests/uri.rs
+++ b/tests/weak/tests/uri.rs
@@ -1,0 +1,24 @@
+use test_weak::Windows::Foundation::Uri;
+use windows::{factory, IActivationFactory, Interface, Result};
+
+// The Uri class supports weak references, so it is used to test a successful downgrade and upgrade.
+#[test]
+fn test_success() -> Result<()> {
+    let strong = Uri::CreateUri("http://kennykerr.ca")?;
+    let weak = strong.downgrade()?;
+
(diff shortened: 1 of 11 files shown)


tests/weak/tests/uri.rs
```diff
@@ -0,0 +1,29 @@
...
+// The Uri class factory does not support weak references, so it is used to test an unsuccessful downgrade.
+#[test]
+fn test_failure() -> Result<()> {
+    let strong = factory::<Uri, IActivationFactory>()?;
+    let result = strong.downgrade();
+
+    assert_eq!(result.is_err(), true);
+
+    // E_NOINTERFACE is returned because IWeakReferenceSource is not implemented.
+    assert_eq!(result.unwrap_err().code(), HRESULT(0x8000_4002));
```
rylev, line 26: Out of curiosity, is there a reason we still don’t define constants for common HRESULTs?

kennykerr: These are defined in the Win32 metadata. We need to be careful not to depend on these as part of code gen to avoid forcing the consuming apps to require the same dependency. In this particular case, its just a local test so we could rely on generated HRESULTs here.

PR #732, opened 2021-04-26
Update metadata

kennykerr:
Updates to version 10.0.19041.5-preview.166 of Windows.Win32.winmd that provides weak reference definitions.


MarijnS95:
@kennykerr Reading through this I only see `10.0.19041.166-preview`, no `10.0.19041.5-preview.166`. Is that the same?

PR #741, opened 2021-04-28
Version 0.9.1

kennykerr:


MarijnS95:
@kennykerr Does every example and test need to have a version, and sometimes a version dependency on the `windows` crate? That's a lot of extra diff noise, if these are not published to crates.io they might be set to `0.0.0` instead?

kennykerr:
Yes, I only publish the `windows`, `windows_gen`, and `windows_macros` crates.  The noise is annoying - I'll see if I can switch the rest to `0.0.0` - I'm assuming that won't impact anyone...

PR #745, opened 2021-04-28
Add support for weak references

### LINKED ISSUE #366: Implementations should support weak references

kennykerr:
Weak references enable COM objects to support non-owning references, allowing an object to be used only if some other strong reference is keeping the object alive. This can be useful for breaking reference cycles. This is required to support Xaml/WinUI and part of the larger support for #81. Implementations based on the `implement` macro now support weak references automatically.

Fixes #366


crates/gen/src/types/delegate.rs
```diff
@@ -140,6 +140,8 @@ impl Delegate {
                             ::std::ptr::null_mut()
                         };
 
+                    // TODO: implement IMarshal
```
rylev, line 143: Nit: it would be nice to track these as issues instead of todos in the code.

src/runtime/weak_ref_count.rs
```diff
@@ -0,0 +1,315 @@
...
+    strong_vtable: *const IWeakReferenceSource_abi,
+    weak_vtable: *const IWeakReference_abi,
+    object: RawPtr,
+    strong_count: RefCount,
+    weak_count: RefCount,
+}
+
+impl TearOff {
+    unsafe fn new(object: RawPtr, strong_count: u32) -> IWeakReferenceSource {
+        std::mem::transmute(::std::boxed::Box::new(TearOff {
```
rylev, line 142: Was this code generated at some point? I don't think we need the fully qualified paths here.

src/runtime/weak_ref_count.rs
```diff
@@ -0,0 +1,315 @@
...
+struct TearOff {
+    strong_vtable: *const IWeakReferenceSource_abi,
+    weak_vtable: *const IWeakReference_abi,
+    object: RawPtr,
+    strong_count: RefCount,
+    weak_count: RefCount,
+}
+
+impl TearOff {
+    unsafe fn new(object: RawPtr, strong_count: u32) -> IWeakReferenceSource {
```
rylev, line 141: It would be really great to document these unsafe functions invariants.

crates/gen/src/types/delegate.rs
kennykerr: https://github.com/microsoft/windows-rs/issues/753

PR #746, opened 2021-04-28
gen/function: Emit consistent stub signatures on unsupported platforms

MarijnS95:
Callers of these functions - whether they're used on unsupported platforms or not, like `create_instance` with `CoCreateInstance` - fail to compile when the function signature does not use the new return pattern for `QueryInterface`-like generic functions.  This change makes sure the function signature stays the same, only the implementation changes to `panic!` when the it is not supported on a given platform.

Fixes: fe8955c ("Transform QueryInterface-like functions into generic functions (#735)")

---

Not sure if this could be written even more concisely if the function signature (without body) is stored in a variable instead.

### DIFF
--- a/crates/gen/src/types/function.rs
+++ b/crates/gen/src/types/function.rs
@@ -51,37 +51,46 @@ impl Function {
             link = "onecoreuap";
         }
 
-        if cfg!(windows) {
+        let body = if cfg!(windows) {
             if signature.has_query_interface() {
                 let leading_params = &signature.params[..signature.params.len() - 2];
-                let params = signature.gen_win32_params(leading_params, gen);
                 let args = leading_params.iter().map(|p| p.gen_win32_abi_arg());
 
                 quote! {
-                    pub unsafe fn #name<#constraints T: ::windows::Interface>(#params) -> ::windows::Result<T> {
-                        #[link(name = #link)]
-                        extern "system" {
-                            pub fn #name(#(#abi_params),*) #abi_return_type;
-                        }
-                        let mut result__ = ::std::option::Option::None;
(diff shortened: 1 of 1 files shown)


kennykerr:
I was wandering whether anyone would notice this. 😉

MarijnS95:
@kennykerr I'm pleasantly surprised this was the only thing breaking Linux support, so that we can hopefully get on with https://github.com/microsoft/windows-rs/discussions/566 using `windows-rs` for clean, up-to-date, autogenerated bindings instead of the legacy com crate or migrating to `com-rs` first :grimacing: 

The only thing we'll need at some point is `src/bindings.rs` for Linux, without `#[link]` and where possible a Rust implementation like https://github.com/Traverse-Research/hassle-rs/blob/5deb75cb7b2eb037c7e6e3e63bd390a8d83172c5/src/os.rs. I'll keep that for a different topic/issue/time though :)

crates/gen/src/types/function.rs
```diff
@@ -51,37 +51,46 @@ impl Function {
...
+                    #[link(name = #link)]
+                    extern "system" {
+                        pub fn #name(#(#abi_params),*) #abi_return_type;
                     }
+                    #name(#(#args),*)
+                }
+            }
+        } else {
+            quote! {
+                panic!("Unsupported target OS");
```
rylev, line 78: nit: I wonder if the `unimplemented!` is a slightly better choice rather than `panic`. It's the same thing, but it's more clear in code and adds "unimplemented" to the panic message.

MarijnS95: Check, I agree :)

kennykerr (approved):
Thanks!

PR #748, opened 2021-04-28
interfaces: Use named struct members for vtable functions

MarijnS95:
Instead of using seemingly magic numbers in a tuple this is clearer to read.  I am not aware of any ABI differences between tuples and structs under `repr(C)`, [1] states: "Tuple structs are like structs with regards to `repr(C)`, as the only difference from a struct is that the fields aren’t named."

Also reuse "parent" vtables instead of replicating the same function signatures, again improving readability and reusing existing code.

[1]: https://doc.rust-lang.org/nomicon/other-reprs.html#reprc

---

Is there a specific reason it wasn't done like this in the first place?  And is it worth applying the same changes to generated code?

### DIFF
--- a/src/interfaces/activation_factory.rs
+++ b/src/interfaces/activation_factory.rs
@@ -1,4 +1,4 @@
-use crate::*;
+use super::*;
 
 /// WinRT classes have a supporting factory object that implements `IActivationFactory` to create a new
 /// instance of the WinRT class with some default state. `IActivationFactory` represents the
@@ -18,7 +18,7 @@ impl IActivationFactory {
 
             // Even though the factory will generally return the WinRT default interface, this isn't guaranteed
             // so a cast is required to convert the `Object` into `I`, or the class type.
-            (self.vtable().6)(self.abi(), &mut object)
+            (self.vtable().activate_instance)(self.abi(), &mut object)
                 .and_some(object)?
                 .cast()
         }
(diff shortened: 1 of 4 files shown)


kennykerr:
So this is done intentionally to mirror what the generated code does. The generated code does this to avoid various collisions that occur in hierarchical COM APIs like DirectX and generic WinRT interfaces where the same interface member has different specialized implementations. It also makes it simpler to generate the implementations via the `implement` macro and just generally get vtable ordering correct. The fast ABI also rewrites vtables and I plan to support that soon, as I've done with C++/WinRT.

MarijnS95:
@kennykerr So it's just a case of being able to reference these functions directly by index, instead of having to go through the various parents before reaching the desired field?

kennykerr:
The vfptrs are [called here](https://github.com/microsoft/windows-rs/blob/master/crates/gen/src/types/com_interface.rs#L128-L149) and [implemented here](https://github.com/microsoft/windows-rs/blob/master/crates/macros/src/implement.rs#L143-L161). This is more complex for COM interfaces today, but the fast ABI adds interface inheritance to WinRT interfaces as well (under the hood).

MarijnS95:
@kennykerr That's just referencing the function pointers by index, nothing that would be complicated when the functions are referenced by their name instead? This already generates the same functions over and over for every subclass of a parent, which would normally be solved with a `Deref` to the parent class (and would solve the parent naming issue, too). Is it possible to inherit multiple interfaces in COM (and/or diamond inheritance) breaking that?

I don't what "fast ABI" (fastcall) does for vtable rewriting.

MarijnS95:
In hindsight I guess it's only fair to close this then. I don't have much of an understanding of "fast ABI" nor will be maintaining this code. It just so happened to bother me to see both duplicate pointers and magic numbers (what functions are these?!?). It would have been nice to make an exception for these few "core" types but alas.

PR #749, opened 2021-04-28
Update to new namespaces

MarijnS95:
Following https://github.com/microsoft/win32metadata/commit/df833c5f4d51659aea57926f0947308a4a9c1fef

I wasn't planning on PR'ing this at all. Well, I took care of the few renames to get the types for `HRESULT` etc. proper again after using newer bindings... and why not PR those few lines in advance.... turns out that required much more effort than expected, and someone else has probably done the same work already. Here it is anyway, in case it is useful to someone when the next `Windows.Win32.winmd` is pushed :grimacing: 

Note that `bindings.rs` needs to be regenerated from Windows - I'm currently playing around with this crate on Linux :wink:.

### DIFF
--- a/crates/gen/default/readme.md
+++ b/crates/gen/default/readme.md
@@ -3,7 +3,7 @@ dependent crate or workspace has an empty or non-existent `.windows/winmd` direc
 
 ## Windows.Win32.winmd
 - Source: https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/
-- Version: 10.0.19041.5-preview.166
+- Version: 10.0.19041.202-preview
 
 ## Windows.WinRT.winmd
 - Source: https://www.nuget.org/packages/Microsoft.Windows.SDK.Contracts
--- a/crates/bindings/src/main.rs
+++ b/crates/bindings/src/main.rs
@@ -3,17 +3,24 @@ use std::io::prelude::*;
 fn main() -> std::io::Result<()> {
     let tokens = windows_macros::generate!(
         Windows::Foundation::{IReference, IStringable, PropertyValue},
-        Windows::Win32::Automation::{BSTR, GetErrorInfo, IErrorInfo, SetErrorInfo},
-        Windows::Win32::WinRT::{IRestrictedErrorInfo, ILanguageExceptionErrorInfo2, IWeakReference, IWeakReferenceSource},
-        Windows::Win32::Debug::{GetLastError, FormatMessageW},
-        Windows::Win32::WindowsProgramming::CloseHandle,
-        Windows::Win32::Com::{
+        Windows::Win32::System::Com::{
             CoCreateGuid, CoTaskMemAlloc, CoTaskMemFree, CLSIDFromProgID, CoInitializeEx, CoCreateInstance,
             IAgileObject,
         },
-        Windows::Win32::SystemServices::{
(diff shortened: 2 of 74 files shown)


kennykerr:
Thanks, I'm not looking forward to this churn... 🤪

kennykerr:
No need to maintain this PR - I'll refresh everything when a new winmd is published.

MarijnS95:
@kennykerr I'm updating this for local purposes, need the new bindings with DXC support :stuck_out_tongue_winking_eye:

kennykerr:
Thanks for driving those API additions!

MarijnS95:
I hope it's still useful though, feel free reopen/cherrypick/force-push once the new winmd is published - I checked `Allow edits by maintainers` (wasn't checked by default this time) for that reason :)

MarijnS95:
@kennykerr I've been maintaining this because of needing it locally, and it looks like the next release containing the changes were just released. Would you mind reopening this so that I can force-push my changes?

kennykerr:
I plan to start working on getting the new version incorporated asap. There's a lot in there that I'd like to light up in Rust.

MarijnS95:
@kennykerr Understood. Looks like more namespaces were shuffled around later, fixing them now.

MarijnS95:
Perhaps for next time: sorting and merging/regrouping imports. Need to double-check the README too but I'm in a hurry now - @kennykerr please push any needed changes to this PR for the time being, or take over the branch locally :)

kennykerr:
Unfortunately, `cargo fmt` doesn't sort them...

kennykerr:
Well it looks like you've done all the heavy lifting here so I don't see why I shouldn't just merge this in as a first step - thanks!

MarijnS95:
@kennykerr Fortunately `rustfmt` reorders and reindents `use` groups, those can be copied back into the macro.

Have you thought about how to group imports - and `generate!` macro - yet? Nightly `rustfmt` knows about these: https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#imports_granularity

And a single `use` statement is on its way to do as much nesting as possible: https://github.com/rust-lang/rustfmt/commit/509ac2e87f12a1bfae693109223db01b93c93897

MarijnS95:
@kennykerr sounds good! Note that `README` checking was removed in 2edcccd, but the `doc-comment` dependency is still there. Is that intended?

crates/gen/default/readme.md
```diff
@@ -3,7 +3,7 @@ dependent crate or workspace has an empty or non-existent `.windows/winmd` direc
 
 ## Windows.Win32.winmd
 - Source: https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/
-- Version: 10.0.19041.5-preview.166
+- Version: 10.0.19041.202-preview
```
MarijnS95, line 6: @kennykerr I wasn't really sure what to do with the version. The previous `.166` release used this "new" format too, is this correct?

kennykerr:
> @kennykerr sounds good! Note that README checking was removed in 2edcccd, but the doc-comment dependency is still there. Is that intended?

No, that was left by mistake.

crates/gen/default/readme.md
kennykerr: Yes, I basically just copy the format from the win32metadata release verbatim.

kennykerr (approved):

PR #751, opened 2021-04-29
Fix link in getting-started.md

TiggeZaki:
Fixed the getting-started.md link because the image was not displaying correctly in the browser.


rylev (approved):
Thanks! 🎉

PR #756, opened 2021-04-29
Add MemoryBuffer example

kennykerr:
This is a simple example illustrating how to use the WinRT `MemoryBuffer` class with the COM `IMemoryBufferByteAccess` interop interface.

### DIFF
--- a/examples/buffer/src/main.rs
+++ b/examples/buffer/src/main.rs
@@ -1,0 +1,33 @@
+use bindings::{Windows::Foundation::*, Windows::Win32::WinRT::IMemoryBufferByteAccess};
+use windows::*;
+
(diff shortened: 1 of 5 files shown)


examples/buffer/src/main.rs
```diff
@@ -0,0 +1,33 @@
+use bindings::{Windows::Foundation::*, Windows::Win32::WinRT::IMemoryBufferByteAccess};
+use windows::*;
+
+fn as_slice(buffer: &IMemoryBufferReference) -> Result<&mut [u8]> {
```
rylev, line 4: Wondering where small little helpers like this could potentially live so that users don't have to write them themselves. Not something we need to address right now, but at some point it'd be nice for these little helpers to live somewhere.

PR #778, opened 2021-05-07
Remove dead code

kennykerr:
`Guid` is no longer defined in Win32 metadata and the `doc-comment` dependency is unused.


MarijnS95 (approved):
^^

PR #780, opened 2021-05-08
Fix build error in the example in readme.md

cnwzhjs:
The example in the readme.md causes proc error:
error: proc macro panicked                                                                          
  --> build.rs:2:5
   |
2  | /     windows::build!(
3  | |         Windows::Data::Xml::Dom::*,
4  | |         Windows::Win32::System::WindowsProgramming::CloseHandle,
5  | |         Windows::Win32::UI::WindowsAndMessaging::MessageBoxA,
...  |
11 | |         },
12 | |     );
   | |______^
   |
   = help: message: Could not find namespace `Windows.Win32.System.WindowsProgramming`

which is introduced by the namespace change.

This PR fixes this compile error

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -30,13 +30,10 @@ This will allow Cargo to download, build, and cache Windows support as a package
 fn main() {
     windows::build!(
         Windows::Data::Xml::Dom::*,
-        Windows::Win32::System::WindowsProgramming::CloseHandle,
-        Windows::Win32::UI::WindowsAndMessaging::MessageBoxA,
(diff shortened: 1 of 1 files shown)


MarijnS95 (changes requested):
I don't think this is correct, these are the old namespaces. Please see #782, all that needed to change is moving `WaitForSingleObject` under `Threading` (and a missing comma between the `Threading::{}` and `SystemServices::{}` block, that has been solved by reordering and not having `SystemServices` imports at all anymore).

PR #791, opened 2021-05-11
Add d3d12 sample, based on D3D12HelloTriangle

damyanp:


kennykerr:
You need to handle both x86 and x64 builds. I may just provide helper functions for this in the Windows crate as everyone runs into this and its tedious. Anyway, here's how:

https://github.com/microsoft/windows-rs/blob/e366c7a93a3732ecf4f76264183951d2ed5df797/examples/clock/src/main.rs#L698-L720

examples/d3d12/src/main.rs
```diff
@@ -0,0 +1,837 @@
...
+            D3D12SerializeRootSignature(
+                &desc,
+                D3D_ROOT_SIGNATURE_VERSION_1,
+                &mut signature,
+                &mut error,
+            )
+            .ok()?;
+
+            assert!(signature.is_some());
+            let signature = signature.unwrap();
```
MarijnS95, line 587: Also `.and_some` :)

MarijnS95 (commented):
Interesting stuff!

examples/d3d12/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "d3d12"
+version = "0.9.0"
```
kennykerr, line 3:
```suggestion
version = "0.0.0"
```

examples/d3d12/bindings/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "d3d12_bindings"
+version = "0.9.0"
+authors = ["Microsoft"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
kennykerr, line 7:
```suggestion
```

examples/d3d12/src/main.rs
```diff
@@ -0,0 +1,837 @@
...
+                        &command_queue,
+                        hwnd,
+                        &swap_chain_desc,
+                        std::ptr::null(),
+                        None,
+                        &mut swap_chain_1,
+                    )
+                    .ok()?;
+
+                swap_chain_1.unwrap().cast()?
```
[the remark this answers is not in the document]
kennykerr: Here's an example:

https://github.com/microsoft/windows-rs/blob/e366c7a93a3732ecf4f76264183951d2ed5df797/examples/clock/src/main.rs#L580-L595

damyanp:
Thank you for the feedback @MarijnS95  and @kennykerr!

kennykerr (approved):

kennykerr:
image

😎

examples/d3d12/bindings/build.rs
```diff
@@ -0,0 +1,60 @@
...
+            CS_VREDRAW,
+            CW_USEDEFAULT,
+            DefWindowProcA,
+            DispatchMessageA,
+            GetWindowLongA,
+            GetWindowLongPtrA,
+            GWLP_USERDATA,
+            IDC_ARROW,
+            LoadCursorW,
+            MSG,
```
kennykerr, line 24:
```suggestion
            MSG,
```

examples/d3d12/src/main.rs
```diff
@@ -0,0 +1,855 @@
...
+                None,
+                "VSMain",
+                "vs_5_0",
+                compile_flags,
+                0,
+                &mut vertex_shader,
+                std::ptr::null_mut(),
+            )
+            .and_some(vertex_shader)?
+        };
```
MarijnS95, line 642: Nit: I personally always try to keep as much safe code outside unsafe blocks as possible. Since `vertex_shader` is defined (and then shadowed) outside the `unsafe` block:


```suggestion
        }
        .and_some(vertex_shader)?;
```

kennykerr: Multiline works well in the browser.

damyanp: Thanks - I am keen to iterate on style before this PR gets completed, so please do keep the suggestions coming!

MarijnS95: > Multiline works well in the browser.

Yup, using it all the time but wanted to try out the plugin for once. A PR to add that feature is on its way apparently :crossed_fingers:

damyanp: (oh, that's almost exactly the same example you posted below :) )

examples/d3d12/src/main.rs
```diff
@@ -0,0 +1,855 @@
...
+                    &ID3D12Device::IID,
+                    std::ptr::null_mut(),
+                )
+                .is_ok()
+                {
+                    return Ok(adapter);
+                }
+            }
+        }
+    }
```
MarijnS95, line 257: You could save two indent levels here: use `.and_some()?` for the pointer check - afaik `EnumAdapters1` should _always_ return a valid pointer when its result is not `DXGI_ERROR_NOT_FOUND`, because `.and_some` would turn this into `E_POINTER`. The second level is removed by slimming down `unsafe {}` to the three function calls that actually need it. You can also remove the `: Option<IDXGIAdapter1>` and `: DXGI_ADAPTER_DESC1` types, `EnumAdapters1`/`GetDesc1` are already strongly typed.

```suggestion
    for i in 0.. {
        let mut adapter = None;
        let hr = unsafe { factory.EnumAdapters1(i, &mut adapter) };
        if hr == DXGI_ERROR_NOT_FOUND {
            break;
        }

        let adapter = hr.and_some(adapter)?;
        let mut desc = DXGI_ADAPTER_DESC1::default();
        unsafe { adapter.GetDesc1(&mut desc) }.ok()?;

        if (DXGI_ADAPTER_FLAG::from(desc.Flags) & DXGI_ADAPTER_FLAG_SOFTWARE)
            != DXGI_ADAPTER_FLAG_NONE
        {
            // Don't select the Basic Render Driver adapter.
            // If you want a software adapter, pass in "/warp" on the command line.
            continue;
        }

        // We need the variant where we pass in NULL for the outparam.
        #[link(name = "d3d12")]
        extern "system" {
            pub fn D3D12CreateDevice(
                padapter: ::windows::RawPtr,
                minimumfeaturelevel: D3D_FEATURE_LEVEL,
                riid: *const ::windows::Guid,
                ppdevice: *mut *mut ::std::ffi::c_void,
            ) -> ::windows::HRESULT;
        }

        // Check to see whether the adapter supports Direct3D 12, but
        // don't create the actual device yet.
        if unsafe {
            D3D12CreateDevice(
                adapter.abi(),
                D3D_FEATURE_LEVEL_11_0,
                &ID3D12Device::IID,
                std::ptr::null_mut(),
            )
        }
        .is_ok()
        {
            return Ok(adapter);
        }
    }
```

examples/d3d12/src/main.rs
MarijnS95: That's a tricky one with the indentation change like that. It looks unusual but I'd probably stick with less-code-in-unsafe-is-better for consistency. Your pick though :)

examples/d3d12/src/main.rs
MarijnS95: Hehe reducing `unsafe` and `mut` are two of the things I look out for while reviewing PRs :)

MarijnS95: Oh yeah, the compiler can't infer that `for i in 0..` is an infinite iterator... For that you can place `unreachable!()` at the end of the loop.

Anyway, it's not really infinite though, it will run out at some point:

```rust
fn main() {
    for i in (std::u32::MAX - 3).. {
        println!("{}", i);
    }
}
````

```console
me:~/ $ rustc test.sh && ./test
4294967292
4294967293
4294967294
thread 'main' panicked at 'attempt to add with overflow', /home/marijn/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/range.rs:393:1
```

examples/d3d12/src/main.rs
```diff
@@ -0,0 +1,844 @@
...
+
+        let signature = unsafe {
+            D3D12SerializeRootSignature(
+                &desc,
+                D3D_ROOT_SIGNATURE_VERSION_1,
+                &mut signature,
+                &mut error,
+            )
+        }
+        .and_some(signature)?;
```
MarijnS95, line 598: That's a tricky one. I don't know if error-pointers for context as last argument are common (like they are in glib) but if they are it'd be nice if context could be added automagically in the bindings.

Also, would you need to check whether it's not `None` and panic instead? DirectXShaderCompiler's `Compile` method usually returns `S_OK` and then sets the `error` blob appropriately with a status code (not available on `ID3DBlob`) and extra error message in the data pointer+size.

Maybe this needs something of the form:
```suggestion
                &mut error,
            )
        };

        if let Some(error) = error {
            panic!(
                "Root signature serialization failed: {}",
                std::str::from_utf8(unsafe {
                    std::slice::from_raw_parts(
                        error.GetBufferPointer() as *const _,
                        error.GetBufferSize(),
                    )
                })
                .unwrap()
            );
        }
        let signature = hr.and_some(signature)?;
```

MarijnS95 (commented):
That's a lot of cleanup, thanks so much for the effort! Few more nits, hope that's okay given that you want this to be a complete example :grimacing:

examples/d3d12/src/main.rs
```diff
@@ -170,100 +172,94 @@ extern "system" fn wndproc<S: DXSample>(
...
+        }
+        WM_DESTROY => {
+            unsafe { PostQuitMessage(0) };
+            LRESULT::default()
+        }
+        _ => {
+            let handled = match sample {
+                Some(s) => sample_wndproc(s, message, wparam),
+                None => false,
+            };
```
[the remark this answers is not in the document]
kennykerr: Watch out, you'll never want to use C++ again. 😋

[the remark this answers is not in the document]
damyanp: Clippy didn't highlight this, but it did find some other things to simplify.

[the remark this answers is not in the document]
MarijnS95: Oh nice, `clippy` points out right about the same things I did, and I overlooked some too! :grimacing:

I assumed it was already ran in the CI. Quite a few valuable suggestions, perhaps it's worth running this repo-wide?

examples/d3d12/src/main.rs
```diff
@@ -170,100 +172,94 @@ extern "system" fn wndproc<S: DXSample>(
...
-                }
+    for i in 0.. {
+        let mut adapter = None;
+        let adapter = unsafe { factory.EnumAdapters1(i, &mut adapter) }.and_some(adapter)?;
+
+        let mut desc: DXGI_ADAPTER_DESC1 = DXGI_ADAPTER_DESC1::default();
+        unsafe { adapter.GetDesc1(&mut desc) }.ok()?;
+
+        if (DXGI_ADAPTER_FLAG::from(desc.Flags) & DXGI_ADAPTER_FLAG_SOFTWARE)
+            != DXGI_ADAPTER_FLAG_NONE
```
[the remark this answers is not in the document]
damyanp: Is this a suggestion for something I can change here?  Or is this more an observation of something that we could improve in windows-rs?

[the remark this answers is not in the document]
kennykerr: Feel free to create an issue for this.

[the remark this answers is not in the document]
MarijnS95: Done, see https://github.com/microsoft/windows-rs/issues/793.

examples/d3d12/src/main.rs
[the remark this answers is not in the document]
MarijnS95: That'd be nice - worth the effort to fix all outstanding issues and enable it in the CI, disallowing any violating contributions from that point on. Most suggestions are really nice but unfortunately a couple good ones aren't enabled by default - and will cause noise in future Rust releases when they're eventually turned on.

examples/d3d12/src/main.rs
```diff
@@ -466,60 +459,59 @@ mod d3d12_hello_triangle {
...
 
-            command_list.OMSetRenderTargets(1, &rtv_handle, false, std::ptr::null());
+        unsafe { command_list.OMSetRenderTargets(1, &rtv_handle, false, std::ptr::null()) };
 
-            // Record commands.
-            let clear_color: [f32; 4] = [0.0, 0.2, 0.4, 1.0]; // https://github.com/microsoft/windows-rs/issues/790
+        // Record commands.
+        let clear_color: [f32; 4] = [0.0, 0.2, 0.4, 1.0]; // https://github.com/microsoft/windows-rs/issues/790
+        unsafe {
             command_list.ClearRenderTargetView(rtv_handle, &clear_color[0], 0, std::ptr::null());
```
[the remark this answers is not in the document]
damyanp: Oh cool - it turns out that I can move the entire array inline, which makes me happy :)

examples/d3d12/src/main.rs
damyanp: Hey, thanks for reviewing this so carefully :)  It's not _hugely_ common for these APIs to include error output like this.  It it is optional, so I'll just make it null for the purposes of this example.

examples/d3d12/src/main.rs
```diff
@@ -0,0 +1,844 @@
...
+    };
+
+    match message {
+        WM_CREATE => {
+            unsafe {
+                let create_struct: &CREATESTRUCTA = transmute(lparam);
+                SetWindowLong(
+                    window,
+                    GWLP_USERDATA,
+                    transmute(create_struct.lpCreateParams),
```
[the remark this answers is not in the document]
damyanp: Thoughts on this being explicit `as isize` rather than relying on inference?  The rust docs seem to discourage `as _`.  Of course the previous transmute version was also relying on inference, but it was that way because I didn't realize I could use `as isize` here.

examples/d3d12/src/main.rs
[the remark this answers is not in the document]
MarijnS95: Ah yes, because the lifetime of the array runs until the end of the statement, so until after the function call :)

examples/d3d12/src/main.rs
MarijnS95: Yes that's true, I was writing this as if it was for the `compile` function where the log is interesting to debug typos etc. Maybe worth adding it over there for that reason though?

examples/d3d12/src/main.rs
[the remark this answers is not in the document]
MarijnS95: Ah yes, it was clippy complaining about this:

```rust
warning: transmute from `*mut std::ffi::c_void` to `isize` which could be expressed as a pointer cast instead
   --> examples\d3d12\src\main.rs:179:54
    |
179 |                 SetWindowLong(window, GWLP_USERDATA, transmute(create_struct.lpCreateParams));
    |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `create_struct.lpCreateParams as isize`
    |
    = note: `#[warn(clippy::transmutes_expressible_as_ptr_casts)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#transmutes_expressible_as_ptr_casts
```

Through `transmute` is much better here because it does not allow changing the size and potentially truncating pointers from 64-bit to 32-bit. Consider if `lpCreateParams` was a `u32` and pointer sizes are 64-bit (yes, that's not truncation but the easiest way to demonstrate this), transmute disallows this:

```rust
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
   --> examples\d3d12\src\main.rs:179:54
    |
179 |                 SetWindowLong(window, GWLP_USERDATA, transmute(create_struct.lpCreateParams as u32));
    |                                                      ^^^^^^^^^
    |
    = note: source type: `u32` (32 bits)
    = note: target type: `isize` (64 bits)
```

But a pointer cast like `create_struct.lpCreateParams as u32 as isize` compiles correctly and horribly fails at runtime.

I guess using an explicit `isize` type should be fine to match the function wrappers though (which do the same `as _` casting. Clock does the same `as _` when calling this.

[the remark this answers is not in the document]
damyanp: I'll follow the other examples here and go with `as _`.

damyanp:
How are we all feeling about this change now?

kennykerr (approved):
Looks good. We can always tweak it further as needed. I'll go ahead and merge it in. Thanks again for the contribution!

MarijnS95 (approved):

damyanp:
Thank you both for all your feedback, it's been a great learning experience!

PR #795, opened 2021-05-12
Implement macro parsing complete

kennykerr:
On the way to #81, this update adds the necessary support for parsing the `implement` macro input in order to handle extending a base class and overriding methods. Next up is plumbing in the code gen to actually take advantage of this extra info.


crates/gen/src/tables/type_def.rs
```diff
@@ -118,6 +118,20 @@ impl TypeDef {
...
+        self.attributes().any(|attribute| {
+            if attribute.name() == "ComposableAttribute" {
+                for arg in attribute.args() {
+                    if let (_, ConstantValue::I32(2)) = arg {
+                        return true;
+                    }
+                }
+            }
+
+            false
```
MarijnS95, line 131:
```suggestion
            attribute.name() == "ComposableAttribute"
                && attribute
                    .args()
                    .iter()
                    .any(|arg| matches!(arg, (_, ConstantValue::I32(2))))
```
Alternatively you could destructure the tuple in the closure parameters.

crates/macros/src/build_macro.rs
```diff
@@ -70,10 +70,7 @@ impl Parse for BuildMacro {
                         if let Some((namespace, name)) = reader.get_type_name(namespace, &name) {
                             build.import(namespace, name);
                         } else {
-                            return Err(Error::new_spanned(
-                                input,
-                                format!("`{}.{}` not found in metadata", namespace, name),
-                            ));
+                            return Err(Error::new_spanned(input, "Type not found in metadata"));
```
MarijnS95, line 73: Despite spanning, I think the name should still be printed as part of the error.

Take for example a similar error in Rust code:

```log
error[E0433]: failed to resolve: use of undeclared crate or module `foo`
```

That way an annotations processor - like GH CI - picks out `^error` lines and prints them in full; with this change the name context is now lost.

crates/gen/src/tables/type_def.rs
```diff
@@ -181,6 +195,21 @@ impl TypeDef {
...
+        for interface in self
+            .interface_impls()
+            .filter(|interface| interface.is_overridable())
+        {
+            for method in interface.interface().resolve().methods() {
+                methods.insert(method.name());
+            }
+        }
+
+        methods
```
MarijnS95, line 210: If `fn methods()` doesn't return `+ '_` (it doesn't seem to need to because everything is moved?), this works as well:

```suggestion
        self.interface_impls()
            .filter(|interface| interface.is_overridable())
            .flat_map(|interface| {
                interface
                    .interface()
                    .resolve()
                    .methods()
                    .map(|method| method.name())
            })
            .collect()
```

crates/macros/src/implement_macro.rs
```diff
@@ -0,0 +1,167 @@
...
+                    namespace.push('.');
+                }
+
+                namespace.push_str(&input.ident.to_string());
+                self.walk_implement(reader, &*input.tree, namespace)?;
+            }
+            UseTree::Name(input) => {
+                let name = input.ident.to_string();
+
+                if let Some((namespace, name)) = reader.get_type_name(namespace, &name) {
```
MarijnS95, line 44: You can also write this like:

```suggestion
                let (namespace, name) = reader
                    .get_type_name(namespace, &name)
                    .ok_or_else(|| Error::new_spanned(input, "Type not found in metadata"))?;
```

And ditch the else-return case.

crates/macros/src/implement_macro.rs
```diff
@@ -0,0 +1,167 @@
...
+            UseTree::Rename(input) => {
+                return Err(Error::new_spanned(input, "Rename syntax not supported"));
+            }
+        }
+
+        Ok(())
+    }
+
+    fn parse_override(&mut self, reader: &'static TypeReader, cursor: ParseStream) -> Result<()> {
+        // Any number of methods may be overridden but only if a class is being overriden.
```
MarijnS95, line 74:
```suggestion
        // Any number of methods may be overridden but only if a class is being overridden.
```

crates/macros/src/implement_macro.rs
```diff
@@ -0,0 +1,167 @@
...
+            }
+            UseTree::Name(input) => {
+                let name = input.ident.to_string();
+
+                if let Some((namespace, name)) = reader.get_type_name(namespace, &name) {
+                    if reader
+                        .resolve_type_def(namespace, name)
+                        .is_public_composable()
+                    {
+                        self.extend = Some((namespace, name));
```
MarijnS95, line 137: You could possibly use `fn replace()` here and assert that the result `.is_none()` to enforce the constraint above.

crates/macros/src/implement_macro.rs
kennykerr: Going to leave this one for now - its nice but with the extra error info I've since added its a bit cumbersome so I'm torn. 😕

kennykerr:
@MarijnS95 thanks for the feedback!

crates/gen/src/tables/type_def.rs
kennykerr: Yes, the `+ '_` is a holdover from when I was still returning rows with non-static references. There's probably a lot of opportunity to clean things up - a lot of places I'd like to return iterators but end up having to return a container.

kennykerr: But I'll tackle this separately.

MarijnS95: Ah, good to know. Sounds like some nice cleanup :+1:

crates/macros/src/implement_macro.rs
```diff
@@ -142,7 +142,7 @@ impl ImplementMacro {
                         .resolve_type_def(namespace, name)
                         .is_public_composable()
                     {
-                        self.extend = Some((namespace, name));
+                        self.extend.replace((namespace, name));
```
MarijnS95, line 145: @kennykerr This one isn't complete yet, I intended to use `replace()` to check that the original value of `self.extend` is `None`:

```suggestion
                        assert!(self.extend.replace((namespace, name)).is_none());
```

Unfortunately `.expect_none("msg")` is still unstable.

crates/gen/src/tables/type_def.rs
MarijnS95: Did you plan to pick this one up in (or now, after, since it's been merged) #796?

crates/macros/src/implement_macro.rs
kennykerr: Yes, I wasn't yet sure whether I wanted to assert.

MarijnS95: Then it's probably fine to leave the `=`, that might be more representative.

crates/gen/src/tables/type_def.rs
kennykerr: Feel free to contribute. Otherwise, I'll get to it. I'm still running into a few issues with moving to iterators more broadly.

MarijnS95: #798. There doesn't seem to be much more opportunity to clean up local `let mut` + `for` + `push/insert` loops, it's pretty clean already :)

PR #798, opened 2021-05-12
gen/tables: Use iterator and collect instead of mut and .insert()

MarijnS95:
Suggestion from https://github.com/microsoft/windows-rs/pull/795#discussion_r631240362


kennykerr (approved):
Thanks!

PR #747, opened 2021-04-28
Generate QueryInterface-like helper function for callbacks

MarijnS95:
Not sure if this is warranted.  I longed to use the same neat pattern that landed just today with DXC's `DxcCreateInstance` function, which is simply a callback on a native function.  There's quite a bit of duplication with the implementation in `function.rs`, too.  Curious to hear your thoughts!

PR #802, opened 2021-05-13
Add DirectXShaderCompiler example

MarijnS95:
Depends on #747, #800

Following the discussion in https://github.com/microsoft/windows-rs/discussions/566#discussioncomment-467792 and the acknowledgement to have something in the tree that could possibly run on the Linux CI (more requirements in https://github.com/microsoft/windows-rs/discussions/566#discussioncomment-701806). Now there's a use for #747 too :)

### TODO
- Should I put `dxcompiler.dll` (and later `libdxcompiler.so`) in `.windows/<arch>`?
- Should I remove the include handler (last commit) until proper `COM` overriding/implementing is finished? It's a complete hack anyway.

### DIFF
--- a/examples/dxc/Cargo.toml
+++ b/examples/dxc/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "dxc"
+version = "0.0.0"
+authors = ["Microsoft"]
+edition = "2018"
+
+[dependencies]
+bindings = { package = "dxc_bindings", path = "bindings" }
+libloading = "0.7"
+windows = { path = "../.." }
--- a/examples/dxc/src/main.rs
+++ b/examples/dxc/src/main.rs
@@ -1,8 +1,14 @@
 use bindings::Windows::Win32::{
-    Globalization::CP_UTF8, Graphics::Hlsl::*, System::SystemServices::BOOL,
+    Globalization::CP_UTF8,
+    Graphics::Hlsl::*,
+    System::{
+        Diagnostics::Debug::ERROR_FILE_NOT_FOUND,
+        SystemServices::{BOOL, PWSTR},
+    },
 };
 use libloading::{Library, Symbol};
 use std::path::Path;
+use std::rc::Rc;
 use windows::*;
 
 // Only exists in newer DXC headers
@@ -38,6 +44,20 @@ fn blob_encoding_as_str(blob: &IDxcBlobEncoding) -> &str {
     }
 }
 
+fn create_blob(library: &IDxcLibrary, data: &str) -> windows::Result<IDxcBlobEncoding> {
+    let mut blob = None;
+    unsafe {
+        library.CreateBlobWithEncodingFromPinned(
+            data.as_ptr() as *const _,
+            data.len() as u32,
+            DXC_CP_UTF8,
+            &mut blob,
+        )
+    }
+    .and_some(blob)
+}
+
+#[allow(non_snake_case)]
 fn main() -> windows::Result<()> {
     let lib = unsafe { Library::new(dxcompiler_lib_name()) }.unwrap();
     let create: Symbol<DxcCreateInstanceProc> = unsafe { lib.get(b"DxcCreateInstance\0") }.unwrap();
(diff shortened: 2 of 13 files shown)


examples/dxc/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "dxc"
+version = "0.9.0"
+authors = ["Microsoft"]
+edition = "2018"
+
+[dependencies]
+bindings = { package = "dxc_bindings", path = "bindings" }
+libloading = "0.7"
```
MarijnS95, line 9: Unfortunately this adds a dependency on `winapi` :/

examples/dxc/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+        SystemServices::{BOOL, PWSTR},
+    },
+};
+use libloading::{Library, Symbol};
+use std::path::Path;
+use std::rc::Rc;
+use windows::*;
+
+// Only exists in newer DXC headers
+const DXC_CP_UTF8: u32 = CP_UTF8;
```
MarijnS95, line 15: This will be cleaned/gone when https://github.com/microsoft/win32metadata/pull/474 lands, like: https://github.com/MarijnS95/windows-rs/commit/6e7d02d907abb88e1694b75840ec5bcff6bfed35

examples/dxc/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+    let lib = unsafe { Library::new(dxcompiler_lib_name()) }.unwrap();
+    let create: Symbol<DxcCreateInstanceProc> = unsafe { lib.get(b"DxcCreateInstance\0") }.unwrap();
+    dbg!(&create);
+
+    // let mut compiler: Option<IDxcCompiler2> = None;
+    // let compiler = unsafe { create(&CLSID_DxcCompiler, &IDxcCompiler2::IID, compiler.set_abi()) }
+    //     .and_some(compiler)
+    //     .unwrap();
+    let compiler: IDxcCompiler2 = unsafe { DxcCreateInstanceProc(&create, &CLSID_DxcCompiler) }?;
+    let library: IDxcLibrary = unsafe { DxcCreateInstanceProc(&create, &CLSID_DxcLibrary) }?;
```
MarijnS95, line 71: #747 ^^

examples/dxc/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    unsafe extern "system" fn LoadSource(
+        this: RawPtr,
+        pfilename: PWSTR,
+        ppincludesource: *mut RawPtr,
+    ) -> HRESULT {
+        let this = &mut *(this as *mut IncludeHandlerData);
+        dbg!(&this, pfilename, ppincludesource);
+
+        if pfilename != "./foo/bar/copy.hlsl" {
```
MarijnS95, line 86: This comparison needs #800.

examples/dxc/src/main.rs
```diff
@@ -0,0 +1,226 @@
...
+    let include_handler = IDxcIncludeHandler_abi(
+        // IUnknown
+        QueryInterface,
+        AddRef,
+        Release,
+        // Virtual destructors breaking the vtable, hack for !windows DXC
+        #[cfg(not(windows))]
+        dtor,
+        #[cfg(not(windows))]
+        dtor,
```
MarijnS95, line 128: This should probably be removed - I'll have to take this up with DXC but it's a tricky one. For more details: https://github.com/Traverse-Research/hassle-rs/pull/5/commits/c37cb037be072cac43f2706b89f1388b9deee19a#diff-8202c69531938d4c80a8f8a9190be8b98a026d2bd0364b97a0c7a32e8ae21ac6

examples/dxc/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "dxc"
+version = "0.9.0"
```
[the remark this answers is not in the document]
MarijnS95: How shaming, fell for the issue I reported! Benefit-of-the-doubt: I reported this after getting started on the example 😬 - but I should have noticed it in final checking.

examples/dxc/src/main.rs
MarijnS95: The commented out code would disappear, in favour of a single call to `DxcCreateInstanceProc`.

examples/dxc/src/main.rs
MarijnS95: Thanks, that'd be nice. Non-Windows support is fairly recent and it looks mostly used for the `./dxc` binary, not the library. I've already submitted some fixes over time to make the library workable, and this seems to be the final step to make it truly cross-compatible. I [mentioned this issue](https://github.com/microsoft/DirectXShaderCompiler/issues/2680#issuecomment-667676936) a while ago but didn't get to report a standalone issue yet nor dig and fix it myself.

> Looks like IUnknown explicitly has a virtual destructor on non-Windows.

Correct: https://github.com/microsoft/DirectXShaderCompiler/blob/aa4f5176ebba010a860a214d73f33bdc5b5042ce/include/dxc/Support/WinAdapter.h#L627

I already tried removing it but that results in tons of compiler warnings, though it _should_ work since on Windows these objects aren't cleaned up through a virtual DTOR either... I'll revisit this again on both GCC and Clang, and see what happens. My C++ knowledge isn't up there so I don't know if we can simply disable the warning without breaking anything. Note that `Release` calls `delete` on itself which would call the destructor and deallocate: https://github.com/microsoft/DirectXShaderCompiler/blob/aa4f5176ebba010a860a214d73f33bdc5b5042ce/lib/DxcSupport/WinAdapter.cpp#L24. Aside, the vtable clearly shows a destructor specific to `DxcLibrary` even though it has no members that need destructing - that already happens the release function. I vaguely recall reading something about passing the pointer _and size_ to `delete` under the hood, that might be why `DxcLibrary` has its own `deleting destructor` in place.

kennykerr:
> Should I put `dxcompiler.dll` (and later `libdxcompiler.so`) in .windows/<arch>?

@damyanp are these Microsoft binaries your team distributes?

damyanp:
They're distributed by a team I work closely with.

Can you explain more what it means to put them in "./windows/"?  Does this involve somehow redistributing them?  Or is this something that would happen at build time?

MarijnS95:
> Can you explain more what it means to put them in "./windows/"? Does this involve somehow redistributing them? Or is this something that would happen at build time?

Redistributing in the [`./windows` directory in this repository](https://github.com/microsoft/windows-rs/tree/master/.windows) so that the CI can run the example for testing, preferably both Linux and Windows.

kennykerr:
The Windows crate doesn't distribute these but any crate may place dependencies in the '.windows' folder of their crate or workspace and the Windows crate will copy them to the target directory for convenience.

damyanp:
Would this only be used for the tests?  End users would still need to obtain the dll through the official distribution?  What we want to avoid doing is making this crate become another vector for distributing the compiler dll.  We should expect users to want control over which version of the dll they're using, and not have the Windows crate dictate which version they use.

kennykerr:
Yes, this is only to enable examples/tests in this repo. You can see [existing binaries here](https://github.com/microsoft/windows-rs/tree/master/.windows). I'm not saying I want to add these necessarily. I'm just explaining how it works.

damyanp:
Would an option be to have the CI build pick up the [binaries](https://github.com/microsoft/DirectXShaderCompiler/releases/tag/v1.6.2104) that are built from the shader compiler project?

kennykerr:
That's an option. I'd rather not complicate the tests and examples with such build dependencies here.

MarijnS95:
That would also complicate local testing if a script is necessary to fetch and unpack the sources. Note that the GH releases don't include Linux binaries, those are only provided on AppVeyor (https://github.com/microsoft/DirectXShaderCompiler/issues/3686).

On the other hand these binaries would add quite a few megabytes to this repo.

kennykerr:
Probably not worth adding such an example to this repo. It's also not something I have any experience with so I'm a little hesitant to take it on.

damyanp:
I'm going to work with the dxcompiler team to try and figure out how they work with these efforts.  Although a version of the compiler is included in the SDK, most users get their version from the github project.

MarijnS95:
> Probably not worth adding such an example to this repo. It's also not something I have any experience with so I'm a little hesitant to take it on.

We can have the example without the library checked in, assuming those wanting to test it need the dll from the SDK or downloaded from GH releases/AppVeyor.

examples/dxc/src/main.rs
MarijnS95: Finally found some time to report this at 2AM: https://github.com/microsoft/DirectXShaderCompiler/issues/3783

kennykerr:
Feel free to share this example in a dedicate repo. I'm sure it will benefit developers in interested in DirectX.

MarijnS95:
No further discussion whether this can make it in without having the library checked in?

kennykerr:
As I'm not a DirectX expert, I don't want to carry examples that I can't verify myself simply by running `cargo run`.

MarijnS95:
> As I'm not a DirectX expert, I don't want to carry examples that I can't verify myself simply by running `cargo run`.

No-one else to carry that effort? As said the library should be included with the SDK thus available on most developer machines?

I had plans to use DXC through windows-rs but all my efforts thus far have been nullified :disappointed:

examples/dxc/src/main.rs
MarijnS95: @damyanp For the record (in case you're still asking around) there's now a PR open: https://github.com/microsoft/DirectXShaderCompiler/pull/3793

Removing the destructor is totally fine, and even MSVC complains about missing virtual dtors on the "official" `IUnknown` from the Windows headers, if enough warnings are enabled: https://godbolt.org/z/hKPT6ThEf

PR #806, opened 2021-05-16
Fix groups inside namespaces in `build!`

fghzxm:
Currently `BuildMacro::parse` tracks the current namespace-path with a
single mutable string throughout the walking of the `UseTree`.  When the
walking of a `UseTree::Path` is completed, it incorrectly forgets to pop
that sub-namespace off the namespace-path string.  As a result, when a
`UseTree::Group` contains more than one sub-`UseTree`s, latter
sub-`UseTree`s are walked with a namespace-path that has the leaf
sub-namespaces of the previous sub-`UseTree`s incorrectly infixed into
it:

```rust
// builds A.B.C and A.B.D.E, should build A.B.C and A.D.E
windows::build!(
    A::{B::C, D::E}
);
```

This commit fixes that by cloning the namespace-path string when walking
a `UseTree::Group`.

It also changes the diagnostics when trying to importing a nonexistent
item in the global namespace from `.Foo` to `(global namespace).Foo`.


kennykerr (changes requested):
Thanks for the fix! Please run `cargo fmt` to let the build succeed.

kennykerr (approved):

PR #815, opened 2021-05-21
Add basic string functions to `BSTR`

kennykerr:
A larger effort is coming to provide more comprehensive and uniform string support for the various historical string types in the Windows API. For now, this just makes using `BSTR` a little easier to use.


lespea:
@kennykerr realize you probably didn't see my previous message since it was after the pr was merged.

Issue #706, opened 2021-04-14
Possible link attributes generated for things outside Windows SDK


Awfa:
`build!`ing `Windows::Win32::FileSystem::NetShareEnum` generates:
```rust
#[link(name = "srvcli")]
extern "system" {
    pub fn NetShareEnum(
        servername: super::SystemServices::PWSTR,
        level: super::NetManagement::Net_level,
        bufptr: *mut *mut u8,
        prefmaxlen: u32,
        entriesread: *mut u32,
        totalentries: *mut u32,
        resume_handle: *mut u32,
    ) -> u32;
}
```
which tells Rust to link to `srvcli.lib`.

But it can't find it in my window's sdk, which has the following libs:
[...]

PR #818, opened 2021-05-22
Add dll example

Jakobzs:
Hello,

I have created an example for a Windows DLL that displays a messagebox upon being loaded.

Extra notes:
Looking into the window-rs documentation I was not able to locate _DWORD_ nor _LPVOID_ for _DllMain_ so I opted to check the [winapi](https://docs.rs/winapi/0.3.9/winapi/) crate, and found:

1. [DWORD](https://tyleo.github.io/sharedlib/doc/winapi/minwindef/type.DWORD.html) is equivalent to [c_ulong](https://tyleo.github.io/sharedlib/doc/winapi/type.c_ulong.html) which is equivalent to `u32`
2. [LPVOID](https://tyleo.github.io/sharedlib/doc/winapi/minwindef/type.LPVOID.html) is equivalent to `*mut c_void`

### DIFF
--- a/examples/dll/src/lib.rs
+++ b/examples/dll/src/lib.rs
@@ -1,22 +1,18 @@
 use bindings::Windows::Win32::System::SystemServices::{
-    DLL_PROCESS_ATTACH, DLL_PROCESS_DETACH, DLL_THREAD_ATTACH, DLL_THREAD_DETACH, HINSTANCE,
+    BOOL, DLL_PROCESS_ATTACH, DLL_PROCESS_DETACH, DLL_THREAD_ATTACH, DLL_THREAD_DETACH, HINSTANCE,
 };
-use bindings::Windows::Win32::UI::WindowsAndMessaging::{MessageBoxA, MB_OK};
 use std::ffi::c_void;
 
 #[no_mangle]
-#[allow(non_snake_case)]
 #[allow(unused_variables)]
-pub extern "stdcall" fn DllMain(module: HINSTANCE, reason: u32, reserved: *mut c_void) -> bool {
+pub extern "stdcall" fn DllMain(module: HINSTANCE, reason: u32, reserved: *mut c_void) -> BOOL {
     match reason {
-        DLL_PROCESS_ATTACH => unsafe {
-            MessageBoxA(None, "Hello from the DLL!", "Hello World", MB_OK);
-        },
+        DLL_PROCESS_ATTACH => println!("Hello from the dll!"),
         DLL_THREAD_ATTACH => (),
         DLL_THREAD_DETACH => (),
-        DLL_PROCESS_DETACH => (),
+        DLL_PROCESS_DETACH => println!("Goodbye from the dll!"),
         _ => (),
     }
 
-    true
+    BOOL::from(true)
 }
--- a/examples/dll/src/lib.rs
+++ b/examples/dll/src/lib.rs
@@ -1,0 +1,22 @@
+use bindings::Windows::Win32::System::SystemServices::{
+    DLL_PROCESS_ATTACH, DLL_PROCESS_DETACH, DLL_THREAD_ATTACH, DLL_THREAD_DETACH, HINSTANCE,
+};
+use bindings::Windows::Win32::UI::WindowsAndMessaging::{MessageBoxA, MB_OK};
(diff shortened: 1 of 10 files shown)


examples/dll/src/lib.rs
```diff
@@ -0,0 +1,22 @@
+use bindings::Windows::Win32::System::SystemServices::{
+    DLL_PROCESS_ATTACH, DLL_PROCESS_DETACH, DLL_THREAD_ATTACH, DLL_THREAD_DETACH, HINSTANCE,
+};
+use bindings::Windows::Win32::UI::WindowsAndMessaging::{MessageBoxA, MB_OK};
+use std::ffi::c_void;
+
+#[no_mangle]
+#[allow(non_snake_case)]
+#[allow(unused_variables)]
+pub extern "stdcall" fn DllMain(module: HINSTANCE, reason: u32, reserved: *mut c_void) -> bool {
```
tim-weis, line 10: [DllMain](https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain) returns a `BOOL`, which [maps to](https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types) an `int` in C, corresponding to an `i32` in Rust.

examples/dll/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+    match reason {
+        DLL_PROCESS_ATTACH => unsafe {
+            MessageBoxA(None, "Hello from the DLL!", "Hello World", MB_OK);
```
tim-weis, line 13: `DllMain` runs under the loader lock, making [lots of things](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices) problematic, like *"call\[ing\] functions in User32.dll or Gdi32.dll"*. A safe alternative for `MessageBoxA` would be [OutputDebugStringA](https://docs.microsoft.com/en-us/windows/win32/api/debugapi/nf-debugapi-outputdebugstringa).

examples/dll/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+        },
+        DLL_THREAD_ATTACH => (),
+        DLL_THREAD_DETACH => (),
+        DLL_PROCESS_DETACH => (),
+        _ => (),
+    }
+
+    true
```
tim-weis, line 21: Needs to be adjusted to match the correct return type (see above).

tim-weis:
I'm not convinced that this example is complete enough. The primary reason to produce a DLL is to have clients consume it. In its current shape, the DLL doesn't export any useful functionality, and there is no client code to consume the library either. The latter is particularly interesting, and how this should be done is [still under investigation](https://github.com/microsoft/windows-rs/issues/706#issuecomment-820013846).

It's usually a good idea to submit an issue to discuss the scope and implementation of an example before pushing a PR.

Jakobzs:
> I'm not convinced that this example is complete enough. The primary reason to produce a DLL is to have clients consume it. In its current shape, the DLL doesn't export any useful functionality, and there is no client code to consume the library either. The latter is particularly interesting, and how this should be done is [still under investigation](https://github.com/microsoft/windows-rs/issues/706#issuecomment-820013846).
> 
> It's usually a good idea to submit an issue to discuss the scope and implementation of an example before pushing a PR.

Thanks for the feedback Tim.

You are indeed right about BOOL and `bool` being a mistake since `bool` only occupies 1 byte, whereas BOOL being an i32 occupies 4 bytes. I have made use of the BOOL struct as found in windows-rs to resolve it.

Previously for testing I used a basic program that called `LoadLibrary`, but instead now I have created a project `dll_loader` that takes a filename (or path) as argument and calls `LoadLibrary` on it.

Lastly I switched from MessageBox to `println!` based on the reasoning you also specified. I tried `OutputDebugStringA` but it did not work for me, although I also feel it may not be needed for the example.

In hindsight: Yes, an issue should have been created to discuss the details.

kennykerr:
You can of course create a DLL with Rust, but that does not require anything from the Windows crate. Once I have #81 completed, we should be able to add a component example that illustrates how to build a DLL that actually implements various kinds of APIs. Until then I don't think it will be beneficial to add a DLL example.  

By the way, writing your own `DllMain` is almost always a bad idea.

PR #822, opened 2021-05-26
Buffer example should be unsafe

### LINKED ISSUE #809: MemoryBuffer example is potentially unsafe
I see some potential issues in the MemoryBuffer example which was added in https://github.com/microsoft/windows-rs/pull/756:

1) If I get a mutable slice (like it's done in the example) then I must ensure that no one else accesses the memory at the same time. Nobody stops me from calling `as_slice` twice on the same `IMemoryBufferReference` ... so the `as_slice` function should be marked unsafe. Unfortunately even changing the input to `&mut IMemoryBufferReference` doesn't help, because WinRT types themselves are just references and I can easily get another one referencing the same underlying object ...
2) Even if I change the output to an immutable slice (and in my case I only need read access, so I modified the code accordingly), I would need to make sure that no one else writes to the buffer at the same time while my immutable slice exists. Is there a guarantee for `IMemoryBufferReference` that the contents don't change while I hold the reference? Does `IMemoryBufferReference` come with any guarantees by itself, or does that depend on where I got the `IMemoryBufferReference` from? (Note that there is a similar issue with `mmap`-like APIs, so e.g. [this wrapper function](https://docs.rs/memmap/0.7.0/memmap/struct.Mmap.html#method.map) is marked `unsafe`.)
3) Lastly, I'm not sure from the documentation of https://docs.microsoft.com/en-us/windows/win32/winrt/imemorybufferbyteaccess-getbuffer whether the pointer to the data might be `NULL` if the buffer is empty. But I don't see any indication that it cannot, so you need to make sure that `NULL` is not passed to `std::slice::from_raw_parts[_mut]` because [that is not allowed](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html). `NonNull::dangling()` can be used instead (when the length is 0).

kennykerr:
Fixes #809

### DIFF
--- a/examples/buffer/src/main.rs
+++ b/examples/buffer/src/main.rs
@@ -1,6 +1,11 @@
 use bindings::{Windows::Foundation::*, Windows::Win32::System::WinRT::IMemoryBufferByteAccess};
 use windows::*;
 
+// This example illustrates how to use IMemoryBufferByteAccess to access the underlying buffer
+// owned by the MemoryBuffer/IMemoryBufferReference. Note that this is inherently unsafe as
+// this function does not perform borrow/lifetime checking. The caller must ensure that the
+// IMemoryBufferReference remains alive and that that buffer is not shared across threads.
+
 unsafe fn as_slice(buffer: &IMemoryBufferReference) -> Result<&mut [u8]> {
     let interop = buffer.cast::<IMemoryBufferByteAccess>()?;
     let mut data = std::ptr::null_mut();
@@ -16,14 +21,14 @@ fn main() -> Result<()> {
     assert_eq!(reference.Capacity()?, 11);
 
     // Write to buffer...
-    unsafe {
-        let slice = as_slice(&reference)?;
+    {
+        let slice = unsafe { as_slice(&reference)? };
         slice.copy_from_slice(b"hello world");
     }
 
     // Read from buffer...
-    unsafe {
-        let slice = as_slice(&reference)?;
+    {
+        let slice = unsafe { as_slice(&reference)? };
         assert_eq!(slice, b"hello world");
     }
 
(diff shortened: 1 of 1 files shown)


examples/buffer/src/main.rs
```diff
@@ -18,13 +16,13 @@ fn main() -> Result<()> {
     assert_eq!(reference.Capacity()?, 11);
 
     // Write to buffer...
-    {
+    unsafe {
         let slice = as_slice(&reference)?;
         slice.copy_from_slice(b"hello world");
```
MarijnS95, line 21: As with the D3D12 example, should we aim for as little as possible code in an unsafe block?

```suggestion
    {
        let slice = unsafe { as_slice(&reference) }?;
        slice.copy_from_slice(b"hello world");
```

Same below if you think this is worth it.

examples/buffer/src/main.rs
```diff
@@ -1,15 +1,13 @@
 use bindings::{Windows::Foundation::*, Windows::Win32::System::WinRT::IMemoryBufferByteAccess};
 use windows::*;
 
-fn as_slice(buffer: &IMemoryBufferReference) -> Result<&mut [u8]> {
+unsafe fn as_slice(buffer: &IMemoryBufferReference) -> Result<&mut [u8]> {
```
MarijnS95, line 4: Could perhaps use `/// # Safety` documentation, explaining that `IMemoryBufferReference` performs no borrow checking (by means of refcounting) and allows multiple mutators accessing the data at the same time (which is forbidden by `&mut`, which pretty much means having unique ownership of the data
).

MarijnS95: In addition, should this function be called `as_mut_slice` to match similar functions in `std` more closely?

kennykerr:
@MarijnS95 thanks for the feedback!

PR #826, opened 2021-05-28
update winmd

PhoNoTheSky:
Windows.Win32.winmd comes from: https://github.com/microsoft/win32metadata/blob/3895e772aa70fc8f0f5f81d003d0651d7f251065/scripts/BaselineWinmd/Windows.Win32.winmd

Examples that don't work:
- clock
- d3d12
- overlapped

Tests that don't work:
- interop
- matrix3x2
- unions
- win32
- winrt

The errors are:
- Could not find type def `Windows.Win32.Graphics.Direct2D.D2D_MATRIX_3X2_F`
- Could not find type def `Windows.Win32.Graphics.Direct3D12._Anonymous_e__Union`
- Could not find type def `Windows.Win32.System.Diagnostics.Debug._Anonymous_e__Union`
- Could not find type def `Windows.Win32.System.SystemServices._Anonymous_e__Union`
- Could not find type def `Windows.Win32.System.WinRT.HSTRING`

Looking in dnSpy at OVERLAPPED for both winmd's
new:
image
old:
image

Similar situation case for D2D_MATRIX_3X2_F,
new:
image
old:
image

fix maybe by replacing the namespace with the type if field name contains Anonymous, including Anoynmous1

### DIFF
--- a/crates/bindings/src/main.rs
+++ b/crates/bindings/src/main.rs
@@ -3,6 +3,7 @@ use std::io::prelude::*;
 fn main() -> std::io::Result<()> {
     let tokens = windows_macros::generate!(
         Windows::Foundation::{IReference, IStringable, PropertyValue},
+        Windows::Win32::Foundation::{CloseHandle, BSTR, CO_E_NOTINITIALIZED, E_POINTER},
         Windows::Win32::System::Com::{
             CoCreateGuid, CoTaskMemAlloc, CoTaskMemFree, CLSIDFromProgID, CoInitializeEx, CoCreateInstance,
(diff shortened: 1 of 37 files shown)


kennykerr:
Thanks for taking a look. I only ingest releases: https://github.com/microsoft/win32metadata/releases

We'll have to wait until another release is published. Perhaps you can ask them to publish another release. 

Regarding the breaks, I'll have a take a closer look to see what's changed. It may be bugs in the metadata that would prevent ingestion.

riverar:
Yep, ran into the same issues. Filed a bug at microsoft/win32metadata#511

Issue #638, opened 2021-03-23
Linux -> Windows Cross Compilation broken


abbec:
This change https://github.com/microsoft/windows-rs/commit/198621c5588a1482eeec4077b113e80f304f2e28#diff-26e81435323b9d11b39a44aa6f5adf5b3d863e3b2b149d6c2777b925ca8694b9R53 seems to break cross-compilation from Linux. It seems the build script is run with the equivalent of `cfg!(unix)`, not `cfg!(windows)`.

kennykerr:
Have you actually tried to build for Windows on Linux? I'm not sure that's ever worked. That would require libs for Windows that a Linux-based linker could use to resolve imports, not to mention a C++ compiler that can cross compile for Windows on Linux. Not saying that's not possible, but I'd be surprised if that worked before this change. Also, I would imagine this problem would eventually go away once [DLL imports](https://github.com/rust-lang/rust/issues/58713) are implemented.

kennykerr:
Anyway, I deal with the issue of the build script not necessarily using the target compiler by checking the `CARGO_CFG_TARGET_ARCH` environment variable when copying architecture-specific dependencies. We could in theory use the 
`CARGO_CFG_TARGET_OS` environment variable in a similar way to identify the target OS if we end up supporting something like this, but I'm not sure that's going to be something we need to support.

abbec:
Hmm what do you mean? We use a mingw cross compiler on Linux to build for Windows and that works fine (except for this crate). Using the winapi crate works but I thought I would give this one a try but if it is not supported for this crate then so be it :)

kennykerr:
That works because the `winapi` crate generates its own lib files. The `windows` crate doesn't do that, but as we have folks working on [DLL imports](https://github.com/rust-lang/rust/issues/58713) we will soon not need lib files at all.

kennykerr:
I'll leave this open for now while noting that it is blocked on https://github.com/rust-lang/rust/issues/58713.

abbec:
Cool, thanks! :)

ke1satsu:
Just wanted to continue my comments from issue #1073, I confirmed that in my particular case I was able to cross-compile a simple MessageBoxA testing project. (Compiled on Debian with mingw)

I symlinked like so:
ln -s /usr/x86_64-w64-mingw32/lib/libole32.a /usr/x86_64-w64-mingw32/lib/libOLE32.a
[...]

Issue #614, opened 2021-03-18
Rust docs broken


kennykerr:
Just leaving this issue here as a few people have asked. There are two separate issues. 

The [Windows crate docs](https://docs.rs/crate/windows/0.5.0) are broken due to [this rustc bug](https://github.com/rust-lang/rust/issues/82991). 

The [generated documentation](https://microsoft.github.io/windows-docs-rs/doc/bindings/windows/) hasn't been updated because `cargo` hangs trying to build [windows-docs-rs](https://github.com/microsoft/windows-docs-rs). I seem to have hit some kind of limit, but its not clear what's wrong. `rustc` allocates about 5 GB and then `cargo` hangs inside `git_refdb_backend_fs`. This happens reliably. [Here's a repro](https://github.com/microsoft/windows-rs/files/6165754/build.zip) if you want to play along.

Running out of memory while the CPU idles is kind of sad. 😢 Here is stable/night/nightly-2021-03-16 running in parallel:

![image](https://user-images.githubusercontent.com/9845234/111663622-3ceb5300-87ce-11eb-9c43-4fbd0e498267.png)

In all three cases, the hang happens here:

![image](https://user-images.githubusercontent.com/9845234/111663756-58eef480-87ce-11eb-8e38-5b7aa11552dc.png)

kennykerr:
FYI @jyn514 and @GuillaumeGomez if you have any insights.

kennykerr:
I tried applying a workaround (#616) that @rylev suggested for the docs.rs build issue but it now fails for a different reason:

https://docs.rs/crate/windows/0.5.0/builds/363871

jyn514:
@kennykerr have you been able to bisect this to a nightly toolchain? Or was this caused by a change to windows-rs?

I still can't build windows-rs on linux, so I can't be much help unfortunately.

GuillaumeGomez:
Is `rustc.exe` supposed to be rustdoc binary?

kennykerr:
> I still can't build windows-rs on linux, so I can't be much help unfortunately.

Fine! I'm going to get this project building on Linux. 😎

kennykerr:
> have you been able to bisect this to a nightly toolchain? Or was this caused by a change to windows-rs?

I don't know - quite a lot changed on the `windows-rs` side since the docs last compiled that it's hard to tell where the problem originated.

GuillaumeGomez:
I might be able to test next week if not fixed until then.

[...]

PR #830, opened 2021-06-02
Generate proper function impl when cross-compiling

roblabla:
When testing `cfg!(windows)` within a proc macro/build script, it will resolve to true if the *build host* is windows, and false otherwise.

Here, what we want is to check if the *target* is windows or not. ~~In order to achieve this, the environment variables `CARGO_CFG_*` should be used. See https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts~~.

It turns out checking the environment does not work for proc macros. So instead, we delegate the selection from codegen-time to build-time, by moving the cfg inside the `quote!` call.

I have a setup that cross-compiles windows-rs using an LLVM toolchain and the Windows SDK libraries. Because of this check, the resulting binaries all panic with "Unsupported target os".

### DIFF
--- a/crates/gen/src/types/function.rs
+++ b/crates/gen/src/types/function.rs
@@ -42,30 +42,36 @@ impl Function {
             link = "onecoreuap";
         }
 
-        let body = if cfg!(windows) {
-            if signature.has_query_interface() {
-                let leading_params = &signature.params[..signature.params.len() - 2];
-                let args = leading_params.iter().map(|p| p.gen_win32_abi_arg());
+        let body = if signature.has_query_interface() {
+            let leading_params = &signature.params[..signature.params.len() - 2];
+            let args = leading_params.iter().map(|p| p.gen_win32_abi_arg());
 
-                quote! {
-                    #[link(name = #link)]
-                    extern "system" {
-                        fn #name(#(#abi_params),*) #abi_return_type;
-                    }
-                    let mut result__ = ::std::option::Option::None;
-                    #name(#(#args,)* &<T as ::windows::Interface>::IID, ::windows::Abi::set_abi(&mut result__)).and_some(result__)
+            quote! {
+                #[link(name = #link)]
+                extern "system" {
+                    fn #name(#(#abi_params),*) #abi_return_type;
                 }
-            } else {
-                quote! {
-                    #[link(name = #link)]
-                    extern "system" {
-                        fn #name(#(#abi_params),*) #abi_return_type;
-                    }
-                    #name(#(#args),*)
-                }
+                let mut result__ = ::std::option::Option::None;
+                #name(#(#args,)* &<T as ::windows::Interface>::IID, ::windows::Abi::set_abi(&mut result__)).and_some(result__)
             }
         } else {
             quote! {
+                #[link(name = #link)]
+                extern "system" {
+                    fn #name(#(#abi_params),*) #abi_return_type;
+                }
+                #name(#(#args),*)
+            }
+        };
+
+        // Don't link on windows dlls when generating code for non-windows:
+        let body = quote! {
+            #[cfg(windows)]
+            {
+                #body
+            }
+            #[cfg(not(windows))]
+            {
                 unimplemented!("Unsupported target OS");
             }
         };
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -2060,11 +2060,18 @@ pub mod Windows {
                 }
             }
             pub unsafe fn CloseHandle<'a>(hobject: impl ::windows::IntoParam<'a, HANDLE>) -> BOOL {
-                #[link(name = "KERNEL32")]
-                extern "system" {
-                    fn CloseHandle(hobject: HANDLE) -> BOOL;
+                #[cfg(windows)]
+                {
+                    #[link(name = "KERNEL32")]
+                    extern "system" {
+                        fn CloseHandle(hobject: HANDLE) -> BOOL;
+                    }
+                    CloseHandle(hobject.into_param().abi())
                 }
-                CloseHandle(hobject.into_param().abi())
+                #[cfg(not(windows))]
+                {
+                    unimplemented!("Unsupported target OS");
+                }
             }
             pub const E_POINTER: ::windows::HRESULT = ::windows::HRESULT(-2147467261i32 as _);
             #[repr(transparent)]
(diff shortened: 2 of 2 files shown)


MarijnS95:
It looks like you have force-pushed [my approach](https://github.com/MarijnS95/windows-rs/compare/unix-bindings-cfg), do note that this [has already been rejected](https://github.com/microsoft/windows-rs/issues/770).

roblabla:
Interesting. I really hope this can be reconsidered however. The change here is extremely simple: it only consists of moving the platform selection from codegen-time to compile-time. Without it, the code as it exists is simply wrong.

Rust is inherently capable of cross-compilation, and doing so is very important. At current $WORKPLACE, we use linux-based CI to build our windows codebase, as the linux CI machines are much faster than the equivalent windows machine. On github actions, the windows machines can be much, much slower than the equivalent linux machines too, and are additionally more expensive.

Having windows-rs unable to cross-compile to windows correctly would make it a non-starter as a `winapi-rs` replacement, which I understand is the long-term goal.

MarijnS95:
> Interesting. I really hope this can be reconsidered however. The change here is extremely simple: it only consists of moving the platform selection from codegen-time to compile-time. Without it, the code as it exists is simply wrong.

Exactly. I would have submitted my approach if it wasn't outright rejected. It's a trivial fix at the expense of larger generated code.

> Rust is inherently capable of cross-compilation, and doing so is very important. At current $WORKPLACE, we use linux-based CI to build our windows codebase, as the linux CI machines are much faster than the equivalent windows machine. On github actions, the windows machines can be much, much slower than the equivalent linux machines too, and are additionally more expensive.

For me it's not only about CI, but also about being able to use the vtables and helpers for COM when interfacing with cross-platform libraries like DXC.

> Having windows-rs unable to cross-compile to windows correctly would make it a non-starter as a `winapi-rs` replacement, which I understand is the long-term goal.

Let's hope this extra use-case helps push the topic of `windows-rs` on non-Windows forward.

kennykerr:
Note that cross-compilation is an explicit goal (#638) whereas cross-platform is a [non-goal](https://github.com/microsoft/windows-rs/issues/770#issuecomment-844444920) (but may be considered in future). 

Is this PR aiming to fix #638?

roblabla:
Yes, my PR aims at making the code cross-compilable (specifically, I'm interested in compiling from a linux host to a windows target), not making the code cross-platform.

This PR does help toward fixing #638. Of course, until we have `raw-dylib`, cross-compilation will still require a toolchain with the `.lib` from the windows SDK (for anyone interested, my setup can be found [here](https://github.com/roblabla/msvc-wine) - it uses LLVM and Windows SDK).

MarijnS95:
In its current form this PR goes out of its way to **finalize** _cross platform_ support by adding many more lines to the generated output, when all that is necessary for cross-compilation is to remove this cfg check altogether. After all it seems you're cross-compiling and running fine without regenerating `src/bindings.rs`, where these runtime cfgs would end up too.

roblabla:
> when all that is necessary for cross-compilation is to remove this cfg check altogether.

I kept the check because it was already there. If it's unnecessary, why yes I could remove it entirely. But I assume it's here for a reason :shrug: .

kennykerr:
The `cfg!(windows)` check exists to allow the code to build on non-Windows platforms [for compiler testing](https://github.com/microsoft/windows-rs/issues/614#issuecomment-802251291). It produces binaries that are obviously inoperative. For cross-compilation (e.g. build on Linux targeting Windows) you'd still want to ensure that the Windows target is used that includes the actual linker import directive. But this is probably unnecessarily complicated. I don't think there's much compiler testing going on anymore, so perhaps we should just remove the `unimplemented` branch and focus on getting cross-compilation working.

MarijnS95:
`cargo build` on Linux only builds the root crate, which is an rlib hence does no linking which is why the various `#[link]` attributes in `src/bindings.rs` are of no harm to the CI. If you actually want to test the entire chain (eg. `--workspace`, eventually `--all-targets`) this runtime `cfg!(windows)` (or one of the other solutions as alluded to before) is the way to fix that. But then that implicitly allows Windows-rs to be used on Linux :unamused:

kennykerr:
The compiler testing was for the [windows-docs-rs](https://github.com/microsoft/windows-docs-rs) repo's bindings crate for stress testing - hence the need for the `cfg`.

kennykerr:
I'm a little rusty 😉 on how rlibs work. I found that the linker complains even when the imports aren't actually used.

MarijnS95:
> The compiler testing was for the [windows-docs-rs](https://github.com/microsoft/windows-docs-rs) repo's bindings crate for stress testing - hence the need for the `cfg`.

I'm curious how that works. If any of those crates link against the `windows` crate `src/bindings.rs` is inevitably pulled in and - without locally rebuilding that file for Linux - succeed the build at compile-time.

> I'm a little rusty  on how rlibs work. I found that the linker complains even when the imports aren't actually used. Unlike in C++.

https://doc.rust-lang.org/reference/linkage.html: rlibs are just an intermediate for Rust, linking only happens when an actual binary consumed by the system is generated (executable, cdylib).

It indeed seems like linker directives are followed even if the functions could have been omitted from the binary entirely. Maybe that's a `debug` profile artifact, maybe it's because most functions are used one way or another through string and error types.

kennykerr:
> When testing cfg!(windows) within a proc macro/build script, it will resolve to true if the build host is windows, and false otherwise.

That is the key. There are now a few things that don't quite work under the proc macro that I have to workaround. This is another one of those. The solution here seems reasonable. 

You just have to regenerate the bindings crate to satisfy the CI build:

```
cargo run -p windows_bindings
```

Hopefully we can get some cross-compiling CI builds going some time soon to ensure this keeps working.

MarijnS95:
@kennykerr It would indeed be great to have the current solution (akin to the proposed fix for #770) as it allows me to run `cargo check` on Linux successfully instead of getting linker errors. That makes it much easier to contribute and test the various PRs submitted over the past few days.

I can submit https://github.com/microsoft/windows-rs/compare/master...MarijnS95:unix-bindings-cfg if that makes things easier/quicker, too.

kennykerr:
Sure I don't mind, but let's give @roblabla a chance to chime in.

roblabla:
Regenerated the bindings.rs file. This should be good to go I think?

kennykerr (approved):
Thanks!

PR #834, opened 2021-06-05
Update metadata to 10.2.84-preview and regroup UseTrees

MarijnS95:
The [10.2.84-preview](https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/10.2.84-preview) `win32metadata` just dropped and only required few extra modifications since starting on this a few weeks ago, to regroup and clean up `UseTree`s. Not all are done yet, but it should look a lot better already (also keeping in mind #827/#828).

~Note that this uses the file from https://github.com/microsoft/win32metadata/tree/v10.2.84-preview/scripts/BaselineWinmd because the nuget package appears to miss some types from the foundation namespace :thinking: - still investigating.~ The broken types switched from `TypeRef` to `TypeDef`.

### DIFF
--- a/crates/gen/src/types/bstr.rs
+++ b/crates/gen/src/types/bstr.rs
@@ -24,7 +24,7 @@ pub fn gen_bstr() -> TokenStream {
                     return 0;
                 }
 
-                unsafe { SysStringLen(self) as usize }
+                unsafe { super::System::OleAutomation::SysStringLen(self) as usize }
             }
 
             /// Create a `BSTR` from a slice of 16-bit characters.
@@ -33,7 +33,12 @@ pub fn gen_bstr() -> TokenStream {
                     return Self(::std::ptr::null_mut());
                 }
 
-                unsafe { SysAllocStringLen(super::SystemServices::PWSTR(value.as_ptr() as _), value.len() as u32) }
+                unsafe {
+                    super::System::OleAutomation::SysAllocStringLen(
+                        PWSTR(value.as_ptr() as _),
+                        value.len() as u32,
+                    )
+                }
             }
 
             /// Get the string as 16-bit characters.
@@ -42,7 +47,7 @@ pub fn gen_bstr() -> TokenStream {
                     return &[];
                 }
 
-                unsafe { ::std::slice::from_raw_parts(self.0 as *const u16, SysStringLen(self) as usize) }
+                unsafe { ::std::slice::from_raw_parts(self.0 as *const u16, self.len()) }
             }
         }
         impl ::std::clone::Clone for BSTR {
@@ -131,7 +136,7 @@ pub fn gen_bstr() -> TokenStream {
         impl ::std::ops::Drop for BSTR {
             fn drop(&mut self) {
                 if !self.0.is_null() {
-                    unsafe { SysFreeString(self as &Self); }
+                    unsafe { super::System::OleAutomation::SysFreeString(self as &Self) }
                 }
             }
         }
--- a/crates/bindings/src/main.rs
+++ b/crates/bindings/src/main.rs
@@ -2,23 +2,27 @@ use std::io::prelude::*;
 
 fn main() -> std::io::Result<()> {
     let tokens = windows_macros::generate! {
-        Windows::Foundation::{IReference, IStringable, PropertyValue},
-        Windows::Win32::System::Com::{
-            CLSIDFromProgID, CoCreateGuid, CoCreateInstance, CoInitializeEx, CoTaskMemAlloc,
-            CoTaskMemFree, IAgileObject,
+        Windows::{
+            Foundation::{IReference, IStringable, PropertyValue},
+            Win32::{
+                Foundation::{CloseHandle, BSTR, CO_E_NOTINITIALIZED, E_POINTER},
+                System::{
+                    Com::{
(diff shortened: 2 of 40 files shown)


riverar:
I didn't run into missing type issues; if there's anything here worth scavenging, feel free https://github.com/riverar/windows-rs/tree/rafael/new-metadata

MarijnS95:
@riverar I get the same `` Could not find type def `Windows.Win32.Foundation.HRESULT` `` error on your branch when running `windows_bindings`, have you ran that too?

riverar:
@MarijnS95 Are you on rustc `nightly`? I think they broke macro ABI recently, had to switch to `stable`.

MarijnS95:
> @MarijnS95 Are you on rustc `nightly`? I think they broke macro ABI recently, had to switch to `stable`. (Was fighting those same errors!)

I'd be surprised if it's that, given that the baseline `.winmd` from the repo _does_ work correctly. But no, the compiler is on stable 1.52.1. For good measure `beta` and `nightly` fail in the same way - likely something is wrong in the `.winmd` instead even though the types are successfully found with IlSpy.

riverar:
@MarijnS95 All my tests/build steps pass locally. Hm.

riverar:
@MarijnS95 Do we know why HRESULT is on the type exclusion list here? Removing this resolves the problem, but I'm wondering if that's just a hack covering up a problem.

https://github.com/microsoft/windows-rs/blob/9bca15616eb51856beded961d84faa9712a2cd63/crates/gen/src/parser/type_reader.rs#L125

kennykerr:
> Do we know why HRESULT is on the type exclusion list here

Types that are represented both in WinRT and Win32 must be excluded and defined by the Windows crate directly.

MarijnS95:
Found the culrpit: `HSTRING` and `HRESULT` are a `TypeDef` now, not a `TypeRef`. Interesting how that changed between the baseline and the nuget release.

riverar:
@MarijnS95 Nice, looks good!

crates/gen/src/parser/element_type.rs
```diff
@@ -110,24 +110,25 @@ impl ElementType {
...
-                    TypeDefOrRef::TypeDef(type_def) =>
-                    // Need to "re-resolve" the TypeDef as it may point to an arch-specific
-                    // definition. This lets the TypeTree be built for a specific architecture
-                    // without accidentally pulling in the wrong definition.
-                    {
-                        TypeReader::get()
-                            .resolve_type_def(type_def.namespace(), type_def.name())
-                            .into()
-                    }
+                    TypeDefOrRef::TypeDef(type_def) => match type_def.full_name() {
```
kennykerr, line 120: This is something I still need to fix. There's no guarantee the types will be TypeDefs or TypeRefs. I was hoping in practice they'd only be TypeRefs as that's simpler but as long as the TypeDef resides in the same PE file then they can be TypeDefs. Ideally the arch-specific types would be TyepRefs though. That's a bug in the metadata (see comment below).

kennykerr:
> Note that this uses the file from https://github.com/microsoft/win32metadata/tree/v10.2.84-preview/scripts/BaselineWinmd

I can only accept the winmd from the release.

MarijnS95:
> > Note that this uses the file from https://github.com/microsoft/win32metadata/tree/v10.2.84-preview/scripts/BaselineWinmd
> 
> I can only accept the winmd from the release.

With aforementioned fix in the second commit this is now using the release blob - I'll remove that note from the mssage.

crates/gen/src/types/bstr.rs
```diff
@@ -24,7 +24,7 @@ pub fn gen_bstr() -> TokenStream {
                     return 0;
                 }
 
-                unsafe { SysStringLen(self) as usize }
+                unsafe { super::System::OleAutomation::SysStringLen(self) as usize }
```
kennykerr, line 27: This seems wrong. Created an issue: https://github.com/microsoft/win32metadata/issues/517

kennykerr:
This looks good - I need to get the update in. Can you take care of the conflicts and then I'll merge this in.

MarijnS95:
> This looks good - I need to get the update in. Can you take care of the conflicts and then I'll merge this in.

Self-induced conflicts with all the formatting :grimacing:. Can we get #844 in first so that I can tackle them all at once?

kennykerr:
Done.

MarijnS95:
@kennykerr Updated. Please give this some thorough scrutineering, most changes didn't survive autoformatting nor the removal of enum constants.

MarijnS95:
Note that some common paths in `use` trees have been merged, but not all/everywhere. What format is desired? Someone should do a pass to unify/normalize them all, both in `build.rs` as in the `use bindings::` blocks in the source.

crates/gen/src/parser/element_type.rs
kennykerr: #845 deals with these issues.

kennykerr: OK, will complete 845.

kennykerr (approved):
Looks good!

crates/gen/src/parser/element_type.rs
MarijnS95: Ah, by kicking off the second commit the metadata import from nuget got lost as well, apologies. Thanks for pushing that back @kennykerr :)

PR #842, opened 2021-06-07
Add build macro format tool

kennykerr:
Adds a simple tool that can format build macros. I'll create a separate PR to update the build macros to use `{}` instead of `()` and perhaps we can use the yml trick from #828 to verify in CI builds. What do you think @MarijnS95?

### DIFF
--- a/crates/fmt/Cargo.toml
+++ b/crates/fmt/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "fmt"
+version = "0.0.0"
+edition = "2018"
+
+[dependencies]
+windows_gen = { path = "../gen" }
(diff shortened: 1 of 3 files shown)


MarijnS95:
> What do you think @MarijnS95?

Disappointed. I just started writing the same to update #828 - not sure why you're doing duplicate work, is there _that much_ priority to get this in? I don't think I stated anywhere to need help or not have time to "rewrite" the script.

kennykerr:
Sorry, I didn't realize you were working on that. I didn't know how else to describe what I was looking for. It took a few minutes so I thought I'd just take a stab at it. No hurry, so if you've got something in the works I'm happy to wait for that.

MarijnS95:
@kennykerr Sorry, no harm done - I might have overreacted a bit. The PR is now updated with the script in its entirety ported to Rust, dealing with the `generate!` macro as well. Let me know and if so how to split it up:

1. Use `{}` everywhere;
2. Add the script;
3. format all macros with it;
4. Enforce the script in the CI (yml ""trick"" :grinning:).

PR #843, opened 2021-06-07
Simplify use of unscoped enums

### LINKED ISSUE #831: Improve usability of scoped enums
https://github.com/microsoft/windows-rs/issues/727#issuecomment-853167528

kennykerr:
Fixes #831 

A few things to note about this change. Build macros are generally less cluttered as only the enum type needs to be mentioned (not individual constants). Generated bindings are generally longer but large numbers of constants has a minimal effect on build time (thus worthwhile tradeoff for usability). 

cc @roblabla


examples/message_box/bindings/build.rs
```diff
@@ -1,5 +1,5 @@
 fn main() {
     windows::build! {
-        Windows::Win32::UI::WindowsAndMessaging::{MessageBoxA, MB_OK}
+        Windows::Win32::UI::WindowsAndMessaging::{MessageBoxA}
```
tim-weis, line 3: Shouldn't #828 have collapsed that into `Windows::Win32::UI::WindowsAndMessaging::MessageBoxA`, observed a non-empty `git diff`, and failed the CI?

MarijnS95: @tim-weis This is only a single use-statement. #828 turns that into:

```rust
use format_build_macro::{
    Windows::Win32::UI::WindowsAndMessaging::{MessageBoxA}
};
```
Which is then formatted to:

```rust
use format_build_macro::Windows::Win32::UI::WindowsAndMessaging::MessageBoxA;
```

The original `use format_build_macro::{` pattern is now gone (missing trailing curly brace) meaning we cannot easily go back to `windows::build! {`. I'll be testing some ideas locally to deal with these files locally, and submit the cleanup for these cases (of which there seem to be quite a few) later.

PR #844, opened 2021-06-08
crates/fmt: Reformat `build!` macro calls with a single top-level import

MarijnS95:
`build!` macro calls containing a single top-level import will have their curly braces removed, making it a little more complicated to substitute the `use` (used to trick `rustfmt` into formatting our macro-call) back with a macro call and surrounding it in curly braces again.

---

CC @tim-weis for https://github.com/microsoft/windows-rs/pull/843#discussion_r647174799

### DIFF
--- a/tests/bstr/build.rs
+++ b/tests/bstr/build.rs
@@ -1,7 +1,5 @@
 fn main() {
-    windows::build! {
-        // The Windows crate manually injects various functions needed to implement BSTR.
-        // This test validates these are included.
-        Windows::Win32::System::OleAutomation::BSTR
-    };
+    // The Windows crate manually injects various functions needed to implement BSTR.
+    // This test validates these are included.
+    windows::build! {Windows::Win32::System::OleAutomation::BSTR};
 }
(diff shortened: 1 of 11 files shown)


tests/bstr/build.rs
```diff
@@ -1,7 +1,7 @@
 fn main() {
+    // The Windows crate manually injects various functions needed to implement BSTR.
+    // This test validates these are included.
     windows::build! {
-        // The Windows crate manually injects various functions needed to implement BSTR.
-        // This test validates these are included.
-        Windows::Win32::System::OleAutomation::BSTR
+        Windows::Win32::System::OleAutomation::BSTR,
     };
```
MarijnS95, line 6: Note that `rustfmt` removed these comments from:

```rust
use format_build_macro::{
    // The Windows crate manually injects various functions needed to implement BSTR.
    // This test validates these are included.
    Windows::Win32::System::OleAutomation::BSTR,
};
```
In order to omit the curly braces, instead of leaving the use path and comment within an extra pair (or move the comment above) :/

kennykerr (approved):
Looks good!

PR #847, opened 2021-06-09
Annotate lifetime for `::windows::Param`

avitex:
Brings `::windows::Param<Self>` usage in line with `rust_2018_idioms`.

### DIFF
--- a/crates/gen/src/types/pstr.rs
+++ b/crates/gen/src/types/pstr.rs
@@ -26,7 +26,7 @@ pub fn gen_pstr() -> TokenStream {
         unsafe impl ::windows::Abi for PSTR {
             type Abi = Self;
 
-            fn drop_param(param: &mut ::windows::Param<Self>) {
+            fn drop_param(param: &mut ::windows::Param<'_, Self>) {
                 if let ::windows::Param::Boxed(value) = param {
                     if !value.0.is_null() {
                         unsafe { ::std::boxed::Box::from_raw(value.0); }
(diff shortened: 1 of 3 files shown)


kennykerr:
@avitex thanks for the contribution. Please start by creating an issue describing the motivation for the PR. By the way, most of the `windows::Param` usage is in generated code so that will take some discussion and more potential code changes to apply uniformly.

avitex:
I understand most `Param` usage is used in generation. Not specifying a lifetime parameter on a type is unidiomatic from `2018` edition onward (deprecated in nightly) and as such this was a quick fix to bring that generation inline with the `rust_2018_idioms` lint.

kennykerr:
This sounds like tooling I'm not familiar with. How would I validate this myself?

avitex:
You can check this via adding the `#![deny(rust_2018_idioms)]` `rustc` lint to the top level module for the crate calling `include_bindings!()`.

Example lints which will generate errors with the current generation:

```rust
#![deny(
    // These are rustc lints
    trivial_casts,
    trivial_numeric_casts,
    rust_2018_idioms,
    // These are clippy lints
    clippy::all,
)]
```

See for details on the `rustc` lint groups: https://doc.rust-lang.org/rustc/lints/groups.html
Also see clippy lint groups: https://github.com/rust-lang/rust-clippy#clippy

I can `#[allow(trivial_casts)]` on the module with `include_bindings!()` but I can't `#[allow(rust_2018_idioms)]` which prevents me from using the `rust_2018_idioms` lint at this stage.

Also note the difference between `#![` and `#[` https://doc.rust-lang.org/reference/attributes.html

kennykerr (approved):
Thanks!

PR #848, opened 2021-06-09
Replace E_NOINTERFACE magic number with constant from metadata

MarijnS95:
Just like `E_POINTER` this constant can be generated and used from the bindings, instead of being hardcoded across the source.

### DIFF
--- a/src/result/error.rs
+++ b/src/result/error.rs
@@ -40,21 +40,22 @@ impl Error {
         Self { code, info }
     }
 
-    // Use internally to create an Error object without error info. Typically used with QueryInterface
-    // (E_NOINTERFACE) or the absense of an object to return (E_POINTER) to avoid the code gen overhead
-    // for casts that should be cheap (few instructions). Think of it as a way to create recoverable errors
-    // that don't need th overhead of debugging origination info.
-    pub fn fast_error(code: HRESULT) -> Self {
+    #[doc(hidden)]
+    /// Used internally to create an Error object without error info. Typically used with [`Interface::query`]
+    /// (`E_NOINTERFACE`) or the absence of an object to return (`E_POINTER`) to avoid the
+    /// code gen overhead for casts that should be cheap (few instructions). Think of it as a way to create
+    /// recoverable errors that don't need the overhead of debugging origination info.
+    pub const fn fast_error(code: HRESULT) -> Self {
         Self { code, info: None }
     }
 
     /// The error code describing the error.
-    pub fn code(&self) -> HRESULT {
+    pub const fn code(&self) -> HRESULT {
         self.code
     }
 
     /// The error information describing the error.
-    pub fn info(&self) -> &Option<IRestrictedErrorInfo> {
+    pub const fn info(&self) -> &Option<IRestrictedErrorInfo> {
         &self.info
     }
 
--- a/crates/bindings/src/main.rs
+++ b/crates/bindings/src/main.rs
@@ -5,7 +5,7 @@ fn main() -> std::io::Result<()> {
         Windows::{
             Foundation::{IReference, IStringable, PropertyValue},
             Win32::{
-                Foundation::{CloseHandle, BSTR, CO_E_NOTINITIALIZED, E_POINTER},
+                Foundation::{CloseHandle, BSTR, CO_E_NOTINITIALIZED, E_NOINTERFACE, E_POINTER},
                 System::{
                     Com::{
                         CLSIDFromProgID, CoCreateGuid, CoCreateInstance, CoInitializeEx,
--- a/src/bindings.rs
+++ b/src/bindings.rs
@@ -2073,6 +2073,7 @@ pub mod Windows {
                     unimplemented!("Unsupported target OS");
                 }
             }
+            pub const E_NOINTERFACE: ::windows::HRESULT = ::windows::HRESULT(-2147467262i32 as _);
             pub const E_POINTER: ::windows::HRESULT = ::windows::HRESULT(-2147467261i32 as _);
             #[repr(transparent)]
             #[derive(:: std :: clone :: Clone, :: std :: marker :: Copy)]
--- a/src/runtime/weak_ref_count.rs
+++ b/src/runtime/weak_ref_count.rs
@@ -7,8 +7,11 @@
 )]
 
 use crate::*;
-use bindings::Windows::Win32::System::WinRT::{
-    IWeakReference, IWeakReferenceSource, IWeakReferenceSource_abi, IWeakReference_abi,
+use bindings::Windows::Win32::{
+    Foundation::E_NOINTERFACE,
+    System::WinRT::{
+        IWeakReference, IWeakReferenceSource, IWeakReferenceSource_abi, IWeakReference_abi,
+    },
 };
 use std::sync::atomic::{AtomicIsize, Ordering};
 
@@ -225,7 +228,7 @@ impl TearOff {
         // TODO: implement IMarshal
 
         if (*interface).is_null() {
-            HRESULT(0x8000_4002) // E_NOINTERFACE
+            E_NOINTERFACE
         } else {
             this.weak_count.add_ref();
             HRESULT(0)
--- a/tests/weak/build.rs
+++ b/tests/weak/build.rs
@@ -1,6 +1,9 @@
 fn main() {
     windows::build! {
         Windows::Foundation::Uri,
-        Windows::Win32::System::WinRT::{IWeakReference, IWeakReferenceSource},
+        Windows::Win32::{
+            Foundation::E_NOINTERFACE,
+            System::WinRT::{IWeakReference, IWeakReferenceSource},
+        },
     };
 }
--- a/tests/weak/tests/uri.rs
+++ b/tests/weak/tests/uri.rs
@@ -1,5 +1,5 @@
-use test_weak::Windows::Foundation::Uri;
-use windows::{factory, IActivationFactory, Interface, Result, HRESULT};
+use test_weak::Windows::{Foundation::Uri, Win32::Foundation::E_NOINTERFACE};
+use windows::{factory, IActivationFactory, Interface, Result};
 
 // The Uri class supports weak references, so it is used to test a successful downgrade and upgrade.
 #[test]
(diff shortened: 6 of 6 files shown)


src/result/error_code.rs
```diff
@@ -13,6 +13,9 @@ use bindings::{
 pub struct HRESULT(pub u32);
 
 impl HRESULT {
+    pub const E_NOINTERFACE: Self = E_NOINTERFACE;
+    pub const E_POINTER: Self = E_POINTER;
```
MarijnS95, line 17: Marked this PR as draft because I'm not sure what the best place is to reexport these constants, without requiring every app to include `E_NOINTERFACE`/`E_POINTER` on their own regard.

src/result/error.rs
```diff
@@ -40,11 +40,11 @@ impl Error {
...
-    // Use internally to create an Error object without error info. Typically used with QueryInterface
-    // (E_NOINTERFACE) or the absense of an object to return (E_POINTER) to avoid the code gen overhead
-    // for casts that should be cheap (few instructions). Think of it as a way to create recoverable errors
-    // that don't need th overhead of debugging origination info.
-    pub fn fast_error(code: HRESULT) -> Self {
+    /// Use internally to create an Error object without error info. Typically used with [`Interface::query`]
+    /// ([`HRESULT::E_NOINTERFACE`]) or the absence of an object to return ([`HRESULT::E_POINTER`]) to avoid the
+    /// code gen overhead for casts that should be cheap (few instructions). Think of it as a way to create
+    /// recoverable errors that don't need the overhead of debugging origination info.
+    pub const fn fast_error(code: HRESULT) -> Self {
```
MarijnS95, line 47: Not sure if this should be made const just yet, but it allows to have nice things like:

```rust
impl Error {
    pub const E_POINTER: Self = Self::fast_error(HRESULT::E_POINTER);
    pub const E_NOINTERFACE: Self = Self::fast_error(HRESULT::E_NOINTERFACE);
}
```

kennykerr:
I decided not to export HRESULT constants from the Windows crate even though that would be useful to me. The Windows crate should be the bare minimum to support the definition of generated API definitions. 

There's always "just one more" HRESULT that someone deems important to their project. Instead, the Windows crate is kept to a minimum and projects can define these constants for themselves using the build macro. There's also no guarantee that the definitions are even available, which is why the generated code uses the magic numbers. 

Once I have pre-generated bindings working, that will become the primary source for such things.

MarijnS95:
@kennykerr Not aiming to export "just one more" `HRESULT` here - but replacing the **magic constants** used _within the `windows` crate itself_. That should be enough reason to generate a binding for them, just like `E_POINTER`.

If you prefer the reexports can be removed, everything within `windows` can still import it from `bindings::` instead of redefining the constant.

> There's also no guarantee that the definitions are even available, which is why the generated code uses the magic numbers.

I don't really understand what you mean here: they're in the metadata checked in to the repo, in `Foundation` even. That file is totally under our control and already has quite a few imports in `bindings/src/main.rs`. The generated result is checked in to this repository in `src/bindings.rs` and won't magically disappear.

kennykerr:
There's a distinction between what is generated in `src/bindings.rs` purely for the internal use of the Windows crate and what is generated in `windows.rs` by the `build` macro for public consumption. While we could certainly add more `HRESULT` constants to `src/bindings.rs` for internal use, we cannot refer to those in `windows.rs` because they may not be available. They may not be available because `src/bindings.rs` and `windows.rs` can potentially be generated from an entirely different set of winmd files. Yes, the metadata used to build the Windows crate is known, but the metadata used to generate `windows.rs` is not. The `build` macro will use whatever metadata is in the workspace's `.windows/winmd` folder and only fall back to the default metadata if no `Windows.*` metadata is found. This allows developers to use newer or entirely different metadata for their project.

MarijnS95:
> There's a distinction between what is generated in `src/bindings.rs` purely for the internal use of the Windows crate and what is generated in `windows.rs` by the `build` macro for public consumption. While we could certainly add more `HRESULT` constants to `src/bindings.rs` for internal use, we cannot refer to those in `windows.rs` because they may not be available.

Correct. I'm totally aware of this distinction, and this patch does nothing to - nor would it ever be my intention to - shove `HRESULT`s into `windows.rs`. All it does is add them to `src/bindings.rs` (and reuse the reexport in some crates).

I guess you want to revert the changes in the tests to use the constants from their own `windows.rs` bindings, and remove the reexport from `HRESULT` (and potential const reexport from `Result`?

src/result/error.rs
kennykerr: No, I'm aggressively using the latest versions of Rust to take advantage of new features (like `const`). Once the Windows crate reaches 1.0 I'll probably start supporting a min version.

crates/macros/src/implement.rs
```diff
@@ -221,7 +221,7 @@ pub fn gen(
                         if let Some(base) = &self.base {
                             ::windows::Interface::query(base, iid, interface)
                         } else {
-                            ::windows::HRESULT(0x8000_4002) // E_NOINTERFACE
+                            ::windows::HRESULT::E_NOINTERFACE
```
kennykerr, line 224: This is very tempting - indeed I did this at first - but having these constants bleed through just encourages developers to take advantage of the limited set and then complain that E_MYFAVE is missing.

src/result/error.rs
MarijnS95: Nice!

In other words, this could have been a welcome change if it weren't for not really wanting to reexport these constants. I guess we can still annotate all applicable functions as `const` for versatility.

crates/macros/src/implement.rs
MarijnS95: I don't think we have access to `::windows::bindings` here? Any other solution that doesn't involve hardcoding the constant?

kennykerr: No, I'm not aware of another solution.

src/result/error.rs
kennykerr: Yep, feel free to suggest other improvements based on the bleeding edge of Rust. 😉

MarijnS95:
> Yes, at which point I don't think there's much left?

Still quite a few opportunities. There are two magic constants left in emitted generated code, the other two are either inside the `windows` crate (`weak_ref_count`) and can use the pregenerated `bindings` module directly, or a test (`weak/uri`) that can add the constant to its `build.rs` instead.

src/result/error.rs
```diff
@@ -40,21 +40,24 @@ impl Error {
...
-    // for casts that should be cheap (few instructions). Think of it as a way to create recoverable errors
-    // that don't need th overhead of debugging origination info.
-    pub fn fast_error(code: HRESULT) -> Self {
+    /// Use internally to create an Error object without error info. Typically used with [`Interface::query`]
+    /// ([`E_NOINTERFACE`]) or the absence of an object to return ([`E_POINTER`]) to avoid the
+    /// code gen overhead for casts that should be cheap (few instructions). Think of it as a way to create
+    /// recoverable errors that don't need the overhead of debugging origination info.
+    ///
+    /// [`E_NOINTERFACE`]: bindings::Windows::Win32::Foundation::E_NOINTERFACE
+    /// [`E_POINTER`]: bindings::Windows::Win32::Foundation::E_POINTER
```
MarijnS95, line 49: Unfortunately:

```rust
   |
48 |     /// [`E_NOINTERFACE`]: bindings::Windows::Win32::Foundation::E_NOINTERFACE
   |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this item is private
```
I love intradoc :/

src/result/error.rs
```diff
@@ -40,21 +40,21 @@ impl Error {
         Self { code, info }
     }
 
-    // Use internally to create an Error object without error info. Typically used with QueryInterface
-    // (E_NOINTERFACE) or the absense of an object to return (E_POINTER) to avoid the code gen overhead
-    // for casts that should be cheap (few instructions). Think of it as a way to create recoverable errors
-    // that don't need th overhead of debugging origination info.
-    pub fn fast_error(code: HRESULT) -> Self {
+    /// Use internally to create an Error object without error info. Typically used with [`Interface::query`]
```
kennykerr, line 43:
```suggestion
    /// Used internally to create an Error object without error info. Typically used with [`Interface::query`]
```

src/result/error.rs
```diff
@@ -40,21 +40,21 @@ impl Error {
...
+    /// (`E_NOINTERFACE`) or the absence of an object to return (`E_POINTER`) to avoid the
+    /// code gen overhead for casts that should be cheap (few instructions). Think of it as a way to create
+    /// recoverable errors that don't need the overhead of debugging origination info.
+    pub const fn fast_error(code: HRESULT) -> Self {
         Self { code, info: None }
     }
 
     /// The error code describing the error.
-    pub fn code(&self) -> HRESULT {
+    pub const fn code(&self) -> HRESULT {
```
kennykerr, line 52: I'm not sure what value this adds as self will never be a compile-time constant.

src/result/error.rs
MarijnS95: I don't think this mention makes much sense anyway as the function is public.

kennykerr: Yes, `fast_error` should just be hidden from docs.

src/result/error.rs
MarijnS95: I had `self` as compile-time constant thanks to `pub const fn fast_error` and the example above:

```rust
impl Error {
    pub const E_POINTER: Self = Self::fast_error(HRESULT::E_POINTER);
    pub const E_NOINTERFACE: Self = Self::fast_error(HRESULT::E_NOINTERFACE);
}
```

However, it seems these are useless given that Rust starts complaining about not being able to run the destructor for `Error`, thanks to `Option<IRestrictedErrorInfo>`. It doesn't need to drop the consts as they're global/static constants, but I guess that's fallout from them being unused before?

src/result/error.rs
MarijnS95: Added `#[doc(hidden)]` while still keeping it as doc-comment for internal development purposes like what rust-analyzer is capable of displaying.

kennykerr: I don't see the `doc(hidden` attribute. Missing commit?

MarijnS95: I pushed it into #855 only, by accident. Now it should be here, too.

kennykerr (approved):

PR #855, opened 2021-06-10
Use intradoc links where applicable

MarijnS95:
Depends on #848 

This replaces some monospace inline code with links, clarifying that the documentation refers to an actual item somewhere in scope and makes it easier to click through to the item.

Additionally adds `const` on some trivial `HRESULT` functions.


kennykerr (approved):
Thanks!

PR #869, opened 2021-06-16
correct wrong description of parameter

astraw:
According to the documentation linked (for `IShellFolder::BindToObject`), the pointer to an `IBindCtx` interface is the *second*, not third, parameter. This fixes the incorrect statement.


kennykerr (approved):
Thanks!

Issue #509, opened 2021-02-03
Map trailing out params as Result<T> for COM interface methods 


kennykerr:
This should rely on the `retval` attribute, but this is not currently included in the win32 metadata. https://github.com/microsoft/win32metadata/issues/207

kennykerr:
Also transform trailing `GUID*, void**` parameters as a generic return value.

kennykerr:
This has now been completed as of #909. Most COM methods and functions now provide a far superior calling experience.

PR #870, opened 2021-06-16
Fixes memory leaks

### LINKED ISSUE #865: The SpellChecker sample appears to be leaking memory
It is my understanding that [`ISpellChecker::Suggest`](https://docs.microsoft.com/en-us/windows/win32/api/spellcheck/nf-spellcheck-ispellchecker-suggest) returns an [`IEnumString`](https://docs.microsoft.com/en-us/windows/win32/api/objidl/nn-objidl-ienumstring) interface. The official documentation for the interface's [`IEnumString::Next`](https://docs.microsoft.com/en-us/windows/win32/api/objidl/nf-objidl-ienumstring-next) method is a [fair bit cryptic](https://stackoverflow.com/q/67999012/1889329).

Under the premise that callers of `IEnumString::Next` do acquire ownership of the pointed-to memory, this [section of code](https://github.com/microsoft/windows-rs/blob/fdf168ddd6f4e22b21f3940228e6b2ac800be22e/examples/spellchecker/src/main.rs#L83-L97) would need to be updated with a call to [`CoTaskMemFree`](https://docs.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-cotaskmemfree) (I guess) to free resources.

tim-weis:
Fixes #865

This PR adds calls to `CoTaskMemFree` to clean up memory allocated on behalf of the caller. I'm not aware of a convenient way to turn `PWSTR`'s into `*mut c_void`'s, so the calls are a bit clunky: `CoTaskMemFree(suggestion.0 as *mut _)`. If there is a better, more conventional way to go about it, feel free to apply changes.

There's another issue with the code that I'd like to address: The calls to `unwrap()`. After I had discovered the [and_some](https://docs.rs/windows/0.11.0/windows/struct.HRESULT.html#method.and_some) associated function for `HRESULT`, my code changed for the better:

```rust
let mut suggestions = None;
unsafe { checker.Suggest(substring, &mut suggestions).ok()? };
let suggestions = suggestions.unwrap();
```

became

```rust
let mut suggestions = None;
let suggestions = unsafe { checker.Suggest(substring, &mut suggestions) }
    .and_some(suggestions)?;
```

I can apply similar changes, either in this PR or in a separate PR. Or not at all, if there is a reason not to. I'm fine either way, just let me know.

Thanks

### DIFF
--- a/examples/spellchecker/src/main.rs
+++ b/examples/spellchecker/src/main.rs
@@ -22,18 +22,13 @@ fn main() -> windows::Result<()> {
 
     // Create a ISpellChecker
     let mut checker = None;
-    unsafe { factory.CreateSpellChecker(locale, &mut checker).ok()? };
-    let checker = checker.unwrap();
+    let checker = unsafe { factory.CreateSpellChecker(locale, &mut checker) }.and_some(checker)?;
 
     // Get errors enumerator for the supplied string
+    println!("Checking the text: '{}'", input);
     let mut errors = None;
-    unsafe {
-        println!("Checking the text: '{}'", input);
-        checker
-            .ComprehensiveCheck(input.clone(), &mut errors)
-            .ok()?
-    };
-    let errors = errors.unwrap();
+    let errors =
+        unsafe { checker.ComprehensiveCheck(input.clone(), &mut errors) }.and_some(errors)?;
 
     // Loop through all the errors
     loop {
@@ -43,8 +38,7 @@ fn main() -> windows::Result<()> {
         if result == S_FALSE {
             break;
         }
-        result.ok()?;
-        let error = error.unwrap();
+        let error = result.and_some(error)?;
 
         // Get the start index and length of the error
         let mut start_index = 0u32;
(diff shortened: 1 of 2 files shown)


kennykerr:
Thanks Tim! This looks good. Unfortunately, there's no clean way to express this right now. We could consider enlightening `IEnumString` to do this automatically, but that's a bigger job. 

Yes, the `and_some` trick is preferred. Feel free to fix any that you come across.

examples/spellchecker/src/main.rs
```diff
@@ -42,8 +38,7 @@ fn main() -> windows::Result<()> {
         if result == S_FALSE {
             break;
         }
-        result.ok()?;
-        let error = error.unwrap();
+        let error = result.and_some(error)?;
```
tim-weis, line 41: To me, this change made the code a bit harder to comprehend, though I still wanted to get rid of the `unwrap()`. Any suggestions on getting the readability back and dropping the `unwrap()` at the same time?

kennykerr: I wouldn't worry too much about this. I plan to make these work more like WinRT methods where a `Result<T>` is returned and trailing `retval` parameters are turned into return types. #509 - that should simplify this a bunch and make the whole `and_some` pattern less essential.

rylev: Just as a thought - would renaming `and_some` to something else potentially make this easier? The closest equivalent in `std` IMO is `Option::and`. Is `result.and(error)?` any clearer?

tim-weis: Thanks for the feedback, Kenny. With that I believe you can merge this PR as you see fit.

Looking forward to a more compact and ergonomic interface over COM's `retval`'s, although I'm not sure how this is going to pan out. As far as I understand, the Windows Runtime uses a single success value (`S_OK`), making this easy. With COM there are more success codes in use (such as `S_FALSE` as in this case, the reason why this part of the code needs to manually check the return value in the first place).

kennykerr: Dealing with other success codes will be an issue, but as long as the original HRESULT can still be retrieved I think we should be fine.

PR #873, opened 2021-06-17
Make HRESULT derive Eq as well as PartialEq

### LINKED ISSUE #872: Cannot use HRESULTs in match expressions
```
fn is_device_removed(e: &Error) -> bool {
    match e.code() {
        DXGI_ERROR_DEVICE_REMOVED => true,
        DXGI_ERROR_DEVICE_RESET => true,
        _ => false,
    }
}
```

Fails with:

```
error: to use a constant of type `HRESULT` in a pattern, `HRESULT` must be annotated with `#[derive(PartialEq, Eq)]`
  --> samples\d3d12-multithreading\src\main.rs:72:9
   |
72 |         DXGI_ERROR_DEVICE_REMOVED => true,
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^
```

Any concerns with deriving these traits for HRESULT?

damyanp:
This allows HRESULTs to be used as constants in match expressions.

Resolves #872

### DIFF
--- a/src/result/error_code.rs
+++ b/src/result/error_code.rs
@@ -7,7 +7,7 @@ use bindings::{
 
 /// A primitive error code value returned by most COM functions.
 #[repr(transparent)]
-#[derive(Copy, Clone, Default, Debug, PartialEq)]
+#[derive(Copy, Clone, Default, Debug, Eq, PartialEq)]
 #[must_use]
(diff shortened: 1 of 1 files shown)


kennykerr (approved):
Thanks!

PR #886, opened 2021-06-20
Display win32 error code if HRESULT's facility is win32

### LINKED ISSUE #885: Display win32 error code when HRESULT's facility is win32
Many APIs deal with raw win32 Error codes. When displaying error codes, we currently show only the `HRESULT` even when the API in question usually does not deal in `HRESULT`s (as is the case with win32 APIs). When users search online for that `HRESULT` they are unlikely to find documentation on that error. 

When the `HRESULT` is a win32 error, we should also display the plain win32 error to user.

rylev:
Fixes #885

### DIFF
--- a/src/result/error.rs
+++ b/src/result/error.rs
@@ -87,6 +87,16 @@ impl Error {
 
         self.code.message()
     }
+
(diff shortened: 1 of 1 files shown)


kennykerr (approved):
Thanks!

PR #894, opened 2021-06-22
Explicitly document MSRV

rylev:
Somewhere along the way, we stopped tracking the minimum supported Rust version (MSRV). This adds it back to CI and adds a note about it in the FAQ. 

New CI tests run in parallel and the new pass for the MSRV is faster than other tests we're doing so this does not slow down CI builds.


kennykerr (approved):
Thanks, I wish there was a way to set this in the crate somehow to avoid the mysterious errors when older compilers are used by mistake.

PR #898, opened 2021-06-22
Sample cleanup

kennykerr:


damyanp (approved):

examples/d3d12/bindings/build.rs
```diff
@@ -1,7 +1,7 @@
 fn main() {
     windows::build! {
         Windows::Win32::Foundation::{HINSTANCE, PSTR, RECT},
-        Windows::Win32::Graphics::Direct3D11::*,
+        Windows::Win32::Graphics::Direct3D11::ID3DBlob,
```
kennykerr, line 4: @damyanp the new depedency tracker (#896) highlighted the fact that this sample depends on an interface in the Direct3D11 namespace. Was that intentional? Perhaps a bug in the metadata/namespace placement.

damyanp: I'm actually surprised that this is the only thing from Direct3D11 that it depends on - I think there some things like D3D_FEATURE_LEVEL as well.   Anyway, https://github.com/microsoft/win32metadata/issues/439 is tracking sorting this all out.

kennykerr: Right, that's the only that must be explicitly imported but others are imported automatically by the dependency tracker. Good, I forgot about that bug.

PR #908, opened 2021-06-24
Add XAML Island example

fghzxm:


kennykerr:
Thanks @fghzxm - I'm in the process of moving the examples to a new home:

https://github.com/microsoft/windows-samples-rs

That will allow the examples to be a lot more reusable as they can refer to the windows crate via crates.io and each have their own workspace for dependencies. I should have the samples moved over today. I'd love for you to contribute the Xaml Islands example. I would only ask that you not include any dependencies for things that can be achieved with the Windows crate. I realize windowing is a little hard right now, but I'd like to keep the examples focused on the Windows crate usage.

PR #909, opened 2021-06-24
Simpler COM function signatures

### LINKED ISSUE #509: Map trailing out params as Result<T> for COM interface methods

### LINKED ISSUE #880: Consider renaming `and_some` to something else potentially make this easier?
Just as a thought - would renaming `and_some` to something else potentially make this easier? The closest equivalent in `std` IMO is `Option::and`. Is `result.and(error)?` any clearer?

_Originally posted by @rylev in https://github.com/microsoft/windows-rs/pull/870#discussion_r654240786_

### LINKED ISSUE #910: DXGI object creation methods are kinda difficult to use
I'm pretty new to Rust, but this is the best way I've found to create an IDXGISwapChain3. It's also how the [d3d12 sample](https://github.com/microsoft/windows-rs/blob/d29f56f18c0a0d246f77361dee2d54da1cc0e856/examples/d3d12/src/main.rs#L322) does it: 

```
        let mut swap_chain = None;
        let swap_chain : IDXGISwapChain3 = unsafe {
            dxgi_factory.CreateSwapChainForHwnd(
            command_queue,
            hwnd,
            &swap_chain_desc,
            std::ptr::null(),
            None,
            &mut swap_chain
        )}
        .and_some(swap_chain)?
        .cast()?;
```

It would be great if I could do this instead:

```
        let swap_chain : IDXGISwapChain3 = unsafe {
            dxgi_factory.CreateSwapChainForHwnd(
            command_queue,
            hwnd,
            &swap_chain_desc,
            std::ptr::null(),
            None,
        )}?;
```

This would be similar to D3D12 :

```
        let device : ID3D12Device5 = unsafe { D3D12CreateDevice(None, D3D_FEATURE_LEVEL_11_0) }?;
```

Is this something that you would be interested in tidying up in the Rust projections? 

I suspect this is a consequence of the underlying functions' C++ definitions (CreateSwapChainForHwnd has an IDXGISwapChain1* out parameter, while D3D12CreateDevice uses an IID/void** pair), but perhaps the Rust projections could tidy this up?

kennykerr:
Along the way to #81, I needed to harmonize COM and WinRT signatures so I finally implemented logical return values for COM methods and functions. This dramatically simplifies using COM APIs in Rust. This is best illustrated with an example. Previously the [com_uri](https://github.com/microsoft/windows-rs/blob/master/examples/com_uri/src/main.rs) example required the following lines of code:

```rust
let mut uri = None;
let uri = CreateUri("http://kennykerr.ca", Default::default(), 0, &mut uri).and_some(uri)?;

let mut domain = BSTR::default();
uri.GetDomain(&mut domain).ok()?;

let mut port = 0;
uri.GetPort(&mut port).ok()?;
```

With this update, you can now write it far more simply:

```rust
let uri = CreateUri("http://kennykerr.ca", Default::default(), 0)?;

let domain = uri.GetDomain()?;
let port = uri.GetPort()?;
```

Fixes #509 
Fixes #880
Fixes #910


MRahilly:
Does this mean that WinRT and now COM function signatures will have a smooth appearance but (the massive set of) Win32 signatures have an unsafe rougher look?

knopp:
This seems to be causing problem for us. The original `hresult` is now lost if not an error (`.ok()`), but for some functions (i.e. `DoDragDrop` we actually need to know the `hresult` to see if drag succeeded (https://github.com/microsoft/windows-rs/issues/922).

PR #927, opened 2021-06-29
Examples

kennykerr:
Add readme pointing to samples repo.


damyanp (approved):

rylev:
I'd just like to add that while I'm in favor of moving _most_ of the examples to the example repo, IMO it is preferable to have at least one example that can run within the repo (and not requiring contributors to download a separate repo just to run one example). 

Especially given that the docs refer to the `ISpellChecker` example, I would advocate for that being present in this repo (I don't see an issue with it also living in the samples repo).

kennykerr:
Ah, I had forgotten about the spell checker tutorial - that's a good point. But I don't see why the location of the examples matters all that much. I've used various crates without ever cloning their repos. If I have to clone a repo just to try an example I might as well clone the samples repo, but perhaps I'm missing something. Having the examples in the Windows repo itself complicates the examples by forcing them to share a workspace and I would prefer to keep the maintenance to a minimum by avoiding redundant examples in different locations.

rylev:
Examples are a good way for contributors (especially newer ones) to try out how the crate works. This is why cargo has built in support for examples.

damyanp:
It might be worth considering the longer term plans here?  Once #432 is resolved then doesn't the purpose of this crate become generating the crates that actually contain the APIs.  Maybe the examples for the various slices of the API surface would really belong with the prebuilt crates, and the example in this crate should be more focused on the basics of setting up a project that generates its own bindings.

kennykerr:
Agreed, it's still early days and a lot will change over the next year or two as this project matures.

PR #947, opened 2021-07-06
Add support for static libs

wravery:
This is using the new winmd `StaticLibraryAttribute` added in https://github.com/microsoft/win32metadata/pull/553, which hasn't been merged yet in `win32metadata`.

Since `windows-rs` uses string comparisons to parse attributes in the winmd and doesn't actually need to reference the attribute definitions in `Windows.Win32.Interop.dll`, it'll work against any winmd attribute which matches the pattern `[StaticLibrary("...")]`. So to be safe, this should not be merged until https://github.com/microsoft/win32metadata/pull/553 is merged and `win32metadata` supports that attribute, but in the meantime, the `tests/static_lib` module I added can already validate the `windows-rs` support for the attribute using a locally generated winmd file and the static libs from the WebView2 SDK.


crates/macros/src/lib.rs
```diff
@@ -90,6 +90,7 @@ pub fn build(stream: proc_macro::TokenStream) -> proc_macro::TokenStream {
                                 if let ::std::option::Option::Some(filename) = path.file_name() {
                                     if filename == profile {
                                         copy(source, &mut path);
+                                        println!("cargo:rustc-link-search=native={}", path.to_str().expect("Invalid profile dir"));
```
kennykerr, line 93: This is already supported. See https://github.com/microsoft/windows-app-rs/ for an example.

wravery: The `cargo:rustc-link-search=native=...` output specifically? Or do you mean static libs in general?

kennykerr: Sorry, just the link search.

riverar: Throw your libs into `.windows/[arch]`, with [arch] being x86, x64, or arm64 (see below). That path is automatically added to the link search path.

Real world example: https://github.com/microsoft/windows-app-rs/tree/master/.windows

Impl for reference:
https://github.com/microsoft/windows-rs/blob/d3870edf5bdc39e14af4d4144015626c959ad7b8/crates/macros/src/lib.rs#L108-L117

wravery: Cool, I'll undo that part.

kennykerr:
The binary dependencies are a little problematic. I take it the libs are from the WebView2 nuget and the winmd you created yourself? 

I'm planning to add a C++ DLL project that can house Rust-specific API tests to avoid the various test dependencies. That should make it easier add targeted tests.

crates/gen/src/types/function.rs
```diff
@@ -34,6 +34,18 @@ impl Function {
...
         }
 
+        let static_lib = def
+            .attributes()
+            .filter_map(|attribute| match attribute.name() {
+                "StaticLibraryAttribute" => Some(attribute.args()[0].1.gen_value()),
+                _ => None,
+            })
+            .next();
+        let link_attr = match static_lib {
```
kennykerr, line 44: Can we imagine a scenario where a developer actually wants dynamic loading even though a static lib is available?

tests/static_lib/build.rs
```diff
@@ -0,0 +1,9 @@
+fn main() {
+    windows::build! {
+        Microsoft::Web::WebView2::Win32::*,
+        Windows::Win32::Storage::FileSystem::{
+            GetFileVersionInfoSizeW, GetFileVersionInfoW, VerQueryValueW,
+        },
+        Windows::Win32::UI::Shell::GetCurrentProcessExplicitAppUserModelID,
```
kennykerr, line 7: These don't seem to be needed for the test.

tests/static_lib/build.rs
```diff
@@ -0,0 +1,9 @@
+fn main() {
+    windows::build! {
+        Microsoft::Web::WebView2::Win32::*,
```
kennykerr, line 3: Can we be more specific and just request `CompareBrowserVersions`?
```suggestion
        Microsoft::Web::WebView2::Win32::CompareBrowserVersions,
```

tests/static_lib/tests/static_lib.rs
```diff
@@ -0,0 +1,29 @@
+use test_static_lib::Microsoft::Web::WebView2::Win32::CompareBrowserVersions;
+use windows::{initialize_sta, Result};
+
+#[test]
+fn test_less_than() -> Result<()> {
+    initialize_sta()?;
+    let mut result: i32 = 0;
+    unsafe { CompareBrowserVersions("86.0.622.0", "86.0.623.0", &mut result)? };
```
kennykerr, line 8: A little surprised that the Windows crate didn't turn `result` into a return value. Is the last parameter marked `in` and `out`?

crates/gen/src/types/function.rs
wravery: I can't think of any reason to do that, other than maybe avoiding transitive linker dependencies (see below). Right now it's up to the developer/implementer who builds the winmd. As long as the win32metadata SDK works well enough to build a custom winmd, a user of the Windows crate who prefers to use DLLs can build their own without the `--staticLibs` flag/`<StaticLibrary />` prop in the winmd project. A project that provides both binaries (like WebView2) could even provide a static library version and a DLL version of the winmd.

Do you prefer adding an option somewhere in `windows-rs` to suppress this downstream?

kennykerr: No, I'd prefer not to add an option for this. Just making sure we've at least considered this.

tests/static_lib/build.rs
wravery: The link fails with just `Microsoft::Web::WebView2::Win32::*` because the static library has these missing imports. It probably only needs 1 from each DLL/import lib to add it to the linker command line, but I figured it was better to be specific about which ones we're pulling in.

I guess the `build.rs` script could also just add them to the link with a cargo directive on `stdout`. That would have the advantage of building a smaller bindings crate.

tests/static_lib/tests/static_lib.rs
wravery: I'm surprised too, but it's not annotated in the WebView2.h header:
```cpp
STDAPI CompareBrowserVersions(PCWSTR version1, PCWSTR version2, int* result);
```
That comes straight from the IDL as a `cpp_quote`:
```idl
/// This method is for anyone want to compare version correctly to determine
/// which version is newer, older or same.  Use it to determine whether
/// to use webview2 or certain feature based upon version.  Sets the value of
/// result to `-1`, `0` or `1` if `version1` is less than, equal or greater
/// than `version2` respectively.  Returns `E_INVALIDARG` if it fails to parse
/// any of the version strings or any input parameter is `null`.  Directly use
/// the `versionInfo` obtained from
/// `GetAvailableCoreWebView2BrowserVersionString` with input, channel
/// information is ignored.

cpp_quote("STDAPI CompareBrowserVersions(PCWSTR version1, PCWSTR version2, int* result);")
```

kennykerr: And that's why - if you can convince the parameter in the winmd to be `out` then it will transform this into `-> Result<u23>`.

wravery:
> The binary dependencies are a little problematic. I take it the libs are from the WebView2 nuget and the winmd you created yourself?

Yes, exactly.

> I'm planning to add a C++ DLL project that can house Rust-specific API tests to avoid the various test dependencies. That should make it easier add targeted tests.

Sounds good. If it can build a static lib with its own winmd then it could replace the test using WebView2. This is just the only winmd I have with the new attribute so far, and it depends on the `win32metadata` SDK with https://github.com/microsoft/win32metadata/pull/553, so it's still manually generated.

crates/gen/src/types/function.rs
riverar: Pros and cons for sure. What's the suggested change?

kennykerr: I'd leave it the way it is - prefer lib if available - other languages (that don't have a linker) ignore. Component authors that wish to offer both could always include two function definitions.

tests/static_lib/tests/static_lib.rs
wravery: Might need to go back to `win32metadata` and spin up another PR to allow specifying a `--remap` response file in the `EmmitWinmd` build task for the SDK. I think it should be possible to inject an `[Out]` attribute on the parameter that way.

kennykerr (approved):
Thanks!

Issue #968, opened 2021-07-10
std::fmt format specifiers not working


ben-lichtman-msft:
For at least some structs, the format specifiers seem to not work correctly.

For example - trying to debug print an instance of MONMSGSTRUCT with:
`println!("{:#x?}", data);`

yields:

```rust
MONMSGSTRUCT {
    cb: 112,
    hwndTo: HWND { Value: 787156 },
    dwTime: 192557406,
    hTask: HANDLE { Value: 13412 },
    wMsg: 992,
    wParam: WPARAM { Value: 1117204 },
    lParam: LPARAM { Value: -1065369362 },
    dmhd: DDEML_MSG_HOOK_DATA { uiLo: 49390, uiHi: 49279, cbData: 0, Data: [0, 0, 0, 0, 0, 0, 0, 0] },
}
```

(notice that the numbers are in decimal and not hexidecimal)

I'm guessing this is due to the Debug trait implementations not using the normal debug format builders?

I have not tested other format specifiers but I expect a similar result...

kennykerr:
Thanks for reporting, this is a duplicate of #439.

riverar:
Fixed in master, give it a whirl!

PR #969, opened 2021-07-10
Pass through params, traits in struct Debug impl

### LINKED ISSUE #439: Generated Debug impls for structs do not support hex formatting
Currently the generated implementation of the `Debug` trait for structures uses an approach that means hex formatting (via `{:x?}` and similar) is not supported.

In particular, [this code](https://github.com/microsoft/windows-rs/blob/master/crates/gen/src/struct.rs#L189-L195) explicitly generates calls to `std::fmt::DebugStruct::field` with the value parameter being the result of a `format!` macro using the default debug formatting.

If using the nested `format!` call is not intentional, I think it can be removed as any value that can be passed to it with the debug format specifier must implement `Debug` and so can be passed directly to the `field` method.

riverar:
Fixes #439, #968


kennykerr (commented):
Thanks for fixing this! Just need to rerun `cargo run -p windows_bindings` now that the bindings have changed.

Minor nit: please put the issues in the description rather than the title - that way they become links that can be clicked and GitHub should automatically close them when the PR is completed.  e.g. #945

kennykerr (approved):

riverar:
The issue numbers are actually clickable in the shortmsg too and GitHub will close the issues as expected.

Happy to move them though!

kennykerr (approved):

PR #977, opened 2021-07-14
Enables experimental support for raw-dylib

### LINKED ISSUE #463: Linker error calling certain functions
The metadata indicates the name of the DLL that exports the given function, but Rust doesn't yet support DLL imports and expects a LIB file to resolve those imports. If you have the Windows SDK installed, there will usually be a LIB file by the right name that Rust can use but this is not guaranteed. There are a few edge cases where this doesn't quite line up. I considered generating [my own LIB files](https://github.com/microsoft/windows-rs/issues/142) for the `windows` crate but we have some folks looking into implementing [DLL imports](https://github.com/rust-lang/rfcs/blob/master/text/2627-raw-dylib-kind.md) for Rust and I'm hopeful that will arrive soon. 

One workaround is to generate and provide your own lib files. This process is [briefly described here](https://github.com/microsoft/windows-rs/issues/142#issuecomment-741054934).

Another workaround is to copy the declaration just for the function in question and change the name of the lib to match a suitable lib that does support the function. The msdn docs usually have the correct name of a lib.

kennykerr:
The `raw-dylib` linking support that @ricobbe has been working on has made it into nightly builds. This update conditionally takes advantage of `raw-dylib` for Win32 functions when the `raw_dylib` feature is selected. Unfortunately, a test could not be added because cargo ends up sharing the feature-enabled `windows` crate with all tests and I'm not yet ready to turn it on everywhere. Still, this lets us start experimenting with the feature in other crates. Once this is published I'll update [some of the samples](https://github.com/microsoft/windows-samples-rs) to use this feature. 

Fixes #463

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,3 +31,4 @@ targets = ["x86_64-pc-windows-msvc"]
 [features]
 default = ["macros"]
 macros = ["gen", "windows_macros"]
+raw_dylib = ["gen/raw_dylib"] # TODO: remove feature once raw-dylib has stablized
--- a/crates/gen/Cargo.toml
+++ b/crates/gen/Cargo.toml
@@ -7,3 +7,6 @@ license = "MIT OR Apache-2.0"
 description = "Code generation for the windows crate"
 
 [dependencies]
+
+[features]
+raw_dylib = []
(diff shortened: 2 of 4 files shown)


rylev:
@kennykerr Can you try using the [v2 feature resolver](https://doc.rust-lang.org/cargo/reference/features.html#feature-resolver-version-2) (which will be the default feature resolver in Rust 2021, to see if you can get the tests working?

kennykerr:
@rylev I gave it a try - adding `resolver = "2"` - but it didn't solve the problem.

PR #978, opened 2021-07-14
Document the usage of `HRESULT::from_thread()`

DoctorNefario:
This small amount of documentation should be helpful to anyone looking through the windows-rs docs for a way to handle Win32 errors.

This example is tested when running `cargo test --all`, so it shouldn't be at risk of becoming outdated.

<details>
<summary>Screenshot of generated documentation</summary>

image
</details>


kennykerr (approved):
Thanks!

Issue #983, opened 2021-07-16
NTSTATUS error codes


tim-weis:
At this time, both standard [system error codes](https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes) and [HRESULT](https://docs.microsoft.com/en-us/office/client-developer/outlook/mapi/hresult)s are well covered by the Windows crate. This is not the case for [NTSTATUS](https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/using-ntstatus-values) based error codes. Those are rarely used in the Windows API surface, though I just ran into a set of functions that do, namely in the [Cryptography API: Next Generation](https://docs.microsoft.com/en-us/windows/win32/seccng/cng-portal) (e.g. [`BCryptOpenAlgorithmProvider`](https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptopenalgorithmprovider)).

Error handling with those APIs is manual, and with [NTSTATUS](https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/Foundation/struct.NTSTATUS.html) being a generated type, interoperability across crate boundaries is stifled. Are there any plans to lift `NTSTATUS` out of the generated code into Windows' [Error](https://docs.rs/windows/0.17.2/windows/struct.Error.html) type?

kennykerr:
Ooh, [I love BCrypt](https://docs.microsoft.com/en-us/archive/msdn-magazine/2007/july/applying-cryptography-using-the-cng-api-in-windows-vista). 😉 

So there are a few different angles to this. 

The general principle is that if a type exists in both the WinRT and Win32 type systems then it must necessarily be lifted out and defined canonically in the Windows crate. `NTSTATUS` is not such a type so it can remain where it is. 

However, `NTSTATUS` is a common error type so it makes sense to enlighten it so that callers can easily convert it to a `Result` if needed. This is what I did for `BOOL` and I think that works well enough. In the case of `NTSTATUS`, there's no ambiguity over whether it is in fact representing a status code so it could even provide a more seamless experience. 

As far as interoperability across crates, that should be solved with #432 however I would suggest that crates should prefer to convert error codes like `WIN32_ERROR` and `NTSTATUS` into an `HRESULT` so that it can be portably and uniformly treated as a `windows::Result`. An idiomatic Rust wrapper library for BCrypt would not expose `NTSTATUS` error codes as at all.

tim-weis:
[...]

PR #989, opened 2021-07-20
Prepares HRESULT to transparently store NTSTATUS values

tim-weis:
Introducing idiomatic Rust error handling for `NTSTATUS` values requires two steps:

1. Allowing `HRESULT` to transparently encode `NTSTATUS` values.
2. Extending/augmenting/enlightening the generated `NTSTATUS` struct.

This PR proposes a solution to solving the first (for the moment). The core idea revolves around the fact that any [`NTSTATUS`](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/87fba13e-bf06-450e-83b1-9241dc81e781) value can be mapped to an [`HRESULT`](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/0642cb2f-2075-4469-918c-4441e69c548a) value, that is distinct from any other `HRESULT` value. This makes `HRESULT` the perfect container to store any Win32 system error code, `NTSTATUS`, or (regular) `HRESULT`.

The following changes were done:

* Added `HRESULT::from_ntstatus()`:  
  This turns `NTSTATUS` values into `HRESULT`s, similar to `from_win32()`. Outside of testing, this is not used, but rather is in preparation of a future `impl From<NTSTATUS> for windows::Result`.
* Added `HRESULT::ntstatus_code()`:  
  Similar to `win32_code()` verifies whether the stored value is an `NTSTATUS` value and, if so, returns the decoded `NTSTATUS` value.
* Adjusted `HRESULT::message()`:  
  Since `FormatMessage` can only be used with (regular) `HRESULT`s, this function produces a generic message for `NTSTATUS` values (e.g. `"NTSTATUS: 0xC0000005"`).
* `std::fmt::Debug for Error` did not need adjustments. In case of an `NTSTATUS`, the `message` already contains the decoded value and an additional debug field is not required.

If that all looks reasonable, I would try working on the second part. For that I would need a bit of guidance, though (e.g. would this be best done in `gen_replacement()` or `gen_extensions()`, and so on).

### DIFF
--- a/src/result/error.rs
+++ b/src/result/error.rs
@@ -81,16 +81,6 @@ impl Error {
 
         self.code.message()
     }
-
-    /// Returns the win32 error code if the underlying HRESULT's facility is win32
-    fn win32_code(&self) -> Option<u32> {
-        let hresult = self.code.0;
-        if ((hresult >> 16) & 0x7FF) == 7 {
-            Some(hresult & 0xFFFF)
-        } else {
-            None
-        }
-    }
 }
 
 impl std::convert::From<Error> for HRESULT {
@@ -143,7 +133,7 @@ impl std::fmt::Debug for Error {
         debug
             .field("code", &format_args!("{:#010X}", self.code.0))
             .field("message", &self.message());
-        if let Some(win32) = self.win32_code() {
+        if let Some(win32) = self.code.win32_code() {
             debug.field("win32_code", &format_args!("{}", win32));
         }
         debug.finish()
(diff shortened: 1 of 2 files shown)


kennykerr:
Sorry I didn't get a chance to comment further on #983 - been quite sick the last few days - will take a look at this soon.

tim-weis:
I'm sad to hear that. Hope you get all the sleep and rest you need for a quick recovery.

PR #991, opened 2021-07-21
Improved implementation support

kennykerr:
More on #81, this makes it easier to write more interesting implementations that need to hop between implementation and projected types. 

I also added a more interesting test for a fully functioning `IIterable<T>` implementation.


src/traits/to_impl.rs
```diff
@@ -0,0 +1,5 @@
+use super::*;
+
+pub unsafe trait ToImpl<T: Interface> {
```
rylev, line 3: Can we document this? (especially what is needed to use this safely - even if it's not meant to be used directly by end users)

kennykerr: Added docs.

PR #996, opened 2021-07-23
Run cargo-clippy in the CI for every push

MarijnS95:
`clippy` is extremely handy in pointing out code smells, and has a hoist of suggestions to (usually) write the same Rust code in a more concise way. Seeing some earlier `clippy` passes committed to this repo I can't help but request for lints to be checked in every PR rather than procrastinating them to larger batches. It's easier to fix a few to appease the CI every now and then - and keep the code free of apparent smells at all times - than it is to deal with large swaths the fact. I hope you agree.

This PR is pretty much draft (but draft status prevents CI from running) to show what that'd look like, but keep in mind that `-Dwarnings` stops clippy after the first crate with a non-zero exit-code; there are far more warnings than shown here.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,15 @@ jobs:
     - name: run cargo fmt
       run: cargo fmt --all -- --check
 
+  cargo_clippy:
+    runs-on: windows-latest
+    name: check cargo clippy
+    steps:
+    - uses: actions/checkout@v2
+    - name: run cargo clippy
+      # TODO: Add  --all-targets when the tests adhere to clippy lints
+      run: cargo clippy --all -- -D warnings
+
   cargo_doc:
     runs-on: ubuntu-latest
     name: check cargo docs


kennykerr:
Although I like the idea of running it, there are some stubborn warnings that I haven not yet been able to get rid of. I would have to tackle those before contemplating a change like this.

MarijnS95:
@kennykerr That is totally understandable, and it makes no sense to merge this before all current warnings and errors are fixed.

Anything in particular that's annoying but doesn't have a proper solution? Disabling insensible (for the context) warnings shouldn't be too big of a deal, as long as we do it in code. I don't like projects running around with complicated argument strings in their CI, I just wish to run `cargo clippy --all --all-targets` locally before submitting changes, without being presented with lots of noise.

kennykerr:
> Anything in particular that's annoying but doesn't have a proper solution?

Probably just my inexperience. There's been a recurring complaint about replacing a `loop` with a `while let` the syntax of which I've not been able to get my head around.

MarijnS95:
> Probably just my inexperience. There's been a recurring complaint about replacing a `loop` with a `while let` the syntax of which I've not been able to get my head around.

Good to know that the lints are not the problem. I've only found one occurence of `clippy::while_let_loop` that seems rather trivial to solve (solution pushed to this PR) - is this the one you were fighting with?

I guess all that remains then is for someone to fix all the lints, so that this can be enabled. Note that it _will_ give temporary CI "flukes" around every 6 weeks when Rust updates and enables new/improved lints, but that's only beneficial to the codebase in the end.

kennykerr:
Will reopen when I get a chance to work on the clippy issues again.

MarijnS95:
@kennykerr Sounds like a good idea. Note that one of the downsides of running with `-Dwarnings` (it's needed to get a non-zero exit code and actually fail the CI) is that it stops processing on the first crate in our workspace with warnings (that are considered errors/denials at that point). The list seen in the CI isn't representative, unfortunately.

When/where do you want to tackle the safety docs? Is that something you are working on and merging separately, after which I can drop `--all-targets` from this PR allowing the CI to succeed for now?

kennykerr:
I should get to that later today.

kennykerr:
All done.

MarijnS95:
> All done.

I wasn't aware that clippy allowed empty `# Safety` sections :hand_over_mouth:, seems like a bug or oversight that should be solved at some point. Either way this allows us to slowly bring the safety docs online while having linting at the same time, neat!

CI should now succeed on clippy, and we can add `--all-targets` later to clean up the tests. I do consider this a non-trivial task though.

PR #997, opened 2021-07-23
Minor usability improvements for NTSTATUS

kennykerr:
Based on feedback here https://github.com/microsoft/windows-rs/pull/994

### DIFF
--- a/crates/gen/src/types/ntstatus.rs
+++ b/crates/gen/src/types/ntstatus.rs
@@ -9,7 +9,7 @@ pub fn gen_ntstatus() -> TokenStream {
         impl NTSTATUS {
             #[inline]
             pub const fn is_ok(self) -> bool {
-                self.0 == 0
+                self.0 & 0x8000_0000 == 0
             }
 
             #[inline]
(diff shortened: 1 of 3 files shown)


tim-weis:
This looks better, and addresses all concerns that were raised. It supports convenient `.ok()?` error propagation, makes it possible to preserve non-zero success codes, and provides a way to translate from `NTSTATUS` codes to corresponding `HRESULT` values for easy comparison.

It's better than what we currently have, and I would be happy to try how well this works out in practice.

Honestly, though, I'm growing increasingly skeptical, that solving this entirely outside the core library is the correct solution. The binary protocol for [HRESULT](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/0642cb2f-2075-4469-918c-4441e69c548a) expressly includes representation of `NTSTATUS` values, where Windows' [HRESULT](https://docs.rs/windows/0.17.2/windows/struct.HRESULT.html) implementation only covers part of it. It would feel natural to have an implementation that covers the entire binary protocol, especially with `HRESULT` being the pivot point of Windows' error representation.

Maybe this will need to be re-evaluated at some point. Not at this time, I'm more than happy with what this change delivers as it stands. Thanks!

PR #998, opened 2021-07-23
Remove unecessary helper functions

### LINKED ISSUE #980: Refer to create_instance in CoCreateInstance documentation
Doing so will help users learn the easier helper function exists. 👍

kennykerr:
Thanks to #909, there is no longer any need for the Windows crate to provide COM-specific helpers as the APIs themselves are now quite straightforward to use directly. 

Closes #980


filips123:
I think you also need to update [getting started](https://github.com/microsoft/windows-rs/blob/master/docs/getting-started.md#initialization-code) docs which still say we should use `windows::create_instance` and `windows::initialize_mta`.

PR #1025, opened 2021-08-09
Additional COM signatures supported for implementations

kennykerr:
Adds support for implementing methods with retval signatures.


MarijnS95:
That's truly cool stuff, I really like this :tada: !

https://github.com/MarijnS95/windows-rs/blob/e48206c8dc178777e8bfd5029e71c4e119bba901/examples/dxc/src/main.rs#L53-L69

PR #1026, opened 2021-08-09
Support IVectorView

j0shuams:
@kennykerr


kennykerr:
Running 'cargo fmt' and pushing the results should unblock the build to proceed.

kennykerr:
Also, it won't build with warnings. Best to keep it clean, even if its still early in development.

tests/implement/tests/vector_view.rs
```diff
@@ -0,0 +1,141 @@
+use std::vec;
+use test_implement::*;
+use windows::*;
+use Windows::Foundation::Collections::*;
+
+#[implement(
+    Windows::Foundation::Collections::IVectorView<T>,
```
kennykerr, line 7: `IVectorView<T>` requires `IIterable<T>`. Soon the `implement` macro will pick this up automatically, but for now you'll just have to list that here as well.

tests/implement/tests/vector_view.rs
```diff
@@ -0,0 +1,141 @@
...
+)]
+struct TestView<T>(Vec<T>)
```
kennykerr, line 9: This is fine for now, but we'll also need a non-owning version eventually.

tests/implement/tests/vector_view.rs
```diff
@@ -0,0 +1,141 @@
...
+where 
+    T: ::windows::RuntimeType + 'static;
+
+#[allow(non_snake_case)]
+impl<T: ::windows::RuntimeType + 'static> TestView<T> {
+    /* GetAt returns the value at the given index, and panic if given an invalid index */
+    fn GetAt(&self, index : u32) -> Result<T> {
+        let vec_size = self.0.len() as u32;
+        assert!(index < vec_size, "Given index ({}) was out of bounds for the IVectorView (length {})", index, vec_size);
```
kennykerr, line 18: Since this is intended for library use, we can't have the implementation panic. Instead, this should return `E_BOUNDS`. [Here's an example](https://github.com/microsoft/windows-rs/blob/6116afb9e8c426dd90ca44e654aede9208e5400f/tests/implement/tests/into_impl.rs#L25).

tests/implement/tests/vector_view.rs
```diff
@@ -0,0 +1,141 @@
...
+        self.0.get(index as usize).cloned().ok_or_else(|| panic!())
+    }
+   
+    /* Size returns the length of the underlying vector */
+    fn Size(&self) -> Result<u32> {
+        let size : u32 = self.0.len() as u32;
+        Ok(size)
```
kennykerr, line 25:
```suggestion
        Ok(self.0.len() as _)
```

Type inference is helpful here.

tests/implement/tests/vector_view.rs
j0shuams: Added the type, and we need to implement First, which requires an implementation of an IIterator. Used the same implementation as in `into_impl.rs`. Is that expected?

kennykerr: Yes, that's the basic outline - `IIterable<T>` must be implemented on the same object while `IIterator<T>` must be implemented on a seperate object.

tests/implement/tests/vector_view.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    // test_index_of_three_vec ALWAYS FAILS until we fix the `==` issue 
+    #[test]
+    fn test_index_of_three_vec() {
+        let three_vec: IVectorView<i32> = TestView(vec![5, 120, 625]).into();
+ 
+        // ? correct way to get a pointer to a u32 ?
+        let mut idx = &mut 1;
+        
+        let index_of_five = three_vec.IndexOf(5, idx).unwrap();
```
kennykerr, line 94:
```suggestion
        let mut idx = 0;
        
        let index_of_five = three_vec.IndexOf(5, &mut idx).unwrap();
```

kennykerr:
Feel free to reopen once this is further along.

PR #1027, opened 2021-08-10
Fix clippy warnings in core crates

MarijnS95:
For #996

`cargo clippy --all -- -A clippy::missing_safety_doc` can now run without any warnings.  Note that this excludes every `#[test]` unless `--all-targets` is passed, which does not lint cleanly yet.

---

@kennykerr What do you think about gradually introducing `clippy` to windows-rs instead? It's quite the burden to fix tons of warnings in the tests (that usually intentionally write things a certain "inefficient" way for validation) or to write appropriate `# safety` docs, when it is anyway much more relevant to keep the published crates clean and held to a certain standard instead.
After this PR we can reopen #996 and I can subsequently change the command to the one listed above. I'll document the desire to enable `--all-targets` at some point, as well as removing `-A clippy::missing_safety_doc`.

### DIFF
--- a/crates/fmt/src/main.rs
+++ b/crates/fmt/src/main.rs
@@ -28,7 +28,7 @@ fn update_file(path: &std::path::Path, pattern: &str) -> std::io::Result<()> {
                 .take_while(|&c| c == ' ')
                 .count();
 
-            let indent = std::iter::repeat(' ').take(indent).collect::<String>();
+            let indent = " ".repeat(indent);
 
             // Replace `use` with macro call and insert curly braces around the UseTree
             let macro_ = format!(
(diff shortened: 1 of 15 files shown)


kennykerr:
Let's get what we can building. I don't know what's involved with getting the tests clean but if that's minimal let's do it. Otherwise we can at least get everything else clean as a start.

MarijnS95:
@kennykerr There's a lot of noise from comparing against boolean literals with `assert_eq!()`, a bunch of `.len() == 0` and a bunch of nontrivial operations with copies for the tests; I don't dare to touch those as I'm not entirely certain what you're trying to test.

Shall I push as many fixes as possible to this PR and get #996 reinstantiated, then you can look at the warnings and fix the remaining issues that require some more thought?

kennykerr:
Sounds good - thanks!

kennykerr (approved):

PR #1047, opened 2021-08-13
Find winmd files without reading `PATH` nor copying to `target/`

### LINKED ISSUE #979: Remove binary packaging support
Cross-compiling a Windows-rs project from Linux (ie. one of the test crates within this repo) result in the following build error:

```console
$ cargo b -p test_bstr --target x86_64-pc-windows-gnu
...
error: proc macro panicked
 --> tests/bstr/build.rs:4:5
  |
4 | /     windows::build! {
5 | |         Windows::Win32::Foundation::BSTR,
6 | |     };
  | |______^
  |
  = help: message: Path not ending in `;`
```

The `PATH` variable on Linux is separated with colons, not semicolons.

It is worrying to see the same undocumented logic and error duplicated across three distinct places in the codebase. What is the first path in `PATH` supposed to point to? Why are the winmd files copied to `../../.windows/winmd` relative to it? Do I need to worry about the bit that copies to `%PROFILE%`, too?

On a related note, is it worth turning the `cargo build` for `ubuntu-latest` in the CI into `cargo check --all --target x86_64-pc-windows-gnu`, to ensure cross-compiling remains in working order?

MarijnS95:
Fixes #979

Non-Windows platforms - which are supported for cross-compiling - do not set the output directory in `PATH` nor use semicolons to separate this variable, resulting in errors Split out `winmd` "discovery" by emitting the path to these files in a build step, that is subsequently compiled into `windows_gen`.  This is more efficient than `include_bytes!` which was used prior to the `PATH` system as no copies or binary includes are required at all.  Copying of DLL targets to the `target/` dir for easy consumption and running of crates is still performed, but only on Windows for the same `PATH` reason.

In the future, if a chain of crates is required to export `winmd` files for downstream crates to consume, this can be extended to also export their `$CARGO_MANIFEST_DIR/.winmd/windows` in such a way that the next crate can pick it up, again without copying any files around.

### TODO

@riverar [Linked to](https://github.com/microsoft/windows-rs/issues/979#issuecomment-890223809) the copy function on an older commit, but I think he intended to link to the way the target directory for those copies was derived:

https://github.com/microsoft/windows-rs/blob/49ecfbc2f99cbc4bac32d6d771eeb589fe9d4e76/crates/macros/src/lib.rs#L121

Should be migrate back to this and drop any and all `PATH` handling entirely?


riverar:
My link was just meant to show that we copy artifacts out of the .windows subfolders. No PATH handling should remain.

MarijnS95:
@riverar I interpreted it as pointing to a specific commit, conveniently right before `PATH` handling was introduced. Will remove!

crates/gen/build.rs
```diff
@@ -1,36 +1,24 @@
-#[cfg(windows)]
 fn main() {
-    let mut source: ::std::path::PathBuf = ::std::env::var("CARGO_MANIFEST_DIR")
-        .expect("No `CARGO_MANIFEST_DIR` env var")
-        .into();
+    let path_file =
```
kennykerr, line 2: We can probably just revert to using `include_bytes` and avoid all of this build trickery. That macro is slow but not that slow to warrant all the extra complexity. And hopefully it will eventually improve. 

https://github.com/rust-lang/rust/issues/65818

MarijnS95: Indeed, I don't recall it being noticeable but then I haven't (had a reason to) performed a profile to find out either way. Will revert to it as we don't need this build trickery (unless we'd like to support "scenario 3", though even that might be possible with `include_bytes` one way or another).

MarijnS95: @kennykerr 098ef4926b2db7aa4afdf033dc04048cd313918e dropped the `if !result.iter().any(|file| file.name.starts_with("Windows."))` before adding "builtin" winmds too, that should be fine because these files are added at the end and would anyway be visited last.

MarijnS95:
https://github.com/microsoft/windows-rs/blob/49ecfbc2f99cbc4bac32d6d771eeb589fe9d4e76/crates/macros/src/lib.rs#L121-L132

I don't understand this loop. That's wrong if someone builds with their target folder in a different location.

kennykerr:
Yep, that loop should not be needed anymore. It was an attempt to find dirs across crates which we no longer need.

MarijnS95:
> Yep, that loop should not be needed anymore. It was an attempt to find dirs across crates which we no longer need.

It's gone now, let me know if this is okay. I'm happy because this has now unblocked building on Linux (I'm not cross-compiling at all :flushed:), but it should work for your Windows (and rustdoc) purposes as well.

kennykerr:
@MarijnS95 🤣 I've actually had some internal interest for cross-platform support, so you never know.

crates/gen/build.rs
kennykerr: The `if` is needed in the event that a project provides its own `.windows/winmd` folder with its own winmd files for the Windows API.

MarijnS95: Yeah, but it was removed at some point. Wrongly? I was under the assumption that winmds were linearly scanned for types, returning on the first match. In that sense a custom winmd would have precedence?

MarijnS95:
> @MarijnS95  I've actually had some internal interest for cross-platform support, so you never know.

Just shoot and I'll throw the cross-platform working DirectXShaderCompiler example back online :grin: 

Joking aside, really glad to hear this.

crates/gen/build.rs
riverar: I believe this check exists for several reasons, like
* to strongly discourage use of the reserved (by the OS) `Windows.*` namespace, and
* to prevent developer confusion. Folks may grab a bunch of (older) Windows metadata from the SDK (they are misused often) and plop them into the winmd folder creating a mess

Hackers™️ can override this behavior by renaming their winmd.

I don't have a strong opinion here.

riverar (approved):

riverar (commented):
Looks OK to me. I defer to @kennykerr due to high impact. 😂

crates/gen/build.rs
MarijnS95: @riverar Yes it would definitely be odd to pick up builtin winmds if the user explicitly uses different ones. But then filenames are at the same time fragile to solely base this on.

kennykerr: The removal was accidental. Projects may need to use newer/different metadata. For example, building the next version of Windows requires the new metadata that isn't yet publicly available. Other projects based on Windows, like Xbox, often have their own set of Windows APIs that may look a bit different. We also have various internal teams that need to provide their own metadata for various reason like provenance. An of course the hackers. Anyway, this is important. 😉

kennykerr: Obviously I'll need to add a test for this.

MarijnS95: Sure, it's added back now.

kennykerr:
Haven't had a chance to dig in, but a quick smoke testing compiling https://github.com/microsoft/windows-app-rs fails unfortunately.

kennykerr:
For a change as invasive as this, it may also help to smoke out issues by building https://github.com/microsoft/windows-samples-rs

MarijnS95:
@kennykerr Can you clarify what exact issue you're seeing? For `windows-app-rs`, after fixing capitalization issues in `cargo.toml`, I'm left with the following:

```rust
error: `Microsoft.UI.Xaml.Application` not extendable
  --> src/main.rs:27:41
   |
27 | #[implement(extend Microsoft::UI::Xaml::Application, override OnLaunched)]
   |          
```
Is this a result of building with the wrong metadata?

Same for the samples; what errors should I be looking out for? Do I need to be on Windows to test specific things (need to revive a Windows install first that decided to corrupt itself for no apparent reason)?

kennykerr:
Yes, that's the error I saw. I would its misleading because the type is not actually found in metadata at all.

MarijnS95:
> Yes, that's the error I saw. I think its misleading because the type is not actually found in metadata at all and because its not found its not extendable.

And that error isn't present when compiling against windows-rs master? This is hard to test for me currently due to the original "Path doesn't contain `;`" errors.

kennykerr:
Yes that error is not present in master.

MarijnS95:
@kennykerr It's pretty obvious what happens: `bindings/` has the `.windows/winmd` directory used to generate bindings, but `#[implement]` in the root crate needs to use these files as well. We have various ways to solve this including `ln -s .windows/winmd` from the crate root, or "exporting" the winmd path from `bindings` as suggested in https://github.com/microsoft/windows-rs/issues/979#issuecomment-880198391 and pointing `windows-rs` to that one way or another, so that it uses the winmds directed by the `bindings` crate.

Let's discuss before jumping on an implementation detail.

riverar:
> `ln -s .windows/winmd` from the crate root

Tip: Symlinks in Windows require administrative privileges in normal scenarios, so that's out.

MarijnS95:
@riverar I'm unfortunately aware, and it nauseates me. IMO this should be fully solved from Rust, no messing around with paths and links/copies by the user.

kennykerr:
> (EDIT: This requirement is pretty much your "scenario 3" we were trying to avoid 😛)

This worked before because the `.windows` folder was relative to the workspace rather than the crate so both crates shared the same `.windows/winmd` folder.

kennykerr:
Basically, the `TypeReader` should 

1. look in `<workspace>/.windows/winmd` and 
2. add the `gen` crate's default winmd files if the previous step didn't have any `windows.*` winmd files.

MarijnS95:
> This worked before because the `.windows` folder was relative to the workspace rather than the crate so both crates shared the same `.windows/winmd` folder.

"Before" referring to the situation before, `PATH` changes, correct? Now on `master` it works because of unregulated copies to the target folder.

> Basically, the `TypeReader` should
> 
> 1. look in `<workspace>/.windows/winmd` and
> 2. add the `gen` crate's default winmd files if the previous step didn't have any `windows.*` winmd files.

That's what it's doing already, but not relative to `<workspace>` but the `<crate>` (`CARGO_MANIFEST_DIR`), as you said in your previous comment. This `<crate>` changes between building a "toplevel" package and its `bindings` crate.

Should we use cargo metadata to point windows-rs to a different directory, so that we can just parse `$CARGO_MANIFEST_DIR/Cargo.toml`?

kennykerr:
Yes, that's the way it worked before the binary packaging support was added.

kennykerr:
I need to get this wrapped up. It sounds like you still have some open questions about how to proceed. Do you mind terribly if I just take this over and get it done. I'd like to get this fixed up and publish a new version of the crate.

kennykerr:
Closing in favor of #1067

MarijnS95:
> It sounds like you still have some open questions about how to proceed. Do you mind terribly if I just take this over and get it done. I'd like to get this fixed up and publish a new version of the crate.

I don't have time for this back-and-forth so feel free. Looks like #1067 is a simple "yes" to my question inhttps://github.com/microsoft/windows-rs/pull/1047#issuecomment-900512379, but in code form.

MarijnS95:
@kennykerr Do you mind awfully if I rebase this branch and open a new PR after yours is merged, so that we at least keep some cleanups/simplifications?

kennykerr:
That would be great.

PR #1049, opened 2021-08-13
Splitting out quote support into a separate module

kennykerr:
In preparation for the next big push for #81 I'm refactoring the gen crate into more manageable pieces. This is step one. Splitting it up improves build performance, because crates can build in parallel but also enables me to add more functionality (like metadata generation) as a distinct module without slowing down compilation when only consumption is required.


rylev:
Would you like to take ownership of https://crates.io/crates/squote so you don't have to have a different crate for this? I only created squote for windows-rs

PR #1067, opened 2021-08-19
Remove binary packaging support

### LINKED ISSUE #979: Remove binary packaging support

kennykerr:
This change moves to the simpler model of having the .windows folder relative to the workspace and avoids all of the file copies and cross-crate copying of files as Cargo just isn't up to that at the moment. 

Fixes #979


MarijnS95 (approved):
Works for me (on grey-area "Linux" support), thanks!

PR #1068, opened 2021-08-19
macros/lib: Don't copy DLL files outside of the workspace target folder

MarijnS95:
The previous commit now uses a different folder for `target_dir`, namely `target/` directly instead of `target/<profile>/deps`.  Popping this twice would go outside of the directory containing `target/` and copy DLL files into _every_ folder matching the `<profile>` name (ie. every `debug` folder inside my `Rust/` projects folder).

Fixes: 1291b32 ("Remove binary packaging support (#1067)")


kennykerr (approved):
Good catch.

PR #1069, opened 2021-08-19
Simply workspace path logic

MarijnS95:
Simplifies some code left and right, and solidifies `fn target_dir()` by expecting a `"` to start the JSON value for `"target_directory"` instead of assuming there's no space here with `+1`.

Also print reruns before checking if a directory exists: this way creating the directory after the build script has already ran triggers a rerun, instead of doing nothing.

### DIFF
--- a/crates/reader/src/workspace.rs
+++ b/crates/reader/src/workspace.rs
@@ -10,38 +10,31 @@ pub fn workspace_winmds() -> &'static [File] {
     unsafe { &*VALUE.as_ptr() }
 }
 
-pub fn workspace_dir() -> String {
-    const JSON_KEY: &str = r#""workspace_root":"#;
+fn json_value(key: &str) -> String {
     let json = cargo_metadata();
+    let json_key = format!(r#""{}":""#, key);
 
     let beginning_index = json
-        .rfind(JSON_KEY)
-        .expect("Cargo metadata did not contain `workspace_root` key.")
-        + JSON_KEY.len()
-        + 1;
+        .rfind(&json_key)
+        .unwrap_or_else(|| panic!("Cargo metadata did not contain `{}` key.", key))
+        + json_key.len();
 
-    let ending_index = json[beginning_index..]
-        .find('"')
-        .expect("Cargo metadata ended before closing '\"' in `workspace_root` value");
+    let ending_index = json[beginning_index..].find('"').unwrap_or_else(|| {
+        panic!(
+            "Cargo metadata ended before closing `\"` in `{}` value",
+            key
+        )
+    });
 
     json[beginning_index..beginning_index + ending_index].replace("\\\\", "\\")
 }
 
+pub fn workspace_dir() -> String {
+    json_value("workspace_root")
+}
+
 pub fn target_dir() -> String {
-    const JSON_KEY: &str = r#""target_directory":"#;
-    let json = cargo_metadata();
-
-    let beginning_index = json
-        .rfind(JSON_KEY)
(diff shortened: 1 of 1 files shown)


crates/macros/src/lib.rs
```diff
@@ -90,28 +90,25 @@ pub fn build(stream: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
                             }
                         }
                     }
                 }
             }
 
-            let mut source : ::std::path::PathBuf = #workspace_dir.into();
+            let mut source: ::std::path::PathBuf = #workspace_dir.into();
             source.push(".windows");
+            println!("cargo:rerun-if-changed={}", source.to_str().expect("`CARGO_MANIFEST_DIR` not a UTF-8 string"));
```
kennykerr, line 109: I think moving this out the existence check below forces Cargo to rerun every time as it gets confused by the absence of the "changed" directory.

crates/macros/src/lib.rs
```diff
@@ -60,6 +60,8 @@ pub fn build(stream: proc_macro::TokenStream) -> proc_macro::TokenStream {
             );
 
             path.push("windows.rs");
+            println!("cargo:rerun-if-changed={}", path.to_str().expect("`OUT_DIR` not a UTF-8 string"));
```
kennykerr, line 63: I'm not clear why we need this. If the .windows folder changes the build script reruns and causes the OUT_DIR to be updated. That's the output. We watch the input, not the output.

crates/macros/src/lib.rs
MarijnS95: Ouch, that's still not fixed?

crates/macros/src/lib.rs
MarijnS95: It makes it easier to regenerate `windows.rs` without a full `cargo clean` (loosing potentially large intermediates); just modifying or deleting `windows.rs` (`rm target/**/windows.rs`) is enough to trigger a regen. However, this doesn't seem to debounce properly with the build script writing to the file at the same time, and always reruns it instead (also when printing after writing). Removed it, should be fine to modify the file containing `build!()` too.

crates/macros/src/lib.rs
kennykerr: Not as of a month or two when I last checked.

crates/reader/src/workspace.rs
```diff
@@ -28,14 +28,13 @@ pub fn workspace_dir() -> String {
 }
 
 pub fn target_dir() -> String {
-    const JSON_KEY: &str = r#""target_directory":"#;
+    const JSON_KEY: &str = r#""target_directory":""#;
```
riverar, line 31: Is this valid? This is performing a search for `"target_directory":""` which will fail right?

kennykerr: I think the raw string literal is enclosed by `r#"` and `"#` so the resulting string is merely `"target_directory":"` instead of `"target_directory":`.

kennykerr: But let's be consistent with `workspace_dir`. Should probably collapse the two into a single function.

kennykerr:
@MarijnS95 I moved the code for the build macro to the gen crate with #1070 and to avoid you having to merge that with this PR, I just copied your changes to that PR. So this PR can just focus on the workspace improvements.

crates/reader/src/workspace.rs
MarijnS95: Fair point, I hadn't even spotted `fn workspace_dir()` :P

kennykerr (approved):

PR #1071, opened 2021-08-20
Use explicit version in cargo.toml (avoid cargo-deny errors)

Guiguiprim:
Small improvement to make sure we are cargo-deny friendly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ gen = { package = "windows_gen", path = "crates/gen",  version = "0.18.0", optio
 const-sha1 = "0.2"
 
 [dev-dependencies]
-gen = { package = "windows_gen", path = "crates/gen" }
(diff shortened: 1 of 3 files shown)


kennykerr:
First time I've heard of cargo-deny. What's that all about? Just wondering how and whether to support it.

Guiguiprim:
[Cargo-deny](https://github.com/EmbarkStudios/cargo-deny) is a tool often use in CI, to do linting on a project dependencies. It can check for licenses type (making sure that they fit your requirement), that you do not include different versions of the same crates, and more.

One of the checks ensures that every dependencies as an explicit version (even within a same workspace).

This PR makes sure that the crate can be use without breaking the CI of people using cargo-deny.

kennykerr (approved):

kennykerr:
Sounds good, thanks for the explanation! Maybe we should consider running it as part of the CI build here as well at some point.

Issue #760, opened 2021-04-30
Support more standard types


Legorooj:
Currently, `windows-rs` doesn't support using types like `OsString` or `PathBuf`/`Path`. These types seem natural to use with the windows API, so could we get support for them? Happy to implement it if you can point me in the right direction.

ibigbug:
Good idea.

I recently started building on `windows-rs` and realized that it seems to require something like this https://docs.rs/widestring/0.4.0/widestring/type.U16CString.html to pass strings.

Issue #710, opened 2021-04-17
Request for implementations of From/Into to convert between PWSTR/PCWSTR and OsString/&OsStr


enebo:
Similar to #645.  I am working with raw command-line info as OsStr/OsString and I want to pass that to CreateProcessW.  Here is mostly the gist of how I am dealing with PWSTR now:

```rust
    let command_line = quote_vec(command_line); // OsString
    let mut command_line_wide: Vec<u16> = OsStr::new(&command_line).encode_wide().chain(once(0)).collect();
    let mut c = PWSTR::default();
    c.0 = command_line_wide.as_mut_ptr();
```

c is passed as second param to CreateProcessW but it would be nice to just to pass command_line itself.

rylev:
This should be safe to do so I think we should it. We do allow compilation on Linux, so we should be careful to document things properly as if we ever allow for the use of this crate on all platforms, this will not work on Linux.

enebo:
@rylev Does windows-rs work on non-windows platforms at all?  I can see you may allow cross-compilation so you can build the crate on linux but I assumed it would still have OsStr as u16 for the sake of that compilation.

rylev:
@enebo currently the crate is only useable on Windows (though it builds on Linux). This might be the case forever, but it might not be. I personally don't want to get in a situation where we have very subtle reliance on platform specific behavior where a comment could easily make that apparent.

ibigbug:
I had the same question today and found this example https://github.com/microsoft/windows-rs/blob/master/examples/clock/src/main.rs#L414

So you probably don't need to do the conversion and can pass in the `&str` directly

MarijnS95:
> currently the crate is only useable on Windows (though it builds on Linux)

Following https://github.com/microsoft/windows-rs/discussions/566 this crate is perfectly usable on Linux and allows to retrieve an interface from DXC without hassle. I hope to share a working example in the coming days :)

Wide strings are tricky because they are 32-bit on Linux platforms, and even DXC treats them like that. The `widestring` crate could be used to abstract this away, and otherwise `windows-rs` implementations should be careful to add `#[cfg]` so that it won't compile on Linux instead of silently doing the wrong thing.
[...]

PR #1081, opened 2021-08-22
`IntoParam<PWSTR>` for OString & OsStr

sabbyX:
Initially when you got [`PathBuf`](https://doc.rust-lang.org/std/path/struct.PathBuf.html) or [`Path`](https://doc.rust-lang.org/std/path/struct.Path.html) (Both are `OsString` and `OsStr` in behind), you would have to convert to UTF-8, which is unnecessary (PWSTR is UTF-16 too) and lossy. This PR fixes it by implementing `IntoParam<PWSTR` for OsString & OsStr.
There's inconsistency between functions, some takes `IntoParam<PWSTR>`, others just take `PWSTR`. The function I encountered (`SHGetFileInfoW`) was taking `IntoParam<PWSTR>`, perhaps I should implement `From<OsString/OsStr>` too?

### DIFF
--- a/crates/gen/src/pwstr.rs
+++ b/crates/gen/src/pwstr.rs
@@ -46,5 +46,17 @@ pub fn gen_pwstr() -> TokenStream {
                 ::windows::Param::Boxed(PWSTR(::std::boxed::Box::<[u16]>::into_raw(self.encode_utf16().chain(::std::iter::once(0)).collect::<std::vec::Vec<u16>>().into_boxed_slice()) as _))
             }
         }
+        impl<'a> ::windows::IntoParam<'a, PWSTR> for &'a ::std::ffi::OsStr {
+            fn into_param(self) -> ::windows::Param<'a, PWSTR> {
+                use std::os::windows::ffi::OsStrExt;
+                ::windows::Param::Boxed(PWSTR(::std::boxed::Box::<[u16]>::into_raw(self.encode_wide().chain(::std::iter::once(0)).collect::<std::vec::Vec<u16>>().into_boxed_slice()) as _))
+            }
+        }
+        impl<'a> ::windows::IntoParam<'a, PWSTR> for ::std::ffi::OsString {
+            fn into_param(self) -> ::windows::Param<'a, PWSTR> {
+                use std::os::windows::ffi::OsStrExt;
+                ::windows::Param::Boxed(PWSTR(::std::boxed::Box::<[u16]>::into_raw(self.encode_wide().chain(::std::iter::once(0)).collect::<std::vec::Vec<u16>>().into_boxed_slice()) as _))
+            }
+        }
     }
 }


tim-weis:
The comment seems to suggest, that `OsString` and `OsStr` were to use UTF-16 on Windows. I could be reading something into it that's not there, but just to be safe, I'll put this one out:

To my knowledge there is no string type in Rust that uses UTF-16 encoding. Both `OsString` and `OsStr` use an encoding commonly referred to as [WTF-8](https://en.wikipedia.org/wiki/UTF-8#WTF-8). It's a relaxed variant of UTF-8, that allows to encode unmatched UTF-16 surrogate pairs as valid UTF-8. It's still UTF-8, though.

The assumption that `OsString` and `OsStr` were to use UTF-16 and wouldn't need any character encoding conversions seems to have carried forward into the code. The lifetime of the output of either implementation is unrelated to the lifetime of the inputs, yet they use `'a` for either. That doesn't look right.

Though I'm a Rust novice, and could be wrong. On all accounts.

sabbyX:
> To my knowledge there is no string type in Rust that uses UTF-16 encoding. 

It's does not use form native to platform (UTF-16, in case of Windows) internally, its explained pretty well in [docs](https://doc.rust-lang.org/std/ffi/struct.OsString.html). Sorry if my comment seems suggest otherwise, actually meant, the conversion is [lossless](https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStrExt.html#tymethod.encode_wide) unlike the converting OsString -> Rust's strict utf-8 string -> PWSTR.

tim-weis:
That makes sense. Taking the extra conversion through UTF-8 potentially provokes errors that would otherwise be permissable when going straight from `OsString`/`OsStr` to `PWSTR`.

Still, the lifetimes don't appear to be right. Either implementation constructs a temporary `Vec` whose lifetime is unrelated to the lifetime of the input, yet the code subscribes the same lifetime `'a` to both.

sabbyX:
> Still, the lifetimes don't appear to be right. Either implementation constructs a temporary Vec whose lifetime is unrelated to the lifetime of the input, yet the code subscribes the same lifetime 'a to both.

There is no need of lifetime here since there's no borrowing or such happening. Yeah, I should mark the return value as such (`IntoParam<'static, PWSTR>`), But I am concerned that's not how it implemented throughout the crate — return values are tied down to arbitrary lifetimes unrelated to the input.

https://github.com/microsoft/windows-rs/blob/b8be4a8cd3412ea3064fd0db68a8fb79e8f735b4/crates/gen/src/pstr.rs#L38-L50
https://github.com/microsoft/windows-rs/blob/b8be4a8cd3412ea3064fd0db68a8fb79e8f735b4/crates/gen/src/pwstr.rs#L38-L50

kennykerr:
Apologies for the delay. Will look at this soon - just a little swamped at the moment.

crates/gen/src/pwstr.rs
```diff
@@ -46,5 +46,19 @@ pub fn gen_pwstr() -> TokenStream {
                 ::windows::Param::Boxed(PWSTR(::std::boxed::Box::<[u16]>::into_raw(self.encode_utf16().chain(::std::iter::once(0)).collect::<std::vec::Vec<u16>>().into_boxed_slice()) as _))
             }
         }
+        #[cfg(windows)]
+        impl ::windows::IntoParam<'static, PWSTR> for &::std::ffi::OsStr {
```
kennykerr, line 50: If `static` is "more correct" we should at least be consistent within the `pwstr.rs` source for all impls. Can you check that they have all been updated correctly to use the appropriate lifetime? I'm still learning about lifetimes...

kennykerr:
Would you mind adding a simple `/tests/pwstr` test crate to exercise the new impls? I'm planning to clean up the tests soon, but for any new code I'd like to make sure we have a standalone test crate to ease validation.

kennykerr:
Also, this seems to address some of the concerns in #760 and #710 - we don't have to make sure this PR covers everything but we should at least review for completion/gaps.

Issue #737, opened 2021-04-27
Cannot call QueryInterface-like APIs with "NULL" for the ppv parameter


damyanp:
For example, many D3D12 APIs, such as D3D12CreateDevice, operate in a mode where the runtime validates parameters if NULL is passed for the ppv parameter.  Apps often use this to probe whether or not a particular adapter meets their needs, but without performing the heavy-weight, driver-loading, work required to construct and then destruct the object.

See for example https://github.com/microsoft/DirectX-Graphics-Samples/blob/cab614fc3b2ac880f1909c40a485423649940b0d/Samples/Desktop/D3D12HelloWorld/src/HelloWindow/DXSample.cpp#L97

damyanp:
I always forget about this case, sorry for not bringing it up earlier before #735 was completed.  I'd be interested what you think right bindings would look like to support this scenario - it's not obvious to me.

kennykerr:
Yikes, I was not aware of this convention. Thanks for letting me know. I'll noodle on that.

damyanp:
In case anyone else encounters it, here is my workaround:

```rust
    // let mut adapter_: Option<IDXGIAdapter1> = ...

    // We need the variant where we pass in NULL for the outparam.
    #[link(name = "d3d12")]
    extern "system" {
        pub fn D3D12CreateDevice(
            padapter: ::windows::RawPtr,
            minimumfeaturelevel: D3D_FEATURE_LEVEL,
            riid: *const ::windows::Guid,
            ppdevice: *mut *mut ::std::ffi::c_void,
        ) -> ::windows::HRESULT;
    }

    // Check to see whether the adapter supports Direct3D 12, but
    // don't create the actual device yet.
    if D3D12CreateDevice(
        adapter.abi(),
        D3D_FEATURE_LEVEL::D3D_FEATURE_LEVEL_11_0,
        &ID3D12Device::IID,
        std::ptr::null_mut(),
    )
    .is_ok()
    {
        return Ok(adapter);
    }
```

```

kennykerr:
As Rust doesn't support overloads, I'm thinking of adding an "overload" with a suffix like `_query` that simply provides the non-generic parameters. So you would be able to call the generic `D3D12CreateDevice` or the non-generic `D3D12CreateDevice_query`.

damyanp:
Bikeshed: maybe `_validate` for this case?

[...]

PR #1088, opened 2021-08-25
Simpler signature generation

kennykerr:


MarijnS95:
Hi @kennykerr,

This PR unfortunately makes it so that my local bindings to DirectXShaderCompiler revert back to not using the "QueryInterface"-like function signature, thanks to the [detection of `_COM_Outptr`](https://github.com/microsoft/win32metadata/blob/2c83a3893d83d9800a781ddc2066e56fa9523721/sources/ClangSharpSourceToWinmd/MetadataSyntaxTreeCleaner.cs#L706-L710) which is [missing from DXC](https://github.com/microsoft/DirectXShaderCompiler/blob/cb485263b75495015e2a77cc438e72dc813ea287/include/dxc/dxcapi.h#L75-L88). Before I go ahead and PR this annotation to their headers, do you happen to know what is meant by `While this function is similar to CoCreateInstance, there is no COM involvement.` and if this could possibly be the reason for omitting this annotation, since it is generally applied consistently elsewhere?
Also, since it may take quite some time for these headers to go from git repository -> DXC release -> Windows SDK release -> wind32metadata bindings, would it make sense to push the changes as overrides to win32metadata too? Is there a process in place to remove superfluous overrides from win32metadata again when the header changes trickle in?

Secondly, where can I find some documentation on using `_COM_Outptr_` vs `_COM_Outptr_result_maybenull_`? There are many cases where the output can be null, but only if an error is returned; that doesn't count for `result_maybenull_` right? Then there's also `_Outptr_result_nullonfailure_` but that doesn't imply `COM` objects.

PS: The title of this PR doesn't match the commit that was merged, making it harder to find and identify in my list of tabs. Can you update it to reflect that?

MarijnS95:
> Hi @kennykerr,
> 
> This PR unfortunately makes it so that my local bindings to DirectXShaderCompiler revert back to not using the "QueryInterface"-like function signature, thanks to the [detection of `_COM_Outptr`](https://github.com/microsoft/win32metadata/blob/2c83a3893d83d9800a781ddc2066e56fa9523721/sources/ClangSharpSourceToWinmd/MetadataSyntaxTreeCleaner.cs#L706-L710) which is [missing from DXC](https://github.com/microsoft/DirectXShaderCompiler/blob/cb485263b75495015e2a77cc438e72dc813ea287/include/dxc/dxcapi.h#L75-L88). Before I go ahead and PR this annotation to their headers, do you happen to know what is meant by `While this function is similar to CoCreateInstance, there is no COM involvement.` and if this could possibly be the reason for omitting this annotation, since it is generally applied consistently elsewhere?
> Also, since it may take quite some time for these headers to go from git repository -> DXC release -> Windows SDK release -> wind32metadata bindings, would it make sense to push the changes as overrides to win32metadata too? Is there a process in place to remove superfluous overrides from win32metadata again when the header changes trickle in?
> 
> Secondly, where can I find some documentation on using `_COM_Outptr_` vs `_COM_Outptr_result_maybenull_`? There are many cases where the output can be null, but only if an error is returned; that doesn't count for `result_maybenull_` right? Then there's also `_Outptr_result_nullonfailure_` but that doesn't imply `COM` objects.
> 
> PS: The title of this PR doesn't match the commit that was merged, making it harder to find and identify in my list of tabs. Can you update it to reflect that?

Thanks for looking into this!

kennykerr:
As I've steadily improved the accuracy of parsing I discovered that there are some functions that looked like the QueryInterface pattern but were not in fact returning interfaces at all. That's why I had to tighten up the heuristic for signature detection. 

> While this function is similar to CoCreateInstance, there is no COM involvement.

The DirectX APIs don't rely on combase for activation and don't rely on COM apartments at all. That is generally what is meant by such comments. DirectX still relies on COM in the sense that COM interfaces are used and all the normal rules around IUnknown queries and lifetimes are still enforced. So the _COM_Outptr_ annotation is still appropriate. 

I think it would make sense to update the DX headers and also update the win32metadata project in the short term with some kind of fixup so that this works today.

> Secondly, where can I find some documentation on using _COM_Outptr_ vs _COM_Outptr_result_maybenull_? 

I don't know about documentation but DirectX does rely on the latter quite a lot for feature detection e.g. would some expensive object creation succeed with these flags? That's still something I need to support in Rust but may end up dropping the QueryInterface-like support in such cases. #737

I've updated the title to match the resulting commit.

MarijnS95:
> As I've steadily improved the accuracy of parsing I discovered that there are some functions that looked like the QueryInterface pattern but were not in fact returning interfaces at all. That's why I had to tighten up the heuristic for signature detection.

Yeah, that's great :slightly_smiling_face: 

> > While this function is similar to CoCreateInstance, there is no COM involvement.
> 
> The DirectX APIs don't rely on combase for activation and don't rely on COM apartments at all. That is generally what is meant by such comments. DirectX still relies on COM in the sense that COM interfaces are used and all the normal rules around IUnknown queries and lifetimes are still enforced. So the _COM_Outptr_ annotation is still appropriate.

Thanks for confirming my thoughts!

> I think it would make sense to update the DX headers and also update the win32metadata project in the short term with some kind of fixup so that this works today.

Great, I can take care of both over the course of some time; will need to check how many (public-facing) functions are affected.

> > Secondly, where can I find some documentation on using _COM_Outptr_ vs _COM_Outptr_result_maybenull_?
> 
> I don't know about documentation but DirectX does rely on the latter quite a lot for feature detection e.g. would some expensive object creation succeed with these flags? That's still something I need to support in Rust but may end up dropping the QueryInterface-like support in such cases. #737

At least for the most important `DxcCreateInstance` there are no flags. I expect to either receive an error (ie. `REGDB_E_CLASSNOTREG`), or `S_OK` and a valid pointer. If that is the case `result_maybenull` should not be used, right?
For the other cases I'll take this up with the DXC maintainers; they should be able to point me in the right direction if I fail to analyze the functions properly, as long as I've got the conventions right.

Issue #453, opened 2021-01-27
Support COM-like interfaces that don't inherit from IUnknown


kennykerr:
Need a way to deal with COM-like interfaces such as `ID3D11FunctionParameterReflection` that don't inherit from `IUnknown`.

mattrudder:
I ran into this recently when trying to bind to the XAudio2 APIs. Specifically, the `IXAudio2Voice` interfaces and their callback interfaces don't inherit from `IUnknown` either. Maybe useful as another test case for this issue.

kennykerr:
Just to clarify that this issue is about interfaces that look like COM interfaces but don't inherit from `IUnknown` and are thus not COM interfaces at all. They're basically just a vtable without the grounding that `IUnknown` provides for lifetime and polymorphism. They're somewhat rare and mostly seem to be used for callbacks, basically a collection of callback functions. 

Ideally you'd be able to implement such an interface with the `implement` macro as you can actual COM interfaces, but the `implement` macro will need some work to know that such interfaces don't need all the COM plumbing and that something outside of COM will control the interface's lifetime. 

Such interface declarations in the windows crate also need to not have a `Drop` implementation since that currently calls `Release` on `IUnknown` and such types should not implement the `Interface` trait, which requires an `IUnknown` vtable.

wesleywiser:
I also ran into this with the [`IDebugFAEntryTags`](https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/extsfns/nn-extsfns-idebugfaentrytags) interface which doesn't derive from `IUnknown`. 

In my case, I just need to use an instance of the interface which I'm given as a parameter, not actually implement it myself. I *think* wrapping the type in `ManuallyDrop` would prevent windows-rs/Rust from trying to call `IUnknown::Release` from `Drop::drop` ie:

```rust
fn FuncFoo(tags: ManuallyDrop<IDebugFAEntryTags>) { ... }
```

Am I way off base there?

kennykerr:
Yes, the `ManuallyDrop` would suppress the `Release` call but it wouldn't prevent the vtable from including, or expecting, the three `IUnknown` members. So if you're using the `interface` macro to define the interface you'd still end up with the incorrect function pointer offsets. I'll try to find some time this week to get this working.

PR #1096, opened 2021-08-27
Improved signature detection

kennykerr:
Consolidates signature detection across functions/interfaces and improves support for more edge cases around unorthodox signatures. Also adds support for functions with the `AlternateSuccessCodes` attribute by preserving the original function signatures in such cases. 

This PR also brings us closer toward implementing unconventional COM interfaces like `ID3D12FunctionReflection` but still requires #453.

### DIFF
--- a/tests/free_function_return_struct/tests/test.rs
+++ b/tests/free_function_return_struct/tests/test.rs
@@ -1,10 +1,13 @@
-use test_free_function_return_struct::*;
-use Windows::Win32::Graphics::Direct2D::*;
-
 // This test validates that free functions are not affected by the Visual C++ quirk affecting member functions
 // returning structs where the return value is effectively a trailing out parameter.
+//
+// TODO: D2D1ConvertColorSpace is exported by ordinal rather than by name and Rust doesn't appear to support
+// this at the moment. It happens to work on x64 but not on x64, hence the cfg check below.
 #[test]
+#[cfg(target_pointer_width = "64")]
 fn test() {
+    use test_free_function_return_struct::Windows::Win32::Graphics::Direct2D::*;
+
     let before = D2D1_COLOR_F {
         r: 1.0,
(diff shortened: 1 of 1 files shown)


tests/alternate_success_code/tests/test.rs
```diff
@@ -0,0 +1,12 @@
...
+use Windows::Win32::Foundation::CO_E_NOTINITIALIZED;
+use Windows::Win32::System::Com::DoDragDrop;
+
+// This test validates that the AlternateSuccessCodes attribute is being honored by preserving the original signature
+// and not doing any transformation.
+#[test]
+fn test() {
+    let mut effect = 0;
+    let hr = unsafe { DoDragDrop(None, None, 1, &mut effect) };
+    assert!(hr == CO_E_NOTINITIALIZED);
```
tim-weis, line 11: This is likely to cause the test to spuriously fail. While true in isolation, the condition is destined to fall victim to the [curse of the implicit MTA](https://devblogs.microsoft.com/oldnewthing/20130419-00/?p=4613). With tests getting executed in parallel running out of luck is a mere matter of time.

kennykerr: I was under the impression that each test crate is executed in its own process.

tim-weis: That's correct, and at this time there is no opportunity for this test to fail. I was rather concerned about future additions to this test crate with the potential to break unrelated test code.

There's no immediate need to change this test, other than—maybe—leaving a comment stating the implicit assumptions about the test conditional.

kennykerr: The tests are certainly a bit brittle. That's why I am in the process of writing a standalone test component in C++, much like I have for cppwinrt. That should address this concern.

PR #1098, opened 2021-08-28
Fix infinite loop while updating weak reference counter

### LINKED ISSUE #1103: Updating weak reference counter simultaneously leads to infinite loops
The current WeakRefCount implementation for authored WinRT types leads to infinite loops when there are concurrent AddRef requests to the object.

Here is the code snippet used to reproduce this problem. Since `weak_ref_count` is not made public, the test code should be put inside `src/runtime/weak_ref_count.rs`.
```rust
#[cfg(test)]
mod tests {
    use super::*;
    use std::sync::{atomic::AtomicU32, Arc};
    #[test]
    fn test_race_count() {
        const TARGET: u32 = 1_000_000;
        const CONCURRENCY: usize = 16;
        const ATOMIC_U32_ZERO: AtomicU32 = AtomicU32::new(0);
        static PROGRESS: [AtomicU32; CONCURRENCY] = [ATOMIC_U32_ZERO; CONCURRENCY];
        fn run_increment(ref_count: Arc<WeakRefCount>, progress: &AtomicU32) {
            for _ in 0..TARGET {
                ref_count.add_ref();
                progress.fetch_add(1, Ordering::Relaxed);
            }
        }

        let ref_count = Arc::new(WeakRefCount::new());
        let mut threads = Vec::with_capacity(CONCURRENCY);
        for i in 0..CONCURRENCY {
            let ref_count = ref_count.clone();
            threads.push(std::thread::spawn(move || {
                run_increment(ref_count, &PROGRESS[i])
            }));
        }
        let mut last_progress = [0; CONCURRENCY];

        loop {
            std::thread::sleep(std::time::Duration::from_millis(1000));
            let mut new_progress = [0; CONCURRENCY];
            for i in 0..CONCURRENCY {
                new_progress[i] = PROGRESS[i].load(Ordering::Relaxed);
                eprint!("Progress {} = {}, ", i, new_progress[i]);
            }
            eprintln!();

            if new_progress.iter().all(|&p| p == TARGET) {
                break;
            }

            if new_progress
                .iter()
                .zip(last_progress.iter())
                .any(|(&new, &old)| new == old && new != TARGET)
            {
                panic!("Progress did not increase during the last second");
            }
            last_progress.copy_from_slice(&new_progress[..]);
        }

        threads.into_iter().for_each(|t| t.join().unwrap());
    }
}
```

Then, the following commands are used to run the test.

```powershell
# Debug
cargo test --package windows --lib -- runtime::weak_ref_count::tests::test_race_count --exact --nocapture
# Release
cargo test --release --package windows --lib -- runtime::weak_ref_count::tests::test_race_count --exact --nocapture
```

On my laptop with CPU `Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz`, the test fails in both Debug mode and Release mode.

```powershell
PS D:\proj\windows-rs> cargo test --package windows --lib -- runtime::weak_ref_count::tests::test_race_count --exact --nocapture
   Compiling windows v0.19.0 (D:\proj\windows-rs)
    Finished test [unoptimized + debuginfo] target(s) in 1.62s
     Running unittests (target\debug\deps\windows-2960b576f553a6e5.exe)

running 1 test
Progress 0 = 4390, Progress 1 = 0, Progress 2 = 0, Progress 3 = 0, Progress 4 = 467, Progress 5 = 0, Progress 6 = 0, Progress 7 = 83093, Progress 8 = 0, Progress 9 = 15676, Progress 10 = 109, Progress 11 = 298, Progress 12 = 0, Progress 13 = 1000000, Progress 14 = 0, Progress 15 = 96228, 
thread 'runtime::weak_ref_count::tests::test_race_count' panicked at 'Progress did not increase during the last second', src\runtime\weak_ref_count.rs:367:17
test runtime::weak_ref_count::tests::test_race_count ... FAILED

failures:

failures:
    runtime::weak_ref_count::tests::test_race_count

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 12 filtered out; finished in 22.18s

error: test failed, to rerun pass '-p windows --lib'
PS D:\proj\windows-rs> cargo test --release --package windows --lib -- runtime::weak_ref_count::tests::test_race_count --exact --nocapture
   Compiling windows v0.19.0 (D:\proj\windows-rs)
    Finished release [optimized] target(s) in 1.95s
     Running unittests (target\release\deps\windows-eec72b0d7ce64b11.exe)

running 1 test
Progress 0 = 10001, Progress 1 = 0, Progress 2 = 0, Progress 3 = 0, Progress 4 = 379, Progress 5 = 424, Progress 6 = 0, Progress 7 = 2043, Progress 8 = 0, Progress 9 = 121534, Progress 10 = 53, Progress 11 = 309, Progress 12 = 0, Progress 13 = 3, Progress 14 = 1000000, Progress 15 = 0, 
thread 'runtime::weak_ref_count::tests::test_race_count' panicked at 'Progress did not increase during the last second', src\runtime\weak_ref_count.rs:367:17
test runtime::weak_ref_count::tests::test_race_count ... FAILED

failures:

failures:
    runtime::weak_ref_count::tests::test_race_count

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 12 filtered out; finished in 16.04s

error: test failed, to rerun pass '-p windows --lib'
```

Also see #1098 that may fix this issue.

bdbai:
The current `WeakRefCount` implementation for authored WinRT types leads to infinite loops when there are concurrent `AddRef` requests to the object. The issue is caused by a subtle difference of atomic operation APIs between C++ and Rust. In C++ STL, the [`std::atomic<T>::compare_exchange_weak`](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange) function treats the `expected` argument as a reference, and updates it with the actual value. While in Rust, the [`std::sync::atomic::AtomicIsize::compare_exchange_weak`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicIsize.html#method.compare_exchange_weak) function does not change any inputs, and the actual value appears in the `Err` variant of the return value. This means old values should be replaced manually when the operation fails.

This pull request attempts to fix this issue by replacing most CAS operations inside loops with [`fetch_update`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicIsize.html#method.fetch_update), which is a shorthand for loops with CAS. For `query` function, a manual update is used.

Fixes #1103


kennykerr:
This code is based on what I wrote for C++/WinRT and is also some of the most challenging code to reason about. The C++/WinRT code is hardened by years of testing. I would not be surprised if there was a bug in the Rust port, but we need to be very careful in touching this code. Please start with an issue that includes a minimal repro that clearly demonstrates the problem and then once its been established that the Rust implementation deviates from C++/WinRT then we can look at a very targeted fix to bring it in line. Sorry for being overly cautious. I have no doubt my Rust implementation could be improved upon, but I just need to make sure the code remains consistent with C++/WinRT. The reference implementation can be found here:

https://github.com/microsoft/cppwinrt/blob/master/strings/base_implements.h

bdbai:
I agree that code involving concurrency can be hard to make it right, so it is necessary to be careful about it.

#1103 includes a test that reproduces the problem if it helps. Given the implementation in C++/WinRT does not have this problem, is it necessary to create an equivalent test against C++/WinRT?

kennykerr:
No need, this is great - thanks for the explanations. I'll take a close look now.

src/runtime/weak_ref_count.rs
```diff
@@ -90,30 +60,29 @@ impl WeakRefCount {
...
-                    Ordering::Relaxed,
-                )
-                .is_ok()
-            {
-                let result: RawPtr = std::mem::transmute(tear_off);
-                TearOff::from_strong_ptr(result).strong_count.add_ref();
-                return result;
+            let tear_off = TearOff::new(object, count_or_pointer as _);
+            let encoding: usize =
+                ((tear_off.abi() as usize) >> 1) | (1 << (std::mem::size_of::<usize>() * 8 - 1));
```
kennykerr, line 72: Is it necessary to move this inside the loop? This causes the tearoff to be created repeatedly on retry.

kennykerr:
> While in Rust, the `std::sync::atomic::AtomicIsize::compare_exchange_weak` function does not change any inputs, and the actual value appears in the `Err` variant of the return value. This means old values should be replaced manually when the operation fails.

I'm a little confused by this because the following test appears to show that the value is updated with `compare_exchange_weak`:

https://github.com/microsoft/windows-rs/blob/master/tests/winrt/weak/tests/count.rs

Granted there is no concurrency in this test, but I'm now sure how this differs from the C++ `compare_exchange_weak` that this code is based on in C++/WinRT.

kennykerr:
Perhaps we should add the test repro in #1103 to `tests/winrt/weak/tests/count.rs` to verify this change.

src/runtime/weak_ref_count.rs
bdbai: `tear_off` should only be created once. Thanks for pointing that out!

bdbai:
> Perhaps we should add the test repro in #1103 to tests/winrt/weak/tests/count.rs to verify this change.

I have added the repro there in a new test file since it is not easy to integrated it into any existing test. What do you think?

kennykerr:
Thanks @bdbai, a new test file is great.

Note the test failed: https://github.com/microsoft/windows-rs/pull/1098/checks?check_run_id=3479981153

bdbai:
Looks like the CI machine is not able to handle such concurrency. Shall we reduce the loop count and concurrency to, say 100_000 and 8 respectively and see?

kennykerr:
Sure, we'll also have to confirm that the reduced concurrency reliably repros without your fix. We could add just the test to another PR to validate.

kennykerr:
Rerunning the checks to make sure its consistent. 

Also added #1108 to validate that the test fails.

kennykerr (approved):
Looks good - thanks!

PR #1109, opened 2021-09-01
Avoid `Param` abuse

### LINKED ISSUE #1028: The Constructor Pattern for Windows API structures
Hi everyone. 👋
I have a question: 
Are there any plans to extend the generator to create сonstructor methods for structures?
For example, there is a structure WNDCLASSW.
Instead of writing as in the [example](https://github.com/microsoft/windows-samples-rs/blob/master/create_window/src/main.rs#L15):
```rust
let wc = WNDCLASSA {
            hCursor: LoadCursorW(None, IDC_ARROW),
            hInstance: instance,
            lpszClassName: PSTR(b"window\0".as_ptr() as _),
            style: CS_HREDRAW | CS_VREDRAW,
            lpfnWndProc: Some(wndproc),
            ..Default::default()
        };
```

Make it so that you can write something like this:

```rust
let wc = WNDCLASSW::default()
        .hCursor(LoadCursorW(None, IDC_ARROW))
        .lpfnWndProc(wndproc)
        .lpszClassName(window)
        .hInstance(instance)
        .style(CS_HREDRAW | CS_VREDRAW);
```

kennykerr:
Fixes #1028

### DIFF
--- a/src/runtime/param.rs
+++ b/src/runtime/param.rs
@@ -11,14 +11,13 @@ pub enum Param<'a, T: Abi> {
 }
 
 impl<'a, T: Abi> Param<'a, T> {
-    pub fn abi(&mut self) -> T::Abi {
+    /// # Safety
+    pub unsafe fn abi(&self) -> T::Abi {
         match self {
             Param::Borrowed(value) => value.abi(),
             Param::Owned(value) => value.abi(),
             Param::Boxed(value) => value.abi(),
-            // It is always safe to form an `Abi` type's binary representation from an all-zero
-            // byte-pattern as this represents the null or default state for every type.
-            Param::None => unsafe { std::mem::zeroed() },
+            Param::None => std::mem::zeroed(),
         }
     }
 }
--- a/tests/legacy/pwstr/tests/pwstr.rs
+++ b/tests/legacy/pwstr/tests/pwstr.rs
@@ -1,27 +1,29 @@
 use std::ffi::{OsStr, OsString};
 use test_pwstr::Windows::Win32::Foundation::PWSTR;
-use windows::IntoParam;
+use windows::*;
 
 #[test]
 fn test() {
-    let expected = "hello\0";
-    let len = expected.chars().count();
+    unsafe {
+        let expected = "hello\0";
+        let len = expected.chars().count();
 
-    let p = get_pwstr_from("hello");
-    assert_eq!(from_parts(p.0, len), expected);
+        let p = get_pwstr_from("hello");
+        assert_eq!(from_parts(p.abi().0, len), expected);
 
-    let p = get_pwstr_from(String::from("hello"));
-    assert_eq!(from_parts(p.0, len), expected);
+        let p = get_pwstr_from(String::from("hello"));
+        assert_eq!(from_parts(p.abi().0, len), expected);
 
-    let p = get_pwstr_from(OsStr::new("hello"));
-    assert_eq!(from_parts(p.0, len), expected);
+        let p = get_pwstr_from(OsStr::new("hello"));
+        assert_eq!(from_parts(p.abi().0, len), expected);
 
-    let p = get_pwstr_from(OsString::from("hello"));
-    assert_eq!(from_parts(p.0, len), expected);
+        let p = get_pwstr_from(OsString::from("hello"));
+        assert_eq!(from_parts(p.abi().0, len), expected);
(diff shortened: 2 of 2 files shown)


MarijnS95:
@kennykerr Are you sure this fixes things properly? Wasn't/isn't the problem that `IntoParam` has a lifetime generic while it generally "owns" something? Supposedly `into_param()` fully consumes `self`.

kennykerr:
The lifetime is only used (by `Param`) when the input is borrowed. It is ignored in other cases. The lifetime itself is fine, in my understanding. It is the fact that you can call `abi` and use that for something other than a function call that is the problem.

MarijnS95:
Unfortunately I haven't read enough into this code, but it feels like the lifetime should be added by the various type implementations. Calling `into_param()` on `T` should probably give you back a `Param` with unbounded lifetime. Calling `into_param()` on `&'a T` should obviously use `'a`. Afaik none of those need this generic to be part of `IntoParam` (indicated by the `'_` in `impl<'a, T: Abi> IntoParam<'_, T> for &'a T `, but not for the `&'a Option` btw!), which seems to make things confusing here.

In that sense it looks like `Param` replicates functionality from `Cow`.

> It is the fact that you can call `abi` and use that for something other than a function call that is the problem.

Yes, but unfortunately this is not considered `unsafe` as far as I'm aware. Coercing to a raw pointer isn't unsafe, dereferencing is. Since we can't know what happens after a native boundary, calling `extern` functions is unsafe for at least that reason.

MarijnS95:
One thing I had thought of in my own programs is the free coercion from `&` to `*const` (same for `mut`), that way `.abi()` simply returns a borrow (with the right lifetime) and can be passed as-is to a function (or `struct` constructor, which is why this won't solve the problem) without casts. Iirc you can even use `&` types in FFI (`extern`) signatures and things compile fine.

kennykerr:
Yes, `abi` isn't unsafe in the Rust sense but it is easily abused. Ideally, the whole `IntoParam`/`Param` thing was not public but that isn't practical. If there's a better way to express that I'm happy to hear it.

kennykerr:
> One thing I had thought of in my own programs is the free coercion from & to *const (same for mut), that way .abi() simply returns a borrow (with the right lifetime) and can be passed as-is to a function (or struct constructor, which is why this won't solve the problem) without casts. Iirc you can even use & types in FFI (extern) signatures and things compile fine.

The challenge is in cases where (1) the parameter can be null or (2) the type is not blittable. I've also been trying to move away from having such a disparity between the ABI and the projected type. I've come a lot closer with structs where there's no longer a distinct ABI type, but I have a ways to go.

MarijnS95:
> If there's a better way to express that I'm happy to hear it.

I'd love to give this a spin but it seems there are quite a few occurrences to change.

> The challenge is in cases where (1) the parameter can be null or (2) the type is not blittable. I've also been trying to move away from having such a disparity between the ABI and the projected type. I've come a lot closer with structs where there's no longer a distinct ABI type, but I have a ways to go.

Yes, those sound tricky to deal with too, but it's hard to say without fully familiarize with the code. Seems like this bit of windows-rs should just evolve over time, maybe I'll give aforementioned changes a shot some day.

src/runtime/param.rs
```diff
@@ -11,14 +11,13 @@ pub enum Param<'a, T: Abi> {
...
 
 impl<'a, T: Abi> Param<'a, T> {
-    pub fn abi(&mut self) -> T::Abi {
+    /// # Safety
+    pub unsafe fn abi(&self) -> T::Abi {
         match self {
             Param::Borrowed(value) => value.abi(),
             Param::Owned(value) => value.abi(),
             Param::Boxed(value) => value.abi(),
-            // It is always safe to form an `Abi` type's binary representation from an all-zero
```
rylev, line 19: This comment is pretty useful to know... Not sure why it was removed.

rylev:
I think this has been mentioned before, but this isn't really how the Rust std library (and guidelines such as the nomicon) use `unsafe`. Calling the `abi` method is not unsafe just like calling [`Vec::as_ptr`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.as_ptr) is not unsafe. You can easily do the wrong thing with the pointer returned from `Vec::as_ptr` but that would require dereferencing the pointer which itself would require the use of `unsafe`.

Issue #212, opened 2020-06-08
Support WinRT delegates with array parameters


kennykerr:
This has not yet been implemented, but requires a test component (written in C++/WinRT) in order to test since no such API exists in Windows.

https://github.com/microsoft/winrt-rs/blob/0af977075d21f23cd38fc9df6ac04cf92c4cd46c/crates/gen/src/types/param.rs#L157

PR #1116, opened 2021-09-07
Remove unused_variables

kennykerr:
Now that #212 has been fixed, suppressing this warning is no longer necessary.


kennykerr:
Hmm.. turns out we still need this for non-Windows builds.

PR #1120, opened 2021-09-09
Add Win32 support to test component

riverar:
Currently limited to x64 builds due to limitations of the upstream Windows Metadata generator.


kennykerr (approved):
Thanks!

PR #1128, opened 2021-09-13
Lower case lib file names

### LINKED ISSUE #638: Linux -> Windows Cross Compilation broken

kennykerr:
Hopefully partially fixes #638

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,9 +19,6 @@ jobs:
           rust: stable
           other: x86_64-pc-windows-msvc
         - os: windows-latest
-          rust: 1.54.0
-          other: x86_64-pc-windows-msvc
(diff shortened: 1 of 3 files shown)


riverar (approved):

riverar:
@kennykerr Wonder if we can request this be done at the metadata level to trim off the `to_lowercase` here

kennykerr:
@riverar that would be great.

PR #1135, opened 2021-09-14
Add basic WinRT collection and async tests

kennykerr:
And finally remove the test component that was initially used to bootstrap Rust/WinRT.


kennykerr:
@wravery I now have a dedicated test component for the Windows crate. This greatly simplifies systematic testing as I can write discrete tests for each scenario, signature, etc. and not have to rely on arbitrary APIs for test coverage. 

Any chance you can convert the custom webview2 static lib test and get it covered by the new test component so we can get rid of the webview2 dependencies?

@riverar I think you have it set up to publish libs - not sure whether those are static import libs.

wravery:
> @wravery I now have a dedicated test component for the Windows crate. This greatly simplifies systematic testing as I can write discrete tests for each scenario, signature, etc. and not have to rely on arbitrary APIs for test coverage.
> 
> Any chance you can convert the custom webview2 static lib test and get it covered by the new test component so we can get rid of the webview2 dependencies?

Sure, but I think I'd need a static lib version of it, and the winmd would need to have the `[StaticLibrary]` attribute. I'm not sure if that will work with the existing component, whether it can build as a static library or if the IDL-based winmd can include this attribute.

wravery:
BTW, I do have a new [home](https://github.com/wravery/webview2-win32md) for the winmd in that test, but I agree, using a dedicated test library and generating a simple winmd for it would make sense here.

riverar:
> @riverar I think you have it set up to publish libs - not sure whether those are static import libs.

Currently generating dynamic import libs.

I'm a little fuzzy on the need for static lib _output_ for the test component. I think the suggested workflow here is:
1. consume WebView2 static lib in the test component
2. `__declspec(dllexport)` (or equiv.) all WebView2 APIs you wish to test
3. should automatically get picked up in the resultant winmd, ready for test

Am I missing something?

kennykerr:
I was thinking the test component can just produce a static lib in addition to the DLL so that a consumer can either static or dynamic link an exported function like the `SimpleFunction` function. We'd then need to figure out how to produce the .winmd file with the correct attribute to indicate the the particular test function expects static linking.

riverar:
Yep, I think we could quickly add support for static test component output/testing in general, but I'm not quite following how this helps WebView2 testing here yet, unless the idea is that we _enable_ this and those specific tests are run outside this repo.

wravery:
The component this test would need should be a static lib itself. There should not be any DLL imports. It also needs a winmd with the StaticLibrary attribute to direct it to perform static linking instead of dynamic linking through the import lib.

If the test component builds a static lib and a DLL with import lib, the same component could be used to test both. But you'd probably need to build it as a win32 lib with win32metadata to do that, the winmd output from MIDL3/WinRT won't include that attribute.

wravery:
> Yep, I think we could quickly add support for static test component output/testing in general, but I'm not quite following how this helps WebView2 testing here yet, unless the idea is that we _enable_ this and those specific tests are run outside this repo.

Ah, I see. The test in this case isn't for WebView2 specifically, it's testing how `windows-rs` handles the `StaticLibrary` attribute. The WebView2 winmd and static library just happen to be the first component with the `StaticLibrary` attribute in the winmd pointing to a static lib that works as a substitute for the DLL. We could test that `windows-rs` feature with the synthetic test component instead if it had a static lib and generated a winmd with the `StaticLibrary` attribute.

PR #1139, opened 2021-09-15
Simplify code gen for Win32 handle types

### LINKED ISSUE #1131: Convenient getting error when returned zero value (as `0` `BOOL`)
Something like this, but for all nullable types:

```rust
use Windows::Win32::Foundation::HINSTANCE;

windows::include_bindings!();

pub trait NullableToResult {
    #[inline]
    fn is_null(&self) -> bool;
    #[inline]
    fn ok(self) -> ::windows::Result<Self> where Self: Sized {
        if !self.is_null() {
            Ok(self)
        } else {
            Err(::windows::HRESULT::from_thread().into())
        }
    }
}

// HINSTANCE as example:
impl NullableToResult for HINSTANCE {
    #[inline]
    fn is_null(&self) -> bool {
        self.is_null()
    }
}
```

### LINKED ISSUE #1138: Declare the `is_null` methods into a separate trait
I suggesting declare the `is_null` methods into a separate trait (`Nullable` or a similar name) and implementing it for all nullable types. It can be useful.

kennykerr:
Also fixes #1131 and fixes #1138

cc @Zymlex


crates/gen/src/handle.rs
```diff
@@ -2,11 +2,32 @@ use super::*;
...
+        unsafe impl ::windows::Handle for HANDLE {
+            fn is_null(&self) -> bool {
+                *self == unsafe { ::std::mem::zeroed() }
+            }
+
+            fn ok(self) -> ::windows::Result<Self> {
+                if self != Self::NULL && self != Self::INVALID {
+                    Ok(self)
+                } else {
+                    Err(::windows::HRESULT::from_thread().into())
```
riverar, line 24: I think this will return spurious HRESULTs from APIs that don't `SetLastError`, right?

kennykerr: Yes, the caller is responsible for using it as appropriate. That's why it's a method call and not simply woven into the projection.

riverar: The part that bugs me a bit is that checking for null would be common operation but its result on error would probably be _wrong_ most of the time.

Wonder if we should just make this return a bool and the caller can explicitly retrieve the hresult as needed elsewhere?

Just a nit.

kennykerr: You can use is_null for that. I expect that to be more common. The ok method is consistent with methods of the same name across types like HRESULT and BOOL that all return Result.

PR #1144, opened 2021-09-16
Correctly import generic dependencies 

### LINKED ISSUE #772: `IMap` depends on `IKeyValuePair` but build macro doesn't detect indirect generic dependency
Repro:

```rust
fn main() {
    windows::build!(Windows::Foundation::Collections::IMap);
}
```

kennykerr:
Fixes #772 

@riverar can you confirm that the WinUI example you mentioned has also been fixed - you should *not* need to import `IObservableVector`.


riverar:
@kennykerr Will do.

PR #1170, opened 2021-09-22
Support query-optional signatures

### LINKED ISSUE #737: Cannot call QueryInterface-like APIs with "NULL" for the ppv parameter

kennykerr:
Distinguishes between query-like signatures - with trailing `GUID*, void**` - and those that are marked optional in metadata (mostly Direct3D). 

This necessitates a move back to explicit generic type parameters as Rust doesn't allow you to mix `impl Trait` parameters with explicit generic call-sites. 

Fixes #737 cc @damyanp 

For example, you can call `D3D12CreateDevice` as follows:

```rust
D3D12CreateDevice::<_, ID3D12Device>(adapter, D3D_FEATURE_LEVEL_11_0, std::ptr::null_mut())?;
```

Or like this:

```rust
let mut device: Option<ID3D12Device> = None;
D3D12CreateDevice(adapter, D3D_FEATURE_LEVEL_11_0, &mut device)?;
```


MarijnS95:
@kennykerr Note that there's also:

```rust
D3D12CreateDevice(adapter, D3D_FEATURE_LEVEL_11_0, std::ptr::null_mut::<Option<ID3D12Device>>())?;
```

But that's probably uncomfortable enough to "necessitate" explicit type parameters at the call-site instead?

kennykerr:
Yes, I considered that but it just seemed so contrived.

MarijnS95:
The current method at least leaves room to either use a turbofish on the function call, or the `null_mut` allowing to get rid of the `_` substitution for the generic lifetime.

kennykerr:
The `_` is for the first parameter which expects an `impl Trait` - I had to switch from `impl Trait` code gen to explicit type parameters sadly.

kennykerr:
Look at the diff for `bindings.rs` in this PR.

MarijnS95:
Right it reads `_` not `'_`.

PR #1172, opened 2021-09-22
Remove unnecessary target checks in tests

kennykerr:
While the win32 winmd generator doesn't support other arches, the Rust tests still do.


riverar (approved):
Am okay with this. The intent of the gate was to prevent usage of arch-specific structures in Win32 tests. (Structure padding differs between x86/x64 so I believe all structures and related ABI testing would be affected.) The winmdgenerator doesn't support generating the necessary structure metadata at this time (https://github.com/microsoft/win32metadata/issues/626) but does appear to be near per Steve's comment this morning.

kennykerr:
Sounds good - I haven't go any such tests yet and don't expect to get there before the support arrives.

PR #1173, opened 2021-09-22
Add test for functions that return values directly

kennykerr:
This might seem pointless, but the Visual C++ compiler has a quirk when it comes to virtual functions that return structs where the return value is actually packed onto the stack as if it were a trailing out parameter. As a result, other languages must emulate this when interacting with COM interfaces. This behavior is also architecture-specific since stack packing is different between x86 and x64.


kennykerr:
Looks like this is failing due to a Rust compiler bug. https://github.com/kennykerr/link-rs

PR #1176, opened 2021-09-23
Add Win32 static lib tests

riverar:
Changes to Windows Metadata generation and the test component to support static lib scenarios.


tests/static_component/build.cmd
```diff
@@ -0,0 +1,8 @@
+copy "..\target\%2\%1\static_component.lib" ..\..\.windows\%3
+
+if "%3" == "x64" (
```
kennykerr, line 3: I thought Steve fixed the arch issue?

tests/component/.metadata/nuget.config
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+
+    <!--
+      Temporary private nuget feed to serve up winmdgenerator 0.3.2-preview.
```
kennykerr, line 8: I thought Steve fixed the nuget feed?

tests/static_component/win32/simple.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+extern "C"
+{
+    void __stdcall SimpleFunction() noexcept;
```
kennykerr, line 5: Just a guess, but I think the build failure is because Rust doesn't know how to handle multiple imports with the same name that have different link attributes (and thus come from different libs).

kennykerr: Although they're compiled as separate tests so that's probably not the issue.

tests/static_component/build.cmd
riverar: Thanks for the reminder, will pull down the new winmdgenerator and update the PR.

tests/component/.metadata/nuget.config
riverar: Will pull down the new winmdgenerator and update the PR.

tests/static_component/win32/simple.h
riverar: Failing on release-compiled lib. Works fine with a debug-compiled lib. `rustc` is touching the lib in all cases so something _interesting™️_ is going on here. Investigating.

kennykerr (approved):

kennykerr (approved):
Thanks!

kennykerr (approved):

Issue #589, opened 2021-03-11
HWND could implement the Hash trait


emmanuelantony2000:
`win32::windows_and_messaging::HWND` could implement the `Hash` trait.

tim-weis:
The Rust API Guidelines lists the most common traits ([C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits)) to implement. These are:

* `Copy`
* `Clone`
* `Eq`
* `PartialEq`
* `Ord`
* `PartialOrd`
* `Hash`
* `Debug`
* `Display`
* `Default`

I would suggest implementing all of those, particularly for new-type wrapper types, unless there is a striking reason not to.

kennykerr:
This is a little challenging because many structs cannot necessarily implement all of these traits due to language restrictions. Others require specialized code. For example, unions and packed structs pose unique challenges. For those where it is permitted, it can be costly to derive some of these. I'm in favor of providing as many of these as possible, just identifying some of the  reasons why this isn't as simple as it might seem.

kennykerr:
The `windows-sys` crate implements `Copy` and `Clone` (the minimal required) while the `windows` crate implements a few more as applicable. Let me know if additional are required, but I'd like to avoid adding implementations just because "it would be nice to have" as it hurts compile time for so many types.

PR #1190, opened 2021-09-29
Implement Hash for Guid

jrmuizel:


src/runtime/guid.rs
```diff
@@ -131,6 +132,15 @@ impl From<&str> for Guid {
     }
 }
 
+impl Hash for Guid {
```
kennykerr, line 135: Any reason not to simply derive this?

kennykerr:
Thanks for the contribution @jrmuizel - generally it helps to open an issue or at least include a PR comment describing the motivation for the change. 

This particular change seems mostly self-explanatory but there's still the issue of consistency since we don't generate `Hash` for any other Windows structs at present. I think `Hash` could be useful to apply more broadly but it may incur an unexpected compilation cost when applied to all Windows structs. I just haven't done the homework to confirm yet. 

Also #589 for reference.

src/runtime/guid.rs
jrmuizel: Nope, I forgot you could just do that. I've updated the PR.

jrmuizel:
This PR was motivated by https://github.com/n4r1b/ferrisetw/blob/b631bcdd01bc5da250980092d1be192b826e4c02/src/schema.rs#L34

As for the consistency issue, GUIDs are closer to a primitive type on Windows vs other structs. Examples of this would be the `uuid` attribute in C++ and the GUID input type here: https://docs.microsoft.com/en-us/windows/win32/wes/eventmanifestschema-inputtype-complextype

kennykerr (approved):
Thanks!

PR #1208, opened 2021-10-10
Support fast-pass strings

drewcrawford:
APIs such as `Windows::Win32::System::WinRT::WindowsCreateStringReference`
can allocate fast-pass strings as `windows::HSTRING` instances,
however `windows::HSTRING::Drop` is apparently unimplemented for this
case.

Have not tested this extensively but it appears to work.

### DIFF
--- a/src/runtime/hstring.rs
+++ b/src/runtime/hstring.rs
@@ -60,13 +60,13 @@ impl HSTRING {
         }
 
         unsafe {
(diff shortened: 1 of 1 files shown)


kennykerr:
WindowsCreateStringReference is only meant for language projections to use when calling the underlying ABI of a given API.

PR #1236, opened 2021-10-26
gen/win32: Put struct result-pointers in x64 ABI as first parameter

MarijnS95:
As per [1] the MSVC x64 ABI notes that larger returned datastructures are allocated by the caller and passed with a pointer.  This is already implemented in windows-rs, but the return-pointer was always passed as last argument and the tests never validated a case with more than zero parameters.  Now that those tests are extended to include extra parameters and segfault on the functions that return a struct, the `return__` pointer can be moved as first argument (after `this`) and is validated to resolve the issue without breaking other tests.

[1]: https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-160#return-values

---

A couple questions/points that we should process separate from this PR:

- The CI builds `components.sln` in one job, but doesn't reuse its results in the others.  While this makes sure the solution is buildable, it doens't enforce these most recent build result to actually pass the tests;
- `is_udt` still has a `TODO`, and I can't fully match it to the documentation. `SmallStruct` in this PR should be a "user defined type" - a POD even - but it's passed by pointer as first parameter.

Last but not least, thanks a lot for getting the public API done so quickly @kennykerr! This issue is the only thing holding me back from using `GetResourceAllocationInfo` in https://github.com/Traverse-Research/gpu-allocator/compare/windows-rs, other than that it's working perfectly fine!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,9 @@ members = [
 exclude = ["crates/tests/component"]
 
 [package.metadata.docs.rs]
+all-features = true
 default-target = "x86_64-pc-windows-msvc"
-targets = ["x86_64-pc-windows-msvc"]
+targets = []
 
 [dependencies]
 windows_macros = { path = "crates/deps/macros",  version = "0.22.0", optional = true }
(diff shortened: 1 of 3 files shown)


MarijnS95:
@kennykerr Did you find the time to look at https://github.com/MarijnS95/windows-rs/compare/build-artifacts? Shall I open a preliminary PR to further this approach?

PR #1264, opened 2021-10-30
Document feature requirements

### LINKED ISSUE #1246: Some functions do not enable with their corresponding feature
A lot of exports from the `Win32::Storage::CloudFilters` do not export with the `Win32_Storage_CloudFilters` feature enabled.
ex: `CfExecute`
```rs
#[cfg(all(
    feature = "Win32_Foundation",
    feature = "Win32_Storage_FileSystem",
    feature = "Win32_System_SystemServices"
))]
pub unsafe fn CfExecute(
```

kennykerr:
Fixes #1246 

@jyn514 It would be great if it could detect when a feature enables other features and only show the smallest set of features. For example, here `Win32_Storage_CloudFilters` enables `Win32_Storage` which enables `Win32` so it would ideally elide those for brevity and only list the features that must be enabled to actually enable the function. From the Windows crate's Cargo.toml:

```
Win32_Storage = ["Win32"]
Win32_Storage_CloudFilters = ["Win32_Storage"]
```

This is because it is modeling a module hierarchy. 

image

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@ Learn more here: <https://github.com/microsoft/windows-rs>
 */
 
 #![doc(html_no_source)]
+#![feature(doc_cfg)]
 
 extern crate self as windows;
 mod Windows;


kennykerr:
Ah, this is not yet stable so I may opt for a different approach for now.

jyn514:
> would ideally elide those for brevity and only list the features that must be enabled to actually enable the function

This would need major design changes, rustdoc doesn't have knowledge of cargo and doesn't know which features enable other features.

kennykerr:
Fair enough, then I may just generate a simple list with `#[doc = "..."]`

MarijnS95:
> Ah, this is not yet stable so I may opt for a different approach for now.

Fwiw other projects introduce something like a [`dox` feature](https://github.com/gtk-rs/gtk-rs-core/blob/772663b04c2d319ec2e574d9dd119b1222c9ecb3/glib/src/lib.rs#L3) or [`docsrs` cfg](https://github.com/MaikKlein/ash/blob/527b542c69c3d60fac0b6238d2ed9223454a5b1b/ash/src/lib.rs#L7) that is enabled when building for docs.rs ([1] / [2]), which uses nightly (and you can enable it manually locally and on the CI if buildtesting the docs with nightly).

[1]: https://github.com/gtk-rs/gtk-rs-core/blob/772663b04c2d319ec2e574d9dd119b1222c9ecb3/glib/Cargo.toml#L56-L57
[2]: https://github.com/MaikKlein/ash/blob/527b542c69c3d60fac0b6238d2ed9223454a5b1b/ash/Cargo.toml#L28

Secondly you can override the features shown here with `#[doc(cfg(features = "Win32_Storage_CloudFilters", features ...))]` but that requires some extra codegen on your end.

kennykerr:
Thanks, I ended up just generating it myself: https://github.com/microsoft/windows-rs/pull/1268

MarijnS95:
@kennykerr Cool - just keep `doc_cfg` in mind for when it stabilizes, as it's quite a bit nicer than having in-line documentation text (ie. nice, visible bubbles that also show in module-level documentation).

kennykerr:
I'm unlikely to use it, even if it were stable, as the output is not appealing. 

image

vs

image

I like how it "pops", but the wordiness and the redundant features makes it less practical. 

Obviously, if it improves over time I can reconsider.

MarijnS95:
@kennykerr I meant using `#[doc(cfg())]` (the override) to specify exactly _what_ features are shown.

MarijnS95:
@kennykerr So with that [trivial set of changes](https://github.com/MarijnS95/windows-rs/compare/doc-features), here's what the rendered docs look like:

image

And the functions:

image

Note that these bubbles don't repeat what is shown at the top of the page for the entire `CloudFilters` module. However, you may have already noticed that `Win32` made its return, and that also means it unfortunately remains shown on individual types and functions:

image

I think this is `rustdoc` recursively adding all the features it needed to get to this module in the first place, and that the `#[doc(cfg())]` override is only able to change the innermost config requirements set by `#[cfg()]`. The code generated for this functions is as follows:

```rust
#[cfg(all(feature = "Win32_Foundation", feature = "Win32_Storage_FileSystem", feature = "Win32_System_SystemServices"))]
#[cfg_attr(docsrs, doc(cfg(all(feature = "Win32_Storage_CloudFilters", feature = "Win32_Foundation", feature = "Win32_Storage_FileSystem", feature = "Win32_System_SystemServices"))))]
```

jyn514:
> I think this is rustdoc recursively adding all the features it needed to get to this module in the first place, and that the #[doc(cfg())] override is only able to change the innermost config requirements set by #[cfg()]. T

FWIW, I would be ok with changing the behavior of rustdoc here, I think it makes sense not to propagate the `cfg`s if you have a custom `doc(cfg)`. cc @nemo157

MarijnS95:
@jyn514 I think either is fine - I may have even hinted to @GuillaumeGomez to want that at some point while working on gtk/gstreamer/gir docs in the past - especially not showing feature requirements on functions when they are already specified on the surrounding type/impl/mod (but then we're getting in https://github.com/rust-lang/rust/issues/83428 territory).

It's probably more interesting to implement @kennykerr's suggestion above: don't show features that are already required for other (explicitly required/overridden) feature(s). I wasn't expecting rustdoc to not know anything about the cargo environment yet, I guess all crate dependencies coming from it are passed on the cmdline?

riverar:
@MarijnS95 The `and` conjunctions in the feature callout blocks copy/paste scenarios (and is hard to read). Is that customizable too? Just throwing that out there.

jyn514:
> The and conjunctions in the feature callout prevents devs from easily copy/pasting into Cargo.toml (and is hard to read).

It's not customizable (feel free to open an issue for that - https://github.com/rust-lang/rust/issues/43781#issuecomment-903298875), but I think it would be reasonable for rustdoc to make it a comma-separated list instead.

GuillaumeGomez:
This [issue](https://github.com/rust-lang/rust/issues/87957) too. We are talking about how to render the `cfg` information. If you have ideas, please share them. :)

MarijnS95:
Just wondering, does the `and` change to `or` when `any()` instead of `all()` is used? And are these parenthesized to disambiguate when nested combinations of ie. `any(feature = x, all(feature = y, feature = z))` are specified?

PR #1272, opened 2021-11-03
Reduce crate size

kennykerr:
* Avoids adding overridable methods to composable hierarchies, as they are meant to be "protected"
* Derives some common traits for classes and interfaces


yoshuawuyts (approved):
amazing

PR #1282, opened 2021-11-05
Add aarch64-pc-windows-msvc support

riverar:
CI changes have been held back as GitHub Actions has no available runners at this time. (The pipeline does succeed on self-hosted runners.)


kennykerr (approved):
Thanks!

PR #1300, opened 2021-11-10
Don't recurse infinitely when comparing `&HSTRING` and `&str`

### LINKED ISSUE #1286: Infinite recursion comparing &HSTRING with &str
```rust
dbg!(HSTRING::from("foo") == "foo");   // OK
dbg!(&HSTRING::from("foo") == "foo");  // Infinite recursion
```

```
 	xaml_testing_app.exe!core::cmp::impls::impl$9::eq<windows::runtime::hstring::HSTRING,str>(windows::runtime::hstring::HSTRING * * self, str * other) Line 1410	Unknown
>	xaml_testing_app.exe!windows::runtime::hstring::impl$16::eq(windows::runtime::hstring::HSTRING * self, str) Line 185	Unknown
 	xaml_testing_app.exe!core::cmp::impls::impl$9::eq<windows::runtime::hstring::HSTRING,str>(windows::runtime::hstring::HSTRING * * self, str * other) Line 1410	Unknown
 	xaml_testing_app.exe!windows::runtime::hstring::impl$16::eq(windows::runtime::hstring::HSTRING * self, str) Line 185	Unknown
 	xaml_testing_app.exe!core::cmp::impls::impl$9::eq<windows::runtime::hstring::HSTRING,str>(windows::runtime::hstring::HSTRING * * self, str * other) Line 1410	Unknown
 	xaml_testing_app.exe!windows::runtime::hstring::impl$16::eq(windows::runtime::hstring::HSTRING * self, str) Line 185	Unknown
 	xaml_testing_app.exe!core::cmp::impls::impl$9::eq<windows::runtime::hstring::HSTRING,str>(windows::runtime::hstring::HSTRING * * self, str * other) Line 1410	Unknown
 	xaml_testing_app.exe!windows::runtime::hstring::impl$16::eq(windows::runtime::hstring::HSTRING * self, str) Line 185	Unknown
 	xaml_testing_app.exe!core::cmp::impls::impl$9::eq<windows::runtime::hstring::HSTRING,str>(windows::runtime::hstring::HSTRING * * self, str * other) Line 1410	Unknown
 	xaml_testing_app.exe!windows::runtime::hstring::impl$16::eq(windows::runtime::hstring::HSTRING * self, str) Line 185	Unknown
 	xaml_testing_app.exe!core::cmp::impls::impl$9::eq<windows::runtime::hstring::HSTRING,str>(windows::runtime::hstring::HSTRING * * self, str * other) Line 1410	Unknown
 	xaml_testing_app.exe!windows::runtime::hstring::impl$16::eq(windows::runtime::hstring::HSTRING * self, str) Line 185	Unknown
 	xaml_testing_app.exe!core::cmp::impls::impl$9::eq<windows::runtime::hstring::HSTRING,str>(windows::runtime::hstring::HSTRING * * self, str * other) Line 1410	Unknown
...
```

image

mbartlett21:
Fixes #1286.


kennykerr:
Thanks for chipping in! Please add tests for these impls and then we can make sure they don't regress in future. Tests for `HSTRING` go here:

https://github.com/microsoft/windows-rs/blob/master/crates/tests/runtime/tests/hstrings.rs

mbartlett21:
@kennykerr 
I have added in the combinations with `str` and `String`.

kennykerr:
That's better - now the tests are actually running. 😉

mbartlett21:
It seems that less than half of the combinations were covered 😆

src/runtime/hstring.rs
```diff
@@ -171,32 +171,82 @@ impl From<&String> for HSTRING {
...
+    }
+}
+
+impl PartialEq<&HSTRING> for str {
+    fn eq(&self, other: &&HSTRING) -> bool {
+        **other == *self
+    }
+}
+
+impl PartialEq<HSTRING> for String {
```
kennykerr, line 235: Need to add `#[cfg(feature = "std")]` to these three impls.

mbartlett21: @kennykerr 
`String` is available from `alloc` so I have imported it from there. Is that ok?

kennykerr: Hmm.. right now `String` is excluded when `std` is not requested in a lot of other places. How about we just apply `#[cfg(feature = "std")]` here for now and then I can re-enable `String` support consistently throughout as a subsequent change. Even with your change you'd have to also update the `for String` to `for ::alloc::string::String` to compile successfully... If you run `cargo check` locally you should pick that up.

mbartlett21: @kennykerr 
I've re-added the `#[cfg(feature = "std")` annotations back.

kennykerr (approved):

Issue #1292, opened 2021-11-07
GUID helper function with u128?


saschanaz:
Since Rust has 128bit integer I think something like this would be less verbose than the current `from_values()`:

```rust
pub const fn get_guid_from_u128(uuid: u128) -> GUID {
    const fn get_byte(uuid: u128, index: u8) -> u8 {
        (uuid >> (8 * index) & 0xff) as u8
    }

    GUID::from_values(
        (uuid >> (8 * 12)) as u32,
        (uuid >> (8 * 10) & 0xffff) as u16,
        (uuid >> (8 * 8) & 0xffff) as u16,
        [
            get_byte(uuid, 7),
            get_byte(uuid, 6),
            get_byte(uuid, 5),
            get_byte(uuid, 4),
            get_byte(uuid, 3),
            get_byte(uuid, 2),
            get_byte(uuid, 1),
            get_byte(uuid, 0),
        ],
    )
}

get_guid_from_u128(0x01234567_89ab_cdef_0123_4567890abcde);
```

kennykerr:
I love this and even implemented it here:

https://github.com/microsoft/windows-rs/compare/from_u128?expand=1

Only trouble is that the const functions end up being slower to compile. For the [Direct2D sample](https://github.com/microsoft/windows-samples-rs/tree/master/direct2d) it takes the build from 29 sec to 34 sec. 

Anyone know if there are flags I can use to find out where exactly the Rust compiler is spending its time e.g. const functions, derive macros, impls, etc? I'd love to be able to adopt such changes and potentially offset them by paying down the compile time debt in other places.

saschanaz:
It seems this may help: https://blog.rust-lang.org/inside-rust/2020/02/25/intro-rustc-self-profile.html

kennykerr:
Thanks! That's very cool. Unfortunately, most of the time is measured in compiler units (e.g. codegen, link) rather than code units (e.g. functions, structs, impls).

```
+--------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| Item                                             | Self time | % of total time | Time     | Item count | Incremental result hashing time |
+--------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
| run_linker                                       | 381.23ms  | 27.912          | 381.23ms | 1          | 0.00ns                          |
+--------------------------------------------------+-----------+-----------------+----------+------------+---------------------------------+
[...]

PR #1306, opened 2021-11-10
Simpler `const` representation for `GUID`s

kennykerr:
As suggested in #1292.


saschanaz:
Curious, what was affecting the build time then? 👀

kennykerr:
I was comparing `path = "/git/windows-rs"` (my local clone) with `version = "0.26"` and for some reason it is always a few seconds slower using the former. Not sure why. So I compared a few different clones using different branches and confirmed this particular change didn't actually regress compile time and actually produces a smaller package.

PR #1309, opened 2021-11-11
`std` and `alloc` improvements

kennykerr:
* Compiling with `no_std` no longer removes all of the string conversion operations.
* The `alloc` feature was added to enable implicit parameter allocations for strings.

### DIFF
--- a/crates/deps/gen/src/bstr.rs
+++ b/crates/deps/gen/src/bstr.rs
@@ -48,40 +48,34 @@ pub fn gen_bstr() -> TokenStream {
                 Self::from_wide(self.as_wide())
             }
         }
-        // TODO: these str traits should work with no_std
-        #[cfg(feature = "std")]
         impl ::core::convert::From<&str> for BSTR {
             fn from(value: &str) -> Self {
-                let value: ::std::vec::Vec<u16> = value.encode_utf16().collect();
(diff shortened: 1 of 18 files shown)


crates/deps/gen/src/bstr.rs
```diff
@@ -48,40 +48,34 @@ pub fn gen_bstr() -> TokenStream {
                 Self::from_wide(self.as_wide())
             }
         }
-        // TODO: these str traits should work with no_std
-        #[cfg(feature = "std")]
         impl ::core::convert::From<&str> for BSTR {
             fn from(value: &str) -> Self {
-                let value: ::std::vec::Vec<u16> = value.encode_utf16().collect();
+                let value: ::windows::runtime::alloc::vec::Vec<u16> = value.encode_utf16().collect();
```
MarijnS95, line 53: @kennykerr Shouldn't this and other allocations for `BSTR` be wrapped in the `alloc` feature just like some of the other functions?

In other words, perhaps you should hide:
```rust
#[doc(hidden)]
pub extern crate alloc;
```
Behind the `alloc` feature too?

kennykerr: The `alloc` feature is to avoid implicit allocations like `IntoParam`. `From`/`Into` should still work regardless as they are explicit.

PR #1414, opened 2022-01-03
Guid hashmap

Xaeroxe:
Adds the `guid_hashmap` optional feature. It's really common for Windows APIs to return GUID constants. These constants can make debugging pretty hard as there isn't a single queryable source of truth on which constants map to which names. In C++ the accepted solution has been to [refer to a hard coded list of these GUIDs](https://github.com/microsoft/Windows-classic-samples/blob/27ffb0811ca761741502feaefdb591aebf592193/Samples/Win7Samples/multimedia/mediafoundation/common/logmediatype.h#L182). So this PR gives the `windows` crate the ability to generate a `HashMap` which maps the `GUID` values to a `&'static str` containing the name of the GUID in the Windows api.

### DIFF
--- a/crates/libs/windows/src/Windows/Win32/Data/HtmlHelp/mod.rs
+++ b/crates/libs/windows/src/Windows/Win32/Data/HtmlHelp/mod.rs
@@ -1,18 +1,46 @@
 #![allow(non_snake_case, non_camel_case_types, non_upper_case_globals, clashing_extern_declarations, clippy::all)]
 pub const CLSID_IITCmdInt: ::windows::core::GUID = ::windows::core::GUID::from_u128(0x4662daa2_d393_11d0_9a56_00c04fb68bf7);
+#[cfg(feature = "guid_hashmap")]
+inventory::submit! { crate :: core :: GuidConst { name : stringify ! ( CLSID_IITCmdInt ) , guid : :: windows :: core :: GUID::from_u128(0x4662daa2_d393_11d0_9a56_00c04fb68bf7) , } }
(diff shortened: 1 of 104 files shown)


crates/libs/windows/src/Windows/Win32/Data/HtmlHelp/mod.rs
```diff
@@ -1,18 +1,46 @@
 #![allow(non_snake_case, non_camel_case_types, non_upper_case_globals, clashing_extern_declarations, clippy::all)]
 pub const CLSID_IITCmdInt: ::windows::core::GUID = ::windows::core::GUID::from_u128(0x4662daa2_d393_11d0_9a56_00c04fb68bf7);
+#[cfg(feature = "guid_hashmap")]
+inventory::submit! { crate :: core :: GuidConst { name : stringify ! ( CLSID_IITCmdInt ) , guid : :: windows :: core :: GUID::from_u128(0x4662daa2_d393_11d0_9a56_00c04fb68bf7) , } }
```
Xaeroxe, line 4: `rustfmt` seems to be avoiding formatting macro input, which is unfortunate here. I could format it before it gets passed to `quote` to fix this if that'd be preferred.

kennykerr:
Please open an issue before spending time on a PR. We can then discuss the problem you're having and what the right solution may be. As it is, this looks like it would add considerable code to an already large crate for a feature that would not necessarily benefit most developers. Going to close this for now.

Xaeroxe:
Alright, I'll open an issue @kennykerr. For the record building this PR was the easiest solution to my problem, I will use the contents of it whether you merge it or not.

kennykerr:
Sounds great - nothing like a proof of concept to get the ideas flowing. 😃

PR #1449, opened 2022-01-24
Document stability guarantees

yoshuawuyts:
This documents the stability guarantees of `windows`, `windows-sys`, and `windows-bindgen`. Once we agree on the text in this PR, I proceed to file similar PRs for the `lib.rs` documentation of our respective crates.

cc/ @RobJellinghaus, I believe you had thoughts on stability guarantees? How does this read to you?

### DIFF
--- a/.github/readme.md
+++ b/.github/readme.md
@@ -88,3 +88,36 @@ fn main() {
     }
 }
 ```
+
+## Stability
+
+All crates in `windows-rs` interact with [Windows Metadata (`.winmd`)
+files](https://github.com/microsoft/win32metadata) which describe the various
+windows APIs. All Windows APIs can be split into two categories:
+`win32` APIs and `WinRT/COM` APIs. The way we bind to these APIs is different,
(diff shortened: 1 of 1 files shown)


kennykerr:
I think this sort of thing doesn't belong in the readme just yet - it's too vague and aspirational and the readme is prime real estate. 😉 

Perhaps an FAQ, so long as it is maintained and updated. Or better yet a "Stabilization" issue. Even then, I question the wording or value at this stage in light of the fact that we use semver which explicitly reserves 0.x releases for pre-release software. So, there is already the implication that the windows crates are pre-release.

sivadeilra:
I think the standard for stability for representing the Windows API has to be a higher standard than what random OSS libraries do.

RobJellinghaus:
We discussed this in our meeting last week at great length and you were on board with the concept at that time. What changed?

Sent from Samsung Galaxy smartphone.
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #1450, opened 2022-01-24
Trait-based `implement` macro

### LINKED ISSUE #1345: `implement` macro overhaul
I'm starting to limit use of proc macros as much as possible. Part of that includes an overhaul of the `implement` macro that will not depend on metadata and instead rely on traits to determine what interface vtables to implement and what methods to expect from the developer. This should significantly reduce the compile time cost of using the `implement` macro and solve a variety of usability issues with the current macro.

This will require more code to be generated ahead of time to support upcalls from the ABI and traits for implementers to use. This code should be provided via a new `implement` feature to avoid taxing compile time for projects not using this feature. The code itself can be generated in distinct source file and included on demand based on `cfg`.

kennykerr:
Fixes #1345


kennykerr:
Completing PR to work on build issues.

PR #1463, opened 2022-01-26
Improve core/xaml app sample ergonomics

riverar:
Adds an identity check to core/xaml app samples to head off a common pitfall. Also fixes up core_app sample to behave similarly to its peer.


kennykerr (approved):
Thanks!

PR #1466, opened 2022-01-26
Re-enable clippy during build validation

kennykerr:
This is tricky because `clippy` will only work with `nightly` due to the use of non-stable features by the `implement` feature.


kennykerr:
Unfortunately, we run out of memory again.

Issue #1429, opened 2022-01-13
Feature request: validate README.md samples


yoshuawuyts:
### Motivation

Right now we're not testing these, but they show up as the readme on crates.io

kennykerr:
You can add this to the `windows` crate's `lib.rs` file to enable testing:

```rust
#[cfg(doctest)]
#[doc = include_str!("../../../../.github/readme.md")]
pub struct TestReadme;
```

The trouble is that `rustdoc` doesn't know to enable the necessary crate features to allow the code to build successfully so it fails, and I haven't found a way to tell `rustdoc` which features to enable.

kennykerr:
Hey @jyn514 do you happen to know if there's a simple way to do this? Here's the readme:

https://github.com/microsoft/windows-rs/blob/master/.github/readme.md

As you can see, for each example there are various features that need to be enabled.

jyn514:
I am no longer on the rustdoc team. You can ask on Zulip if you like.

jyn514:
but in this case you probably want `cfg_attr(feature = "xxx", include_str!("..."))`

kennykerr:
Thanks @jyn514 I'll give that a try - sorry to bother you. 😊

kennykerr:
Haven't found a way to automate this testing. Will continue to test manually for the time being.

PR #1474, opened 2022-01-28
Add AgileReference

### LINKED ISSUE #236: Support for agile references
Types that aren't Send/Sync can be made so by virtue of an agile reference wrapper.

Nerixyz:
This is my attempt at adding an `AgileReference` to send references across threads, fixes #236.

`AgileReference` is generic over the type it's wrapping as this ensures the type wrapped remains the same. 

In `AgileReference::resolve` I wasn't too sure how to return the correct type. I made the pointer, `IUnknown` is wrapping, public for the module, so i can construct a new instance (maybe transmuting would be fine as well, as long as `ptr` isn't null).
It returns the result of the `cast`. I can't directly transmute to the `T` since the size of `T` isn't known (even though it should always be the size of a pointer).

A better test and documentation is still missing.


crates/tests/agile_reference/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "agile_reference"
```
kennykerr, line 2:
```suggestion
name = "test_agile_reference"
```

Test crates must begin with `test_` and then rerun `cargo -p tool_yml` to regenerate the test yml script.

kennykerr:
Thanks @Nerixyz! I'll take a closer look soon when I have a moment.

crates/libs/windows/src/core/agile_reference.rs
```diff
@@ -0,0 +1,32 @@
...
+impl<T: Interface> AgileReference<T> {
+    pub fn new<'a>(from_ref: &'a T) -> Result<Self>
+    where
+        &'a T: IntoParam<'a, IUnknown>,
+    {
+        let reference = unsafe { RoGetAgileReference(AGILEREFERENCE_DEFAULT, &T::IID, from_ref)? };
+        Ok(Self { reference, _marker: Default::default() })
+    }
+
+    pub fn resolve(&self) -> Result<T> {
```
kennykerr, line 19: This should simply be a call to `Resolve` but is currently blocked by this issue:

https://github.com/microsoft/win32metadata/issues/787

crates/libs/windows/src/core/unknown.rs
```diff
@@ -5,7 +5,7 @@ use super::*;
 /// under the hood to provide reference-counted lifetime management as well as the ability
 /// to query for additional interfaces that the object may implement.
 #[repr(transparent)]
-pub struct IUnknown(core::ptr::NonNull<core::ffi::c_void>);
+pub struct IUnknown(pub(super) core::ptr::NonNull<core::ffi::c_void>);
```
kennykerr, line 8: You should not need this (see above).

crates/libs/windows/src/core/agile_reference.rs
```diff
@@ -0,0 +1,32 @@
...
+    _marker: PhantomData<T>,
+}
+
+impl<T: Interface> AgileReference<T> {
+    pub fn new<'a>(from_ref: &'a T) -> Result<Self>
+    where
+        &'a T: IntoParam<'a, IUnknown>,
+    {
+        let reference = unsafe { RoGetAgileReference(AGILEREFERENCE_DEFAULT, &T::IID, from_ref)? };
+        Ok(Self { reference, _marker: Default::default() })
```
kennykerr, line 16: Would it be simpler to use [map](https://doc.rust-lang.org/std/result/enum.Result.html#method.map)?

Nerixyz: Yeah I didn't see that. Though rustfmt wants to keep it on a single line which makes it a bit harder to read.

kennykerr: Unfortunately, I have the rustfmt line width set to 800 to avoid a rustfmt bug.

riverar: Maybe a good candidate for [`#![rustfmt::skip]`?](https://github.com/rust-lang/rustfmt/blob/master/tests/target/skip.rs)

kennykerr: There are just too many places that I would be tempted to apply that. 🤪

Nerixyz:
I'd like to add an example to the docs but I'm not sure how to do it. Specifically, I'd like to replicate the [C#/WinRT example](https://docs.microsoft.com/en-us/windows/apps/develop/platform/csharp-winrt/agile-objects#create-an-agile-reference).

This is the example:
```rust
use windows::{core::AgileReference, UI::Popups::PopupMenu};

let non_agile = PopupMenu::new().unwrap();
let agile_reference = AgileReference::new(&non_agile).unwrap();

std::thread::spawn(move || {
    let popup = agile_reference.resolve().unwrap();
});
```

This doesn't compile, because the `UI_Popups` feature isn't enabled.

kennykerr:
#1429 discusses the same problem. I'm not yet sure how to resolve that, but let's leave the example for now. I have to think a bit more about how to tackle documentation.

crates/libs/windows/src/core/agile_reference.rs
```diff
@@ -0,0 +1,30 @@
+use super::*;
+use bindings::*;
+use core::marker::PhantomData;
+
+/// A safe wrapper around WinRT interfaces.
+///
+/// Some interfaces are not marked as agile thus then don't implement [`Send`].
+/// These interfaces can be made agile through this `AgileReference`.
```
kennykerr, line 8:
```suggestion
/// A type representing an agile reference to a COM/WinRT object. 
```

crates/libs/windows/src/core/agile_reference.rs
```diff
@@ -0,0 +1,30 @@
...
+pub struct AgileReference<T> {
+    reference: IAgileReference,
+    _marker: PhantomData<T>,
+}
+
+impl<T: Interface> AgileReference<T> {
+    /// Creates a new wrapper around the reference
```
kennykerr, line 15:
```suggestion
    /// Creates an agile reference to the object.
```

crates/libs/windows/src/core/agile_reference.rs
```diff
@@ -0,0 +1,30 @@
...
+    pub fn new<'a>(from_ref: &'a T) -> Result<Self>
+    where
+        &'a T: IntoParam<'a, IUnknown>,
+    {
+        unsafe { RoGetAgileReference(AGILEREFERENCE_DEFAULT, &T::IID, from_ref).map(|reference| Self { reference, _marker: Default::default() }) }
+    }
+
+    /// Resolves the reference for the current thread.
```
kennykerr, line 23:
```suggestion
    /// Retrieves a proxy to the target of the `AgileReference` object that may safely be used within any thread context in which get is called.
```

crates/libs/windows/src/core/agile_reference.rs
```diff
@@ -0,0 +1,30 @@
...
+use core::marker::PhantomData;
+
+/// A safe wrapper around WinRT interfaces.
+///
+/// Some interfaces are not marked as agile thus then don't implement [`Send`].
+/// These interfaces can be made agile through this `AgileReference`.
+pub struct AgileReference<T> {
+    reference: IAgileReference,
+    _marker: PhantomData<T>,
+}
```
kennykerr, line 12:
```suggestion
#[repr(transparent)]
#[derive(Clone, PartialEq, Eq)]
pub struct AgileReference<T> (IAgileReference, PhantomData<T>);
```

For consistency with the rest of `windows::core`, let's define it as such.

crates/libs/windows/src/core/agile_reference.rs
```diff
@@ -0,0 +1,30 @@
...
+
+impl<T: Interface> AgileReference<T> {
+    /// Creates a new wrapper around the reference
+    pub fn new<'a>(from_ref: &'a T) -> Result<Self>
+    where
+        &'a T: IntoParam<'a, IUnknown>,
```
kennykerr, line 18:
```suggestion
    pub fn new<'a, T: IntoParam<'a, IUnknown>>(object: T) -> Result<Self> {
```

kennykerr (dismissed):
Looks great - thanks!

crates/libs/windows/src/core/agile_reference.rs
Nerixyz: This breaks type inference because Rust can't tell that it will return a `Self<T>` but I don't see any way around it.

kennykerr: Hmm - not having the type inferred seems like a recipe for pain and suffering. The test should validate that this (somehow) works.

crates/tests/agile_reference/tests/tests.rs
```diff
@@ -0,0 +1,15 @@
+use windows::core::AgileReference;
+use windows::Media::Control::GlobalSystemMediaTransportControlsSessionManager;
+
+#[test]
+fn test() {
+    let manager = GlobalSystemMediaTransportControlsSessionManager::RequestAsync().unwrap().get().unwrap();
+    let reference = AgileReference::<GlobalSystemMediaTransportControlsSessionManager>::new(&manager).unwrap();
```
kennykerr, line 7:
```suggestion
    let reference = AgileReference::new(&manager).unwrap();
```

Will this work?

crates/tests/agile_reference/tests/tests.rs
```diff
@@ -0,0 +1,15 @@
+use windows::core::AgileReference;
+use windows::Media::Control::GlobalSystemMediaTransportControlsSessionManager;
+
+#[test]
+fn test() {
```
kennykerr, line 5:
```suggestion
fn test() -> windows::core::Result<()> {
```

Should make the test more readable using `?` instead of `unwrap()`.

crates/tests/agile_reference/tests/tests.rs
Nerixyz: This won't work. `T` and `U` don't have any relationship to Rust right now.

crates/tests/agile_reference/tests/tests.rs
kennykerr: OK, probably not all that much given the `spawn`...

Nerixyz: It should simplify it a bit. I could also use a `Result` in the thread.

I'll do that tomorrow.

crates/tests/agile_reference/tests/tests.rs
kennykerr: This seems to work:

```rust
/// A type representing an agile reference to a COM/WinRT object.
#[repr(transparent)]
#[derive(Clone, PartialEq, Eq)]
pub struct AgileReference<T>(IAgileReference, PhantomData<T>);

impl<T: Interface> AgileReference<T> {
    /// Creates an agile reference to the object.
    pub fn new(object: &T) -> Result<Self> {
        unsafe {
            let object: &IUnknown = std::mem::transmute(object);
            RoGetAgileReference(AGILEREFERENCE_DEFAULT, &T::IID, object).map(|reference| Self(reference, Default::default()))
         }
    }

    /// Retrieves a proxy to the target of the `AgileReference` object that may safely be used within any thread context in which get is called.
    pub fn resolve(&self) -> Result<T> {
        unsafe { self.0.Resolve() }
    }
}

unsafe impl<T: Interface> Send for AgileReference<T> {}
unsafe impl<T: Interface> Sync for AgileReference<T> {}

fn main() -> Result<()> {
    let manager = GlobalSystemMediaTransportControlsSessionManager::RequestAsync()?.get()?;
    let reference = AgileReference::new(&manager)?;

    let handle = std::thread::spawn(move || {
        let manager = reference.resolve().unwrap();
        manager.GetSessions().unwrap();
    });
    handle.join().unwrap();
    Ok(())
}
```

Nerixyz: This would be pretty much the same as [it was before](https://github.com/microsoft/windows-rs/pull/1474/commits/ee1aa495970340a7b10ff9ec58d8622a28b4f19c#diff-753f3bff5e476266fac3a79368ba3d184669e90cdc72acd7374fdcfde54aeac4L15-R13), only that it was using `&'a T: IntoParam<'a, IUnknown>` insted of transmuting. I'm not sure if transmuting here is a good idea since there's no constraint that `T` is `repr(transparent)` around `IUnknown`. In my understanding, using `IntoParam` here would result in the same behaviour.

kennykerr: Sure, either way should be fine. Just as long as the inference is preserved. I didn't realize that would be lost with my earlier suggestion.

kennykerr (approved):

PR #1477, opened 2022-01-31
Enable all features on docs.rs

xfix:


riverar:
It's my understanding we can't do this, as it'll consume roughly 30GiB of memory and the pipeline will fail.

xfix:
For me it's more like 14GB which builds in 14 minutes, but this may be still too much for docs.rs. (I'm building on a computer with 32GB of RAM)

kennykerr:
Yes, this won't work until `rustdoc` goes on a diet (or `docs.rs` allows far greater resource utilization). That's why the docs for the `windows` crate are hosted here instead:

https://microsoft.github.io/windows-docs-rs/

PR #1479, opened 2022-01-31
Update win32 metadata

kennykerr:
Pick up [17.0.5](https://github.com/microsoft/win32metadata/releases/tag/v17.0.5-preview) and improve support for string escaping, needed for some of the new string literals in this update.


kennykerr:
This version of the win32 metadata has a few too many bugs - will try again soon.

PR #1503, opened 2022-02-05
Add support for rustc's raw_dylib feature

ricobbe:
Add support for the raw_dylib feature now available in nightly builds of rustc.  Use of raw_dylib is controlled by the new `use_raw_dylib` feature, which is off by default, so clients using stable builds of rustc can continue to use this package.

### DIFF
--- a/crates/libs/bindgen/src/functions.rs
+++ b/crates/libs/bindgen/src/functions.rs
@@ -1,23 +1,32 @@
+use std::collections::BTreeMap;
 use super::*;
 
 pub fn gen_sys_functions(tree: &TypeTree, gen: &Gen) -> TokenStream {
     if gen.sys {
-        let mut tokens = quote! {};
+        // the keys in this map are the libraries from which the various functions
+        // are imported.  The corresponding values are non-empty TokenStreams that contain
+        // the declarations of the functions imported from those libraries.
+        let mut tokens_by_library: BTreeMap<String, TokenStream> = BTreeMap::new();
 
         for entry in tree.types.values() {
-            tokens.combine(&gen_function_if(entry, gen));
+            gen_function_if(&mut tokens_by_library, entry, gen);
         }
 
-        if !tokens.is_empty() {
-            quote! {
-                #[link(name = "windows")]
-                extern "system" {
-                    #tokens
+        let mut tokens = quote! {};
+        // Because tokens_by_library is a BTreeMap, we're guaranteed to visit entries
+        // in ascending order by key.
+        for (library, lib_tokens) in tokens_by_library {
+            tokens.combine(
+                &quote! {
+                    #[cfg_attr(feature = "use_raw_dylib", link(name = #library, kind = "raw-dylib"))]
+                    #[cfg_attr(not(feature = "use_raw_dylib"), link(name = "windows"))]
+                    extern "system" {
+                        #lib_tokens
+                    }
                 }
-            }
-        } else {
-            quote! {}
+            );
         }
+        tokens
     } else {
         quote! {}
     }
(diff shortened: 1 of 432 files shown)


ricobbe:
Note to reviewers: in an attempt to make this easier to review, I've split the changes into 2 commits.  The first (df1f5be) is quite small; it contains only the changes to the code generator.  The other commit, 25c802b, is much larger and contains only the results of rerunning the modified code generator.

I'll be happy to squash these commits once there are no additional issues, if that's what you'd prefer.

kennykerr:
For the failing check to pass, you'll have to split it across stable and nightly runs and exclude the new feature for the stable runs.

ricobbe:
> For the failing check to pass, you'll have to split it across stable and nightly runs and exclude the new feature for the stable runs.

We appear to have reached an impasse, unfortunately.  My latest push attempts to do just this by adding a cargo feature named `all_stable_features` that, basically, turns on all of the features _except_ `use_raw_dylib`.  However, this doesn't work, because the build then tries to run `cargo check -p windows-sys --features all_stable_features`, but that fails because you can't use `--features` when compiling from the root directory of a workspace.

There doesn't seem to be a clean way to make compilation conditional on the version of a compiler.  The best way I've seen (and "best" is a relative term) is to use a build.rs to detect the version of the compiler and set a rustc feature (as distinct from a cargo feature) accordingly.  We can go that route if we need to, but I'd really prefer to avoid adding a build.rs file just for this.  That's entirely up to you, though.

kennykerr:
> However, this doesn't work, because the build then tries to run cargo check -p windows-sys --features all_stable_features, but that fails because you can't use --features when compiling from the root directory of a workspace.

I don't really understand the error because I do that locally all of the time.

ricobbe:
> I don't really understand the error because I do that locally all of the time.

Yes, it works for me from the command line as well.  I don't fully understand what's going on, but I wonder if there may be some sort of configuration difference between cargo as we have installed it vs. as it's installed in github CI?  Either way, I worked around it for now.   Hopefully this is all temporary; once raw-dylib becomes available in stable builds of rustc we can get rid of all this stuff.  For now, though, the tests are passing.

.github/workflows/build.yml
```diff
@@ -44,17 +53,28 @@ jobs:
...
       matrix:
-        rust: [1.46.0, stable, nightly]
+        rust: [1.46.0, stable]
     steps:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Update toolchain
       run: rustup update --no-self-update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
+    - name: Run cargo check
+      run: cd crates/libs/sys && cargo check --features all_stable_features
```
kennykerr, line 68: Since we have to generate the list either way, why don't we generate the list directly into the `build.yml` rather than exposing a *second* feature only to remove it later?

crates/libs/bindgen/src/functions.rs
```diff
@@ -1,23 +1,30 @@
...
-        if !tokens.is_empty() {
-            quote! {
-                #[link(name = "windows")]
+        let mut tokens = quote! {};
+        // Because tokens_by_library is a BTreeMap, we're guaranteed to visit entries
+        // in ascending order by key.
+        for (library, lib_tokens) in tokens_by_library {
+            tokens.combine(&quote! {
+                #[cfg_attr(feature = "use_raw_dylib", link(name = #library, kind = "raw-dylib"))]
+                #[cfg_attr(not(feature = "use_raw_dylib"), link(name = "windows"))]
```
kennykerr, line 21: Can we add a test for this? Perhaps a dylib version of this test:

https://github.com/microsoft/windows-rs/tree/master/crates/tests/sys

.github/workflows/build.yml
ricobbe: I've made this change and pushed it to github; CI is just finishing up.  How do you typically handle things with this project?  Should I click "resolve conversation" as soon as the changes addressing the feedback have been pushed?

kennykerr: Thanks, feel free to resolve.

kennykerr:
Closing for now - we can reopen once the `rustc` compiler is a bit further along.

PR #1511, opened 2022-02-07
Prefer `&self` for COM interface traits

### LINKED ISSUE #1497: COM interface traits using `&mut self` may not be sound
Without the `implement` macro (or supporting shims) providing additional protection, COM methods implemented with `&mut self` may not be sound due to aliasing rules and result in miscompilation. We have a few potential options:

* Use something like `this: *mut Self` to make it clear to the developer and the compiler that all bets are off. 
* Provide automatic locking of some kind behind the scenes.
* Use `&self` as the default, if that can be determined to be safe from miscompilation. 

(Based on some internal feedback from the compiler team)

kennykerr:
The "implement_vector" test illustrates one way to mutate within an implementation using a `RwLock`. 

Fixes #1497


crates/tests/implement/tests/buffer.rs
```diff
@@ -7,8 +7,11 @@ struct TestBuffer(Vec<u8>);
 
 #[allow(non_snake_case)]
 impl IBufferByteAccess_Impl for TestBuffer {
-    fn Buffer(&mut self) -> Result<*mut u8> {
-        Ok(self.0.as_mut_ptr())
+    fn Buffer(&self) -> Result<*mut u8> {
+        unsafe {
+            let writer: &mut Self = &mut *(self as *const _ as *mut Self);
```
rpjohnst: This kind of cast is unsound- because `self` is a shared reference, producing a unique reference from it has immediate undefined behavior. (The same thing applies to the into_impl and implement_data_object tests.)

For this to work you will need some kind of interior mutability to make `*mut` access legal at all, and also some care to avoid invalidating (in the language-lawyer, avoiding-miscompilation sense) the `*mut u8`s you hand out if `Buffer` gets called multiple times. For reference, the current candidate for the actual rules here is [Stacked Borrows](https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md), and you can test compliance with those rules using [Miri](https://github.com/rust-lang/miri#using-miri).

kennykerr: Yep, this is just to let the tests pass without writing a full `RwLock` implementation as I did for [the vector test](https://github.com/microsoft/windows-rs/blob/master/crates/tests/implement_vector/tests/test.rs). 

Is there a simple way to write this test that is not UB even if it is wildly unsafe? `IBufferByteAccess` is wildly unsafe to begin with. I'd just prefer not to have to write each test with something like `RwLock` if I can help it.

rpjohnst: Yes- the lowest level form of interior mutability is `UnsafeCell`, which directly gives you a `*mut` to its contents.

For example, this passes under Miri (Tools > Miri in the upper right): https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=abb63fe3c9e3a30f9d274f6111bdd6d9

kennykerr: Thanks, will update the tests to use `UnsafeCell`.

PR #1517, opened 2022-02-08
Update tests to use `UnsafeCell`

### LINKED ISSUE #1515: Update tests to use `UnsafeCell`
Yes- the lowest level form of interior mutability is `UnsafeCell`, which directly gives you a `*mut` to its contents.

For example, this passes under Miri (Tools > Miri in the upper right): https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=abb63fe3c9e3a30f9d274f6111bdd6d9

_Originally posted by @rpjohnst in https://github.com/microsoft/windows-rs/pull/1511#discussion_r802073208_

kennykerr:
Fixes #1515

### DIFF
--- a/crates/tests/implement/tests/buffer.rs
+++ b/crates/tests/implement/tests/buffer.rs
@@ -3,21 +3,18 @@ use windows::Storage::Streams::Buffer;
 use windows::Win32::System::WinRT::*;
 
 #[implement(IBufferByteAccess)]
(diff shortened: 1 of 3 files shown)


rpjohnst (approved):
Looks reasonable to me- IIUC this fixes the immediate UB (of course without addressing concurrency).

Issue #1521, opened 2022-02-09
Remove workaround for wrong import metadata


kennykerr:
I think we can remove this workaround now. @riverar 

https://github.com/microsoft/windows-rs/blob/c7cbaaafd9781a2aefafbed5b3777f1cf12d9056/crates/tools/msvc/src/main.rs#L86-L88

riverar:
@kennykerr Yup 🥳

PR #1522, opened 2022-02-09
Remove bthprops special handling

riverar:
Removes bthprops special handling (#1257) that is no longer needed. Metadata now explicitly emits extensions for non `.dll` libs.

Closes: #1521 
See also: microsoft/win32metadata#777


kennykerr (approved):
Thanks!

PR #1531, opened 2022-02-11
Correct aarch64 import lib and add tests

riverar:
Temporary comment: Added yaml changes first, will push the new lib after the build fails to validate tests.


.github/workflows/test.yml
```diff
@@ -138,4 +138,27 @@ jobs:
         cargo test --target ${{ matrix.target }} -p test_implement_vector &&
         cargo test --target ${{ matrix.target }} -p test_implement_winrt 
       if: matrix.version == 'nightly'
+
+    - name: Check import libs
```
kennykerr, line 142: Thanks! You'll need to add this to the `tool_yml` crate to generate this yml.

riverar: ✔️

PR #1540, opened 2022-02-15
COM Interface Declaration Support

### LINKED ISSUE #1486: Add `interface` macro to define COM interfaces locally
While I'm working on #1093 and #1094 for complete API and component authoring support, some projects need to implement COM interfaces that are not necessarily described by Windows metadata and don't necessarily warrant public API definitions. A project might just want to implement some locally defined COM interface for interop with another private component. 

I recently overhauled the windows crate's `implement` macro to rely on traits rather than metadata (#1345). What that means is that we could conceivably define interfaces directly in Rust, via a new `interface` macro, and then have the `implement` macro implement them and never actually touch any metadata from start to finish. At that point the `windows` crate could do everything the `com` crate currently does.

* The `implement` proc macro can define the necessary types that the `implement` macro could simply consume and implement (as if it had come from the pre-generated bindings). 
* There should be no material difference between the code that the `windows-bindgen` produces and the `implement` macro produces. The `implement` macro should not care or be able to tell the difference. 
* It can further be limited to COM interfaces and avoid the complexities of describing WinRT interfaces. 
* Support for WinRT would be left to #1093 and #1094. 

Perhaps something like this:

```rust
#[interface("0000010c-0000-0000-C000-000000000046")]
trait IPersist : IUnknown {
    fn GetClassID(clsid: *mut GUID) -> HRESULT;
}
```

rylev:
This PR adds the very beginnings of COM interface declaration support. 

```rust
#[interface("8CEEB155-2849-4ce5-9448-91FF70E1E4D9")]
unsafe trait IUIAnimationVariable: IUnknown {
    fn GetValue(&self, value: *mut f64) -> HRESULT;
}
```

There is still plenty of additional support to add, and many interfaces will not work yet, but I believe it's a good start, and I wanted to get feedback sooner rather than later. 

Fixes #1486


kennykerr:
That's a great start. That's essentially what I had in mind with #1486. A few comments.

Why require an `unsafe` trait? It's not actually making a Rust trait, so it seems a bit superfluous. 

We need to support two key scenarios that should be easy to test:

1. Re-declare an existing Windows interface and call it against an existing Windows implementation. 

2. Use the `implement` macro to implement the newly minted interface and test (call) it through the definition provided by the `windows` crate. Something simple like `CreateUri` and `IUri` from [the com_uri sample](https://github.com/microsoft/windows-rs/tree/master/crates/samples/com_uri) should suffice to smoke out basic issues to start.

Something like this should validate the basic support:

```rust
#![allow(non_snake_case)]

use windows::{
    core::*,
    Win32::Foundation::*,
    Win32::System::Com::*,
};

#[interface("a39ee748-6a27-4817-a6f2-13914bef5890")]
unsafe trait ICustomUri : IUnknown {
    fn GetPropertyBSTR(&self);
    fn GetPropertyLength(&self);
    fn GetPropertyDWORD(&self);
    fn HasProperty(&self);
    fn GetAbsoluteUri(&self);
    fn GetAuthority(&self);
    fn GetDisplayUri(&self);
    fn GetDomain(&self, value: *mut BSTR) -> HRESULT;
    // etc
}

#[implement(ICustomUri)]
struct CustomUri();

impl ICustomUri_Impl for CustomUri {
    fn GetPropertyBSTR(&self) { todo!() }
    fn GetPropertyLength(&self) { todo!() }
    fn GetPropertyDWORD(&self) { todo!() }
    fn HasProperty(&self) { todo!() }
    fn GetAbsoluteUri(&self) { todo!() }
    fn GetAuthority(&self) { todo!() }
    fn GetDisplayUri(&self) { todo!() }
    fn GetDomain(&self, value: *mut BSTR) -> HRESULT {
        *value = "kennykerr.ca".into();
        S_OK
    }
}

    unsafe {
        // Use the OS implementation through the OS interface
        let a: IUri = CreateUri("http://kennykerr.ca", Default::default(), 0)?;
        let domain = a.GetDomain()?;
        assert_eq!(domain, "kennykerr.ca");

        // Call the OS implementation through the custom interface
        let b: ICustomUri = a.cast()?;
        let mut domain = BSTR::new();
        b.GetDomain(&mut domain).ok()?;
        assert_eq!(domain, "kennykerr.ca");

        // Use the custom implementation through the OS interface
        let c: IUri = CustomUri().into();
        let domain = c.GetDomain()?;
        assert_eq!(domain, "kennykerr.ca");

        // Call the custom implementation through the custom interface
        let d: ICustomUri = c.cast()?;
        let mut domain = BSTR::new();
        d.GetDomain(&mut domain).ok()?;
        assert_eq!(domain, "kennykerr.ca");

        Ok(())
    }
}
```

rylev:
@kennykerr thanks for the example code! I got it working (with a few minor changes that I believe were just typos in the provided example). 

As for why I require `unsafe` before the `trait`: this was a source of debate in com-rs. All the methods are als required to be marked unsafe which definitely makes sense as calling a COM method is inherently unsafe (if the interface is declared wrong you might end up with incorrect pointer dereferences and such). Since the declaration of the interface needs to be correct in order for things to work properly (and not potentially end up dereferencing junk memory for example), I think it makes sense to require the user to annotate the trait with `unsafe` as a sort of acknowledgement of this fact. 

I talk a bit more about this in the com-rs [docs](https://github.com/microsoft/com-rs/blob/master/docs/safety.md). 

Speaking of which, I think I should add docs on how `interface` and `implement` work.

crates/tests/interface/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "test_interface"
```
kennykerr, line 2: Currently, if you rely on nightly features (like the `implement` macro) then you need to name your test crate a little differently so that the yml generator groups it acccordingly:

https://github.com/microsoft/windows-rs/blob/324a0c082f2f6019840ad4e30d4920682c0637f9/crates/tools/yml/src/main.rs#L241-L243

I should just make this "nightly" instead of "implement" but that should unblock this for now.

kennykerr: Unfortunately, `tool_yml` exists because things like `cargo test --all` will just run out of memory on GitHub's build servers. So that tool unrolls the `--all` flag.

crates/tests/interface/tests/com.rs
```diff
@@ -0,0 +1,156 @@
...
+    unsafe fn GetDisplayUri(&self) -> HRESULT {
+        todo!()
+    }
+    unsafe fn GetDomain(&self, value: *mut BSTR) -> HRESULT {
+        *value = "kennykerr.ca".into();
+        S_OK
+    }
+}
+
+impl IUri_Impl for CustomUri {
```
kennykerr, line 51: Not needed.

crates/tests/interface/tests/com.rs
```diff
@@ -0,0 +1,156 @@
...
+    unsafe fn GetPropertyDWORD(&self) -> HRESULT;
+    unsafe fn HasProperty(&self) -> HRESULT;
+    unsafe fn GetAbsoluteUri(&self) -> HRESULT;
+    unsafe fn GetAuthority(&self) -> HRESULT;
+    unsafe fn GetDisplayUri(&self) -> HRESULT;
+    unsafe fn GetDomain(&self, value: *mut BSTR) -> HRESULT;
+    // etc
+}
+
+#[implement(ICustomUri, IUri)]
```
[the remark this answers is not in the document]
rylev: This is needed for `let c: IUri = CustomUri.into();` to work. I suppose we could just transmute but that's pretty ugly. While this is verbose, it does get the point across pretty clearly.

kennykerr (approved):

PR #1542, opened 2022-02-16
Make NTSTATUS::ok a const fn

roblabla:
is_ok, to_hresult and fast_error are all const fn, so `ok` should be a fine candidate for const-ification.


kennykerr (approved):
Thanks!

PR #1559, opened 2022-02-22
Fixed win32 array support

kennykerr:
This is the first step toward full Win32 array support. This provides support for parsing win32 array metadata as well as limited application for those parameters that model fixed win32 arrays. These are modeled as either mutable or immutable slices. Next up is relative win32 array support.


kennykerr:
Looks like some of the APIs that are marked with this metadata will return memory allocated via a custom allocator, which makes using a slice invalid. Going to have to figure out how to exclude those before proceeding.

kennykerr:
Added a workaround (https://github.com/microsoft/windows-rs/pull/1559/commits/efb0c914b15f6fa33834928a5fd5862385afc23c) which assumes such parameters are attributed correctly.

kennykerr:
Going to round out the remainder of the win32 array support for completing.

PR #1560, opened 2022-02-23
Document stability guarantees

yoshuawuyts:
Continuation of https://github.com/microsoft/windows-rs/pull/1449. Kenny and I followed up offline and agreed it was useful to document the stability guarantees in a separate file, but having a link to it from the README.md.

I've added a section on MSRV based on customer feedback as well. And added a section about our current release schedule too (next release is on Thursday (tomorrow), but after that we're switching to Tuesday releases).

I hope this is helpful, thanks!

---

cc/ @robjellinghaus I'm sorry last time you didn't have the opportunity to review the PR. Pinging you here in case you're still interested in giving it a read and sharing your thoughts. Thanks!

### DIFF
--- a/.github/stability.md
+++ b/.github/stability.md
@@ -5,7 +5,7 @@ under the org are considered "stable". Some crates such as `windows-sys` might
 be closer than others, but until a crate reaches `1.0.0` it may be subject to
 change.
 
-## Release schedule and versioning
+## Release Schedule and Versioning
 
 We're set to release new versions of crates every 3 weeks on Tuesdays. Currently all crates
 are pre-1.0 and have their semver-minor version bumped on release. We are
@@ -16,14 +16,17 @@ update this section with our versioning strategy.
 
 All crates in `windows-rs` interact with [Windows Metadata (`.winmd`)
 files](https://github.com/microsoft/win32metadata) which describe the various
-windows APIs. All Windows APIs can be split into two categories:
-`win32` APIs and `WinRT/COM` APIs. The way we bind to these APIs is different,
+Windows APIs. All Windows APIs can be split into two categories:
+`win32/COM` APIs and `WinRT` APIs. The way we bind to these APIs is different,
 but in the metadata they also have different stability guarantees:
 
 | Metadata category | Stable? |
 |-------------------|---------|
-| `Win32`           | ❌       |
-| `WinRT/COM`       | ✅       |
+| `Win32/COM`       | ❌       |
+| `WinRT`           | ✅       |
+
+When metadata is marked as "stable" it means that it will only ever be extended.
+Information will not be changed or removed.
 
 ## Crate Stability
 
@@ -40,16 +43,14 @@ two factors which both must be stable for the crate to be stable:
 | **`windows-sys`**     | ✅                | ❌                    | ❌ (but we're close) | ❌             |
 | **`windows-bindgen`** | ✅                | ✅                    | ❌                   | ❌             |
 
-Because of the scope of the project (`230.000+` unique types), marking anything
-as "stable" is a difficult task. Despite that, it is our explicit intent to
-eventually achieve that. Over time we hope to gradually mark more APIs as
-stable, propose concrete plans how to get to `1.0.0` stable releases for all
-crates, and ensure that any breakage before then is well-communicated to reduce
-friction.
+Because of the scope of the project (`230_000+` unique types), marking anything
+as "stable" is a difficult task. While we aspire to eventually provide stable
+versions of crates in the future, it is still too early to say anything
+meaningful of what that might look like, or when that might happen.
 
 ## Minimum Supported Rust Version (MSRV)
 
-The crates in `windows-rs` are tested to work on the latest _stable_ version of
+The crates in `windows-rs` are only guaranteed to work on the latest _stable_ version of
 Rust. We may introduce MSRV requirements once individual crates reach 1.0
 versions. But right now we're expecting to be able to significantly improve our
 APIs based on features expected to land in upcoming versions of Rust, and
--- a/.github/stability.md
+++ b/.github/stability.md
@@ -1,0 +1,58 @@
+## Stability
+
+In short: `windows-rs` is still under active development, and none of the crates
+under the org are considered "stable". Some crates such as `windows-sys` might
+be closer than others, but until a crate reaches `1.0.0` it may be subject to
+change.
+
+## Release schedule and versioning
+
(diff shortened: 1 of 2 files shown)


.github/stability.md
```diff
@@ -0,0 +1,58 @@
...
+change.
+
+## Release schedule and versioning
+
+We're set to release new versions of crates every 3 weeks on Tuesdays. Currently all crates
+are pre-1.0 and have their semver-minor version bumped on release. We are
+looking for ways to version and release crates individually. Once we do, we will
+update this section with our versioning strategy.
+
+## Metadata Stability
```
kennykerr, line 15: COM APIs are included with the Win32 metadata, not the WinRT metadata.

.github/stability.md
```diff
@@ -0,0 +1,58 @@
...
+All crates in `windows-rs` interact with [Windows Metadata (`.winmd`)
+files](https://github.com/microsoft/win32metadata) which describe the various
+windows APIs. All Windows APIs can be split into two categories:
+`win32` APIs and `WinRT/COM` APIs. The way we bind to these APIs is different,
+but in the metadata they also have different stability guarantees:
+
+| Metadata category | Stable? |
+|-------------------|---------|
+| `Win32`           | ❌       |
+| `WinRT/COM`       | ✅       |
```
rylev, line 26: Is the WinRT stable in the semver sense or only stable in the "eh, it doesn't change that often". We might want to define what we mean by stable metadata.

.github/stability.md
```diff
@@ -0,0 +1,58 @@
...
+
+| Crate name            | `Win32` support? | `WinRT/COM` support? | Projection Stable?  | Crate Stable? |
+|-----------------------|------------------|----------------------|---------------------|---------------|
+| **`windows`**         | ✅                | ✅                    | ❌                   | ❌             |
+| **`windows-sys`**     | ✅                | ❌                    | ❌ (but we're close) | ❌             |
+| **`windows-bindgen`** | ✅                | ✅                    | ❌                   | ❌             |
+
+Because of the scope of the project (`230.000+` unique types), marking anything
+as "stable" is a difficult task. Despite that, it is our explicit intent to
+eventually achieve that. Over time we hope to gradually mark more APIs as
```
kennykerr, line 45: I'd suggest toning this down or just removing this comment. While we aspire to stabilized all APIs, it promises too much. We also have no plans or means to "mark more APIs as stable" so until we can define what that means, I'd also avoid making such claims.

yoshuawuyts:
Updated with feedback from review!

.github/stability.md
```diff
@@ -0,0 +1,59 @@
...
+| Crate name            | `Win32` support? | `WinRT/COM` support? | Projection Stable?  | Crate Stable? |
+|-----------------------|------------------|----------------------|---------------------|---------------|
+| **`windows`**         | ✅                | ✅                    | ❌                   | ❌             |
+| **`windows-sys`**     | ✅                | ❌                    | ❌ (but we're close) | ❌             |
+| **`windows-bindgen`** | ✅                | ✅                    | ❌                   | ❌             |
+
+Because of the scope of the project (`230_000+` unique types), marking anything
+as "stable" is a difficult task. While we aspire to eventually provide stable
+versions of crates in the future, it is still too early to say anything
+meaningful of what that might look like, or when that might happen.
```
kennykerr, line 49: > it is still too early to say anything meaningful of what that might look like

Ironically, this nicely captures why I don't think I'm ready to add a document describing stability guarantees. There are some useful explainers here but perhaps they belong in an FAQ or similar format.

.github/stability.md
```diff
@@ -0,0 +1,59 @@
...
+`win32/COM` APIs and `WinRT` APIs. The way we bind to these APIs is different,
+but in the metadata they also have different stability guarantees:
+
+| Metadata category | Stable? |
+|-------------------|---------|
+| `Win32/COM`       | ❌       |
+| `WinRT`           | ✅       |
+
+When metadata is marked as "stable" it means that it will only ever be extended.
+Information will not be changed or removed.
```
riverar, line 29: With this understanding of _stable_ (thanks for defining it!), I don't think we can say the WinRT bucket is stable.

WinRT APIs get added or get marked deprecated regularly, typically during OS release windows. This results in changes to the metadata this crate relies upon that could then critically alter the crate feature map, causing breakage.

---

A somewhat contrived but real world example:

Rewind your clocks to Windows 8.1 timeframe. Let's envision this shiny new API shipped: `GattDeviceService.GetIncludedServices`

It's currently described in Windows Metadata as:
```
[MethodImpl(MethodImplOptions.InternalCall)]
public extern IReadOnlyList<GattDeviceService> GetIncludedServices([In] Guid serviceUuid);
```

Looks good, I slam `features = ["Windows_Devices_Bluetooth"]`" into my crate and life is good.

Some time passes, Microsoft sweeps Windows 8 under the carpet and puts out Windows 10. This Windows 8-era API is subsequently deprecated.

It is now described in Windows Metadata as:

```
[MethodImpl(MethodImplOptions.InternalCall)]
[Deprecated("Use GetIncludedServicesAsync instead of GetAllIncludedServices.  For more information, see MSDN.", DeprecationType.Deprecate, 262144u, "Windows.Foundation.UniversalApiContract")]
public extern IReadOnlyList<GattDeviceService> GetAllIncludedServices();
```

The `windows` crate then picks up newer metadata at some point and pushes out a point release. Now because this API is lumped into the `deprecated` feature bucket, my crate breaks.

---

I think your stability goal is reachable. But we're going to need to sit down and answer a few more questions before we can begin to really make any promises. (Forgive me if you already have these answers, add them to doc. if so!)

Some questions:

* Which versions of Windows does the crate target? How do APIs deprecated mid-OS-major-release-cycle get handled? (An increasing number of Windows components are shipping/updating out of band...)
* Do we have a clear understanding of the upstream sources and their release schedules? Any stability guarantees there? win32metadata, to my (limited external) knowledge, has no non-prerelease release in sight. And OS metadata is a toss up (see earlier example). Are we going to need to slice and dice up upstream metadata?
* What's the expected service level agreement (SLA)? What's my recourse if Microsoft breaks its promise? Will it fix the crate immediately? Drag its feet until next release? Can I open a ticket with Product Support and pay money? What if Kenny is out for vacation?

kennykerr:
> Kenny and I followed up offline and agreed it was useful to document the stability guarantees in a separate file

While I agree that such a document would be useful, I don't think we are close to having a sufficient understanding of this topic and a much more detailed analysis remains to be done. @riverar also raises some good points that need to be explored.

PR #1562, opened 2022-02-23
Add support for Win32 array parameters

### LINKED ISSUE #479: Support for win32 array parameters
Hi, I'm trying out the Direct 3D Api in conjunction with VLC. I'm pretty much following an example provided by VLC [here](https://code.videolan.org/videolan/vlc/-/blob/master/doc/libvlc/d3d11_player.cpp).

When creating the Direct 3D related things, I need to use the method/function `ID3D11DeviceContext::OMSetRenderTargets`. In the [documentation from Microsoft](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-omsetrendertargets) in C++, the second parameter `ppRenderTargetViews` is a **pointer to an array**. 

[This](https://microsoft.github.io/windows-docs-rs/doc/bindings/windows/win32/direct3d11/struct.ID3D11DeviceContext.html#method.OMSetRenderTargets) is the generated documentation for this method.

<details><summary>Generated bindings</summary> 

```rust
pub fn OMSetRenderTargets<
    'a,
    T1__: ::std::convert::Into<::windows::Param<'a, ID3D11RenderTargetView>>,
    T2__: ::std::convert::Into<::windows::Param<'a, ID3D11DepthStencilView>>,
>(
    &self,
    num_views: u32,
    pp_render_target_views: T1__,
    p_depth_stencil_view: T2__,
) {
    unsafe {
        (::windows::Interface::vtable(self).33)(
            ::windows::Abi::abi(self),
            num_views,
            pp_render_target_views.into().abi(),
            p_depth_stencil_view.into().abi(),
        )
    }
}
```
 </details>

In the [C++ version](https://code.videolan.org/videolan/vlc/-/blob/master/doc/libvlc/d3d11_player.cpp#L191), the method is called like this:
```c++
// swapchainRenderTarget is declared like this: ID3D11RenderTargetView *swapchainRenderTarget
 ctx->d3dctx->OMSetRenderTargets(1, &ctx->swapchainRenderTarget, NULL);
```

Now when I try to call the method like in the following example, the program just sleeps/waits until it's stopped. So there's something wrong.

```rust
// swapchain_render_target: ID3D11RenderTargetView
device_context.OMSetRenderTargets(1, &swapchain_render_target, None);
```

How do I call the method with the correct argument?

<details><summary>My code</summary>

```rust
let scd = DXGI_SWAP_CHAIN_DESC {
    buffer_count: 1,
    buffer_desc: DXGI_MODE_DESC {
        width: dimensions.x as u32,
        height: dimensions.y as u32,
        format: DXGI_FORMAT::DXGI_FORMAT_R8G8B8A8_UNORM,
        ..Default::default()
    },
    buffer_usage: DXGI_USAGE_RENDER_TARGET_OUTPUT,
    output_window: hwnd,
    sample_desc: DXGI_SAMPLE_DESC {
        count: 1,
        ..Default::default()
    },
    windowed: true.into(),
    flags: DXGI_SWAP_CHAIN_FLAG::DXGI_SWAP_CHAIN_FLAG_ALLOW_MODE_SWITCH.0 as u32,
    ..Default::default()
};

let (device, device_context, swapchain) = unsafe {
    let mut device = None;
    let mut swapchain = None;
    let mut device_context = None;
    let hr = D3D11CreateDeviceAndSwapChain(
        None,
        D3D_DRIVER_TYPE::D3D_DRIVER_TYPE_HARDWARE,
        0,
         D3D11_CREATE_DEVICE_FLAG::D3D11_CREATE_DEVICE_DEBUG.0 as u32,
        ptr::null(),
        0,
        D3D11_SDK_VERSION as u32,
        &scd,
        &mut swapchain,
        &mut device,
        ptr::null_mut(),
        &mut device_context,
    );
    match (device, device_context, swapchain) {
        (Some(device), Some(ctx), Some(swap)) => (device, ctx, swap),
        _ => return Err(bindings::windows::Error::new(
            hr,
            "Failed to create device/ctx/swapchain",
        ))
    }
};
println!("Created Device & Ctx & Swapchain");

if let Ok(mt) = device.cast::<ID3D10Multithread>() {
    mt.SetMultithreadProtected(true.into());
}
println!("Set Multithread protection");

let viewport = D3D11_VIEWPORT {
    width: dimensions.x,
    height: dimensions.y,
    max_depth: 0.0,
    min_depth: 0.0,
    top_leftx: 0.0,
    top_lefty: 0.0,
};

device_context.RSSetViewports(1, &viewport);
println!("Set Viewport");

let mut back_buffer: Option<ID3D11Texture2D> = None;
let hr = swapchain.GetBuffer(0, &ID3D11Texture2D::IID, &mut back_buffer as *mut _ as _);
let back_buffer = back_buffer.ok_or_else(|| bindings::windows::Error::new(
    hr,
    "Failed to create back buffer",
))?;
println!("Created BackBuffer");

let mut swapchain_render_target = None;
let hr = device.CreateRenderTargetView(back_buffer, ptr::null(), &mut swapchain_render_target);
let swapchain_render_target = swapchain_render_target.ok_or_else(|| bindings::windows::Error::new(
    hr,
    "Failed to create render target",
))?;
println!("Created RenderTarget");

device_context.OMSetRenderTargets(1, &swapchain_render_target, None);

// This is never logged
println!("Set RenderTarget");
```
 </details>

### LINKED ISSUE #790: ID3D12GraphicsCommandList1::ClearRenderTargetView - array params could possibly be projected better
I'm not sure what the best rust projection would be here - and I don't think that win32metadata has everything needed yet, but I figured we should figure out what can be possible in rust first.

Here's the declaration of ClearRenderTargetView:

```c++
virtual void STDMETHODCALLTYPE ClearRenderTargetView( 
            _In_  D3D12_CPU_DESCRIPTOR_HANDLE RenderTargetView,
            _In_  const FLOAT ColorRGBA[ 4 ],
            _In_  UINT NumRects,
            _In_reads_(NumRects)  const D3D12_RECT *pRects) = 0;
```

Notice the ColorRGBA is explicitly an array for 4 elements.  pRects reads the number of rects specified by NumRects.

Here's the rust binding:

```rust
pub unsafe fn ClearRenderTargetView<'a>(
    &self,
    rendertargetview: impl ::windows::IntoParam<'a, D3D12_CPU_DESCRIPTOR_HANDLE>,
    colorrgba: *const f32,
    numrects: u32,
    prects: *const super::super::UI::DisplayDevices::RECT,
)
```

Ideally, I'd be able to specify the color inline.  Maybe that's possible?  This is what I came up with:
```rust
let clear_color: [f32; 4] = [0.0, 0.2, 0.4, 1.0];
command_list.ClearRenderTargetView(rtv_handle, &clear_color[0], 0, std::ptr::null());
```

It's a bit clunky having to take a reference to the first element of the array.

For pRects, I notice that the rust binding isn't make any use of the relationship between prects and numrects.  Maybe there's not anything it could do with it, but I thought I'd mention it while I was here.

### LINKED ISSUE #972: SendInput takes only one INPUT
I have just started learning Rust and was playing around.

I found out that [SendInput](https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/UI/KeyboardAndMouseInput/fn.SendInput.html) takes only one [INPUT](https://microsoft.github.io/windows-docs-rs/doc/bindings/Windows/Win32/UI/KeyboardAndMouseInput/struct.INPUT.html) but according to the [Microsoft Windows API documentation ](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-sendinput) it should take an array of INPUTs.

I think this is a bug but please let me know if I'm wrong and if so please tell me how to make it work!

This code compiles:
```
use bindings::{
    Windows::Win32::UI::KeyboardAndMouseInput::{INPUT, INPUT_0, INPUT_TYPE, KEYBDINPUT, KEYBD_EVENT_FLAGS, SendInput},
    Windows::Win32::UI::WindowsAndMessaging::GetMessageExtraInfo,
};
use std::mem::size_of;

fn main(){
    const CBSIZE: i32 = size_of::<INPUT>() as i32;

    for i in 1..10 {
        unsafe {
            let extra_info = GetMessageExtraInfo();
            let extra_info = extra_info.0.unsigned_abs();
            let cinputs: u32 = 1;
            let mut pinputs = INPUT {
                r#type: INPUT_TYPE {0: 1},
                Anonymous: INPUT_0 {
                    ki: KEYBDINPUT {
                        wVk: 0x41,
                        wScan: 1,
                        dwFlags: KEYBD_EVENT_FLAGS(0),
                        time: 0,
                        dwExtraInfo: extra_info,
                    }
                }
            };
            let res = SendInput(cinputs, &mut pinputs, CBSIZE);
            println!("{}. {}", i, res);
        }
    }
}
```

But this code dosn't:
```
use bindings::{
    Windows::Win32::UI::KeyboardAndMouseInput::{INPUT, INPUT_0, INPUT_TYPE, KEYBDINPUT, KEYBD_EVENT_FLAGS, SendInput},
    Windows::Win32::UI::WindowsAndMessaging::GetMessageExtraInfo,
};
use std::mem::size_of;

fn main(){
    const CBSIZE: i32 = size_of::<INPUT>() as i32;

    for i in 1..10 {
        unsafe {
            let extra_info = GetMessageExtraInfo();
            let extra_info = extra_info.0.unsigned_abs();
            let cinputs: u32 = 1;
            let mut pinputs = [
                INPUT {
                    r#type: INPUT_TYPE {0: 1},
                    Anonymous: INPUT_0 {
                        ki: KEYBDINPUT {
                            wVk: 0x41,
                            wScan: 1,
                            dwFlags: KEYBD_EVENT_FLAGS(0),
                            time: 0,
                            dwExtraInfo: extra_info,
                        }
                    }
                },
            ];
            let res = SendInput(cinputs, &mut pinputs, CBSIZE);
            println!("{}. {}", i, res);
        }
    }
}
```
Error:
```
error[E0308]: mismatched types
  --> src\main.rs:29:42
   |
29 |             let res = SendInput(cinputs, &mut pinputs, CBSIZE);
   |                                          ^^^^^^^^^^^^ expected struct `INPUT`, found array of 1 element
   |
   = note:    expected raw pointer `*mut INPUT`
           found mutable reference `&mut [INPUT; 1]`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
error: could not compile `wintest`

To learn more, run the command again with --verbose.
```

You can find the working code [here](https://github.com/CrumblyLiquid/WinSnippets/tree/main/press-key)

### LINKED ISSUE #1339: Add a way to convert from an `Interface` to `Option<Interface>`
Some APIs like [`VSSetConstantBuffers`](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Graphics/Direct3D11/struct.ID3D11DeviceContext.html#method.VSSetConstantBuffers) take an `*const Option<ID3D11Buffer>`. Given a `&ID3D11Buffer` I'd like to be able to construct a `&Option<ID3D11Buffer>` to pass to a function like this.

kennykerr:
Adds support for both relative and fixed size Win32 array parameters. 

Here's an example of an API with a fixed-size array parameter:

```rust
let mut buffer = [0; 260];
let a = GetTempFileNameA(".", "test", 0x7b, &mut buffer);
assert_eq!(a, 0x7b);
assert_eq!(&buffer[..12], b".\\tes7B.tmp\0");
```

Previously, the last parameter was merely a pointer and could not check whether the buffer was large enough. Now the size is checked at compile time. 

And here is an example of an API with a relative array parameter:

```rust
let input = b"hello";
let mut output = [0; 5];
let len = MultiByteToWideChar(CP_UTF8, Default::default(), input, &mut output);
assert_eq!(len, 5);
```

Previously, both input and output buffers consisted of a pair of parameters for each pointer and length. Now, a pair of slices is expected and the underlying argument pairs are derived from those.

Fixes #1339 
Fixes #972 
Fixes #790 
Fixes #479


kennykerr:
@damyanp could you look over the DirectX APIs - many changes there - they look good to me, but I'd appreciate another pair of eyes.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
```diff
@@ -48,14 +48,14 @@ pub unsafe fn D3D12CreateVersionedRootSignatureDeserializer(psrcdata: *const ::c
...
+pub unsafe fn D3D12EnableExperimentalFeatures(piids: &[::windows::core::GUID], pconfigurationstructs: &[u8], pconfigurationstructsizes: &[u32]) -> ::windows::core::Result<()> {
     #[cfg(windows)]
     {
         #[link(name = "windows")]
         extern "system" {
             fn D3D12EnableExperimentalFeatures(numfeatures: u32, piids: *const ::windows::core::GUID, pconfigurationstructs: *const ::core::ffi::c_void, pconfigurationstructsizes: *const u32) -> ::windows::core::HRESULT;
         }
-        D3D12EnableExperimentalFeatures(::core::mem::transmute(numfeatures), ::core::mem::transmute(piids), ::core::mem::transmute(pconfigurationstructs), ::core::mem::transmute(pconfigurationstructsizes)).ok()
+        D3D12EnableExperimentalFeatures(pconfigurationstructsizes.len() as _, ::core::mem::transmute(::windows::core::as_ptr_or_null(piids)), ::core::mem::transmute(::windows::core::as_ptr_or_null(pconfigurationstructs)), ::core::mem::transmute(::windows::core::as_ptr_or_null(pconfigurationstructsizes))).ok()
```
damyanp: The three arrays passed in to this all have to be the same length.  As it's also valid to pass in null for the configuration structs, relying on that to get the length could be a problem.  This pIIDs parameter is non-optional and should really be authoritative about the size of the arrays.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
```diff
@@ -18269,17 +18269,29 @@ impl ID3D12CommandQueue {
...
-        (::windows::core::Interface::vtable(self).UpdateTileMappings)(::core::mem::transmute_copy(self), presource.into_param().abi(), ::core::mem::transmute(numresourceregions), ::core::mem::transmute(presourceregionstartcoordinates), ::core::mem::transmute(presourceregionsizes), pheap.into_param().abi(), ::core::mem::transmute(numranges), ::core::mem::transmute(prangeflags), ::core::mem::transmute(pheaprangestartoffsets), ::core::mem::transmute(prangetilecounts), ::core::mem::transmute(flags))
+    pub unsafe fn UpdateTileMappings<'a, Param0: ::windows::core::IntoParam<'a, ID3D12Resource>, Param4: ::windows::core::IntoParam<'a, ID3D12Heap>>(&self, presource: Param0, presourceregionstartcoordinates: &[D3D12_TILED_RESOURCE_COORDINATE], presourceregionsizes: &[D3D12_TILE_REGION_SIZE], pheap: Param4, prangeflags: &[D3D12_TILE_RANGE_FLAGS], pheaprangestartoffsets: &[u32], prangetilecounts: &[u32], flags: D3D12_TILE_MAPPING_FLAGS) {
+        (::windows::core::Interface::vtable(self).UpdateTileMappings)(
+            ::core::mem::transmute_copy(self),
+            presource.into_param().abi(),
+            presourceregionsizes.len() as _,
```
damyanp: presourceregionsizes is allowed to be null.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
```diff
@@ -18269,17 +18269,29 @@ impl ID3D12CommandQueue {
...
+            ::core::mem::transmute(::windows::core::as_ptr_or_null(presourceregionstartcoordinates)),
+            ::core::mem::transmute(::windows::core::as_ptr_or_null(presourceregionsizes)),
+            pheap.into_param().abi(),
+            prangetilecounts.len() as _,
```
damyanp: prangetilecounts is allowed to be null.

kennykerr: `as_ptr_or_null` will pass a nullptr to the API if the slice is empty.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
kennykerr: `as_ptr_or_null` will pass a nullptr to the API if the slice is empty.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
```diff
@@ -19830,9 +19842,9 @@ impl ID3D12Device {
         ::core::mem::transmute((::windows::core::Interface::vtable(self).GetDescriptorHandleIncrementSize)(::core::mem::transmute_copy(self), ::core::mem::transmute(descriptorheaptype)))
     }
     #[doc = "*Required features: 'Win32_Graphics_Direct3D12'*"]
-    pub unsafe fn CreateRootSignature<T: ::windows::core::Interface>(&self, nodemask: u32, pblobwithrootsignature: *const ::core::ffi::c_void, bloblengthinbytes: usize) -> ::windows::core::Result<T> {
+    pub unsafe fn CreateRootSignature<T: ::windows::core::Interface>(&self, nodemask: u32, pblobwithrootsignature: &[u8]) -> ::windows::core::Result<T> {
```
damyanp: This function is usually used in conjunction with ID3DBlob with GetBufferPointer() and GetBufferSize().  Will this interop nicely with the array parameter?

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
kennykerr: Good catch! If there's a shared length parameter, I'll have to look for a "winner" and the non-optional parameter can be the winner by default.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
damyanp: I'm looking more at line 18280 that does `prangetilecounts.len()`

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
kennykerr: I updated your Direct3D sample (in this repo) to use it as follows:

```rust
device.CreateRootSignature(0, std::slice::from_raw_parts(signature.GetBufferPointer() as _, signature.GetBufferSize())) 
```

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
kennykerr: Hmm, I see. They're all optional. How do we pick the "winner" in this case?

kennykerr: Otherwise, I may have to abandon the slice transformation in cases like this where there is ambiguity.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
damyanp: hmm....not beautiful, but could probably be fixed with a simple helper to build a slice from a ID3DBlob directly.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
damyanp: I'm not sure what else you could do, unless you write a bunch of code to verify that the slices are all the same size and ignore the null ones.

crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
kennykerr: I updated the algorithm to detect shared length parameters and to avoid the transformation if there any optional parameters among them. That's probably more conservative than we need to be, but we can tweak it later if it becomes an issue.

kennykerr: This function is no longer transformed.

PR #1564, opened 2022-02-24
Minor doc improvements

kennykerr:


kennykerr:
Build is broken because https://crates.io/ is down.

PR #1569, opened 2022-02-25
Anchor CI to Windows Server 2019 virtual image

riverar:
Quick fix until I file an issue on GitHub for undocumented MSYS2 package regression. This image is still fully supported by GitHub.


kennykerr:
Need to update https://github.com/microsoft/windows-rs/blob/master/crates/tools/yml/src/main.rs

kennykerr (approved):
Thanks!

PR #1570, opened 2022-02-26
Improved Win32 array transformation

### LINKED ISSUE #1567: ID3D11DeviceContext::IAGetVertexBuffers/IASetVertexBuffers does not take array parameters.
Using [feb7d25](https://github.com/microsoft/windows-rs/commit/feb7d258fa8957945eabe1ce18e57795f211f2ee) after #1562 got merged, it seems like 

`ID3D11DeviceContext::IAGetVertexBuffers` still has signature

```rust
pub unsafe fn IAGetVertexBuffers(&self, startslot: u32, numbuffers: u32, ppvertexbuffers: *mut Option<ID3D11Buffer>, pstrides: *mut u32, poffsets: *mut u32);

// Same with IASetVertexBuffer

pub unsafe fn IASetVertexBuffers(&self, startslot: u32, numbuffers: u32, ppvertexbuffers: *const Option<ID3D11Buffer>, pstrides: *const u32, poffsets: *const u32);
```

Therefore code like this does not work

```rust
#[derive(Default)]
struct IABackup<'ctx> {
    vertex_buffers: [Option<ID3D11Buffer>; D3D11_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT as usize],
    vertex_buffer_strides: [u32; D3D11_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT as usize],
    vertex_buffer_offs: [u32; D3D11_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT as usize],
    _pd: PhantomData<&'ctx ()>
}

impl <'ctx> IABackup<'ctx> {
    unsafe fn backup(context: &'ctx ID3D11DeviceContext) {
        let mut backup : IABackup<'ctx> = Default::default();
        context.IAGetVertexBuffers(0, D3D11_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT, &mut backup.vertex_buffers,
                                   &mut backup.vertex_buffer_strides, &mut backup.vertex_buffer_offs);
    }
}
```

Following #1562, I expected signatures like

```rust
pub unsafe fn IAGetVertexBuffers(&self, startslot: u32, ppvertexbuffers: &mut [Option<ID3D11Buffer>], pstrides: &mut [u32], poffsets: &mut [u32]);

pub unsafe fn IASetVertexBuffers(&self, startslot: u32, ppvertexbuffers: &[Option<ID3D11Buffer>], pstrides: &[u32], poffsets: &[u32]);
```

It seems like this is the only associated function affected in `ID3D11DeviceContext` as everything else seems to take slices correctly.

The only issue I see with the above signatures is that the input/output slices must all be of the same size. I'm not sure if there are other cases like this still being projected with pointers, but the array size restrictions can/should be projected idiomatically with min const generics (MSRV 1.51)

```rust
pub unsafe fn IAGetVertexBuffers<const N: usize>(&self, startslot: u32, ppvertexbuffers: &mut [Option<ID3D11Buffer>; N], pstrides: &mut [u32; N], poffsets: &mut [u32; N]);

pub unsafe fn IASetVertexBuffers<const N: usize>(&self, startslot: u32, ppvertexbuffers: &[Option<ID3D11Buffer>; N], pstrides: &[u32; N], poffsets: &[u32; N]);
```

The documentation for [`IASetVertexBuffers`](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-iasetvertexbuffers) doesn't specify what happens if the slice lengths are mismatched so it's highly likely unsound in that scenario. However since the functions are marked unsafe anyways, I don't think it's too big of a deal if adding support for const generics in `bindgen` is too much of an undertaking, or if you don't want to bump the MSRV that high.

kennykerr:
Improves win32 array parsing and fixes a few minor correctness issues with shared array lengths.


kennykerr:
@damyanp this should also address some of your other feedback from #1562

samlh:
I would love to be able to use this signature to be able to avoid the possibility of runtime failures, but I'm not sure if this is good idea because this makes it impossible to call the APIs with dynamic length slices - you can only call the APIs with compile-time fixed length arrays.

If the restriction is a problem, I think the API wrapper may be forced to add run-time checks. I'm not sure how the runtime checks would emit failures - perhaps panic, as failing them would be a programmer error.

Perhaps 2 different signatures could be provided to allow for trading off compile-time checks for runtime flexibility?

kennykerr:
@samlh I wondered about that, but you can still use a slice if you so desire, you just have to use a run time check to confirm that it has the right length. For example, the `GetTempFileNameA` function requires an array with a length of exactly 260 characters. Here's how you can use it with anything that can produce a slice, including dynamic containers like `Vec`.

```rust
fn wrapper(buffer: &mut [u8]) {
    unsafe {
        GetTempFileNameA(".", "test", 0x7b, buffer.try_into().expect("slice length not 260"));
    }
}
```

samlh:
Right, that works if you know _at compile time_ what length you need to cast to.

However, for a function like ID3D12CommandQueue::UpdateTileMappings, you might want to call it with 2 regions on one frame, and 3 regions on another frame, so the caller doesn't have a single compile-time length to cast to.

To put it another way `const PARAM1: usize` isn't necessarily a constant in practice :).

kennykerr:
Got it - that's a good example.

samlh:
One approach to this problem is to make a new type to represent the invariant, with multiple constructors (untested):
```rust
struct SameLenSlices2<T1,T2>(t1: &[T1], t2: &[T2]);

impl<T,U> TryInto<SameLenSlices2<T,U> for (&[T],&[U]) { ... }

impl<const LEN: usize,T,U> Into<SameLenSlices2<T,U> for (&[T;LEN],&[U;LEN]) { ... }

// Repeat for SameLenSlices3, etc

fn UpdateTileMappings(regions: SameLenSlices2<D3D12_TILED_RESOURCE_COORDINATE, D3D12_TILE_REGION_SIZE>, ...)
```

Either way, callers would need to do `(coords,sizes).into()` or `(coords,sizes).try_into().unwrap()` depending on whether the values were arrays or slices.

I'm throwing this out as an idea, but I do think it is likely overkill...

chyyran:
For what its worth, just tested this out and it is working great for the specific use case in #1567 but what @samlh brought up seems like a pretty big dealbreaker.

I wonder if it would be appropriate to just not length check the slices and allowing the length parameter to be passed as part of the call signature while still going through with the `Option<&(mut) [Option<T>]>` transform for the rest of the params.

Not familiar enough with bindgen to know if such a transform will require much more extra code at runtime. If the caller is not careful there is obvious unsoundness here but as its `unsafe` its up to the caller to maintain invariants regardless.

kennykerr:
Thanks for the feedback! I'm just taking a few days off, but I'll likely scale this back to avoid such ambiguities for now. Separately I'm hoping to reduce the number of methods that must be `unsafe` so I'd like to avoid introducing code that makes unverifiable assumptions.

kennykerr:
Ended up scaling back the changes to only include the parsing improvements and very limited code gen changes to improve correctness. It avoids all parameters that share a length and have any optional parameters.

PR #1576, opened 2022-03-02
doc: expose Interface trait and Weak object in documentation

### LINKED ISSUE #1502: Expose cast method in docs and update documentation on Interface trait
I want to use QueryInterface in a safe way to get a derived interface from a base COM interface. From what i've seen, the cast method in the Interface trait is exactly that, and works perfectly. However:
- finding it is not easy as the interface module is hidden from the documentations (but the cast method itself is not doc(hidden), which seems to indicate it is intended to be exposed).
- The Interface trait is documented with this sentence:
```
/// This trait is automatically used by the generated bindings and should not used directly.
```
which seems to indicate that this method should not be used.

AFAICT this cast method is intended to be the documented, official way to use QueryInterface (and if it isn't, it should be :) ). So the doc should probably be updated:
- removing the doc hidden from the interface module in core
- removing this "should not be used directly" comment from the trait documentation, and moving it to all the doc(hidden) methods of the trait.

I can do a PR for this if you are OK with those changes.

vthib:
The Interface trait exposes two public functions that are quite useful
when using COM objects.

The Weak struct is also returned by the downgrade function, so expose it
as well.

Closes #1502

### DIFF
--- a/crates/libs/windows/src/core/interface.rs
+++ b/crates/libs/windows/src/core/interface.rs
@@ -3,8 +3,9 @@ use bindings::*;
 
 /// Provides low-level access to a COM interface.
 ///
-/// This trait is automatically used by the generated bindings and should not be
-/// used directly.
+/// This trait is automatically implemented by the generated bindings and should not be
+/// implemented manually.
+///
 /// # Safety
 pub unsafe trait Interface: Sized {
     #[doc(hidden)]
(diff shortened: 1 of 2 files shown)


kennykerr:
Did you test this locally?

```
cargo doc -p windows --open
```

vthib:
Yes I did, it looks fine. The build issues are afaict not linked to my PR

kennykerr:
OK, will investigate.

kennykerr:
Fixed in #1580 - then I'll rerun this build.

kennykerr:
@vthib this issue has now been fixed but I can't seem to update your branch to take the fix - if you can merge `master` we should be good to go.

vthib:
Thanks for the fix, I rebased the PR

kennykerr (approved):
Thanks!

PR #1590, opened 2022-03-09
Provide hex output for `Debug` impl of `HRESULT` and `NTSTATUS`

### LINKED ISSUE #1588: Feature request: show hex code for NTSTATUS
### Motivation

Printing an `NTSTATUS` with `{:?}` shows `NTSTATUS(-1073741772)`.

The [official doc](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/596a1078-e883-4972-9bbc-49e60bebca55) for NTSTATUS lists these codes as hex, this would save me a conversion. This would be good that NTSTATUS would implement Debug, and that would show them as hex (and possibly also as decimal, since googling them is also useful sometimes?)

I haven't checked, but maybe the same could be done for HRESULT and win32 error codes.

I'm relatively new to the windows-rs crate, I hope these suggestions make sense.
Thanks

### Drawbacks

_No response_

_No response_

### Additional context

_No response_

kennykerr:
This should now be consistent with most documented `HRESULT` and `NTSTATUS` values. 

Fixes #1588

### DIFF
--- a/crates/libs/bindgen/src/replacements/ntstatus.rs
+++ b/crates/libs/bindgen/src/replacements/ntstatus.rs
@@ -55,7 +55,7 @@ pub fn gen() -> TokenStream {
         impl ::core::cmp::Eq for NTSTATUS {}
         impl ::core::fmt::Debug for NTSTATUS {
             fn fmt(&self, f: &mut ::core::fmt::Formatter<'_>) -> ::core::fmt::Result {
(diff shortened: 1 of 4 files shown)


kennykerr:
Blocked on #1591

eholk (approved):
Looks good and seems useful!

PR #1595, opened 2022-03-11
Allow for non-HRESULT return types in COM

rylev:
This allows for return types in COM interfaces that are not `HRESULT`


kennykerr (approved):
A comm

kennykerr (dismissed):

PR #1598, opened 2022-03-14
Double quote features in documentation for easier consumption

### LINKED ISSUE #1597: Feature request: In the documentation, present required features with double quotes instead of single quotes
### Motivation

It would be more convenient if the documentation showed the required features using double quotes. So for example, instead of:

‘Win32_Storage_FileSystem’, ‘Win32_Foundation’

Do this:

"Win32_Storage_FileSystem", "Win32_Foundation"

Then you could just simply copy and paste into cargo.toml without having to do any editing.

### Drawbacks

Only if it would somehow break with some sort of norm in how rust documentation is normally presented

Doing this only provides convenience, nothing more.

### Additional context

None

riverar:
Double quotes feature names for easier cut/paste into Cargo.toml.

Closes #1597.

### DIFF
--- a/crates/libs/bindgen/src/gen.rs
+++ b/crates/libs/bindgen/src/gen.rs
@@ -54,14 +54,14 @@ impl Gen<'_> {
         if !self.doc {
             quote! {}
         } else {
-            let mut tokens = format!("'{}'", to_feature(self.namespace));
+            let mut tokens = format!(r#"\"{}\""#, to_feature(self.namespace));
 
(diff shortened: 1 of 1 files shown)


kennykerr:
I tried to do this before. The trouble is that the `"` ends up as `“` instead and you still end up having to replace those characters.

riverar:
Yep, this isn't fixable without a rustdoc change or tons of html/css hackery which isn't worth it. How unfortunate.

PR #1602, opened 2022-03-15
Add license files to crate packages

### LINKED ISSUE #1587: Add a LICENSE file
Hi, would it be possible to add a LICENSE file(s) to the repository? IIUC from the various crates' manifests, e.g. for [windows-sys](https://github.com/microsoft/windows-rs/blob/master/crates/libs/sys/Cargo.toml#L7), the crates are uniformly licensed under "MIT OR Apache-2.0". It would be great to have either a project-wide LICENSE file and/or crate-specific ones to make that more clear.

kennykerr:
I have to copy them to the crate dirs but at least my build tools will ensure they are in sync.

Fixes #1587

### DIFF
--- a/crates/libs/sys/license-apache
+++ b/crates/libs/sys/license-apache
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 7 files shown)


riverar (approved):
Looks good, do we need to also handle dependencies like bindgen, metadata, et al.?

riverar (dismissed):

kennykerr:
> Looks good, do we need to also handle dependencies like bindgen, metadata, et al.?

The `windows` and `windows-sys` crates are the main dependencies. We can add it to the others as there is need/demand.

PR #1604, opened 2022-03-15
Update Windows metadata, import libraries

riverar:
Updated Windows.Win32 metadata to 18.0.1-preview, Windows metadata to 10.0.22000.196, and regenerated umbrella import libs.


crates/libs/sys/src/Windows/Win32/Graphics/Printing/mod.rs
```diff
@@ -3877,7 +3877,7 @@ pub struct JOB_INFO_2A {
     pub pDriverName: ::windows_sys::core::PSTR,
     pub pDevMode: *mut super::Gdi::DEVMODEA,
     pub pStatus: ::windows_sys::core::PSTR,
-    pub pSecurityDescriptor: *mut super::super::Security::SECURITY_DESCRIPTOR,
+    pub pSecurityDescriptor: *mut ::core::ffi::c_void,
```
kennykerr, line 3880: I wonder what happened here...

crates/libs/sys/src/Windows/Win32/Graphics/Printing/mod.rs
```diff
@@ -3913,7 +3913,7 @@ pub struct JOB_INFO_2W {
     pub pDriverName: ::windows_sys::core::PWSTR,
     pub pDevMode: *mut super::Gdi::DEVMODEW,
     pub pStatus: ::windows_sys::core::PWSTR,
-    pub pSecurityDescriptor: *mut super::super::Security::SECURITY_DESCRIPTOR,
+    pub pSecurityDescriptor: *mut ::core::ffi::c_void,
```
kennykerr, line 3916: Same here.

crates/libs/sys/src/Windows/Win32/Graphics/Printing/mod.rs
```diff
@@ -3962,7 +3962,7 @@ pub struct JOB_INFO_4A {
     pub pDriverName: ::windows_sys::core::PSTR,
     pub pDevMode: *mut super::Gdi::DEVMODEA,
     pub pStatus: ::windows_sys::core::PSTR,
-    pub pSecurityDescriptor: *mut super::super::Security::SECURITY_DESCRIPTOR,
+    pub pSecurityDescriptor: *mut ::core::ffi::c_void,
```
kennykerr, line 3965: Ditto.

crates/libs/sys/src/Windows/Win32/NetworkManagement/WindowsFilteringPlatform/mod.rs
```diff
@@ -1,9 +1,9 @@
 #![allow(non_snake_case, non_camel_case_types, non_upper_case_globals, clashing_extern_declarations, clippy::all)]
 #[link(name = "windows")]
 extern "system" {
-    #[doc = "*Required features: `\"Win32_NetworkManagement_WindowsFilteringPlatform\"`, `\"Win32_Foundation\"`, `\"Win32_Security\"`*"]
-    #[cfg(all(feature = "Win32_Foundation", feature = "Win32_Security"))]
-    pub fn FwpmCalloutAdd0(enginehandle: super::super::Foundation::HANDLE, callout: *const FWPM_CALLOUT0, sd: *const super::super::Security::SECURITY_DESCRIPTOR, id: *mut u32) -> u32;
+    #[doc = "*Required features: `\"Win32_NetworkManagement_WindowsFilteringPlatform\"`, `\"Win32_Foundation\"`*"]
+    #[cfg(feature = "Win32_Foundation")]
+    pub fn FwpmCalloutAdd0(enginehandle: super::super::Foundation::HANDLE, callout: *const FWPM_CALLOUT0, sd: *const ::core::ffi::c_void, id: *mut u32) -> u32;
```
kennykerr, line 6: It's almost as if SECURITY_DESCRIPTOR just vanished. Is it in the latest win32 winmd?

crates/libs/sys/src/Windows/Win32/Networking/HttpServer/mod.rs
```diff
@@ -1043,7 +1040,7 @@ pub const HTTP_REQUEST_AUTH_FLAG_TOKEN_FOR_CACHED_CRED: u32 = 1u32;
 #[cfg(feature = "Win32_Foundation")]
 pub struct HTTP_REQUEST_AUTH_INFO {
     pub AuthStatus: HTTP_AUTH_STATUS,
-    pub SecStatus: i32,
+    pub SecStatus: ::windows_sys::core::HRESULT,
```
kennykerr, line 1043: SECURITY_STATUS is always an HRESULT?

kennykerr: It seems to be, at least in some compilation modes, just a raw status code.

crates/tests/win32/tests/win32.rs
```diff
@@ -126,28 +126,31 @@ fn com() -> windows::core::Result<()> {
     unsafe {
         let stream = CreateStreamOnHGlobal(0, true)?;
         let values = vec![1, 20, 300, 4000];
-        let copied = stream.Write(values.as_ptr() as _, (values.len() * core::mem::size_of::<i32>()) as u32)?;
 
+        let mut copied = 0;
+        let hr = stream.Write(values.as_ptr() as _, (values.len() * core::mem::size_of::<i32>()) as u32, &mut copied);
+        assert!(hr.is_ok());
```
kennykerr, line 132: Generally if a function previously returned `Result` and now returns `HRESULT` you can just turn the `HRESULT` back into a `Result` using `ok` so `stream.Write(...).ok()?` and the error propagation will work as before.

crates/libs/sys/src/Windows/Win32/Networking/HttpServer/mod.rs
riverar: Additional context @ https://github.com/microsoft/win32metadata/issues/603

It appears `SECURITY_STATUS` grew up from its 16-bit Windows days and is now just a HRESULT. Metadata already mapped SEC_xx_xx constants to HRESULTs; I simply finished the job, so to speak, by making the two types equivalent so we get richer code gen. I spot checked a few API impls (ncrypt, sspi, etc.), particularly `AcceptSecurityContext`, `SslOpenPrivateKey`, and `NCryptGetProperty`, and they all appear to return HRESULTs.

If you're seeing otherwise, holler and we'll get that change reverted.

crates/tests/win32/tests/win32.rs
riverar: Ah great, will clean that up. Thanks!

crates/libs/sys/src/Windows/Win32/Graphics/Printing/mod.rs
riverar: Agree. Looks like it reverted to `void*` in metadata, investigating.

riverar: Oh, now I remember why this sounded familiar. I made this change!

This was an incorrect mapping in metadata. In this specific case, `JOB_INFO_2` looks like so:

```cpp
typedef struct _JOB_INFO_2 {
  ...
  LPDEVMODE            pDevMode;
  LPTSTR               pStatus;
  PSECURITY_DESCRIPTOR pSecurityDescriptor;
  ...
```

`PSECURITY_DESCRIPTOR` [is not a pointer to a SECURITY_DESCRIPTOR](https://devblogs.microsoft.com/oldnewthing/20151223-00/?p=92701). It's `typedef PVOID PSECURITY_DESCRIPTOR` behind the scenes.

So this is correct.

kennykerr (approved):

crates/libs/sys/src/Windows/Win32/Graphics/Printing/mod.rs
kennykerr: https://github.com/microsoft/win32metadata/issues/835

qarmin:
Removal of SqlLite was intentional?

PR #1608, opened 2022-03-16
Support richer COM interface hierarchies

rylev:
As part of #1486, this allows for COM interfaces that inherit from something other than `IUnknown`. 

This also adds a test with a simple but rich COM hierarchy.

### DIFF
--- a/crates/libs/interface/src/lib.rs
+++ b/crates/libs/interface/src/lib.rs
@@ -127,7 +127,7 @@ impl Interface {
     }
 
     fn get_com_trait(&self) -> proc_macro2::TokenStream {
-        let name = quote::format_ident!("{}Impl", self.name);
+        let name = quote::format_ident!("{}_Impl", self.name);
         let vis = &self.visibility;
         let methods = self
             .methods
@@ -170,7 +170,7 @@ impl Interface {
                 }
             })
             .collect::<Vec<_>>();
-        let trait_name = quote::format_ident!("{}Impl", name);
+        let trait_name = quote::format_ident!("{}_Impl", name);
         let functions = self
             .methods
             .iter()
@@ -295,7 +295,7 @@ impl Interface {
         if i == "IUnknown" {
             return quote!();
         }
-        let i = quote::format_ident!("{}Impl", i);
+        let i = quote::format_ident!("{}_Impl", i);
         quote!(#i +)
     }
 }
--- a/crates/libs/interface/src/lib.rs
+++ b/crates/libs/interface/src/lib.rs
@@ -292,7 +292,7 @@ impl Interface {
     /// Gets the parent trait constrait which is nothing if the parent is IUnknown
     fn parent_trait_constraint(&self) -> proc_macro2::TokenStream {
         let i = self.parent_ident();
-        if i.clone().to_string() == "IUnknown" {
+        if i == "IUnknown" {
             return quote!();
         }
         let i = quote::format_ident!("{}Impl", i);
--- a/crates/libs/interface/src/lib.rs
+++ b/crates/libs/interface/src/lib.rs
@@ -56,7 +56,7 @@ macro_rules! expected_token {
 struct Interface {
     visibility: syn::Visibility,
     name: syn::Ident,
-    parent: Option<syn::Path>,
+    parent: syn::Path,
     methods: Vec<InterfaceMethod>,
     docs: Vec<syn::Attribute>,
 }
@@ -127,7 +127,7 @@ impl Interface {
     }
 
     fn get_com_trait(&self) -> proc_macro2::TokenStream {
-        let name = quote::format_ident!("{}_Impl", self.name);
+        let name = quote::format_ident!("{}Impl", self.name);
         let vis = &self.visibility;
         let methods = self
             .methods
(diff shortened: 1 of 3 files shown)


crates/libs/interface/src/lib.rs
```diff
@@ -67,8 +67,8 @@ impl Interface {
         let vis = &self.visibility;
         let name = &self.name;
         let docs = &self.docs;
-        let parent = self.parent.as_ref().map(|p| quote!(#p)).unwrap_or_else(|| quote!(::windows::core::IUnknown));
-        let vtable_name = quote::format_ident!("{}_Vtbl", name);
+        let parent = self.parent();
+        let vtable_name = quote::format_ident!("{}Vtbl", name);
```
kennykerr, line 71: This may not support the pre-generated metadata that uses `_Vtbl`.

crates/libs/interface/src/lib.rs
```diff
@@ -127,7 +127,7 @@ impl Interface {
     }
 
     fn get_com_trait(&self) -> proc_macro2::TokenStream {
-        let name = quote::format_ident!("{}_Impl", self.name);
+        let name = quote::format_ident!("{}Impl", self.name);
```
kennykerr, line 130: This may not support the pre-generated metadata that uses `_Impl`.

kennykerr:
`IPeristMemory` is a simple existing interface that inherits from `IPersist` so should make a great test for the `interface` macro, much like the existing `IUri` test. Here's a simple example of implementing and exercising it. You could then add a declaration using the `interface` macro and test the matrix of implementations as before. That should ensure we have consistency between the interface macro and existing metadata-based interfaces. 

```rust
use std::ffi::*;
use std::sync::*;
use windows::{core::*, Win32::Foundation::*, Win32::System::Com::*};

#[implement(IPersistMemory, IPersist)]
#[derive(Default)]
struct Persist(RwLock<PersistState>);

#[derive(Default)]
struct PersistState {
    memory: [u8; 10],
    dirty: bool,
}

impl IPersist_Impl for Persist {
    fn GetClassID(&self) -> Result<GUID> {
        Ok("117fb826-2155-483a-b50d-bc99a2c7cca3".into())
    }
}

impl IPersistMemory_Impl for Persist {
    fn IsDirty(&self) -> Result<()> {
        let reader = self.0.read().unwrap();
        if reader.dirty {
            Ok(())
        } else {
            Err(S_FALSE.into())
        }
    }
    fn Load(&self, input: *const c_void, size: u32) -> Result<()> {
        unsafe {
            let mut writer = self.0.write().unwrap();
            if size <= writer.memory.len() as _ {
                std::ptr::copy(input, writer.memory.as_mut_ptr() as _, size as _);
                writer.dirty = true;
                Ok(())
            } else {
                Err(E_OUTOFMEMORY.into())
            }
        }
    }
    fn Save(&self, output: *mut c_void, clear_dirty: BOOL, size: u32) -> Result<()> {
        unsafe {
            let mut writer = self.0.write().unwrap();
            if size <= writer.memory.len() as _ {
                std::ptr::copy(writer.memory.as_mut_ptr() as _, output, size as _);
                if clear_dirty.as_bool() {
                    writer.dirty = false;
                }
                Ok(())
            } else {
                Err(E_OUTOFMEMORY.into())
            }
        }
    }
    fn GetSizeMax(&self) -> Result<u32> {
        let reader = self.0.read().unwrap();
        Ok(reader.memory.len() as _)
    }
    fn InitNew(&self) -> Result<()> {
        let mut writer = self.0.write().unwrap();
        writer.memory = Default::default();
        writer.dirty = false;
        Ok(())
    }
}

fn main() -> Result<()> {
    unsafe {
        let p: IPersistMemory = Persist::default().into();
        assert_eq!(
            p.GetClassID()?,
            "117fb826-2155-483a-b50d-bc99a2c7cca3".into()
        );
        // TODO: can't test IsDirty until this is fixed: https://github.com/microsoft/win32metadata/issues/838
        assert_eq!(p.GetSizeMax()?, 10);
        p.Load(&[0xAA, 0xBB, 0xCC])?;
        let mut memory = [0x00, 0x00, 0x00, 0x00];
        p.Save(&mut memory, true)?;
        assert_eq!(memory, [0xAA, 0xBB, 0xCC, 0x00]);
        Ok(())
    }
}
```

kennykerr:
Perhaps something like this:

```rust
#[interface("0000010c-0000-0000-C000-000000000046")]
pub unsafe trait ICustomPersist : IUnknown {
    unsafe fn GetClassID(&self, clsid: *mut GUID) -> HRESULT;
}

#[interface("BD1AE5E0-A6AE-11CE-BD37-504200C10000")]
pub unsafe trait ICustomPersistMemory : ICustomPersist {
    unsafe fn IsDirty(&self) -> HRESULT;
    unsafe fn Load(&self, input: *const c_void, size: u32) -> HRESULT;
    unsafe fn Save(&self, output: *mut c_void, clear_dirty: BOOL, size: u32) -> HRESULT;
    unsafe fn GetSizeMax(&self, len: *mut u32) -> HRESULT;
    unsafe fn InitNew(&self) -> HRESULT;
}
```

kennykerr (approved):

PR #1611, opened 2022-03-17
Fix bug where COM interfaces needed to be declared pub

rylev:
Fixes another issue found in https://github.com/microsoft/windows-rs/issues/1486 where it was always necessary to mark an interface as `pub`.


kennykerr (approved):
Thanks!

Issue #1618, opened 2022-03-18
`interface` macro should implement `IntoParam`


cherryleafroad:
Some methods have signatures like
```rust
windows::Windows::Win32::System::Wmi::IUnsecuredApartment
pub unsafe fn CreateObjectStub<'a, Param0>(&self, pobject: Param0) -> ::windows::core::Result<::windows::core::IUnknown>
where
    Param0: ::windows::core::IntoParam<'a, ::windows::core::IUnknown>,
```
Where the current custom interfaces defined using `#[interface]` cannot be sent as a parameter to this since they don't seem to implement `IntoParam`. It would be convenient if they implemented this so they can be easily passed to functions that require it.

Workaround isn't super difficult, something like this works
`pUnsecApp.CreateObjectStub::<IUnknown>(event_sink.into())`

kennykerr:
Yes that's missing. Thanks for the reminder. @rylev is still working on that macro.

kennykerr:
Correction, this seems to work. Here's a complete repro:

```
[dependencies.windows]
version = "0.34"
features = [
    "implement",
    "interface",
    "Win32_System_WinRT",
]
```

```rust
use windows::{core::*, Win32::System::WinRT::*};

#[interface("bdbec94c-ae60-49e3-894f-009b1ce9af71")]
pub unsafe trait ISample: IUnknown {
    unsafe fn method(&self);
}

#[implement(ISample)]
struct Sample();

impl ISample_Impl for Sample {
    unsafe fn method(&self) {}
}

fn main() {
    let sample: ISample = Sample().into();
    unsafe {
        let _ = RoGetAgileReference(AGILEREFERENCE_DEFAULT, &IUnknown::IID, sample);
    }
}
```

PR #1620, opened 2022-03-18
Make test checking for conversion to param

rylev:
@cherryleafroad @kennykerr perhaps I'm not understanding #1618, but I believe that the desired `impl` is already present and this PR adds a test that confirms that.


cherryleafroad:
I'm using 0.34.0. Maybe it has changed?

rylev:
Yes, passing the implementation (in this case `Persist`) would be incorrect since it is a backing implementation and not itself an interface.

kennykerr:
Ah yes, I forgot that was there already. Thanks for adding the test! 😜

kennykerr (approved):

PR #1648, opened 2022-03-31
Type-specific handle validity checking

### LINKED ISSUE #1630: Bug: FindVolumeHandle::is_invalid() does not check for a value of -1

windows

### Crate version

0.34.0

### Summary

``FindVolumeHandle::is_invalid()`` only considers the value ``0`` to represent an invalid handle which leads to incorrect branches being followed if a function returns a handle value of ``-1`` as its invalid handle code.

Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\Ian\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc
nightly-2022-01-27-x86_64-pc-windows-msvc
nightly-x86_64-pc-windows-msvc (default)

installed targets for active toolchain
--------------------------------------

i686-pc-windows-msvc
x86_64-pc-windows-msvc

active toolchain
----------------

nightly-x86_64-pc-windows-msvc (default)
rustc 1.61.0-nightly (d3ad51b48 2022-02-25)

### Reproducible example

```rust
use windows::Win32::Storage::FileSystem::{FindFirstVolumeW, FindVolumeClose};

fn main() {
    unsafe {
        // will return `-1` due to empty output buffer.
        let handle = FindFirstVolumeW(&mut []);
        if !handle.is_invalid() {
            // heap corruption due to handle being `-1`
            FindVolumeClose(handle);
        }
    }
}
```

### Crate manifest

```toml
[package]
name = "find_volume_bug"
version = "0.1.0"
edition = "2021"

[target.'cfg(windows)'.dependencies]
windows = { version = "0.34.0", features = ["Win32_Storage_FileSystem", "Win32_Foundation"] }
```

### Expected behavior

Body of ``if`` block is goes unentered because the handle is invalid and the program exits gracefully.

### Actual behavior

Body of ``if`` block is gets entered because the ``is_invalid`` fails to catch the value ``-1`` and the program exits due to heap corruption.

### Additional comments

_No response_

kennykerr:
The win32 metadata now includes information about what constitutes invalid handle values. This update plumbs this through so that you can accurately check whether a handle returned by some function is invalid. Handles will have an `is_invalid` method if their underlying type is a pointer or have at least one invalid value.

With this information in hand, the `bindgen` crate can now combine this with metadata that indicates whether a particular function set the last error so that the caller should use `GetLastError` on failure. This means that functions that return a handle and provide error information in this way will now be transformed to return `-> Result<HANDLE>` instead of just `-> HANDLE`. This should make error handling more predictable. 

Fixes #1630

### DIFF
--- a/crates/tests/handles/tests/win.rs
+++ b/crates/tests/handles/tests/win.rs
@@ -22,7 +22,7 @@ fn handle() {
 fn psid() {
     let underlying: *mut std::ffi::c_void = std::ptr::null_mut();
     let handle: PSID = PSID(underlying);
-    assert!(handle.is_null());
+    assert!(handle.is_invalid());
 
     let copy = handle;
     assert!(copy == handle);
@@ -31,7 +31,7 @@ fn psid() {
     assert!(clone == handle);
 
     let default = PSID::default();
-    assert!(default.is_null());
+    assert!(default.is_invalid());
 
     assert_eq!(format!("{:?}", handle), "PSID(0x0)");
 }


crates/libs/bindgen/src/handles.rs
```diff
@@ -20,26 +20,42 @@ pub fn gen_sys_handle(def: &TypeDef, gen: &Gen) -> TokenStream {
...
     let ident = gen_ident(def.name());
-    let signature = gen_signature(def, gen);
+    let underlying_type = underlying_type(def);
+    let signature = gen_default_type(&underlying_type, gen);
+    let check = if underlying_type.is_pointer() {
+        quote! {
+            impl #ident {
+                pub fn is_null(&self) -> bool {
+                    self.0.is_null()
+                }
```
kennykerr, line 30: Types like `PSID` will get an `is_null` method instead of `is_invalid` when the underlying type is a pointer rather than an integer.

crates/libs/bindgen/src/replacements/mod.rs
```diff
@@ -1,15 +1,13 @@
...
 mod bstr;
-mod handle;
 mod ntstatus;
 
 pub fn gen(def: &TypeDef) -> Option<TokenStream> {
     match def.type_name() {
         TypeName::BOOL => Some(bool32::gen()),
         TypeName::BSTR => Some(bstr::gen()),
         TypeName::NTSTATUS => Some(ntstatus::gen()),
-        TypeName::HANDLE => Some(handle::gen()),
```
kennykerr, line 12: `HANDLE` no longer needs to specialized! 😉

crates/libs/bindgen/src/functions.rs
```diff
@@ -202,28 +202,78 @@ fn gen_win_function(def: &MethodDef, gen: &Gen) -> TokenStream {
...
+                    #[inline]
+                    pub unsafe fn #name<#constraints>(#params) -> ::windows::core::Result<#return_type> {
+                        #[cfg(windows)]
+                        {
+                            #link_attr
+                            extern "system" {
+                                fn #name(#(#abi_params),*) -> #return_type;
+                            }
+                            let result__ = #name(#args);
+                            (!result__.is_invalid()).then(||result__).ok_or_else(::windows::core::Error::from_win32)
```
kennykerr, line 269: Here we check if the returned handle is valid and call `from_win32` if it is invalid to return an error `Result`.

riverar (approved):
Nice! Really cool to see all this light up with a sprinkle of metadata 🎉

PR #1658, opened 2022-04-04
Support explicit activatable class libs

riverar:
Some projects, like the Windows App SDK, make use of activatable classes with IDs that are not resolvable to their host libraries using simple probing semantics. (That is, chopping off the fully-qualified type name suffix in a loop and probing libraries along the way.)

This PR adds a method to the `FactoryCache` to bypass activation factory probing and instead use an explicitly provided library. (It also adds the code generator plumbing needed to tap into this new method.) The idea here is that upstream crates would figure out the activatable class<->library mapping themselves (e.g. by parsing a side-by-side manifest) and feed that in to the code generator.

This unblocks adopting Windows App SDK 1.1 xcopy deployment and could potentially speed up activation for others.


crates/libs/bindgen/src/classes.rs
```diff
@@ -51,14 +52,20 @@ fn gen_class(def: &TypeDef, gen: &Gen) -> TokenStream {
                     quote! {}
                 };
 
+                let factory_cache_method = if let Some(library) = gen.class_map.get(&class_name.as_str()) {
+                    format_token!(r#"new_with_library("{library}")"#)
+                } else {
+                    format_token!("new()")
+                };
+
```
kennykerr, line 60: Maybe hoist this out of this scope so we can share the code gen and avoid the redundancy below.

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -6,13 +6,18 @@ use core::sync::atomic::{AtomicPtr, Ordering};
 #[doc(hidden)]
 pub struct FactoryCache<C, I> {
     shared: AtomicPtr<core::ffi::c_void>,
+    library: Option<&'static str>,
```
kennykerr, line 9: Maybe just `&'static str` and let an empty string mean the absence of a library to avoid having to check two things at load time.

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -6,13 +6,18 @@ use core::sync::atomic::{AtomicPtr, Ordering};
...
     _i: PhantomData<I>,
 }
 
 impl<C, I> FactoryCache<C, I> {
     pub const fn new() -> Self {
-        Self { shared: AtomicPtr::new(::core::ptr::null_mut()), _c: PhantomData, _i: PhantomData }
+        Self { shared: AtomicPtr::new(::core::ptr::null_mut()), library: None, _c: PhantomData, _i: PhantomData }
+    }
+
+    pub const fn new_with_library(library: &'static str) -> Self {
```
kennykerr, line 19: Perhaps `from_library` is more idiomatic.

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -44,11 +49,20 @@ impl<C: RuntimeName, I: Interface> FactoryCache<C, I> {
     }
 }
 
-/// Attempts to load the factory interface for the given WinRT class.
-pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
+/// Attempts to load the factory interface for the given WinRT class, and library
+/// if provided.
+pub fn factory<C: RuntimeName, I: Interface>(library: Option<&str>) -> Result<I> {
```
kennykerr, line 54: Perhaps split this into two functions: `factory` unchanged with no params and then `factory_from_library` that takes a library.

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -118,6 +128,18 @@ pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
     }
 }
 
+unsafe fn get_activation_factory(library: &[u8], name: &HSTRING) -> Result<*mut std::ffi::c_void> {
```
kennykerr, line 131: This can just be merged into the `factory_from_library` function suggested above.

crates/samples/consent/src/main.rs
```diff
@@ -2,7 +2,7 @@ use windows::{core::*, Foundation::*, Security::Credentials::UI::*, Win32::Found
 
 fn main() -> Result<()> {
     unsafe {
-        let interop = factory::<UserConsentVerifier, IUserConsentVerifierInterop>()?;
+        let interop = factory::<UserConsentVerifier, IUserConsentVerifierInterop>(None)?;
```
kennykerr, line 5: Remove the `None` as that's not going to be norm so we don't want folks to have to `None` all the things. But would be good to add a test for `from_library` if there's a system DLL that would work that way...

crates/libs/windows/src/core/factory_cache.rs
riverar: This is used by both `factory` and `factory_from_library` so left it in place.

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -118,6 +119,29 @@ pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
...
+    let function = delay_load(library, b"DllGetActivationFactory\0");
+    let function: DllGetActivationFactory = core::mem::transmute(function);
+    let mut abi = core::ptr::null_mut();
+    let hr = function(core::mem::transmute_copy(name), &mut abi);
+    if hr.is_err() {
+        Err(hr.into())
+    } else {
+        Ok(abi)
+    }
+}
```
kennykerr, line 143:
```suggestion
unsafe fn get_activation_factory(library: &[u8], name: &HSTRING) -> Result<IActivationFactory> {
    let function = delay_load(library, b"DllGetActivationFactory\0");
    let function: DllGetActivationFactory = core::mem::transmute(function);
    let mut abi = core::ptr::null_mut();
    function(core::mem::transmute_copy(name), &mut abi).from_abi(abi)
}
```

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -118,6 +119,29 @@ pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
...
 }
 
+/// Attempts to load the factory interface for the given WinRT class located
+/// in the specified library.
+pub fn factory_from_library<C: RuntimeName, I: Interface>(library: &str) -> Result<I> {
+    let name = HSTRING::from(C::NAME);
+    unsafe {
+        let abi = get_activation_factory(library.as_bytes(), &name)?;
+        let factory: IActivationFactory = core::mem::transmute(abi);
+        factory.cast()
```
kennykerr, line 129:
```suggestion
        get_activation_factory(library.as_bytes(), &name)?.cast()
```

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -98,18 +103,14 @@ pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
             library[..path.len()].copy_from_slice(path.as_bytes());
             library[path.len()..].copy_from_slice(b".dll\0");
 
-            let function = delay_load(library, b"DllGetActivationFactory\0");
-
             if !function.is_null() {
-                let function: DllGetActivationFactory = core::mem::transmute(function);
-                let mut abi = core::ptr::null_mut();
-                let _ = function(core::mem::transmute_copy(&name), &mut abi);
+                let result = get_activation_factory(library, &name);
```
kennykerr, line 107:
```suggestion
                if let Ok(factory) = get_activation_factory(library, &name) {
                    return factory.cast();
                }
```

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -118,6 +119,29 @@ pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
     }
 }
 
+/// Attempts to load the factory interface for the given WinRT class located
+/// in the specified library.
+pub fn factory_from_library<C: RuntimeName, I: Interface>(library: &str) -> Result<I> {
+    let name = HSTRING::from(C::NAME);
+    unsafe {
+        let abi = get_activation_factory(library.as_bytes(), &name)?;
```
rylev, line 127: Doesn't the library name have to be null terminated? Rust strings are not null terminated.

kennykerr: Yes, the underlying `delay_load` function assumes the byte array is null terminated. The `factory_cache` (in master) ensures that the library name is terminated.

crates/libs/windows/src/core/factory_cache.rs
kennykerr: Note also that this code will never run since the `function` points to the wrong function.

kennykerr: Actually, that's not entirely true by the looks of it but that `is_null` check is still suspect.

crates/tests/nightly_component_client/build.rs
```diff
@@ -7,12 +8,25 @@ fn main() -> std::io::Result<()> {
...
+    bindings.write_all(gen_namespace(&gen).as_bytes())?;
+    bindings.write_all(gen_namespace_impl(&gen).as_bytes())?;
+    drop(bindings);
+
+    Command::new("rustfmt").arg("src/bindings/test_nightly_component.rs").status()?;
+
+    let mut class_map = HashMap::new();
+    let _ = class_map.insert("Microsoft.Windows.System.Power.PowerManager".to_string(), "test_nightly_component.dll".to_string());
+    let gen = Gen { namespace: "Microsoft.Windows.System.Power", component: true, class_map, ..Default::default() };
+    let mut bindings = File::create("src/bindings/microsoft_windows_system_power.rs")?;
```
kennykerr, line 25: Not for this PR, but all of this redundant and boilerplate code should be wrapped in a simpler "gen" library or tool that the builds scripts can run with a single command.

kennykerr (approved):

PR #1665, opened 2022-04-06
Avoid clearing `SetLastError` when allocating string params

### LINKED ISSUE #1664: Bug: 0.35.0 clears GetLastError on somewhere

windows

### Crate version

0.35.0

### Summary

I called QueryDosDeviceW with `&str` and `Vec<u16>`.
with 0.34.0, I can get LastError (with `Error::from_win32`) if the function returns 0, but with 0.35.0 returns No error only.

efault host: x86_64-pc-windows-msvc
rustup home:  D:\data\depot\rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc
nightly-x86_64-pc-windows-msvc

installed targets for active toolchain
--------------------------------------

aarch64-apple-darwin
aarch64-apple-ios
aarch64-unknown-linux-gnu
nvptx64-nvidia-cuda
wasm32-unknown-unknown
wasm32-wasi
x86_64-apple-darwin
x86_64-pc-windows-gnu
x86_64-pc-windows-msvc
x86_64-unknown-linux-gnu
x86_64-unknown-linux-musl

active toolchain
----------------

stable-x86_64-pc-windows-msvc (default)
rustc 1.59.0 (9d1b2106e 2022-02-23)

### Reproducible example

```rust
use windows::{
    Win32::Storage::FileSystem::QueryDosDeviceW,
    core::Error,
};

fn main() {
    let mut buf: Vec<u16> = vec![0; 20 as usize];
    let out_chars = unsafe {
        QueryDosDeviceW("C:", &mut buf)
    };
    let err = Error::from_win32();
    println!("out-chars: {}", out_chars);
    if out_chars == 0 {
        println!("error: {:?}", err);
    } else {
        buf.truncate(out_chars as usize);
        println!("out: {:?}", String::from_utf16_lossy(&buf));
    }
}
```

### Crate manifest

```toml
[dependencies.windows]
version = "0.35.0"
features = [
    "alloc",
    "Win32_Storage_FileSystem"
]
```

### Expected behavior

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target\debug\windows-rs-0-35-bug.exe`
out-chars: 0
error: Error { code: 0x8007007A, message: The data area passed to a system call is too small.
 }

Process finished with exit code 0
```

### Actual behavior

```
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target\debug\windows-rs-0-35-bug.exe`
out-chars: 0
error: Error { code: 0x00000000, message: The operation completed successfully.
 }

Process finished with exit code 0
```

### Additional comments

(update) I believe it is caused by BSTR handling.

Windows Version (if curious): Windows 11 21H2 (Build 22000.593)

kennykerr:
Switched to using the low-level `HeapAlloc` allocator rather than `SysAllocXxx` as the former doesn't clear the thread's last error.

Also added a test to ensure the behavior is persevered. 

Fixes #1664


crates/libs/windows/src/core/heap.rs
```diff
@@ -18,3 +18,12 @@ pub unsafe fn heap_free(ptr: RawPtr) {
         HeapFree(heap, HEAP_NONE, ptr);
     }
 }
+
+pub fn heap_string<T: Copy + Default + Sized>(value: &[T]) -> *const T {
+    let buffer = heap_alloc((value.len() + 1) * std::mem::size_of::<T>()).expect("Could not allocate string") as *mut T;
```
rylev, line 23: Nit: Since the `T` here is uninitialized this should probably be `*mut MaybeUninit<T>`. It's probably fine to leave for now though 🤷

kennykerr: I really only want to use this function for `u8` and `u16`. Be nice if Rust had traits for stuff like that. C++ for example has [the is_unsigned trait](https://en.cppreference.com/w/cpp/types/is_unsigned).

Lokathor: Because the pointer points to uninit memory, the next line (where you make the slice) is UB.

kennykerr: Feel free to suggest a better implementation or shoot me a PR.

rylev: Pull request opened here: https://github.com/microsoft/windows-rs/pull/1667

Lokathor: Kenny that's an unfortunate way to respond to UB reports. I encourage you to consider them a serious bug even if the person who points it out doesn't provide a solution at the same time. There's several ways to deal with uninit data in Rust, and `rylev` had already mentioned one of them for you just above.

kennykerr: I wasn't trying to minimize the issue. I just wasn't sure how to address it. Sorry for the confusion.

PR #1666, opened 2022-04-06
Add test component

kennykerr:
This adds a WinRT test component (DLL) written in Rust. This ensures we test activation fallback logic.


crates/tests/nightly_component/build.rs
```diff
@@ -0,0 +1,23 @@
...
+
+fn main() {
+    println!("cargo:rerun-if-changed=src/component.idl");
+    let metadata_dir = format!("{}\\System32\\WinMetadata", env!("windir"));
+    std::fs::create_dir_all(".windows/winmd").unwrap();
+    let status = Command::new("midlrt.exe").arg("/winrt").arg("/nomidl").arg("/h").arg("nul").arg("/metadata_dir").arg(&metadata_dir).arg("/reference").arg(format!("{}\\Windows.Foundation.winmd", metadata_dir)).arg("/winmd").arg(".windows/winmd/component.winmd").arg("src/component.idl").status().unwrap();
+
+    if !status.success() {
+        panic!();
+    }
```
rylev, line 14:
```suggestion
    assert!(status.success(), "running midlrt.exe did not succeed"); 
```

crates/tests/nightly_component/build.rs
```diff
@@ -0,0 +1,23 @@
...
+
+    let gen = Gen { namespace: "test_nightly_component", component: true, ..Default::default() };
+    let mut bindings = File::create("src/bindings.rs").unwrap();
+    bindings.write_all(gen_namespace(&gen).as_bytes()).unwrap();
+    bindings.write_all(gen_namespace_impl(&gen).as_bytes()).unwrap();
+    drop(bindings);
+
+    Command::new("rustfmt").arg("src/bindings.rs").status().unwrap();
```
rylev, line 22: I think most of these unwraps are probably fine (though getting in the habit of at least using `.expect` would be good). However, considering we've had issues with rustfmt in the past, it might really be helpful for debugging to use `.expect` here instead of `.unwrap`

crates/tests/nightly_component/src/lib.rs
```diff
@@ -0,0 +1,37 @@
+mod bindings;
+use std::mem::*;
+use std::sync::*;
+use windows::core::*;
+use windows::Win32::Foundation::*;
+
+#[implement(bindings::Class)]
+struct Class(RwLock<i32>);
```
rylev, line 8: Using [`AtomicI32`](https://doc.rust-lang.org/stable/std/sync/atomic/struct.AtomicI32.html) instead of `RwLock<i32>` is preferable.

crates/tests/nightly_component/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+    fn SetProperty(&self, value: i32) -> Result<()> {
+        let mut writer = self.0.write().unwrap();
+        *writer = value;
+        Ok(())
+    }
+}
+
+#[implement(IActivationFactory)]
+struct ClassFactory();
```
rylev, line 23: nit:

```suggestion
struct ClassFactory;
```

crates/tests/nightly_component_client/build.rs
```diff
@@ -0,0 +1,23 @@
+use std::fs::*;
+use std::io::prelude::*;
+use std::process::*;
+use windows_bindgen::*;
+
+fn main() {
+    std::fs::create_dir_all(".windows/winmd").unwrap();
+
+    // I would have expected Cargo to wait until the component crate (which is a dependency) is built
```
rylev, line 9: I believe if you mark the component as a build dependency (and not just a regular dependency) this won't happen.

rylev (approved):
Looks good in general! Just a few small comments 😊

crates/tests/nightly_component_client/build.rs
kennykerr: Ah - that's a relief! 🤣

rylev:
This unfortunately puts a dependency on `midlrt.exe` to build windows-rs now. Two concerns:
* This would be nice if it was documented somewhere
* I only have access from `midlrt.exe` from the VS command prompt, but I normally build from powershell. How can I make sure that `midlrt.exe` is available from powershell?

kennykerr:
Yes, I don't like this either. It was quite a slog getting this to work at all. We should probably document some of the pain points around getting this to work and see what we can improve (both with the `midlrt` dependency issues and the `cargo` target dependency issues).

Regarding `midlrt`, the use of this tool is temporary until I'm finished with my `winmd` generator at which point that will simply be yanked out. If there's a way to make this test not run by default with `cargo test --all`, or some other solution, that would be great.

rylev:
I think we can find a solution. For example, we can feature gate it and then run the tests with that feature in CI.

PR #1667, opened 2022-04-07
Remove UB in the heap_string function

rylev:
After some [discussion](https://github.com/microsoft/windows-rs/pull/1665/files/d6d8d56e405200ded57f69415a05ce1324184b81#r844373812) it was concluded that there is UB in the implementation of `heap_string`. This removes that UB.

In particular, `std::slice::from_raw_parts_mut` requires that the pointer points to initialized member which we do not get from `heap_alloc`. The solution to initializing uninitialized memory is to use `std::ptr::write` which does not read or drop values it writes to (and is thus documented specifically as being appropriate for writing to uninitialized memory). 

cc @lokathor

### DIFF
--- a/crates/libs/windows/src/core/heap.rs
+++ b/crates/libs/windows/src/core/heap.rs
@@ -22,13 +22,17 @@ pub unsafe fn heap_free(ptr: RawPtr) {
 /// Copy a slice of `T` into a freshly allocated buffer with an additional default `T` at the end.
 ///
 /// Returns a pointer to the beginning of the buffer
+///
+/// # Panics
+///
+/// This function panics if the heap allocation fails or if the pointer returned from
+/// the heap allocation is not properly aligned to `T`.
 pub fn heap_string<T: Copy + Default + Sized>(slice: &[T]) -> *const T {
     unsafe {
         let buffer = heap_alloc((slice.len() + 1) * std::mem::size_of::<T>()).expect("could not allocate string") as *mut T;
-        for offset in 0..slice.len() {
-            std::ptr::write(buffer.add(offset), slice[offset]);
-        }
-        std::ptr::write(buffer.add(slice.len()), T::default());
+        assert!(buffer.align_offset(std::mem::align_of::<T>()) == 0, "heap allocated buffer is not properly aligned");
+        buffer.copy_from_nonoverlapping(slice.as_ptr(), slice.len());
+        buffer.add(slice.len()).write(T::default());
         buffer
     }
 }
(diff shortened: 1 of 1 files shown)


crates/libs/windows/src/core/heap.rs
```diff
@@ -19,11 +19,16 @@ pub unsafe fn heap_free(ptr: RawPtr) {
...
-    buffer
+/// Copy a slice of `T` into a freshly allocated buffer with an additional default `T` at the end.
+///
+/// Returns a pointer to the beginning of the buffer
+pub fn heap_string<T: Copy + Default + Sized>(slice: &[T]) -> *const T {
+    unsafe {
+        let buffer = heap_alloc((slice.len() + 1) * std::mem::size_of::<T>()).expect("could not allocate string") as *mut T;
+        for offset in 0..slice.len() {
+            std::ptr::write(buffer.add(offset), slice[offset]);
+        }
```
kennykerr, line 30: This feels like it would be slower than just calling `memcpy` but hopefully the optimizer can figure that out (LLVM usually does).

Lokathor: Instead of this loop you probably want `std::ptr::copy_nonoverlapping`, which will lower to a correct memcpy usage.

```rust
std::ptr::copy_nonoverlapping(slice.as_ptr(), buffer, slice.len());
```

rylev:
@kennykerr @Lokathor thanks for the tip on using `copy_nonoverlapping`. I'm using the method `copy_from_nonoverlapping` instead. I also added a check to make sure the allocation is properly aligned since it's also UB if the source or destination are not aligned. This might incur a very small runtime cost for now but `align_offset` should become const soon and then this check will be optimized away.

Lokathor:
You could probably lower that assert to a debug_assert, since if the allocator gives you a bad pointer a million other places will all have similar problems anyway.

Though now that you mention it, checking for a non-null buffer pointer is probably a good idea. The allocation cam fail, and in that case the pointer would be null.

rylev:
@Lokathor We already check for null inside of `heap_alloc` so `heap_alloc` is guaranteed to return a non-null value. It's also [guaranteed to return a pointer that is aligned to either 8 or 16 byte offsets](https://docs.microsoft.com/en-us/windows/win32/api/heapapi/nf-heapapi-heapalloc#remarks). The issue is that `T` is not guaranteed to fall within that alignment (it could be 32 byte aligned for example). It would be preferable if we could put some constraint on `T` such that it was guaranteed that `T` is properly aligned on 8 bit or 16 byte offsets, but I don't believe that's possible.

kennykerr:
This is fine for now - we may eventually replace the generic function with a pair of non-generic functions since this almost seems like more trouble than it's worth.

rylev:
@kennykerr moving to two functions would only really allow removing the trait constraints on the argument type and the alignment check which might not be worth the duplicated code, but we can see 😊

Issue #1675, opened 2022-04-09
Move implementation details into `imp` module


kennykerr:
Currently the `windows::core` module contains a blend of supporting types, some of which are meant to be used directly by developers and some are implementation details. I'm thinking of moving the implementation types into a `windows::imp` module, both to make it clearer what is part of the core library and is off limits and also to more easily distinguish between those parts that should be documented publicly and those that should be hidden. It also avoid name collisions between implementation types and developer code. The `imp` name is inspired by crates like `proc_macro2` that use the name for their internal types.

kennykerr:
Haven't found a good moment to do this and its not pressing enough to leave open. Can revisit in future.

PR #1678, opened 2022-04-11
Replace internal `IActivationFactory` definition with tailored definition

### LINKED ISSUE #1662: Get rid of hand-rolled `IActivationFactory`
The metadata bugs have been fixed so we should be able to rely on the generated version now.

kennykerr:
The `IActivationFactory` interface is implemented by all WinRT class factories and acts as the gateway for DLL-based activation. This is a curious interface whose `ActivateInstance` method isn't very practical for language projections to use since they almost always need to immediately cast the result to something other than what this method returns. As such, the definition used by the `windows` crate is different to the one provided by the generated bindings. In order to avoid name collisions, the version used internally by the factory cache is renamed to avoid this conflict. #1675 should also help to avoid such collisions in a more general way in future but is a bigger change that I'll tackle separately. 

This PR also updates the test component to illustrate how components can implement `IActivationFactory` directly using the bindings generated by the `bindgen` crate rather than depending on the internal definition used by the `windows` crate. 

Fixes #1662


riverar (approved):

crates/libs/windows/src/core/generic_factory.rs
```diff
@@ -0,0 +1,28 @@
+use super::*;
+use std::mem::*;
+
+// A streamlined version of the IActivationFactory interface used by WinRT class factories used internally by the windows crate
```
rylev, line 4: Nit: this should be a `///` doc comment so that it's picked up by tooling

crates/libs/windows/src/core/generic_factory.rs
```diff
@@ -0,0 +1,28 @@
...
+// to simplify code generation. Components should implement the `IActivationFactory` interface published by the windows crate.
+#[repr(transparent)]
+#[derive(Clone, PartialEq, Eq)]
+pub struct IGenericFactory(IUnknown);
+
+impl IGenericFactory {
+    pub fn ActivateInstance<I: Interface>(&self) -> Result<I> {
```
rylev, line 11: I know this is an internal API but some documentation would be helpful. Even if it just points out that this is the `ActivateInstance` from `IActivationFactory` (and then link [here](https://docs.microsoft.com/en-us/windows/win32/api/activation/nf-activation-iactivationfactory-activateinstance))

rylev (commented):
Just a few nits.

One thing to note: the actual change here is pretty small, but the diff is big because so many bindings needed to change. It would be *very* helpful to separate out implementation work in one commit from changes to bindings in another.

crates/libs/windows/src/core/generic_factory.rs
kennykerr: How does this work when the docs are hidden for this type?

rylev: `rust-analyzer` shows hidden docs so those working on Windows (who would be the ones who would benefit from this doc), would be able to see this on hover.

kennykerr: Ah that makes sense. I haven't used rust-analyzer in many months as it seemed to really struggle with the size of the repo.

kennykerr: Anyway, I'll try to make a habit of using `///` for this purpose.

rylev: Yes, I noticed rust-analyzer was really struggling... I was able to get it to work fine by editing the workspace definition and removing some things temporarily. This at least allowed me to use `rust-analyzer` from within `windows::core`.

PR #1693, opened 2022-04-14
Added conversions between HSTRING and OsStr/OsString

kaivol:
I think these conversions are all quite straightforward and self-explanatory.

I'm not sure what exactly is supposed to be hidden behind the `alloc` feature, and whether I have applied this correctly.

### DIFF
--- a/crates/libs/windows/src/core/hstring.rs
+++ b/crates/libs/windows/src/core/hstring.rs
@@ -265,6 +265,78 @@ impl PartialEq<&HSTRING> for alloc::string::String {
     }
 }
 
+impl PartialEq<std::ffi::OsString> for HSTRING {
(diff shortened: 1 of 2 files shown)


kennykerr (approved):
Thanks for the contribution! 

The `alloc` feature unlocks implicit allocations via the `IntoParam` trait. Your addition means that `OsStr` and `OsString` values can be used to call APIs that expect an `HSTRING` as an input parameter. This addresses part of #760. 

I'm a little concerned about the explosion of trait implementations needed to add this support. We may want to consider adding a `macro_rules` at some point to get this under control especially when it comes to adding similar support for other types.

kennykerr:
[0.36.0](https://github.com/microsoft/windows-rs/releases/tag/0.36.0) has now been released and includes your contribution - thanks again!

PR #1703, opened 2022-04-19
Remove explicit factory support

kennykerr:
This was a failed attempt to simplify WinAppSDK deployment, which unfortunately requires a great deal of infrastructure specific to Visual Studio.


riverar (approved):
😢

Issue #1094, opened 2021-08-26
Simplify WinRT and COM class authoring


kennykerr:
Dedicating an issue to this topic. Originally part of #81 (huge thread).

The windows crate now supports implementing COM and WinRT interfaces, but more is required to support classes as as whole. This early sample illustrates what's possible today:

https://github.com/kennykerr/component-rs

I still have much work to do to streamline this experience.

Alovchin91:
@kennykerr Is it possible today to author WinRT components in Rust, either using windows-rs crate or not? 🤔

I suppose it would require the use of Midl compiler + writing COM wiring manually. Could you maybe provide a sample of what can be done today?

Thanks!

kennykerr:
You can do so today but yes you'd need to write some IDL and call the MIDL compiler to produce a winmd. Next year I'm hoping to spend most of my time on improving authoring support and being able to produce a winmd directly from Rust so that IDL will not be required.

Alovchin91:
Thank you for the quick response! 😊 

I don't mind writing IDL and calling Midl manually. Having an ability to author WinRT components would unblock Rust's use for me 🙂

Do I understand correctly that I would need to call the `windows::build!` macro on the resulting WinMD, and then `#[implement]` interfaces from it? 🤔

kennykerr:
Yep, that's it. It's a bit error-prone which is also why I'm planning on automating much of that.

Alovchin91:
Yeah, I can imagine 😅  I'm gonna try this out and if I succeed, I could open a PR to the Samples repo, if you would be interested 🙂

kennykerr:
Sounds good!

Alovchin91:
Hi @kennykerr , you might be interested to take a look:

https://github.com/Alovchin91/winrt-component-rs

Please let me know if you have any suggestions, comments etc. 🙂

I'm not entirely sure about `*mut Option` in `DllGetActivationFactory` signature but that's the only way I managed to make it work.

kennykerr:
Thanks @Alovchin91, that's very helpful. I'm now starting to work on this and #1093 in earnest. We should get to a point where MIDL is no longer required.

hmyan90:
[...]

PR #1705, opened 2022-04-19
Add `Event<T>` type that can be used to implement a WinRT event

kennykerr:
The implementation is based on [the event class](https://github.com/microsoft/cppwinrt/blob/master/strings/base_events.h) in [cppwinrt](https://github.com/microsoft/cppwinrt). The notable difference is that Rust lacks something akin to C++ variadic templates so we can't just create an agile delegate wrapper and must instead use an explicit enum. Beyond that, the implementation is largely the same to simplify review and validation. The Rust `Event<T>` implementation is thus thread-safe and largely equivalent to C++/WinRT and WRL in terms of behavior and performance. 

A part of #1094.


crates/libs/windows/src/core/agile_reference.rs
```diff
@@ -9,11 +9,9 @@ pub struct AgileReference<T>(IAgileReference, PhantomData<T>);
...
 impl<T: Interface> AgileReference<T> {
     /// Creates an agile reference to the object.
-    pub fn new<'a>(object: &'a T) -> Result<Self>
-    where
-        &'a T: IntoParam<'a, IUnknown>,
-    {
-        unsafe { RoGetAgileReference(AGILEREFERENCE_DEFAULT, &T::IID, object).map(|reference| Self(reference, Default::default())) }
+    pub fn new(object: &T) -> Result<Self> {
+        let unknown: &IUnknown = unsafe { std::mem::transmute(object) };
+        unsafe { RoGetAgileReference(AGILEREFERENCE_DEFAULT, &T::IID, unknown).map(|reference| Self(reference, Default::default())) }
```
kennykerr, line 14: This just removes the redundant constraint on the `AgileReference` constructor. The `Interface` trait doesn't "know" that it requires `IUnknown`. That's something I'll try to fix separately.

Scottj1s (approved):

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+        Self::new()
+    }
+}
+
+impl<T: Interface + Clone> Event<T> {
+    /// Creates a new, empty `Event<T>`.
+    pub fn new() -> Self {
+        Self { delegates: Array::new(), swap: Mutex::default(), change: Mutex::default() }
+    }
+    /// Registers a delegate with the event object.
```
rylev, line 26: Nit: it would be nice to have a space between methods. That's not required by rustfmt, but is the more typical way things are formatted.

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+    pub fn add(&mut self, delegate: &T) -> Result<i64> {
+        let mut _lock_free_drop = Array::new();
+        Ok({
+            let change_lock = self.change.lock().unwrap();
```
rylev, line 30: I'm surprised you don't get a warning here for an unused variable. I think this should be `_change_lock`. Similiarly, I think `_lock_free_drop` shouldn't have the `_` prefix since this convention is used for *unused* bindings not bindings that are of secondary importance like here.

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+            let mut new_delegates = Array::with_capacity(self.delegates.len() + 1)?;
+            for delegate in self.delegates.as_slice() {
+                new_delegates.push(delegate.clone());
```
rylev, line 33: Why do the delegates have to be cloned instead of the new delegate just being pushed on to the existing delegates array?

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+        {
+            let change_lock = self.change.lock().unwrap();
+            if self.delegates.is_empty() {
+                return Ok(());
+            }
+            let mut capacity = self.delegates.len() - 1;
+            let mut new_delegates = Array::new();
+            let mut removed = false;
+            if capacity == 0 {
+                if self.delegates.as_slice()[0].to_token() == token {
```
rylev, line 56: nit: we might want to just add [indexing](https://doc.rust-lang.org/std/ops/trait.Index.html) directly to the `Array` type

rylev, line 56:
```rust
removed = self.delegates.as_slice()[0].to_token() == token;
```

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+                }
+            } else {
+                new_delegates = Array::with_capacity(capacity)?;
+                for delegate in self.delegates.as_slice() {
+                    if !removed && delegate.to_token() == token {
+                        removed = true;
+                        continue;
+                    }
+                    if capacity == 0 {
+                        debug_assert!(!removed);
```
rylev, line 67: Nit: I would add a string message to the `debug_assert` so it's a bit more straight forward to debug when that condition is met.

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
+use super::*;
+use bindings::*;
+use std::sync::*;
+
+/// A type that you can use to declare and implement an event of a specified delegate type.
+///
+/// The implementation is thread-safe and designed to avoid contention between events being
+/// raised and delegates being added or removed.
+pub struct Event<T: Interface + Clone> {
+    swap: Mutex<()>,
```
rylev, line 10: I understand the need for the `swap` and the `change` lock being different (to avoid contention), but why do you need the `swap` mutex instead of wrapping `delegates` array itself in a `RWLock`?

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+impl Buffer {
+    /// Creates a new `Buffer` with the specified size in bytes.
+    fn new(size: usize) -> Result<*mut Buffer> {
+        if size == 0 {
+            Ok(std::ptr::null_mut())
+        } else {
+            let alloc_size = std::mem::size_of::<Buffer>() + size;
+            let header = heap_alloc(alloc_size)? as *mut Buffer;
+            unsafe {
+                (*header).0 = RefCount::new(1);
```
rylev, line 205: You should use `header.write(Buffer(RefCount::new(1))` since the `Buffer` will be uninitialized. You should always use `std::ptr::write` for writing to uninitialized memory.

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+            }
+            Ok(header)
+        }
+    }
+    /// Returns a raw pointer to the buffer's contents.
+    fn as_ptr<T>(&self) -> *const T {
```
rylev, line 211: I think it might be useful to note that `T` might be uninitialized.

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+struct Buffer(RefCount);
+
+impl Buffer {
+    /// Creates a new `Buffer` with the specified size in bytes.
+    fn new(size: usize) -> Result<*mut Buffer> {
+        if size == 0 {
+            Ok(std::ptr::null_mut())
+        } else {
+            let alloc_size = std::mem::size_of::<Buffer>() + size;
+            let header = heap_alloc(alloc_size)? as *mut Buffer;
```
rylev, line 203: The other code makes assumptions that this pointer is properly aligned. Since `Buffer` is an `AtomicI32` can we assume that the pointer will always be properly aligned?

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+        self.len == 0
+    }
+    /// Returns the number of delegates in the array.
+    fn len(&self) -> usize {
+        self.len
+    }
+    /// Appends a delegate to the back of the array.
+    fn push(&mut self, delegate: Delegate<T>) {
+        unsafe {
+            std::ptr::write((*self.buffer).as_mut_ptr::<Delegate<T>>().add(self.len) as _, delegate);
```
rylev, line 150: What happens if we push beyond the capacity of the buffer?

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+        }
+    }
+}
+
+impl<T: Interface + Clone> Clone for Array<T> {
+    fn clone(&self) -> Self {
+        if !self.is_empty() {
+            unsafe { (*self.buffer).0.add_ref() };
+        }
+        Self { buffer: self.buffer, len: self.len, _phantom: std::marker::PhantomData }
```
rylev, line 177: If I clone this `Array` I then have two `Array<T>` which I can call `push` on at the same time. However, if they are pointing to the same buffer, I might be overwriting their contents. This would seem to violate memory safety, no?

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+                    self.remove(delegate.to_token())?;
+                }
+            }
+        }
+        Ok(())
+    }
+}
+
+/// A thread-safe reference-counted array of delegates.
+struct Array<T: Interface + Clone> {
```
rylev, line 112: Nit: typically, the bounds on a generic param are only given where they're needed. So, here we would not constrain `T` but we would keep the constraint where it's really needed (on the inherit `impl` and the `Drop` impl).

crates/libs/windows/src/core/event.rs
```diff
@@ -0,0 +1,254 @@
...
+enum Delegate<T: Interface + Clone> {
+    Direct(T),
+    Indirect(AgileReference<T>),
+}
+
+impl<T: Interface + Clone> Delegate<T> {
+    /// Creates a new `Delegate<T>`, containing a suitable reference to the specified delegate.
+    fn new(delegate: &T) -> Self {
+        if delegate.cast::<IAgileObject>().is_err() {
+            if let Ok(delegate) = AgileReference::new(delegate) {
```
rylev, line 232: If `delegate` does not implement `IAgileObject` but we can't create an `AgileReference` shouldn't we error?

rylev (commented):
I have some questions, but overall, I think this has good bones to it 😊

crates/libs/windows/src/core/event.rs
kennykerr: > I think _lock_free_drop shouldn't have the _ prefix

If I remove it, compiler complains with `value assigned to `lock_free_drop` is never read`.

crates/libs/windows/src/core/event.rs
kennykerr: To avoid disrupting a concurrent read of the delegates, a new array needs to be created every time. The tension is between a thread adding/removing a delegate and another thread firing the event and thus cycling through the existing array. The vast majority of event sources have either zero or one handler, and this was determined to be the most efficient implementation.

crates/libs/windows/src/core/event.rs
kennykerr: Because both locks govern the same memory and their lifetimes are overlapping. The one handles changes and the other replacements. I don't think a `RwLock` can easily model this.

crates/libs/windows/src/core/event.rs
kennykerr: This is a back door that C++ uses to handle Xaml's internal delegates which refuse to support agility or marshaling, but I agree it's a hack and Rust doesn't need to support it.

crates/libs/windows/src/core/event.rs
kennykerr: I tried that but Rust complains:

```
    Checking windows v0.37.0 (D:\git\windows-rs\crates\libs\windows)
error[E0367]: `Drop` impl requires `T: interface::Interface` but the struct it is implemented for does not
   --> crates\libs\windows\src\core\event.rs:190:9
    |
190 | impl<T: Interface + Clone> Drop for Array<T> {
    |         ^^^^^^^^^
    |
note: the implementor must specify the same requirement
   --> crates\libs\windows\src\core\event.rs:114:1
    |
114 | / struct Array<T> {
115 | |     buffer: *mut Buffer,
116 | |     len: usize,
117 | |     _phantom: std::marker::PhantomData<T>,
118 | | }
    | |_^

error[E0367]: `Drop` impl requires `T: Clone` but the struct it is implemented for does not
   --> crates\libs\windows\src\core\event.rs:190:21
    |
190 | impl<T: Interface + Clone> Drop for Array<T> {
    |                     ^^^^^
    |
note: the implementor must specify the same requirement
   --> crates\libs\windows\src\core\event.rs:114:1
    |
114 | / struct Array<T> {
115 | |     buffer: *mut Buffer,
116 | |     len: usize,
117 | |     _phantom: std::marker::PhantomData<T>,
118 | | }
    | |_^

For more information about this error, try `rustc --explain E0367`.
error: could not compile `windows` due to 2 previous errors
```

crates/libs/windows/src/core/event.rs
kennykerr: This is just bumping the ref count on the same array. Basically, there are now two smart pointers to the same array to support lock-free calling. The `swap`/`change` locks ensures we don't touch the array being used for calls.

crates/libs/windows/src/core/event.rs
kennykerr: Bad things.

crates/libs/windows/src/core/event.rs
kennykerr: Yes, `HeapAlloc` returns memory is aligned to 8 bytes on x86 and 16 bytes on x64.

Issue #1699, opened 2022-04-18
[Q&A] Is it possible to target `edition=2018` instead?


Byron:
When integrating `gitoxide` with other packages, some of them set a minimum supported rust version (MSRV), which may also constrain the maximum supported edition. However, when integrating with `vergen` in particular this now fails only for the reason of requiring [a higher edition in `windows` than is supported](https://github.com/Byron/vergen/runs/6058258829?check_suite_focus=true#step:12:19) by their MSRV.

Currently the `windows` crates require `edition = 2021` and I wonder if it's technically possible to switch them to edition 2018, maybe because no features that would require edition 2021 are actually used.

If edition 2021 is indeed required, this issue can be closed. Otherwise I think there is value in supporting a lower edition for greater compatibility.

_Note that `gitoxide` was setting higher editions and the latest compiler features for no other reason except for 'novelty', but learned why that would be undesirable when integrating with `vergen`. That's the only reason I post this issue here, there might be a chance :)._

kennykerr:
So yes, I think it might be possible to use 2018 instead. I switched to 2021 because it simplified a few things and because [the docs indicate that interop is seamless](https://doc.rust-lang.org/stable/edition-guide/editions/index.html). Having switched, I'm not sure what impact it would have to switch back... I'm guessing it will be just as seamless. 😄

riverar:
I think the real issue here is that the pipeline in question is using a version of rustc that falls below MSRV. https://github.com/Byron/vergen/runs/6058258829?check_suite_focus=true#step:10:31

Think once that's fixed on the vergen side, everything will be good to go. Will close this as there are no changes to be made to the windows crate, but feel free to re-open if I missed something.

Byron:
Thanks for your responses, let me summarize what I understand.

1. it would indeed be possible to switch back to edition 2018, even though there are concerns about downstream breakage.
2. everyone is expected to use the latest compiler so that markers for editions are understood. Using an older Rust compiler to verify the MSRV like `vergen` does is a bug to be fixed.

Regarding 1) and the concern of breakage when switching to 2018, it's my experience that switching back to older editions is never a problem as backwards compatibility is always maintained. From my experience, breakage when doing that is not a concern. From my interactions with the cargo team I myself trust in this assumption.

As I am not an authority in the matter at all let me share some information I gathered while researching point 2 to work up towards my conclusion.

[...]

PR #1706, opened 2022-04-19
Undo Rust 2021 adoption

kennykerr:
[Adopting Rust 2021](https://github.com/microsoft/windows-rs/pull/1636) had some unintended consequences. Turns outs it's [not quite as seamless](https://github.com/microsoft/windows-rs/issues/1699#issuecomment-1101667920) as [is claimed](https://doc.rust-lang.org/stable/edition-guide/editions/index.html). 

This update just reverts to 2018 so that we can continue to support `windows-sys` on 1.46 rather than forcing it artificially to 1.56 just to satisfy Cargo.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
     runs-on: windows-2019
     strategy:
       matrix:
-        rust: [1.56.0, stable, nightly]
+        rust: [1.46.0, stable, nightly]
     steps:
     - name: Checkout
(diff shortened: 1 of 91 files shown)


rylev:
Is the minimum support Rust version (MSRV) documented anywhere?

kennykerr:
The `rust-version` is set in in Cargo.toml (and verified by the CI build). Be nice if crates.io displayed that info under "Metadata".

PR #1709, opened 2022-04-21
Remove redundant warning suppression

### LINKED ISSUE #1707: Remove redundant warning suppression
Generated modules include redundant warning suppression.

kennykerr:
This will need some more work for WinRT components, but I think this is generally an improvement over blindly suppressing warnings all over the place. I'm just shooting for clean builds with warnings allowed centrally, but this opens the door for code gen improvements as it's much easier to selectively disallow warnings and improve code gen as needed. 

Fixes #1707


.github/workflows/build.yml
```diff
@@ -70,7 +70,7 @@ jobs:
     - name: Update toolchain
       run: rustup update --no-self-update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
     - name: Run cargo check
-      run: cargo check -p windows
+      run: cargo check -p windows --features Foundation,Win32_Foundation,Win32_Graphics_Direct2D
```
kennykerr, line 73: Unfortunately, we can't test `--all-features` as we do for `windows-sys` due to build server limitations (and a very greedy compiler) but at least this adds a bit more smoke testing for some common type system patterns that we can validate on the MSRV.

PR #1717, opened 2022-04-26
Version 0.36.0

kennykerr:


kennykerr:
Looks like something may have regressed in the nightly Rust compiler that has broken both open PRs (#1713 & #1717). 

Unfortunately, these tests are only supported on nightly at the moment so we're stuck. 😬

PR #1722, opened 2022-04-27
Make target crates `no_std`

ChrisDenton:
This allows them to build if the standard library is not present.


kennykerr (approved):
Thanks!

PR #1724, opened 2022-04-28
Correct `compose` offset in `implement` macro

kennykerr:
This affects only Xaml - I have updated the Xaml sample to trigger the offset calculation. 

Once authoring support is further along (#1094), we should add a composition test to the `test_component` to validate this as well.


riverar (approved):
Thanks!

PR #1729, opened 2022-05-03
Correct implementation identity for MBM and .NET compatibility

### LINKED ISSUE #1730: Implementations must return GetRuntimeClassName to satisfy MBM and .NET
Registration-free metadata based marshaling (MBM) always queries for `IInspectable`, so the implementation of `GetRuntimeClassName` from this `IInspectable` implementation must return the class or interface type best describing the implementation. This change also ensures that the .NET interop support will be able to correctly resolve the type of a given Rust implementation. 

The identity is assumed to be the first interface or class declared by the `implement` macro. Previously, it returned an empty string if you explicitly query for `IInspectable` and then call `GetRuntimeClassName` since it doesn't know which interface you're specifically asking about. 

Original repro is here:

https://github.com/kennykerr/component-rs

kennykerr:
Registration-free metadata based marshaling (MBM) always queries for `IInspectable`, so the implementation of `GetRuntimeClassName` from this `IInspectable` implementation must return the class or interface type best describing the implementation. This change also ensures that the .NET interop support will be able to correctly resolve the type of a given Rust implementation. 

The identity is assumed to be the first interface or class declared by the `implement` macro. Previously, it returned an empty string if you explicitly query for `IInspectable` and then call `GetRuntimeClassName` since it doesn't know which interface you're specifically asking about. 

Original repro is here: https://github.com/kennykerr/component-rs 

Fixes #1730

### DIFF
--- a/crates/libs/bindgen/src/helpers.rs
+++ b/crates/libs/bindgen/src/helpers.rs
@@ -310,15 +310,11 @@ pub fn gen_runtime_name(def: &TypeDef, cfg: &Cfg, gen: &Gen) -> TokenStream {
                 const NAME: &'static str = #runtime_name;
             }
         }
-    } else if def.vtable_types().iter().any(|e| e == &Type::IInspectable) {
+    } else {
         quote! {
             #cfg
-            impl ::windows::core::RuntimeName for #name {
-                const NAME: &'static str = "";
-            }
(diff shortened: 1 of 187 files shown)


kennykerr:
This will need a bit more work.

PR #1734, opened 2022-05-04
Improve `IUnknown` Unsafe Annotations

rylev:
This PR does two things:
* Makes the `QueryInterface` and `Release` methods on `IUknownImpl` unsafe. 
  *  These methods require ensuring certain variants for them to be called safely. For example, `Release` can free memory so the user might be sure that once they call `Release` the no longer use `self`. 
* Makes `QueryInterface`, `AddRef`, and `Release` take non-exclusive references since they are effectively methods on aliased references and thus cannot be `&mut self`. 
  * As a side effect I needed to change `QueryInterface`'s out param to be `*mut *const c_void` instead of `*mut *mut c_void`. I believe this more closely models what an interface out param is.  

cc @eholk who has expressed interest in helping review safety related PRs

### DIFF
--- a/crates/libs/implement/src/lib.rs
+++ b/crates/libs/implement/src/lib.rs
@@ -124,14 +124,14 @@ pub fn implement(attributes: proc_macro::TokenStream, original_type: proc_macro:
                     }
                 }
             }
-            fn AddRef(&mut self) -> u32 {
+            fn AddRef(&self) -> u32 {
                 self.count.add_ref()
             }
-            unsafe fn Release(&mut self) -> u32 {
+            unsafe fn Release(&self) -> u32 {
                 let remaining = self.count.release();
                 if remaining == 0 {
                     unsafe {
-                        ::std::boxed::Box::from_raw(self);
+                        ::std::boxed::Box::from_raw(self as *const Self as *mut Self);
                     }
                 }
                 remaining
--- a/crates/libs/implement/src/lib.rs
+++ b/crates/libs/implement/src/lib.rs
@@ -42,7 +42,7 @@ pub fn implement(attributes: proc_macro::TokenStream, original_type: proc_macro:
         let offset: TokenStream = format!("{}", count).into();
         quote! {
             else if #vtbl_ident::matches(iid) {
-                &mut self.vtables.#offset as *mut _ as _
+                &self.vtables.#offset as *const _ as *const _
             }
         }
     });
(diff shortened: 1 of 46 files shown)


kennykerr (approved):

rylev:
I'm just going to check with @eholk before merging.

eholk:
> I'm just going to check with @eholk before merging.

Looks good to me!

crates/libs/implement/src/lib.rs
```diff
@@ -124,14 +124,14 @@ pub fn implement(attributes: proc_macro::TokenStream, original_type: proc_macro:
                     }
                 }
             }
-            fn AddRef(&mut self) -> u32 {
+            fn AddRef(&self) -> u32 {
                 self.count.add_ref()
             }
-            fn Release(&mut self) -> u32 {
+            unsafe fn Release(&self) -> u32 {
```
eholk, line 130: Is it possible to make `Release` take `self`? If so, I think you could make this safe.

I'm guessing that doesn't fit with lots of COM patterns, since I remember making lots of manual calls to `Release` back when I've done COM programming before. Still, using this object after calling `Release` seems suspect, since `Release` might drop itself, but I guess that's why this is unsafe.

crates/libs/windows/src/core/unknown.rs
```diff
@@ -61,15 +61,28 @@ impl core::fmt::Debug for IUnknown {
...
-    fn QueryInterface(&mut self, iid: &GUID, interface: *mut RawPtr) -> HRESULT;
-    fn AddRef(&mut self) -> u32;
-    fn Release(&mut self) -> u32;
+    /// The classic `QueryInterface` method from COM.
+    ///
+    /// # Safety
+    ///
+    /// This function is safe to call as long as the interface pointer is non-null and valid for writes
+    /// of an interface pointer.
+    unsafe fn QueryInterface(&self, iid: &GUID, interface: *mut *const core::ffi::c_void) -> HRESULT;
```
eholk, line 70: You could use [`NonNull`](https://doc.rust-lang.org/std/ptr/struct.NonNull.html) to encode the non-null safety requirement here if you wanted.

crates/tests/nightly_class_factory/tests/test.rs
```diff
@@ -28,7 +28,7 @@ impl IClassFactory_Impl for Factory {
         assert!(outer.is_none());
         let unknown: IInspectable = Object().into();
         // TODO: https://github.com/microsoft/windows-rs/issues/1441
-        unsafe { unknown.query(&*iid, object).ok() }
+        unsafe { unknown.query(&*iid, object as *mut _).ok() }
```
eholk, line 31: It might be worth adding a `SAFETY` comment here, although I don't know that it'd be terribly informative. It looks like you're just calling an unsafe function, `query`, and following its requirements.

eholk (approved):
Looks good! I added a few clarification questions and possible suggestions, but I think it's fine as is.

crates/libs/implement/src/lib.rs
rylev: `Release` gets called in the `Drop` impl of COM objects, and we only have a `&mut self` in `drop` so I don't think it's possible for `Release` to take `self`. This is unfortunately where COM's semantics and Rust semantics don't 100% align and so we're left with a dangling pointer for a short amount of time. I think the unfortunate answer is that we just have to be careful. Luckily, users will almost never interact with this method and instead `Release` will get called automatically on drop.

crates/libs/windows/src/core/unknown.rs
rylev: I consider that, but I'm not sure it's worth it since `NonNull` is a bit awkward to work with, and the user still has to verify the pointer is valid for writing an interface pointer to. Since `NULL` is not valid for writes, the safety check should already tip the caller to check for non-null.

crates/tests/nightly_class_factory/tests/test.rs
rylev: I think that's a good idea, but we can take care of that when we document `query`'s safety needs (which I personally don't yet have a good grasp on).

Issue #1556, opened 2022-02-20
Getting raw pointer out of a COM interface?


chyyran:
I'm trying to convert an `ID3D11ShaderResourceView` struct into a `usize` handle for use with ImGui textures. The equivalent C++ code is 

```cpp
ID3D11ShaderResourceView* srv = ...;
...
ImGui::Image((void*)srv, ImVec2(width, height));
```

However, I can't cast the Rust projection of `ID3D11ShaderResourceView`  into `*const ()`. 

Are any of these approaches correct? None of them seem exactly right to me.

1. Coerce the reference of an `ID3D11ShaderResourceView` to a pointer then to usize

```rust
let tex_ptr = &srv as *const ID3D11ShaderResourceView as usize
imgui::Image::new(TextureId::new(tex_ptr), &[width, height]);
```

2. `transmute` the `ID3D11ShaderResourceView` to a pointer (??)

```rust
let tex_ptr : *const () = std::mem::transmute(srv);
imgui::Image::new(TextureId::new(tex_ptr as usize), &[width, height]);
```

What would be the correct approach to take here?

kennykerr:
`transmute_copy` will give you the pointer value without stealing the ownership. 

```rust
let ptr: usize = std::mem::transmute_copy(&srv);
```

rylev:
Just to push back a bit: `transmute` and `transmute_copy` are extremely unsafe and can at times be hard to use correctly. Your suggestion here is correct, but it does seem strange that the official way to get a raw pointer from what is essentially a new type around a raw pointer would be to use these highly unsafe functions when adding a method is pretty low cost. This is why com-rs provided `as_raw` methods which are safe to use and allow users to turn their COM pointers into raw pointers in a safe way.

kennykerr:
Sure, I'm not opposed to adding a helper to the `Interface` trait for example. I'm just generally concerned about making this operation appear less dangerous than it is because while it is safe from Rust's perspective it's a pretty dangerous thing to do from COM's perspective.

PR #1738, opened 2022-05-05
Improve the `Interface` trait

rylev:
*Note*: since this involves changes to generated code, it's easiest to review this commit by commit. The codegen was run in a separate commit from the other changes.

This improves the safety auditability of the `Interface` trait:
* Adds safety docs to the `Interface` trait. Even though this trait is not meant for public consumption, the docs can really help those reviewing the safety of generated code.
* Adds an `as_raw` method to the trait which simply wraps `std::mem::transmute_copy`. `transmute_copy` is very powerful and easy to use wrong. Having a bespoke method for converting an interface to a raw pointer which is documented as such makes the code easier to read. We've also had several [users ask for this type of functionality before](https://github.com/microsoft/windows-rs/issues/1556). 
* Unmarks `Interface::vtable` as `unsafe` since this method is in fact safe. 

Finally, we use the `Interface::as_raw` method instead of `std::mem::transmute_copy` in the generated code.  

cc @eholk


kennykerr:
> Adds safety docs to the Interface trait. Even though this trait is not meant for public consumption, the docs can really help those reviewing the safety of generated code.

Note that parts of this trait are meant for public consumption. That's why some members were hidden. Notably, `IID` and `cast` are meant to be used by developers. The rest should remain hidden.

rylev:
@kennykerr that sounds fine. I would argue for the newly added `as_raw` to also be publicly accessible as I think it's a better API (with many parallels in the standard library) than having to use `std::mem::transmute_copy`.

crates/libs/windows/src/core/interface.rs
```diff
@@ -7,42 +7,67 @@ use bindings::*;
...
+///
+/// It is only safe to implement this trait if the implementing type is a valid COM interface pointer meaning
+/// the following criteria are met:
+/// * its in-memory representation is equal to `NonNull<NonNull<Self::VTable>>`
+/// * the vtable is correctly structured beginning with the `IUnknown` function pointers
+/// * the vtable must be the correct vtable for the supplied IID
 pub unsafe trait Interface: Sized {
-    #[doc(hidden)]
+    /// The interface's vtable
     type Vtable;
```
kennykerr, line 18: Let's keep this hidden.

kennykerr (dismissed):

eholk (dismissed):

rylev:
@kennykerr the latest nightly of clippy brought some new things to fix which I did, but it made the reviews stale. Can you re-review (only the last commit is new), and merge?

kennykerr (approved):

PR #1742, opened 2022-05-06
Add WinSock conversion utilities

### LINKED ISSUE #1650: Feature request: Add WinSock conversion utilities
### Motivation

Rust has its own network library in std::net, which has its own IpAddr types and SocketAddr types.
So, there would be a good feature to add From and Into traits for matching WinSock structs, so other devs don't have to do it by hand.

```
IN_ADDR <-> Ipv4Addr
IN6_ADDR <-> Ipv6Addr
SOCKADDR_IN <-> SocketAddrV4
SOCKADDR_IN6 <-> SocketAddrV6
SOCKADDR_INET <-> SocketAddr
```

### Drawbacks

None

_No response_

### Additional context

_No response_

ryancerium:
Fixes #1650

~This is a skeleton PR as-is, I'm mostly seeking feedback on whether or
not this is the right place to put the files.~

I didn't implement `TryFrom` for converting the Win32 socket address types to Rust types because they could be something other than IP protocol.  It seemed early just yet.


ryancerium:
Hey @AronParker you can thrash my code now! :-D

crates/libs/windows/src/core/winsock.rs
```diff
@@ -0,0 +1,24 @@
...
+#[cfg(feature = "Win32_Networking_WinSock")]
+impl core::convert::From<&std::net::Ipv4Addr> for IN_ADDR {
+    fn from(addr: &std::net::Ipv4Addr) -> IN_ADDR {
+        IN_ADDR { S_un: IN_ADDR_0 { S_addr: u32::from(addr) } }
+    }
+}
+*/
+
+#[cfg(feature = "Win32_Networking_WinSock")]
+impl core::convert::From<std::net::Ipv4Addr> for IN_ADDR {
```
kennykerr, line 13: We could inject this code into the module that defines `IN_ADDR`. Not sure what the pros and cons are of the two approaches. If there aren't any for the consumer, then this is certainly simpler.

ryancerium: Where is the module that defines `IN_ADDR`?  A cursory glance suggests `crates/libs/bindgen/src/extensions/` is what you're talking about?

I'm not a crate publishing expert, but I think this gets exported everywhere the same.  Simple is generally better, but what if the feature name changes?

kennykerr: Yes, the main advantage is that it travels with the type. I'm now looking more seriously into splitting the `windows` crate into smaller chunks and that would allow such extensions to automatically get generated into the appropriate crate.

ryancerium: I think that's a very worthwhile goal. I'm having to do all this in vim in a terminal because my laptop (only 16 GB RAM) can't get rust-analyzer to finish parsing the project.

ryancerium:
I moved the generation into `bindgen/src/extensions/in_addr.rs`.  I'm a bit at a loss how to run the bindgen and test the resulting code though.

kennykerr:
`cargo run -p tool_windows` will regenerate the `windows` crate locally.

crates/libs/bindgen/src/extensions/in_addr.rs
```diff
@@ -0,0 +1,26 @@
+use super::*;
+
+pub fn gen() -> TokenStream {
+    quote! {
+        impl ::core::convert::From<::std::net::Ipv4Addr> for IN_ADDR {
+            fn from(addr: ::std::net::Ipv4Addr) -> IN_ADDR {
+                IN_ADDR { S_un: IN_ADDR_0 { S_addr: u32::from(addr) } }
```
kennykerr, line 7:
```suggestion
            fn from(addr: ::std::net::Ipv4Addr) -> Self {
                Self { S_un: IN_ADDR_0 { S_addr: u32::from(addr) } }
```

kennykerr: A few more instances worth simplifying in this way.

kennykerr:
Once it's rebuilt the `windows` crate you can test your specific addition using `cargo test -p test_win32`.

crates/libs/bindgen/src/extensions/mod.rs
```diff
@@ -16,6 +16,7 @@ pub fn gen(def: &TypeDef) -> TokenStream {
         TypeName::Matrix3x2 => matrix3x2::gen(),
         TypeName::Matrix4x4 => matrix4x4::gen(),
         TypeName::WIN32_ERROR => win32_error::gen(),
+        TypeName::IN_ADDR => in_addr::gen(),
```
kennykerr, line 19: You forgot to declare the `in_addr` module.

ryancerium: Yeah, I pushed that commit before you taught me how to rebuild and test.  I've added ipv6 already and I'm looking at how to set the `sin_family` in the sockaddr structs now.

AronParker:
> Hey @AronParker you can thrash my code now! :-D

Haha thank you it's a bit late here so I'll review it tomorrow! That being said while taking a first glance, `SOCKADDR_INET` seems to be missing, it's the union version that *can* be equivalent to [`IpAddr`](https://doc.rust-lang.org/std/net/enum.IpAddr.html), so possibly a `From` and `TryFrom` relationship. I also roughly remember that there was another structure in the pre-vista era that encompasses IPv4 and IPv6 that was a bit leaner, but I'll have to look that up tomorrow again. Either way nice to see your contribution!

crates/libs/bindgen/src/extensions/in_addr.rs
```diff
@@ -0,0 +1,26 @@
...
+        impl ::core::convert::Into<::std::net::Ipv4Addr> for &IN_ADDR {
+            fn into(self) -> ::std::net::Ipv4Addr {
+                ::std::net::Ipv4Addr::from( unsafe { self.S_un.S_addr } )
+            }
+        }
+        impl ::core::convert::Into<::std::net::Ipv4Addr> for IN_ADDR {
+            fn into(self) -> ::std::net::Ipv4Addr {
+                ::std::net::Ipv4Addr::from( unsafe { self.S_un.S_addr } )
+            }
+        }
```
kennykerr, line 24: Why implement `Into`? You should only need to implement `From`. 

https://doc.rust-lang.org/std/convert/trait.Into.html

crates/libs/bindgen/src/extensions/mod.rs
```diff
@@ -1,4 +1,6 @@
 use super::*;
+mod in6_addr;
```
kennykerr, line 2: Looks like this is missing.

crates/libs/bindgen/src/extensions/in_addr.rs
```diff
@@ -0,0 +1,26 @@
...
+            fn from(addr: &::std::net::Ipv4Addr) -> Self {
+                Self { S_un: IN_ADDR_0 { S_addr: u32::from(*addr) } }
+            }
+        }
+        impl ::core::convert::Into<::std::net::Ipv4Addr> for &IN_ADDR {
+            fn into(self) -> ::std::net::Ipv4Addr {
+                ::std::net::Ipv4Addr::from( unsafe { self.S_un.S_addr } )
+            }
+        }
+        impl ::core::convert::Into<::std::net::Ipv4Addr> for IN_ADDR {
```
AronParker, line 20:
```suggestion
        impl ::core::convert::From<IN_ADDR> for ::std::net::Ipv4Addr {
```

crates/libs/bindgen/src/extensions/in_addr.rs
```diff
@@ -0,0 +1,26 @@
...
+            fn from(addr: ::std::net::Ipv4Addr) -> Self {
+                Self { S_un: IN_ADDR_0 { S_addr: u32::from(addr) } }
+            }
+        }
+        impl ::core::convert::From<&::std::net::Ipv4Addr> for IN_ADDR {
+            fn from(addr: &::std::net::Ipv4Addr) -> Self {
+                Self { S_un: IN_ADDR_0 { S_addr: u32::from(*addr) } }
+            }
+        }
+        impl ::core::convert::Into<::std::net::Ipv4Addr> for &IN_ADDR {
```
AronParker, line 15:
```suggestion
        impl ::core::convert::From<&IN_ADDR> for ::std::net::Ipv4Addr {
```

As said by kennykerr, it's discouraged to implement Into directly, so best to swap the type types and implement From instead

crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs
```diff
@@ -1833,6 +1833,26 @@ impl ::core::default::Default for IN6_ADDR {
...
+    fn from(addr: ::std::net::Ipv6Addr) -> Self {
+        Self { u: IN6_ADDR_0 { Byte: addr.octets() } }
+    }
+}
+impl ::core::convert::From<&::std::net::Ipv6Addr> for IN6_ADDR {
+    fn from(addr: &::std::net::Ipv6Addr) -> Self {
+        Self { u: IN6_ADDR_0 { Byte: addr.octets() } }
+    }
+}
+impl ::core::convert::Into<::std::net::Ipv6Addr> for &IN6_ADDR {
```
AronParker, line 1846:
```suggestion
impl ::core::convert::From<&IN6_ADDR> for ::std::net::Ipv6Addr {
```

crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs
```diff
@@ -1833,6 +1833,26 @@ impl ::core::default::Default for IN6_ADDR {
...
+    fn into(self) -> ::std::net::Ipv6Addr {
+        ::std::net::Ipv6Addr::from(unsafe { self.u.Byte })
+    }
+}
+impl ::core::convert::Into<::std::net::Ipv6Addr> for IN6_ADDR {
```
AronParker, line 1851:
```suggestion
impl ::core::convert::From<IN6_ADDR> for ::std::net::Ipv6Addr {
```

crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs
```diff
@@ -2122,6 +2142,26 @@ impl ::core::default::Default for IN_ADDR {
...
+    fn from(addr: ::std::net::Ipv4Addr) -> Self {
+        Self { S_un: IN_ADDR_0 { S_addr: u32::from(addr) } }
+    }
+}
+impl ::core::convert::From<&::std::net::Ipv4Addr> for IN_ADDR {
+    fn from(addr: &::std::net::Ipv4Addr) -> Self {
+        Self { S_un: IN_ADDR_0 { S_addr: u32::from(*addr) } }
+    }
+}
+impl ::core::convert::Into<::std::net::Ipv4Addr> for &IN_ADDR {
```
AronParker, line 2155:
```suggestion
impl ::core::convert::From<&IN_ADDR> for ::std::net::Ipv4Addr {
```

crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs
```diff
@@ -2122,6 +2142,26 @@ impl ::core::default::Default for IN_ADDR {
...
+    fn into(self) -> ::std::net::Ipv4Addr {
+        ::std::net::Ipv4Addr::from(unsafe { self.S_un.S_addr })
+    }
+}
+impl ::core::convert::Into<::std::net::Ipv4Addr> for IN_ADDR {
```
AronParker, line 2160:
```suggestion
impl ::core::convert::From<IN_ADDR> for ::std::net::Ipv4Addr {
```

crates/tests/win32/tests/winsock.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[test]
+fn from_ipv4addr() {
+    let ipv4addr = std::net::Ipv4Addr::new(4, 3, 2, 1);
+    let in_addr: IN_ADDR = ipv4addr.into();
+    unsafe {
+        assert!(in_addr.S_un.S_un_b.s_b1 == 1);
+        assert!(in_addr.S_un.S_un_b.s_b2 == 2);
+        assert!(in_addr.S_un.S_un_b.s_b3 == 3);
+        assert!(in_addr.S_un.S_un_b.s_b4 == 4);
```
AronParker, line 11:
```suggestion
        assert_eq!(in_addr.S_un.S_un_b.s_b1, 1);
        assert_eq!(in_addr.S_un.S_un_b.s_b2, 2);
        assert_eq!(in_addr.S_un.S_un_b.s_b3, 3);
        assert_eq!(in_addr.S_un.S_un_b.s_b4, 4);
```

AronParker (changes requested):
We could implement the following as well: SOCKADDR_IN, SOCKADDR_IN6 and SOCKADDR_INET conversions here too. I'd say that could look like so:

```rust
impl From<SocketAddrV4> for SOCKADDR_IN {}
impl From<SocketAddrV6> for SOCKADDR_IN6 {}
impl From<SocketAddr> for SOCKADDR_INET {}
impl TryFrom<SOCKADDR_IN> for SocketAddrV4 {}
impl TryFrom<SOCKADDR_IN6> for SocketAddrV6 {}
impl TryFrom<SOCKADDR_INET> for SocketAddr {}
```

Since rust only supports IP as address family, it wouldn't work both ways so we have to resort to `TryFrom`. I suppose the error would be `io::Error` with `io::ErrorKind::Unsupported` in case the address family is not IP.

kennykerr:
> We could implement the following as well: SOCKADDR_IN, SOCKADDR_IN6 and SOCKADDR_INET conversions here too. I'd say that could look like so:
>
> *snip*
> 
> Since rust only supports IP as address family, it wouldn't work both ways so we have to resort to `TryFrom`. I suppose the error would be `io::Error` with `io::ErrorKind::Unsupported` in case the address family is not IP.

Let's just be conscious of only adding as much as is truly needed backed by solid tests or examples illustrating their universal benefit. I'm generally leery of adding hand-authored bells and whistles as this crate should for the most part remain focused on generated code only. So if there's any doubt, let's proceed very conservatively and minimally.

crates/tests/win32/tests/winsock.rs
ryancerium: Done.

crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs
ryancerium: Done.

crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs
ryancerium: Done.

crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs
ryancerium: Done.

crates/libs/windows/src/Windows/Win32/Networking/WinSock/mod.rs
ryancerium: Done.

crates/libs/bindgen/src/extensions/in_addr.rs
ryancerium: Done.

crates/libs/bindgen/src/extensions/mod.rs
ryancerium: Done.

crates/libs/bindgen/src/extensions/in_addr.rs
ryancerium: 🧠 💨 

Done.

ryancerium:
I skipped the `TryFrom` implementations per @kennykerr 's suggestion.  I agree with him that the size of this crate does not make hand-written stuff in here a great idea, but maybe not a bad idea either.  The orphan rule is gonna be a tough one for this kind of stuff.

crates/libs/bindgen/src/extensions/sockaddr_inet.rs
```diff
@@ -0,0 +1,42 @@
+use super::*;
+
+pub fn gen() -> TokenStream {
+    quote! {
+        impl ::core::convert::From<::std::net::SocketAddrV4> for SOCKADDR_INET {
+            fn from(addr: ::std::net::SocketAddrV4) -> Self {
+                SOCKADDR_INET {                    Ipv4: addr.into()                }
```
AronParker, line 7: You might want to run `cargo fmt` here or is there any particular reason for the number of spaces here?

ryancerium: I ran `cargo fmt`, committed and pushed!  Fixed it.

crates/libs/bindgen/src/extensions/in_addr.rs
```diff
@@ -0,0 +1,26 @@
...
+            }
+        }
+        impl ::core::convert::From<&::std::net::Ipv4Addr> for IN_ADDR {
+            fn from(addr: &::std::net::Ipv4Addr) -> Self {
+                Self { S_un: IN_ADDR_0 { S_addr: u32::from(*addr) } }
+            }
+        }
+        impl ::core::convert::From<IN_ADDR> for ::std::net::Ipv4Addr {
+            fn from(in_addr: IN_ADDR) -> Self {
+                Self::from( unsafe { in_addr.S_un.S_addr } )
```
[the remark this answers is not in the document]
ryancerium: Done.

crates/libs/bindgen/src/extensions/in_addr.rs
```diff
@@ -0,0 +1,28 @@
+use super::*;
+
+pub fn gen() -> TokenStream {
+    quote! {
+        impl ::core::convert::From<::std::net::Ipv4Addr> for IN_ADDR {
+            fn from(addr: ::std::net::Ipv4Addr) -> Self {
+                Self { S_un: IN_ADDR_0 { S_addr: u32::from(addr) } }
```
GamePad64, line 7: [`u32::from`](https://doc.rust-lang.org/std/primitive.u32.html#impl-From%3CIpv4Addr%3E) returns u32 in host byte order, but [`IN_ADDR`](https://docs.microsoft.com/en-us/previous-versions/windows/hardware/drivers/ff556972(v=vs.85)#remarks) awaits the address in network byte order.

crates/tests/win32/tests/winsock.rs
```diff
@@ -0,0 +1,65 @@
+use windows::Win32::Networking::WinSock::*;
+
+#[test]
+fn in_addr() {
+    let ipv4_addr = std::net::Ipv4Addr::new(4, 3, 2, 1);
+    let in_addr: IN_ADDR = ipv4_addr.into();
+
+    assert_eq!(unsafe { in_addr.S_un.S_un_b.s_b1 }, 1);
```
GamePad64, line 8: Byte-wise representation has to be the same as seen in IP address above. So s_b1 will be 4, s_b1 will be 3 and so on.

GamePad64: We can use [`inet_ntop`](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Networking/WinSock/fn.inet_ntop.html) here to get textual representations of IN_ADDR and IN_ADDR6, and then compare them to `.to_string` versions. There may be some issues in ipv6, though. But IN_ADDR6 has no these endianness issues, so this test for ipv6 may be less strict.

ryancerium:
Finally got some spare time this week to have some fun.  I learned something about IPv6 Scope ID too.  It's not all that common to see a C bitfield outside of CPU level stuff.  (Or so I think)

crates/libs/bindgen/src/extensions/in_addr.rs
ryancerium: Good catch.

crates/tests/win32/tests/winsock.rs
ryancerium: Done.

crates/libs/bindgen/src/extensions/in6_addr.rs
```diff
@@ -0,0 +1,28 @@
+use super::*;
+
+pub fn gen() -> TokenStream {
+    quote! {
+        impl ::core::convert::From<::std::net::Ipv6Addr> for IN6_ADDR {
+            fn from(addr: ::std::net::Ipv6Addr) -> Self {
+                Self { u: IN6_ADDR_0 { Byte: addr.octets() } }
+            }
+        }
+        impl ::core::convert::From<&::std::net::Ipv6Addr> for IN6_ADDR {
```
GamePad64, line 10: Here the same comment applies: https://github.com/microsoft/windows-rs/pull/1756#discussion_r871508252. `Ipv4Addr`/`Ipv6Addr`/`IpAddr`/`SocketAddrV4`/`SocketAddrV6` are `Copy`. Also, `IN_ADDR`/`IN6_ADDR`/`SOCKADDR_IN`/`SOCKADDR_IN6`/`SOCKADDR_INET` are `Copy`, too.

So there is no strict necessity to implement `From` for both a reference and a value for these types. Implementing `From` for values would be enough.
For example, in std [`impl From<Ipv4Addr> for u32`](https://doc.rust-lang.org/std/primitive.u32.html#impl-From%3CIpv4Addr%3E) is implemented only for a value, but not for a reference. I think, that the same logic can be applied here.

ryancerium: I'd noticed that for `u32` as well and I'm glad there's consensus on the topic.

ryancerium: `SocketAddrVN.addr()` returns a reference to an `IpvNAddr`, so I had to deref it the ugly way :-D

kennykerr (approved):
Looks good - thanks for adding all of the tests!

PR #1749, opened 2022-05-09
Add `HSTRING` compatibility testing

kennykerr:
In support of #1747, here are some tests to ensure we preserve `HSTRING` compatibility with the OS. Call me paranoid. 😨

### DIFF
--- a/crates/libs/windows/src/core/heap.rs
+++ b/crates/libs/windows/src/core/heap.rs
@@ -14,6 +14,13 @@ pub fn heap_alloc(bytes: usize) -> Result<RawPtr> {
     if ptr.is_null() {
         Err(E_OUTOFMEMORY.into())
     } else {
+        // HeapAlloc is not guaranteed to return zero memory but usually does. This just ensures that
+        // it predictably returns non-zero memory for testing purposes. This is similar to what MSVC's
+        // debug allocator does for the same reason.
+        #[cfg(debug_assertions)]
+        unsafe {
+            std::ptr::write_bytes(ptr, b'?', bytes);
+        }
         Ok(ptr)
     }
 }
--- a/crates/libs/windows/src/core/heap.rs
+++ b/crates/libs/windows/src/core/heap.rs
@@ -19,7 +19,7 @@ pub fn heap_alloc(bytes: usize) -> Result<RawPtr> {
         // debug allocator does for the same reason.
         #[cfg(debug_assertions)]
         unsafe {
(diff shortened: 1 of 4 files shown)


crates/tests/core/tests/hstrings.rs
```diff
@@ -153,3 +152,43 @@ fn hstring_osstring_equality_combinations() {
...
+        assert_eq!(WindowsIsStringEmpty(&world), false);
+        assert_eq!(WindowsIsStringEmpty("World"), false);
+
+        let mut len = 0;
+        let buffer = WindowsGetStringRawBuffer(&world, &mut len);
+        assert_eq!(std::slice::from_raw_parts(buffer.0, len as _), world.as_wide());
+
+        // We need to drop to windows-sys to call the raw WindowsDeleteString function to avoid double-freeing the HSTRING,
+        // but this test is important as it ensures that the allocators match.
+        let hresult = windows_sys::Win32::System::WinRT::WindowsDeleteString(std::mem::transmute_copy(&*std::mem::ManuallyDrop::new(hey)));
```
rylev, line 189: Somewhat unrelated but should `WindowsDeleteString` take a `ManuallyDrop<HSTRING>` instead of a `HSTRING`?

kennykerr: Yes, there's just not metadata for that right now. We could just exclude it from the `windows` crate as there's no correct way to call it and there's no need to call it either.

kennykerr: Yep, there's friction for sure.

eholk (approved):

crates/libs/windows/src/core/heap.rs
```diff
@@ -14,6 +14,13 @@ pub fn heap_alloc(bytes: usize) -> Result<RawPtr> {
     if ptr.is_null() {
         Err(E_OUTOFMEMORY.into())
     } else {
+        // HeapAlloc is not guaranteed to return zero memory but usually does. This just ensures that
+        // it predictably returns non-zero memory for testing purposes. This is similar to what MSVC's
+        // debug allocator does for the same reason.
+        #[cfg(debug_assertions)]
+        unsafe {
+            std::ptr::write_bytes(ptr, b'?', bytes);
```
riverar, line 22: Think it'd be more useful to devs if this returned `asm { int 3 }`
```suggestion
            std::ptr::write_bytes(ptr, 0xCC, bytes);
```

PR #1750, opened 2022-05-10
Use `core::ptr::write` for out params

### LINKED ISSUE #1680: Bug: Generated VTable method calls drop() on potentially uninitialized value

windows

### Crate version

0.35.0

### Summary

I use the `#[implement]` macro to implement [IStringDisplayableConcept](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/Diagnostics/Debug/struct.IStringDisplayableConcept.html). The `ToDisplayString()` method there is a callback that gets called from C++. The C++ side does *not initialize* the `BSTR` that the `BSTR* displayString` out parameter points to, so the following code generated by windows-rs will call `SysFreeString()` on an uninitialized value:

```rust
#[cfg(feature = "Win32_Foundation")]
impl IStringDisplayableConcept_Vtbl {
    pub const fn new<Identity: ::windows::core::IUnknownImpl, Impl: IStringDisplayableConcept_Impl, const OFFSET: isize>() -> IStringDisplayableConcept_Vtbl {
        unsafe extern "system" fn ToDisplayString<Identity: ::windows::core::IUnknownImpl, Impl: IStringDisplayableConcept_Impl, const OFFSET: isize>(this: *mut ::core::ffi::c_void, contextobject: ::windows::core::RawPtr, metadata: ::windows::core::RawPtr, displaystring: *mut super::super::super::Foundation::BSTR) -> ::windows::core::HRESULT {
            let this = (this as *mut ::windows::core::RawPtr).offset(OFFSET) as *mut Identity;
            let this = (*this).get_impl() as *mut Impl;
            match (*this).ToDisplayString(::core::mem::transmute(&contextobject), ::core::mem::transmute(&metadata)) {
                ::core::result::Result::Ok(ok__) => {
                    *displaystring = ::core::mem::transmute(ok__); // <--- THIS WILL TRY TO DROP THE POINTEE BEFORE OVERWRITING
                    ::windows::core::HRESULT(0)
                }
                ::core::result::Result::Err(err) => err.into(),
            }
        }
        Self { base__: ::windows::core::IUnknownVtbl::new::<Identity, OFFSET>(), ToDisplayString: ToDisplayString::<Identity, Impl, OFFSET> }
    }
    pub fn matches(iid: &windows::core::GUID) -> bool {
        iid == &<IStringDisplayableConcept as ::windows::core::Interface>::IID
    }
}
```
(from https://github.com/microsoft/windows-rs/blob/f00461914705ff1673606d38dba8e3a31730775d/crates/libs/windows/src/Windows/Win32/System/Diagnostics/Debug/impl.rs)

The bug is not hit when calling `IStringDisplayableConcept::ToDisplayString()` from Rust because the wrapper there properly initializes the `BSTR` being pointed to.

active toolchain
----------------

nightly-x86_64-pc-windows-msvc (directory override for '...')
rustc 1.62.0-nightly (306ba8357 2022-04-05)

### Reproducible example

```rust
// NOTE: I was not able to reproduce this from pure Rust code. 
//       The following is an example implementation of IStringDisplayableConcept, 
//       but ToDisplayString() is not called with an uninitialized value.

use windows::{
    core::implement,
    Win32::{
        Foundation::BSTR,
        System::Diagnostics::Debug::{
            IKeyStore, IModelObject, IStringDisplayableConcept, IStringDisplayableConcept_Impl,
        },
    },
};

#[implement(IStringDisplayableConcept)]
struct Foo;

impl IStringDisplayableConcept_Impl for Foo {
    fn ToDisplayString(
        &self,
        _: &Option<IModelObject>,
        _: &Option<IKeyStore>,
    ) -> windows::core::Result<BSTR> {
        Ok(BSTR::from("bar"))
    }
}

fn main() {
    let foo: IStringDisplayableConcept = Foo.into();
    let _ = unsafe { foo.ToDisplayString(None, None) };
}
```

### Crate manifest

```toml
[package]
name = "windows-rs-drop-repro"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies.windows]
version = "0.35.0"
features = [
    "alloc",
    "interface",
    "implement",
    "Win32_System_Diagnostics_Debug",
    "Win32_Foundation",
    "Win32_System_Com",
    "Win32_System_Ole",
    "Win32_System_Kernel",
]
```

### Expected behavior

I expected the potentially uninitialized value being pointed to not to be touched, e.g. by using [std::ptr::write](https://doc.rust-lang.org/std/ptr/fn.write.html):

```rust
#[cfg(feature = "Win32_Foundation")]
impl IStringDisplayableConcept_Vtbl {
    pub const fn new<Identity: ::windows::core::IUnknownImpl, Impl: IStringDisplayableConcept_Impl, const OFFSET: isize>() -> IStringDisplayableConcept_Vtbl {
        unsafe extern "system" fn ToDisplayString<Identity: ::windows::core::IUnknownImpl, Impl: IStringDisplayableConcept_Impl, const OFFSET: isize>(this: *mut ::core::ffi::c_void, contextobject: ::windows::core::RawPtr, metadata: ::windows::core::RawPtr, displaystring: *mut super::super::super::Foundation::BSTR) -> ::windows::core::HRESULT {
            let this = (this as *mut ::windows::core::RawPtr).offset(OFFSET) as *mut Identity;
            let this = (*this).get_impl() as *mut Impl;
            match (*this).ToDisplayString(::core::mem::transmute(&contextobject), ::core::mem::transmute(&metadata)) {
                ::core::result::Result::Ok(ok__) => {
                    std::ptr::write(displaystring, ok__); // <--- Use std::ptr::write() 
                    ::windows::core::HRESULT(0)
                }
                ::core::result::Result::Err(err) => err.into(),
            }
        }
        Self { base__: ::windows::core::IUnknownVtbl::new::<Identity, OFFSET>(), ToDisplayString: ToDisplayString::<Identity, Impl, OFFSET> }
    }
    pub fn matches(iid: &windows::core::GUID) -> bool {
        iid == &<IStringDisplayableConcept as ::windows::core::Interface>::IID
    }
}
```

### Actual behavior

Heap corruption 😃 

### Additional comments

The support for implementing COM interfaces in windows-rs is already much more ergonomic than going via com-rs. Thanks for all your hard work!

rylev:
Fixes #1680 

When COM interfaces are used from Rust, out params are initialized before being passed to the vtable. However, this is not always the case from other languages where an uninitialized value might be passed as an out param. Using the `*ptr = value` syntax in Rust causes `ptr` to be read (and dropped) which is obviously not correct for uninitialized values. The docs clearly state that one should prefer using `core::ptr::write` when writing to uninitialized values. 

*Please review commit by commit as the codegen is run in separate commits*

### DIFF
--- a/crates/libs/bindgen/src/helpers.rs
+++ b/crates/libs/bindgen/src/helpers.rs
@@ -329,7 +329,8 @@ pub fn gen_win32_upcall(sig: &Signature, inner: TokenStream) -> TokenStream {
             quote! {
                 match #inner(#(#invoke_args,)*) {
                     ::core::result::Result::Ok(ok__) => {
-                        *#result = ::core::mem::transmute(ok__);
+                        // use `core::ptr::write` since the result could be uninitialized
+                        ::core::ptr::write(#result, ::core::mem::transmute(ok__));
                         ::windows::core::HRESULT(0)
                     }
                     ::core::result::Result::Err(err) => err.into()
@@ -369,8 +370,9 @@ pub fn gen_winrt_upcall(sig: &Signature, inner: TokenStream) -> TokenStream {
                 match #inner(#(#invoke_args,)*) {
                     ::core::result::Result::Ok(ok__) => {
                         let (ok_data__, ok_data_len__) = ok__.into_abi();
-                        *result__ = ok_data__;
-                        *result_size__ = ok_data_len__;
+                        // use `core::ptr::write` since `result` could be uninitialized
+                        ::core::ptr::write(result__, ok_data__);
+                        ::core::ptr::write(result_size__, ok_data_len__);
                         ::windows::core::HRESULT(0)
                     }
                     ::core::result::Result::Err(err) => err.into()
@@ -381,7 +383,8 @@ pub fn gen_winrt_upcall(sig: &Signature, inner: TokenStream) -> TokenStream {
             quote! {
                 match #inner(#(#invoke_args,)*) {
                     ::core::result::Result::Ok(ok__) => {
-                        *result__ = ::core::mem::transmute_copy(&ok__);
+                        // use `core::ptr::write` since `result` could be uninitialized
+                        ::core::ptr::write(result__, ::core::mem::transmute_copy(&ok__));
                         ::core::mem::forget(ok__);
                         ::windows::core::HRESULT(0)
                     }
--- a/crates/libs/windows/src/Windows/AI/MachineLearning/Preview/impl.rs
+++ b/crates/libs/windows/src/Windows/AI/MachineLearning/Preview/impl.rs
@@ -17,7 +17,7 @@ impl ILearningModelVariableDescriptorPreview_Vtbl {
             let this = (*this).get_impl();
             match this.Name() {
                 ::core::result::Result::Ok(ok__) => {
-                    *result__ = ::core::mem::transmute_copy(&ok__);
+                    ::core::ptr::write(result__, ::core::mem::transmute_copy(&ok__));
                     ::core::mem::forget(ok__);
                     ::windows::core::HRESULT(0)
                 }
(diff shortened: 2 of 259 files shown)


riverar (dismissed):
Good catch, defer to others for final merge guidance.

kennykerr (approved):

PR #1752, opened 2022-05-10
Inline `HSTRING::clear` into the `Drop` impl

rylev:
As I've asked in [another PR](https://github.com/microsoft/windows-rs/pull/1747#issuecomment-1122352608), there was some confusion about a `debug_assert!` in the `HSTRING::clear` method which would panic in debug mode if it was detected we were trying to clear a "fast pass" string (i.e., an `HSTRING` which allocates its header on the stake which is used in some languages).

It struck me as odd that we were using a `debug_assert!` since it seemed like something we'd always want to prevent, and @kennykerr let me know that "fast pass" `HSTRING`s are only used as `in` arguments (and only can be since their headers are stack allocated). Because `in` arguments are not "owned" and thus should not have their destructors run, the `Drop` impl where `HSTRING::clear` was (and continues) being called would never run, and if it did it was because of a bad translation from the Rust side (hence the `debug_assert!`). 

However, this means that calling `HSTRING::clear` should not be called directly on a "fast pass" `HSTRING`, but it was possible to do so from otherwise correct and safe code. Since `HSTRING::clear` is not really necessary to expose, I've decided to inline the implementation into the `Drop` impl, and I've made some of the comments around it more clear.

### DIFF
--- a/crates/libs/windows/src/core/hstring.rs
+++ b/crates/libs/windows/src/core/hstring.rs
@@ -1,6 +1,6 @@
 use super::*;
 
-/// A WinRT string ([HSTRING](https://docs.microsoft.com/en-us/windows/win32/winrt/hstring)),
+/// A WinRT string ([HSTRING](https://docs.microsoft.com/en-us/windows/win32/winrt/hstring))
 /// is reference-counted and immutable.
 #[repr(transparent)]
 pub struct HSTRING(*mut Header);
@@ -51,27 +51,6 @@ impl HSTRING {
     /// Get the contents of this `HSTRING` as a OsString.
     pub fn to_os_string(&self) -> std::ffi::OsString {
         std::os::windows::ffi::OsStringExt::from_wide(self.as_wide())
-    }
-
-    /// Clear the contents of the string and free the memory if `self` holds the
-    /// last reference to the string data.
-    pub fn clear(&mut self) {
-        if self.is_empty() {
-            return;
-        }
-
-        unsafe {
-            // This flag indicates a "fast pass" string created by some languages where the
-            // header is allocated on the stack. Such strings must never be freed.
-            let header = self.0;
-            debug_assert!((*header).flags & REFERENCE_FLAG == 0);
-
-            if (*header).count.release() == 0 {
-                heap_free(self.0 as RawPtr);
-            }
-        }
-
-        self.0 = core::ptr::null_mut();
     }
 
     /// # Safety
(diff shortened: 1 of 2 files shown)


kennykerr (approved):

crates/libs/windows/src/core/hstring.rs
```diff
@@ -128,7 +107,23 @@ impl Clone for HSTRING {
...
+            return;
+        }
+
+        unsafe {
+            let header = std::mem::replace(&mut self.0, core::ptr::null_mut());
+            // This flag indicates a "fast pass" string created by some languages where the
+            // header is allocated on the stack. "Fast pass" strings can only ever be used
+            // as non-owned input params and as such `drop` will never be called. This check
+            // just ensures the Rust translation never accidentally modeled "fast pass" input
+            // params as owned `HSTRING` types which would be a bug in the translation.
```
riverar, line 120: Feel free to find your own words for this, or ignore altogether, but I was very confused by the use of "owned" and "params" in a context that isn't strictly dealing with either. (Perhaps also a result of my Rust naivete.)

```suggestion
            // This check ensures we do not attempt to free a "fast pass" HSTRING.
            // "Fast pass" HSTRINGs do not own the memory they describe and as such `drop`
            // should never be called.
```

riverar: Oh, looking at the original, I see I unwittingly brought us full circle. 🤭

rylev: 😊 Unfortunately, we're dealing with APIs and callers that exist outside of Rust's notions of ownership and borrowing, but that will interact with it which makes this confusing. What's more is that we still haven't fully figured out how we map the outside world on to Rust semantics. What we currently have works, but it's inconsistent and sometimes hard to reason about.

In this particular case, the `HSTRING` type defined here in Rust really represents an `HSTRING` that owns its memory and should have its ref count decremented (and memory possibly freed) when it's no longer needed. However, `HSTRING`s in other languages don't always represent handles to memory that is owned - one can also pass around `HSTRING`s that are just meant as temporary references (and thus the ref count should be left alone). One example of this are these "fast pass" strings which (because their headers are stack allocated) can only ever be used as "in" arguments to functions. When these types of `HSTRING`s come into a function, the function must take care to not decrement the ref count, and if the function wants to store the string somewhere that lives past the function call, it must be sure to increment the ref count. 

In our Rust translation we're currently modeling this as `ManuallyDrop<HSTRING>` because essentially these types of `HSTRING`s act just like the `HSTRING`s that own their memory with the exception that when you're done with them you shouldn't decrement the ref count (which in Rust happens in the `Drop` impl. `ManuallyDrop` accomplishes this by essentially turning off the `Drop` impl for the wrapped type. This is not particularly elegant or obvious, but it works. 

This comment is meant to say that if `Drop::drop` gets called on a "fast pass" string, the windows crate has a bug in how code got generated.

It's clear we still have some work to do to make some of this clearer in the code. We have a working solution, but we have some ways to go to make that solution obvious and understandable.

PR #1756, opened 2022-05-11
Add GUID::to_u128

### LINKED ISSUE #1755: Feature request: GUID to u128 conversion function
### Motivation

Well, there is `GUID::from_u128`, so I think, there should be `GUID::to_u128`. Then we could use u128 for public interface in crates and convert it to GUID.

### Drawbacks

A bit slower compilation time?

_No response_

### Additional context

_No response_

GamePad64:
Now GUID can round-trip to and from u128

Relates to #1306
Closes #1755

### DIFF
--- a/crates/libs/windows/src/core/guid.rs
+++ b/crates/libs/windows/src/core/guid.rs
@@ -103,6 +103,24 @@ impl core::convert::From<&str> for GUID {
     }
 }
 
+impl core::convert::From<u128> for GUID {
+    fn from(value: u128) -> Self {
+        Self::from_u128(value)
+    }
+}
+
+impl core::convert::From<GUID> for u128 {
+    fn from(value: GUID) -> Self {
+        value.to_u128()
+    }
+}
+
+impl core::convert::From<&GUID> for u128 {
+    fn from(value: &GUID) -> Self {
+        value.to_u128()
+    }
+}
+
 trait HexReader {
     fn next_u8(&mut self) -> u8;
     fn next_u16(&mut self) -> u16;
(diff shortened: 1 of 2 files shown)


kennykerr:
Would a `From` impl be more appropriate? `from_u128` exists because `From` is not `const` and needs to be for the `GUID` constants in the `windows` crate.

GamePad64:
Hmm, implementing From looks good for me. But having a separate `const fn` is nice, because compiler can optimize it out and it may matter in some cases.

kennykerr:
I think `From` is more idiomatic - how about we drop the explicit `from_u128` function for now - we can always add a `const` function down the road if that becomes necessary. Then we just need a basic test in `crates/tests/core/tests/guid.rs`.

kennykerr:
Oh, I see you have one already. I guess there is some symmetry to it. Let's see what others think.

crates/libs/windows/src/core/guid.rs
```diff
@@ -98,6 +103,24 @@ impl core::convert::From<&str> for GUID {
...
+    }
+}
+
+impl core::convert::From<GUID> for u128 {
+    fn from(value: GUID) -> Self {
+        value.to_u128()
+    }
+}
+
+impl core::convert::From<&GUID> for u128 {
```
rylev, line 118: We don't impl `From<&u128> for GUID` so I'm not sure we need to do the reverse. `GUID` is also `Copy` so it's trivial to go from a `&GUID` to `GUID`.

rylev (commented):
I like the symmetry. Looks good to mem just with one comment

crates/libs/windows/src/core/guid.rs
GamePad64: Ok, removed From<&GUID>

rylev (approved):

rylev:
Thanks for the contribution! Looks great 😊

PR #1758, opened 2022-05-12
Small clean ups to factory caching

rylev:
I was reading the factory caching code and noticed some small issues and ways to simplify. This just cleans up a bit.

### DIFF
--- a/crates/libs/windows/src/core/factory_cache.rs
+++ b/crates/libs/windows/src/core/factory_cache.rs
@@ -12,7 +12,7 @@ pub struct FactoryCache<C, I> {
 
 impl<C, I> FactoryCache<C, I> {
     pub const fn new() -> Self {
-        Self { shared: AtomicPtr::new(::core::ptr::null_mut()), _c: PhantomData, _i: PhantomData }
+        Self { shared: AtomicPtr::new(core::ptr::null_mut()), _c: PhantomData, _i: PhantomData }
     }
 }
 
@@ -32,7 +32,7 @@ impl<C: RuntimeName, I: Interface> FactoryCache<C, I> {
 
             // If the factory is agile, we can safely cache it.
             if factory.cast::<IAgileObject>().is_ok() {
-                if self.shared.compare_exchange_weak(core::ptr::null_mut(), unsafe { core::mem::transmute_copy(&factory) }, Ordering::Relaxed, Ordering::Relaxed).is_ok() {
+                if self.shared.compare_exchange_weak(core::ptr::null_mut(), factory.as_raw(), Ordering::Relaxed, Ordering::Relaxed).is_ok() {
                     core::mem::forget(factory);
                 }
             } else {
@@ -59,8 +59,8 @@ pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
             if code == CO_E_NOTINITIALIZED {
                 if let Ok(mta) = delay_load(b"ole32.dll\0", b"CoIncrementMTAUsage\0") {
                     let mta: CoIncrementMTAUsage = core::mem::transmute(mta);
-                    let mut _cookie = core::ptr::null_mut();
-                    let _ = mta(&mut _cookie);
+                    let mut cookie = core::ptr::null_mut();
+                    let _ = mta(&mut cookie);
                 }
 
                 // Now try a second time to get the activation factory via the OS.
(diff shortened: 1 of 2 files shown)


crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -92,11 +91,12 @@ pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
         while let Some(pos) = path.rfind('.') {
             path = &path[..pos];
 
-            let library = core::slice::from_raw_parts_mut(heap_alloc(path.len() + 5)? as *mut u8, path.len() + 5);
```
rylev, line 95: This was potentially UB as `from_raw_parts_mut` requires the memory to be initialized which `heap_alloc` does not guarantee.  I moved to using a `Vec` because I wasn't sure there was a reason for us to directly allocate memory ourselves vs using the std lib.

riverar: The memory provided by HeapAlloc is initialized, it's just not initialized to zero. So I don't think there's a real UB case here. But this is certainly nicer to look at, so no complaints.

riverar: Suggestion: On a more serious note, I think we should create a separate function for this chunk of code and write a test against it. It is critical that this code iterate through a given set of components in the correct order.

rylev: > The memory provided by HeapAlloc is initialized

Where do you see this? The documentation seems ambigious to me. When the `HEAP_ZERO_MEMORY` is passed "The allocated memory will be initialized to zero. Otherwise, the memory is not initialized to zero." ([link](https://docs.microsoft.com/en-us/windows/win32/api/heapapi/nf-heapapi-heapalloc#parameters)). One could read this as otherwise the memory is "initialized but just not zero" (which seems to be your understanding) or "not initialized at all" (which was my understanding). Is there any way to confirm that `HeapAlloc` *must* return initialized memory? 

>  I think we should create a separate function for this chunk of code and write a test against it

This is a good idea! 😊

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -59,50 +59,62 @@ pub fn factory<C: RuntimeName, I: Interface>() -> Result<I> {
...
 
-            if let Ok(factory) = get_activation_factory(library, &name) {
-                return factory.cast();
-            }
+// Remove the suffix until a match is found appending `.dll\0` at the end
+///
+/// For example, if the class name is
+/// "A.B.TypeName" then the attempted load order will be:
+///   1. A.B.dll
+///   2. A.dll
```
rylev, line 99: @riverar thanks to the test, I found that the implementation of the search heuristic differed from the documentation. The documentation here now matches the behavior. Was the previous documentation correct? Should this first search `A.B.TypeName.dll`?

crates/libs/windows/src/core/factory_cache.rs
riverar: > One could read this as otherwise the memory is "initialized but just not zero" (which seems to be your understanding) or "not initialized at all" (which was my understanding). Is there any way to confirm that `HeapAlloc` _must_ return initialized memory?

HeapAlloc, on success, will return a block of memory. The values stored in that block are arguably irrelevant. I don't think `from_raw_parts_mut` should care about what values we're pointing to. I suspect the `properly initialized` segment in the bullet `data must point to len consecutive properly initialized values of type T` is just an error.

But I still prefer your change!

crates/libs/windows/src/core/factory_cache.rs
riverar: I knew this sounded familiar, hah! I did correct this comment when I was in here https://github.com/microsoft/windows-rs/pull/1658/files#diff-b5b69fd449b1b061e71948bc6db077ea2d34ce8cca5a11aea3de2e12588a2f3eR96-R100

riverar: I believe this is an artifact from my previous explicit activation support. The comment is OK as-is now.

Here's the supporting documentation https://docs.microsoft.com/en-us/windows/uwp/winrt-components/create-a-windows-runtime-component-in-cppwinrt#naming-best-practice-for-windows-runtime-component-dlls

> [...] The logic is to remove the class name (from that fully-qualified class name), and then look for a DLL named for the full namespace that remains. If that's not found, remove the most-specific segment name, and repeat.

riverar (approved):

kennykerr:
I'll be back in a few days to review this - just a little cautious as there's a lot of subtlety to the factory cache.

crates/libs/windows/src/core/factory_cache.rs
rylev: > I suspect the properly initialized segment in the bullet `...` is just an error in Rust docs

This might not be relevant because we're talking about a slice of `u8` which can be any value, but in general this comment is very correct. If we were trying to build a `&mut [bool]` then random bytes would 100% be incorrect as Rust only allows `bool`s to be `0` or `1`. A byte that is not `0` or `1` representing a `bool` is UB and can cause miscompilations. 

For a good introduction in to why it's not correct for invalid data to represent a type even if that type is not read, [check out Ralf Jung's blog post on the topic](https://www.ralfj.de/blog/2020/07/15/unused-data.html).  

I know it feels pedantic, but this crate is foundational, and we should be *very* careful about memory safety.

kennykerr (approved):
Looks good.

Issue #1771, opened 2022-05-20
Remove nightly test distinction


kennykerr:
Now that Rust 1.61 is stable, we can remove all the complicated nightly test bifurcation.

PR #1773, opened 2022-05-20
Set windows-implement min rust-version

wravery:
Both the `#[implement]` macro and the `implement` feature in `windows` require newly stabilized features in `rustc` 1.61 (which just hit stable 🎉). We can specify it on the macro crate so anyone who wants to use the macro knows they need to update their toolchain and gets a clear error message if they don't.

I don't think you can specify a conditional `rust-version` which depends on the feature, but if there's a way to do that it would make sense to add it to the `windows` crate as well.

### DIFF
--- a/crates/libs/implement/Cargo.toml
+++ b/crates/libs/implement/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2018"
 license = "MIT OR Apache-2.0"
 description = "The implement macro for the windows crate"
 repository = "https://github.com/microsoft/windows-rs"
+rust-version = "1.61"
 
 [package.metadata.docs.rs]
 default-target = "x86_64-pc-windows-msvc"


kennykerr:
As part of #1771 I planned to remove the nightly restriction for testing. It is not sufficient to set the `rust-version` (although definitely a good move) as we also need to ensure that the yml scripts actually test that version. It's probably simplest to set the min version for the `windows` crate as a whole and test that via the yml scripts. The yml scripts are currently only testing 1.59 and nightly. The yml scripts are generated here: 

https://github.com/microsoft/windows-rs/blob/master/crates/tools/yml/src/main.rs

wravery:
> As part of #1771 I planned to remove the nightly restriction for testing. It is not sufficient to set the `rust-version` (although definitely a good move) as we also need to ensure that the yml scripts actually test that version. It's probably simplest to set the min version for the `windows` crate as a whole and test that via the yml scripts. The yml scripts are currently only testing 1.59 and nightly. The yml scripts are generated here:
> 
> https://github.com/microsoft/windows-rs/blob/master/crates/tools/yml/src/main.rs

OK, do you mean that you would like me to add `rust-version` to `windows` and update the yml script (presumably to update `stable` and `nightly`, and drop/replace `1.59` with `stable`) in the same PR? Or is that more of a forward-looking suggestion for another PR?

kennykerr:
Updating the yml generation is a much bigger job - happy for you to do it if you're in a hurry - but equally happy for you to wait until I've wrapped that up in a few days. I'm just not comfortable updating the rust-version without testing. I hope to get that done next week.

wravery:
> Updating the yml generation is a much bigger job - happy for you to do it if you're in a hurry - but equally happy for you to wait until I've wrapped that up in a few days. I'm just not comfortable updating the rust-version without testing. I hope to get that done next week.

OK, what about updating just `windows-implement` in this PR first, then? It looks like it passed all the CI checks, but I guess that means they aren't testing the macro crate (or only testing it on `nightly`)?

kennykerr:
Yep, it's only testing with nightly. I think the change is harmless now that 1.61 is stable, so I'll take it but yeah there's more work to be done.

kennykerr (approved):

wravery:
I think this may have already made the conditional MSRV work. When you add the `implement` feature to `windows`, it adds `windows-implement` as a dependency, right? When it tries to build `windows-implement`, it should already error out for its greater MSRV on earlier toolchains. 😄

kennykerr:
Yep... 😄 Although testing that would mean having to have different runs for the different min versions.

PR #1777, opened 2022-05-24
Stateful metadata & bindgen

### LINKED ISSUE #1406: Specify input .winmd files in windows-bindgen
### Motivation

Currently `windows-bindgen` crate seems to always look for .winmd files in .windows subdirectory. There should be a way to specify input files/directories.

### Drawbacks

The need to specify input path, or a more complex logic selecting defaults

When generating a .winmd file (see [WinRT component sample](https://github.com/Alovchin91/winrt-component-rs)), I would like to put it into cargo’s `$OUT_DIR` and then generate a projection from it.

### Additional context

_No response_

kennykerr:
The existing metadata reader is the last holdover from the old `build` macro and the switch to pre-generated bindings. It provided a static type reader that could be shared by multiple invocations of the various code generation macros but was rather difficult to use in a more traditional procedural model since you couldn't give it an explicit set of winmd files or control the lifetime of the metadata reader. Unfortunately, the static nature of the metadata had a rather viral effect on dependencies which took me quite a lot of work to untangle. 

The new metadata reader can now be created and dropped as needed and given an explicit set of winmd files to parse. This also ends up being a lot faster. The port still has a few rough edges, but I'd like to get this in and deal with the stragglers in upcoming PRs as this change has been in the pipeline for long enough. Feel free to comment and I'll address feedback in subsequent PRs. 

Fixes #1406


crates/libs/bindgen/src/implements.rs
```diff
@@ -1,86 +1,96 @@
...
+    if gen.reader.type_def_name(def).starts_with("Disp") && gen.reader.type_def_methods(def).next().is_none() {
         return quote! {};
     }
 
-    let type_ident = gen_ident(def.name());
+    let type_name = gen.reader.type_def_type_name(def);
+    // TODO: workaround for https://github.com/microsoft/win32metadata/issues/814
+    if type_name.name == "INetCfgComponentUpperEdge" {
+        return quote! {};
+    }
```
riverar, line 16: I believe this can be removed now. [Was fixed in a release in March](https://github.com/microsoft/win32metadata/commit/1939efb65d8c19f5301f5bb47fcd4087d51d47b9).

crates/libs/metadata/src/reader/flags.rs
```diff
@@ -1,23 +1,54 @@
...
-
 impl PInvokeAttributes {
     pub fn last_error(&self) -> bool {
-        self.0 & 0x0040 != 0
+        self.0 & 0x40 != 0
+    }
+}
+
+impl TypeAttributes {
+    pub fn union(&self) -> bool {
```
riverar, line 45: Nit: This is really `ExplicitLayout` right, which could or could not be related to unions.

crates/libs/bindgen/src/implements.rs
kennykerr: They claim to have removed it but it's back. Anyway, I excluded it here to make diffing simpler, but I can now remove this exclusion.

crates/libs/metadata/src/reader/flags.rs
kennykerr: Yep, I can use the more explicit :wink: name.

PR #1778, opened 2022-05-24
Remove nightly test distinction

### LINKED ISSUE #1771: Remove nightly test distinction

kennykerr:
Fixes #1771

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,14 +80,36 @@ jobs:
               >> $env:GITHUB_ENV
           }
         }
-    - name: Test stable
+    - name: Test
       run: >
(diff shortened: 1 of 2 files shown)


kennykerr:
Simplifying the yml generation seems to now exceed VM capacity. Perhaps it allows cargo to cache more stuff.

rylev (changes requested):
Seems like the CI takes much longer to run now. It's a shame that for some reason so much build artifact is being created that we run out of disk space. However, I think calling `cargo clean` before every test is probably over kill. It should be relatively straight forward to only add `cargo clean` once half way through.

rylev (approved):

PR #1787, opened 2022-05-27
Switch to LLVM dlltool for GNU lib generation

riverar:
Swaps out GNU dlltool for newer LLVM dlltool.exe-compatible driver. This results in a substantial decrease in import lib sizes for both i686 and x86_64 GNU targets.

| Target       | Before               | After              | % Savings |
|--------------|----------------------|--------------------|-----------|
| i686-gnu     | 14009016 (~13.36MiB) | 4285402 (~4.08MiB) | **69.4097**   |
| x86_64-gnu   | 13879932 (~13.23MiB) | 4103298 (~3.91MiB) | **70.4372**   |
| i686-msvc    | 4650454 (~4.43MiB)   | 4650454 (~4.43MiB) | 0.0000    |
| x86_64-msvc  | 4345164 (~4.14MiB)   | 4345164 (~4.14MiB) | 0.0000    |
| aarch64-msvc | 4345164 (~4.14MiB)   | 4345164 (~4.14MiB) | 0.0000    |

### DIFF
--- a/crates/tools/gnu/readme.md
+++ b/crates/tools/gnu/readme.md
@@ -16,6 +16,8 @@ The Windows umbrella lib (targeting GNU tooling) is generated using the followin
 1. Open `MSYS2 MinGW 32-bit`
 2. Execute: `pacman -Syuu --noconfirm` (repeat until no further updates available)
 3. Repeat step 1 if needed
-4. Navigate to crate root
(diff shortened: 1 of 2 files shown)


riverar:
Will need to potentially work with llvm-dlltool folks on gnu dlltool compat issues here before we can move forward.

Issue #1410, opened 2022-01-02
How to correctly release `ID3D12Resource` for `IDXGISwapChain3::ResizeBuffers`


polymonster:
Hi I'm new to rust, but experienced with win32/directx.

I am trying to make a resizable swap chain from a window using `IDXGISwapChain3::ResizeBuffers`

I get an error in the validation layer `DXGI ERROR: IDXGISwapChain::ResizeBuffers: Swapchain cannot be resized unless all outstanding buffer references have been released. [ MISCELLANEOUS ERROR #19: ]`

I have working c++ code which calls `ID3D12Resource::Release` on buffers which are obtained from the `IDXGISwapChain3::GetBuffer` function before making the call to `IDXGISwapChain3::ResizeBuffers`. 

What would be the equivalent of a `Release` call in rust? because the api doesn't mention `Release`

robmikh:
AddRef and Release are tied to Clone and Drop respectively. Implicitly or explicitly droppings your buffers should be sufficient.

polymonster:
Thanks for your quick response,

In this case I don't have references to buffers that I keep myself, they are created implicitly along with the `IDXGISwapChain3`  so I think I must need to explicitly drop them in this case as the ownership is with the swap chain and I want to retain the swap chain itself.

I am not sure how to explicitly drop the internally owned buffer resources. They can be obtained through the function call `IDXGISwapChain3::GetBuffer` I have tried passing to `std::mem::drop` but still get the same validation error.

```rust
  for i in 0..FRAME_COUNT {
      let buffer: ID3D12Resource = self.swap_chain.GetBuffer(i).unwrap();
      std::mem::drop(buffer);
  }
```

Thanks again for your help, it has given me more to look into.

robmikh:
Have you created render target views for any of the buffers? I believe they'll hold a reference.

polymonster:
Yeah, I have references to render target views which I was also clearing. 

I have been able to successfully call `ResizeBuffers` as long as I don't use the render target views in a command list, so I'm just following back from there now to make sure the waits are being correctly handled and the command lists are no longer in use on the gpu.

I think it should be ok from here on out, at first I wasn't entirely sure how the COM objects interacted with rust lifetimes but it's much clearer now.

Thanks for the help.

polymonster:
The issue seems to be with using a transition barrier, each frame the swap chain buffer leaks references. This issue is also present in the direct3d12 sample of the repository.

[...]

PR #1788, opened 2022-05-27
`Borrowed<'a, T>` RFC

rylev:
This introduces an RFC for a `Borrowed<'a, T>` type which models a type that is logically borrowed but still passed by value. As the motivation explains, this is needed to bridge the gap between Rust's ownership semantics and the looser requirements of C and C++. 

Introducing this would not only simplify codegen considerably but would also address issues like #1410.

[[rendered]](https://github.com/rylev/windows-rs/blob/borrow-rfc/rfcs/0002-borrow-type.md)


rfcs/0002-borrow-type.md
```diff
@@ -0,0 +1,257 @@
...
+[summary]: #summary
+
+This RFC introduces a new type `Borrow<'a, T>` which mirrors the semantics of a `&'a T` but with the difference that it has the same in memory layout as `T`.
+
+# Motivation
+[motivation]: #motivation
+
+Windows APIs often have ownership semantics that differ slightly from Rust's model. Types can often have different ownership semantics depending on the context. In Rust we would normally model these different semantics by using two different types, in C/C++ this is usually left to documentation and convention. 
+
+This difference is only interesting for types that are *not* "blittable" (`Copy` in Rust parlance). Non-"blittable" types have non-trivial destructors and cannot just be arbitrarily mem-copied. "Blittable" types like integers and booleans on other the hand can be, and so their ownership semantics are simple and map directly to Rust's model.
```
rylev, line 15: We should make a note about `PCWSTR` and kin. These types can be considerable "blittable" in this context since they have "trivial" ownership semantics from Rust's perspective (i.e., the user is in full control). This is the same reason `*mut T` and `*const T` are considered `Copy`. 

With this reasoning, I would see no reason for arguments of type `PCWSTR` to be modeled as `Borrow<'a, PCWSTR>` instead of just plain `PCWSTR` or *perhaps* `Into<PCWSTR>`.

ChrisDenton:
Should [I/O Safety](https://github.com/rust-lang/rust/issues/87074)? In particular `BorrowedHandle<'a>` is essentially the equivalent of:

`Borrowed<'a, Win32::Foundation::HANDLE>`.

It looks to be nearing stabilization now so I think it would be good to have an interop story, if possible.

kennykerr:
Since this is the first RFC-style PR, I'll just leave a meta comment that GitHub doesn't have a great markdown review experience. I realize you mitigate the issue with a preview link, but without an easy way to preview the actual content of the PR its less than ideal. The comment granularity also isn't ideal (e.g. large paragraphs).

rylev:
> Since Borrow is a trait in std (with a very different meaning), it is probably a bad name for this type. Borrowed might be good IMO?

This has [already been noted in the RFC](https://github.com/rylev/windows-rs/blob/6e94a752fb7ba8ab991cb016c8235af2311fe94d/rfcs/0002-borrow-type.md?plain=1#L233). We'll definitely want to figure this out before merging.

Josh015:
How about "Loaned"?

rfcs/0002-borrow-type.md
```diff
@@ -0,0 +1,277 @@
...
+# Reference-level Explanation  
+[reference-level-explanation]: #reference-level-explanation
+
+`Borrow<'a, T>` has the following definition:
+
+```rust
+#![warn(unsafe_op_in_unsafe_fn)]
+
+#[repr(transparent)]
+struct Borrow<'a, T> {
```
kennykerr, line 75: I agree with @nrc, for consistency it should probably be called `Borrowed`, primarily used for fields with a `In` type alias used by code gen for input parameters.

rylev: I'm fine with the `Borrowed` nomenclature. I'm unsure if providing an `In` alias would be helpful. While it would make generated code shorter and thus likely a bit easier to parse, it would require the reader to understand that `In` and `Borrowed` are the same thing.

rfcs/0002-borrow-type.md
```diff
@@ -0,0 +1,277 @@
...
+impl<'a, T> std::ops::Deref for Borrow<'a, T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &*self.item
+    }
+}
+```
+
+## Usages
```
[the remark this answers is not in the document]
rylev: I will add this to the document, but `Option<Borrowed<'a, IUnknown>>` would have the same layout as `Option<IUnknown>` (i.e., a pointer), so that should not be a problem.

rylev:
I've renamed the type to `Borrowed`, and added the question of whether an `In` type alias should be provided.

rfcs/0002-borrow-type.md
kennykerr: Sure, we can always add it later if it becomes obviously valuable.

rfcs/0002-borrow-type.md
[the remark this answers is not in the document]
kennykerr: I was just looking for something a little more fluent, perhaps you can construct a Borrowed from a None and it would produce a zeroed Borrowed value. The thing is that practically everywhere you'd use a Borrowed you would also be expected to accept a zeroed/none value.

kennykerr (approved):

PR #1793, opened 2022-06-02
Avoid redundant signature calculations

kennykerr:
Another technical debt around calculating method signature dependencies. This also fixes a few bugs in `cfg` generation where generic dependencies were skipped.

### DIFF
--- a/crates/libs/windows/src/Windows/Web/Http/Headers/mod.rs
+++ b/crates/libs/windows/src/Windows/Web/Http/Headers/mod.rs
@@ -4229,8 +4229,8 @@ impl HttpLanguageHeaderValueCollection {
             (::windows::core::Interface::vtable(this).TryParseAdd)(::windows::core::Interface::as_raw(this), input.into_param().abi(), result__.as_mut_ptr()).from_abi::<bool>(result__)
         }
     }
-    #[doc = "*Required features: `\"Web_Http_Headers\"`, `\"Foundation_Collections\"`*"]
-    #[cfg(feature = "Foundation_Collections")]
+    #[doc = "*Required features: `\"Web_Http_Headers\"`, `\"Foundation_Collections\"`, `\"Globalization\"`*"]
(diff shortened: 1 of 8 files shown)


crates/libs/windows/src/Windows/Web/Http/Headers/mod.rs
```diff
@@ -4319,17 +4319,17 @@ impl HttpLanguageHeaderValueCollection {
...
         let this = &::windows::core::Interface::cast::<super::super::super::Foundation::Collections::IVector<super::super::super::Globalization::Language>>(self)?;
         unsafe {
             let mut result__ = ::core::mem::MaybeUninit::<u32>::zeroed();
             (::windows::core::Interface::vtable(this).GetMany)(::windows::core::Interface::as_raw(this), startindex, items.len() as u32, ::core::mem::transmute_copy(&items), result__.as_mut_ptr()).from_abi::<u32>(result__)
         }
     }
-    #[doc = "*Required features: `\"Web_Http_Headers\"`, `\"Foundation_Collections\"`*"]
-    #[cfg(feature = "Foundation_Collections")]
+    #[doc = "*Required features: `\"Web_Http_Headers\"`, `\"Foundation_Collections\"`, `\"Globalization\"`*"]
+    #[cfg(all(feature = "Foundation_Collections", feature = "Globalization"))]
```
rylev, line 4332: Why does the required features list still not match what's in the `cfg` attribute?

rylev (approved):
I think this ok though I'm not really familiar with the cfg resolution code. I did have one question though...

crates/libs/windows/src/Windows/Web/Http/Headers/mod.rs
kennykerr: Because the docs include the "current" module whereas the `cfg` attribute need not include that since its already enabled in order to reach this code.

PR #1794, opened 2022-06-02
Remove need for allocation inside Reader

rylev:
This should be easier to understand, less code, have less rightward drift, completely eliminate all allocation, and be much less hacky than some alternatives.


kennykerr:
Here's another related case:

https://github.com/microsoft/windows-rs/blob/04217210a2525810e50b29daffe290b42f5b8551/crates/libs/metadata/src/reader/tree.rs#L21-L24

PR #1797, opened 2022-06-02
Improve `Reader` to `Tree` construction

kennykerr:
This includes a minor change to the `Reader` to let the caller decide which namespaces to exclude.

The `windows-sys` crate now also excludes the Xaml API entirely as it is unusable with `windows-sys`.


crates/libs/metadata/src/reader/mod.rs
```diff
@@ -218,14 +216,14 @@ impl<'a> Reader<'a> {
         }
         Self { files, types, nested }
     }
-    pub fn tree(&self, root: &str) -> Option<Tree> {
+    pub fn tree(&'a self, root: &'a str, exclude: &[&str]) -> Option<Tree> {
```
rylev, line 219: These lifetime annotations seems wrong, but I guess if it works it's fine.

rylev (approved):

crates/libs/metadata/src/reader/mod.rs
kennykerr: Yeah, that's the lifetime annotations the Rust compiler's error message suggested. 🤷

PR #1798, opened 2022-06-03
Implement equality and hash for GUID

faern:
In `winapi` we had the function [`IsEqualGUID`](https://docs.rs/winapi/latest/winapi/shared/guiddef/fn.IsEqualGUID.html) to check GUID equality. `windows-sys` currently does not have this.

Is there a reason `Copy` and `Clone` are not implemented with `#[derive]`? If so, I can implement these traits manually also.

I also threw in `Hash`, since I did not see why it would not implement it. Could be useful in some cases I imagine.

### DIFF
--- a/crates/libs/sys/src/core/mod.rs
+++ b/crates/libs/sys/src/core/mod.rs
@@ -1,3 +1,4 @@
+#[derive(PartialEq, Eq, Hash)]
 #[repr(C)]
 pub struct GUID {
     pub data1: u32,


faern:
I realize you probably want me to use the `windows` crate if I want this convenience?... I then realized that `windows::core::GUID` and `windows_sys::core::GUID` are two completely separate structs. I was assuming `windows` just re-exported stuff from `windows-sys`. :scream:

What's the long term plan for the ecosystem here? Crates using `windows` will not be compatible with crates using `windows-sys`. Sounds to me like this will create a great split in the ecosystem.

kennykerr:
Yes, the windows-sys crate intentionally lacks any of these extra features to reduce build time. The `derive` macro is also avoided for traits that are essential because `derive` is very slow. 

The `windows` crate is where you should go for such features. I'm also exploring how the `windows` crate can use more of these definitions from the `windows-sys` crate.

TomzBench:
Kind of an old post, but maybe you can add a `to_128` similar as the `from_128` and then we can do matching on these guid's easier?

PR #1800, opened 2022-06-03
Reduce allocations in Tree::flatten()

### LINKED ISSUE #1795: Optimize `Tree`'s `flatten` method to avoid allocations
Here's another related case:

https://github.com/microsoft/windows-rs/blob/04217210a2525810e50b29daffe290b42f5b8551/crates/libs/metadata/src/reader/tree.rs#L21-L24

_Originally posted by @kennykerr in https://github.com/microsoft/windows-rs/issues/1794#issuecomment-1145060538_

AronParker:
Initial draft for eliminating allocations in `Tree::flatten`.

Warning: this code is untested so far.

Fixes #1795

### DIFF
--- a/crates/libs/metadata/Cargo.toml
+++ b/crates/libs/metadata/Cargo.toml
@@ -18,3 +18,10 @@ features = [
     "Win32_System_SystemServices",
     "Win32_System_Diagnostics_Debug",
 ]
+
+[dev-dependencies]
+criterion = "0.3.5"
+
+[[bench]]
+name = "my_benchmark"
+harness = false
--- a/crates/libs/metadata/benches/my_benchmark.rs
+++ b/crates/libs/metadata/benches/my_benchmark.rs
@@ -1,0 +1,37 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use windows_metadata::reader::File;
+use windows_metadata::reader::Reader;
+
+const EXCLUDE_NAMESPACES: [&str; 1] = ["Windows.Win32.Interop"];
+
+pub fn criterion_benchmark(c: &mut Criterion) {
+    let files = vec![File::new("default/Windows.winmd").unwrap(), File::new("default/Windows.Win32.winmd").unwrap(), File::new("default/Windows.Win32.Interop.winmd").unwrap()];
+    let reader = &Reader::new(&files);
+    let root = reader.tree("Windows", &EXCLUDE_NAMESPACES).expect("`Windows` namespace not found");
+
+    let mut group = c.benchmark_group("flatten");
+
+    group.bench_function("flatten", |b| {
+        b.iter(|| {
+            let root = black_box(&root);
+            root.flatten()
+        })
+    });
+
+    group.bench_function("flatten_iterative", |b| {
+        b.iter(|| {
+            let root = black_box(&root);
+            root.flatten_iter()
+        })
+    });
+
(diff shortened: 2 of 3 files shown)


AronParker:
Here is a recursive alternative that just requires one `Vec`:

```rust
    pub fn flatten(&'a self) -> Vec<&'a Self> {
        fn visit<'a>(this: &'a Tree<'a>, res: &mut Vec<&'a Tree<'a>>) {
            res.push(this);

            for child in this.nested.values() {
                visit(child, res);
            }
        }

        let mut res = Vec::new();
        visit(self, &mut res);
        res
    }```

**Comment by kennykerr** [comment]
Thanks! I'd be interested to see the difference in performance before accepting an alternative.

**Comment by kennykerr** [comment]
Note that I haven't actually measured `flatten` and it may be negligible to begin with in which case the simplest implementation would probably be best. I'd just prefer to return an iterator on principle. 😏

**Comment by AronParker** [comment]
> Thanks! I'd be interested to see the difference in performance before accepting an alternative.

You're welcome! As I am unfamilar on how the API is used, before doing performance testing I'd require some real world data. It would be very helpful If someone could export the biggest flattened list of namespaces so I can conduct some performance testing.

> Note that I haven't actually measured flatten and it may be negligible to begin with in which case the simplest implementation would probably be best. I'd just prefer to return an iterator on principle. 😏

The question is, how many namespaces are there in total? It's probably relatively small to begin with, but considering Windows' large API surface I wouldn't be surprised if it is very large. Yeah I must admit the current approach is "elegant" in a way and probably easier to maintain and update.

**Comment by kennykerr** [comment]
You can test it as follows:

https://github.com/microsoft/windows-rs/blob/75260cab4c2acc3b6c8f3a767ec5fdf9c82e2d89/crates/tools/windows/src/main.rs#L11-L15

**Comment by AronParker** [comment]
I've added a criterion benchmark to the branch and tested all three variations:

* flatten_master - [current upstream solution](https://github.com/microsoft/windows-rs/blob/10414f60d3d2e905867e191f514951d306f0a7e0/crates/libs/metadata/src/reader/tree.rs#L21-L24)
* flatten_iterative - [the Iterative approach](https://github.com/microsoft/windows-rs/blob/10414f60d3d2e905867e191f514951d306f0a7e0/crates/libs/metadata/src/reader/tree.rs#L26-L36)
* flatten_recursive - [the recursive approach](https://github.com/microsoft/windows-rs/blob/10414f60d3d2e905867e191f514951d306f0a7e0/crates/libs/metadata/src/reader/tree.rs#L38-L52)

Results:
```
flatten/flatten_master  time:   [61.644 us 62.214 us 63.127 us]                                   
Found 11 outliers among 100 measurements (11.00%)
  2 (2.00%) high mild
  9 (9.00%) high severe
flatten/flatten_iterative                                                                             
                        time:   [5.7593 us 5.8906 us 6.0513 us]
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) high mild
  9 (9.00%) high severe
flatten/flatten_recursive                                                                             
                        time:   [4.9838 us 5.0445 us 5.1263 us]
Found 13 outliers among 100 measurements (13.00%)
  8 (8.00%) high mild
  5 (5.00%) high severe
```

VIsualization:

image

kennykerr:
Thanks for taking a look - based on the results it isn't worth fixing and we should just remove the comment and move on. 😃

PR #1801, opened 2022-06-06
Remove `flatten` comment

kennykerr:
Based on #1795 and analysis by @AronParker and @rylev, `flatten` cannot be turned into a pure iterator as Rust doesn't appear to support recursive iterators. https://github.com/rust-lang/rust/issues/97686

Here I'm just removing the comment as there's no more work to be done here.

### DIFF
--- a/crates/libs/metadata/src/reader/tree.rs
+++ b/crates/libs/metadata/src/reader/tree.rs
@@ -19,7 +19,6 @@ impl<'a> Tree<'a> {
         }
     }
     pub fn flatten(&self) -> Vec<&Self> {
-        // TODO: surely there's a way to do this without a ton of intermediate Vec's...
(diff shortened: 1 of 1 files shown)


AronParker (approved):
Seems reasonable. At best performance is improved from [62 to 6 *micro*seconds](https://github.com/microsoft/windows-rs/pull/1800#issuecomment-1146684578), which is probably lost in noise as to my understanding the `flatten` call is only done once and not in a tight loop. Also by merging the recursive or iterative approaches I outlined [here](https://github.com/microsoft/windows-rs/pull/1800#issuecomment-1146684578) a lot of readability/simpliticty/maintainability gets lost. So I'm all for it.

rylev (approved):

PR #1803, opened 2022-06-06
Test error propagation

kennykerr:
As I work on splitting the `core` modules out as separate crates, I noticed we don't have a specific test to ensure error propagation does not regress.


riverar:
Another thing I noticed is when we call `message()`, the call to GetErrorDetails/GetDescription/etc. may result in a message that includes a pesky line break. This results in undesirable side effects when printing these messages, e.g.:

```
Error: Error { code: 0x8007007E, message: The specified module could not be found.
 }
```

riverar (approved):

PR #1804, opened 2022-06-07
Trim error messages

kennykerr:
`FormatMessageW` returns strings with trailing white space characters for certain system error codes. This can then propagate through error objects. It is customary to trim these off for display purposes. 

Fixes https://github.com/microsoft/windows-rs/pull/1803#issuecomment-1148068317

### DIFF
--- a/crates/libs/windows/src/core/mod.rs
+++ b/crates/libs/windows/src/core/mod.rs
@@ -113,3 +113,13 @@ pub fn as_mut_ptr_or_null<T>(value: &mut [T]) -> *mut T {
         value.as_mut_ptr()
     }
 }
+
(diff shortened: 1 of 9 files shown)


riverar (approved):

crates/libs/windows/src/core/mod.rs
```diff
@@ -113,3 +113,13 @@ pub fn as_mut_ptr_or_null<T>(value: &mut [T]) -> *mut T {
         value.as_mut_ptr()
     }
 }
+
+fn wide_trim_end(mut wide: &[u16]) -> &[u16] {
```
rylev, line 117: Nit: it would be nice to provide a short doc about why this helper is necessary. It's obvious in this PR, but in the future, contributors may wonder why we need this.

PR #1806, opened 2022-06-07
Make `Type` methods simpler to use

kennykerr:
Most type queries require the `Reader` but some simple queries can be made into methods for a more streamlined experience.

### DIFF
--- a/crates/libs/metadata/src/reader/type.rs
+++ b/crates/libs/metadata/src/reader/type.rs
@@ -38,6 +38,8 @@ pub enum Type {
 }
 
 impl Type {
+    /// Creates a `Type` object from an `ELEMENT_TYPE` (see ECMA-335) type constant, typically
+    /// used to indicate the type of a constant or primitive type signature.
     pub fn from_code(code: usize) -> Option<Self> {
         match code {
             0x01 => Some(Self::Void),
@@ -60,14 +62,20 @@ impl Type {
             _ => None,
         }
     }
+
+    /// Converts the `Type` to an equivalent `const` variant if appropriate, typically used when the
+    /// mutability is informed by something outside of the type signature.
     pub fn to_const(self) -> Self {
         match self {
-            Self::MutPtr((kind, pointers)) => Self::ConstPtr((kind, pointers)),
+            Self::MutPtr(p) => Self::ConstPtr(p),
             Self::PSTR => Self::PCSTR,
             Self::PWSTR => Self::PCWSTR,
             _ => self,
         }
     }
+
+    /// Removes one level of indirection, typically used when transforming a logical return or array parameter
+    /// from its underlying type signature.
     pub fn deref(&self) -> Self {
         match self {
             Self::ConstPtr((kind, 1)) | Self::MutPtr((kind, 1)) => {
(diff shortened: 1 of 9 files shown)


crates/libs/metadata/src/reader/type.rs
```diff
@@ -0,0 +1,113 @@
...
+            0x0c => Some(Self::F32),
+            0x0d => Some(Self::F64),
+            0x18 => Some(Self::ISize),
+            0x19 => Some(Self::USize),
+            0x0e => Some(Self::String),
+            0x1c => Some(Self::IInspectable),
+            _ => None,
+        }
+    }
+    pub fn to_const(self) -> Self {
```
rylev, line 63: nit: it's more common for there to be a blank line between methods.

crates/libs/metadata/src/reader/type.rs
```diff
@@ -0,0 +1,113 @@
...
+        match self {
+            Self::MutPtr((kind, pointers)) => Self::ConstPtr((kind, pointers)),
```
rylev, line 65:
```suggestion
            Self::MutPtr(p) => Self::ConstPtr(p),
```

crates/libs/metadata/src/reader/type.rs
```diff
@@ -0,0 +1,113 @@
...
+                    Self::U8
+                } else {
+                    *kind.clone()
+                }
+            }
+            Self::ConstPtr((kind, pointers)) => Self::ConstPtr((kind.clone(), pointers - 1)),
+            Self::MutPtr((kind, pointers)) => Self::MutPtr((kind.clone(), pointers - 1)),
+            Self::PSTR | Self::PCSTR => Self::U8,
+            Self::PWSTR | Self::PCWSTR => Self::U16,
+            _ => unimplemented!(),
```
rylev, line 84: This should just be a `panic!` with a message indicating that `deref` can only be called on pointer types.

crates/libs/metadata/src/reader/type.rs
```diff
@@ -0,0 +1,113 @@
...
+    MutPtr((Box<Self>, usize)),
+    ConstPtr((Box<Self>, usize)),
+    Win32Array((Box<Self>, usize)),
+    WinrtArray(Box<Self>),
+    WinrtArrayRef(Box<Self>),
+    WinrtConstRef(Box<Self>),
+}
+
+impl Type {
+    pub fn from_code(code: usize) -> Option<Self> {
```
rylev, line 41: It would be nice if these methods had some docs associated with them. For those new to the codebase, it might not be clear at first what `code` is referring to.

rylev (dismissed):
I like this! I do have some nits but nothing blocking.

crates/libs/metadata/src/reader/type.rs
kennykerr: Done.

crates/libs/metadata/src/reader/type.rs
kennykerr: Done.

crates/libs/metadata/src/reader/type.rs
kennykerr: We should get `clippy` to do that...

PR #1809, opened 2022-06-08
Avoid truncating error codes when converting from `windows::core::Error` to `std::io::Error`

### LINKED ISSUE #1807: Bug: Conversion from windows::core::Error to std::io::Error changes the error code

windows

### Crate version

0.37.0

### Summary

Conversion from a `windows::core::Error` to a `std::io::Error` messes up the error code because it is truncated to 2 bytes only:

https://github.com/microsoft/windows-rs/blob/0.37.0/crates/libs/windows/src/core/error.rs#L91=

This is bothersome, some API calls can return error codes that are bigger than 65536, and the conversion transforms the error into another error entirely.

For example, a call to `FilterReplyMessage` can return `ERROR_FLT_NO_WAITER_FOR_REPLY`, which is `0x801F0020`. Conversion to `std::io::Error` generates an error code `0x20`, which is `ERROR_SHARING_VIOLATION`.

What is the purpose of this truncation? Shouldn't the `io::Error` just use the i32 code as is without any modification?

### Reproducible example

```rust
use windows::core::Error;
use windows::Win32::Foundation::ERROR_FLT_NO_WAITER_FOR_REPLY;

fn main() {
    // Lets imagine I call a windows function such as `FilterReplyMessage`, that returns an error:
    let error = Error::new(ERROR_FLT_NO_WAITER_FOR_REPLY, Default::default());
    dbg!(&error);
    println!("initial error: {}", &error);

    let io_error: std::io::Error = error.into();
    dbg!(&io_error);
    println!("after cast into io::Error: {}", &io_error);
}
```

Output:

```
[src/main.rs:7] &error = Error {
    code: 0x801F0020,
    message: No waiter is present for the filter's reply to this message.
    ,
}
initial error: No waiter is present for the filter's reply to this message.

[src/main.rs:11] &io_error = Os {
    code: 32,
    kind: Uncategorized,
    message: "The process cannot access the file because it is being used by another process.",
}
after cast into io::Error: The process cannot access the file because it is being used by another process. (os error 32)
```

### Crate manifest

```
[package]
edition = "2021"
name = "windows-issue-1807"
version = "0.1.0"

[dependencies]
windows = { version = "0.37", features = [ "Win32_Foundation" ] }
```

### Expected behavior

Conversion to `std::io::Error` should not change the error code

### Actual behavior

A `ERROR_FLT_NO_WAITER_FOR_REPLY` error becomes a `ERROR_SHARING_VIOLATION` error when converted to `std::io::Error`.

kennykerr:
Fixes #1807

### DIFF
--- a/crates/libs/windows/src/core/error.rs
+++ b/crates/libs/windows/src/core/error.rs
@@ -77,7 +77,7 @@ impl core::convert::From<Error> for HRESULT {
 
 impl core::convert::From<Error> for std::io::Error {
     fn from(from: Error) -> Self {
-        Self::from_raw_os_error((from.code.0 & 0xFFFF) as _)
+        Self::from_raw_os_error(from.code.0)
     }
 }
 
--- a/crates/tests/error/tests/std.rs
+++ b/crates/tests/error/tests/std.rs
@@ -1,0 +1,34 @@
+use windows::Win32::Foundation::*;
+
+#[test]
+fn conversions() {
+    // Baseline HRESULT
+    assert_eq!(E_INVALIDARG.message(), "The parameter is incorrect.");
+    assert_eq!(E_INVALIDARG.0, -2147024809);
+
+    // Baseline WIN32_ERROR
+    assert_eq!(ERROR_INVALID_DATA.to_hresult().message(), "The data is invalid.");
+    assert_eq!(ERROR_INVALID_DATA.0, 13);
+
+    // std::io::Error from HRESULT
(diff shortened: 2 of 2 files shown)


ChrisDenton:
Looks good to me. Would it be good to document this somewhere? Maybe in the FAQ? How ever the conversion is implemented, it's going to be surprising to someone in some situation.

kennykerr:
Good idea, cc @rylev regarding FAQ.

Issue #816, opened 2021-05-22
Mapping `Option` to nullable types


gladiator:
As far as I'm aware, it isn't possible to reflect the nullable strings that are used in certain APIs with `Option`.  For example, `CreateProcessA` takes many parameters that can be null to represent different configurations for creating the target process.

```rust
unsafe fn create_process(app: Option<&str>, cmd: Option<&str>, curdir: Option<&str>, ...) -> ... {
    let _ = CreateProcessA(
        app,
        cmd,
        ...,
        curdir,
        ...,
    );

    // ...
}
```

As you would expect, the `None` type would be implicitly converted to `PSTR::NULL` which I think might be useful when trying to write idiomatic code, especially libraries.

~~I _think_ something like this should be fairly easy to implement as some string types [seem to have hand-written implementation details](https://github.com/microsoft/windows-rs/blob/master/crates/gen/src/types/pstr.rs).~~

Ideally, this is probably something that could be handled for every nullable type and not just `PSTR` and friends, right?

gladiator:
Sorry for posting this early, accidentally hit enter as I was typing the title...

gladiator:
Sorry for closing and then immediately reopening... had some struggles with my ability to use a computer.

gladiator:
I've just noticed #292, almost definitely related?

kennykerr:
Hi Daniel, #292 is specifically about WinRT reference parameters. For `CreateProcessA` there is some support for `Option` but more work is needed. Here's a complete example:

```rust
use bindings::Windows::Win32::System::Threading::*;

fn main() {
    unsafe {
        CreateProcessA(
            "c:\\windows\\notepad.exe",
            None,
            std::ptr::null_mut(),
            std::ptr::null_mut(),
            false,
            Default::default(),
            std::ptr::null_mut(),
            None,
            &mut STARTUPINFOA {
                cb: std::mem::size_of::<STARTUPINFOA>() as _,
                ..Default::default()
            },
            &mut PROCESS_INFORMATION::default(),
        );
    }
}
```

As you can see, the optional string parameters can be replaced with `None`. I plan to also support this for the remaining optional pointer parameters when I get a moment. There's also https://github.com/microsoft/win32metadata/issues/433 to simplify the `STARTUPINFOA` initialization.

kennykerr:
I'll keep this issue open to track the remaining usability issues. Ideally, we can write it more simply as follows:

```rust
CreateProcessA(
[...]

PR #1811, opened 2022-06-09
Borrowed

rylev:
This implements the [Borrow Type RFC](https://github.com/microsoft/windows-rs/blob/master/rfcs/0002-borrow-type.md) which makes conversion to input parameters much more straightforward.

## Changes

There are some changes in this PR from the RFC:

### Changes to Borrowed

During development, it was discussed how `Borrowed` types are always logically optional. This PR adjusts the definition of `Borrowed` to be a wrapper around `<T as Abi>::Abi`. 

### Addition of `InParam`

Additionally, an `InParam` type is added which is essentially either an owned `T` or a `Borrowed<'a, T>`. This allows us to hold on params that have been converted to the correct type at the last minute dropping them at the end of the function call. 

## Changes to the user experience

The following are the practical changes to the UX of the `windows` crate after this change:
* Most parameters need to be passed as references now instead of being able to pass ownership
* Working with `HSTRING` and `PCWSTR` (and kin) gets worse
  * This now requires the user to do a manual conversion from a string literal to these types before passing them as parameters
* You used to be able to pass `Default::default()` to a lot of places but now you need to be explicit with the type 
  * For example,  `URI_CREATE_FLAGS::default()` 

[To view the changes without codegen, look here](https://github.com/microsoft/windows-rs/pull/1811/files/742a3076bcf8b1c7a38e44b0a5fa39b0dc04b578..4c860b27c20fca100cfa36c7424149a92e56cfb7)


crates/libs/windows/src/core/borrowed.rs
```diff
@@ -1,50 +1,77 @@
...
+
+impl <'a, T: 'a , U> From<&'a U> for Borrowed<'a, T> where T: super::Abi, &'a U: Into<&'a T> {
+    fn from(item: &'a U) -> Self {
+        Borrowed::new(Some(item.into()))
     }
-}
\ No newline at end of file
+}
+
+impl <'a, T: super::Abi> From<Option<&'a T>> for Borrowed<'a, T> where  &'a T: Into<Borrowed<'a, T>> {
```
kennykerr, line 73: Typically, this is going to be `&Option<T>` rather than `Option<&T>`. Not sure if we need both but the former is definitely more useful in practice. e.g. 

https://github.com/microsoft/windows-rs/blob/11a2db2a161c5bef674573ac4e2121e8c045c708/crates/samples/direct2d/src/main.rs#L154-L159

crates/libs/windows/src/core/borrowed.rs
```diff
@@ -1,50 +1,77 @@
...
-/// Many APIs take a params of the form `Into<Borrowed<'a, T>>`. This means you can pass anything that has an `Into` implementation 
-/// for `Borrowed<'a, T>`. Generally, if this is safe to do, the `windows` crate provides an implementation.
+/// ```
+///  the trait bound `SomeType: Into<Borrowed<'_, SomeType>>` is not satisfied
+/// ```
 #[repr(transparent)]
 pub struct Borrowed<'a, T> {
-	item: core::mem::ManuallyDrop<T>,
-    lifetime: core::marker::PhantomData<&'a ()>
+    item: Option<core::mem::ManuallyDrop<T>>,
```
kennykerr, line 34: I'm not sure `Option<ManuallyDrop>` is appropriate since it won't necessarily be the same size as `T` if `T` is not a transparent pointer. `T` and `Borrowed<T>` must have the same size.

crates/libs/windows/src/core/borrowed.rs
kennykerr: Getting the Direct2D sample to compile would be a good smoke test.

crates/libs/windows/src/core/borrowed.rs
kennykerr: But just storing a `ManuallyDrop` that uses `zeroed` for the `None` case should work fine but implementing `as_ref` might be tricky.

rylev: We're now storing `T::Abi` instead.

crates/libs/windows/src/core/borrowed.rs
rylev: Here are the changes to get d2d working: https://github.com/microsoft/windows-rs/pull/1811/commits/792a1e975a77f0261fc5bc457a58e7fe653cd3c8

crates/samples/direct2d/src/main.rs
```diff
@@ -156,13 +156,13 @@ impl Window {
             target.SetTarget(clock);
             target.Clear(std::ptr::null());
             self.draw_clock()?;
-            target.SetTarget(previous);
+            target.SetTarget(previous.as_ref());
```
kennykerr, line 159: Seems to have lost my previous comments on the other commit. Ideally this would be `&previous` if we can make `Borrowed` creatable from `&Option<T>`.

crates/samples/direct2d/src/main.rs
```diff
@@ -209,15 +209,15 @@ impl Window {
         unsafe {
             target.SetTransform(&(Matrix3x2::rotation(angles.second, 0.0, 0.0) * translation));
 
-            target.DrawLine(D2D_POINT_2F::default(), D2D_POINT_2F { x: 0.0, y: -(radius * 0.75) }, brush, radius / 25.0, &self.style);
+            target.DrawLine(&D2D_POINT_2F::default(), &D2D_POINT_2F { x: 0.0, y: -(radius * 0.75) }, brush, radius / 25.0, &self.style);
```
kennykerr, line 212: These should be passed by value (when the underlying metadata passes structs by value). It's a fine line because you don't want to pass non-blittable structs by value even if the underlying metadata does.

crates/samples/direct2d/src/main.rs
```diff
@@ -451,7 +451,7 @@ fn create_device() -> Result<ID3D11Device> {
 
 fn create_render_target(factory: &ID2D1Factory1, device: &ID3D11Device) -> Result<ID2D1DeviceContext> {
     unsafe {
-        let d2device = factory.CreateDevice(device.cast::<IDXGIDevice>()?)?;
+        let d2device = factory.CreateDevice(&device.cast::<IDXGIDevice>()?)?;
```
kennykerr, line 454: Can this now be inferred? `&device.cast()?`

crates/samples/direct3d12/src/main.rs
```diff
@@ -73,18 +73,20 @@ where
...
 
+    title.push('\0');
+
     let hwnd = unsafe {
         CreateWindowExA(
-            Default::default(),
-            "RustWindowClass",
-            title,
+            WINDOW_EX_STYLE::default(),
+            PCSTR(b"RustWindowClass".as_ptr()),
```
kennykerr, line 81: Missing a null terminator. `HSTRING` would work if we used `CreateWindowExW` instead...

crates/samples/kernel_event/src/main.rs
```diff
@@ -1,11 +1,12 @@
 use windows::{
+    core::PCWSTR,
     Win32::Foundation::CloseHandle,
     Win32::System::Threading::{CreateEventW, SetEvent, WaitForSingleObject, WAIT_OBJECT_0},
 };
 
 fn main() -> windows::core::Result<()> {
     unsafe {
-        let event = CreateEventW(std::ptr::null(), true, false, None)?;
+        let event = CreateEventW(std::ptr::null(), true, false, PCWSTR::default())?;
```
kennykerr, line 9: Ideally, we'd have more support for `Option` rather than less. https://github.com/microsoft/windows-rs/issues/816

crates/samples/message_box/src/main.rs
```diff
@@ -1,7 +1,9 @@
+use windows::core::PCSTR;
+use windows::Win32::Foundation::HWND;
 use windows::Win32::UI::WindowsAndMessaging::{MessageBoxA, MB_OK};
 
 fn main() {
     unsafe {
-        MessageBoxA(None, "Hello", "World", MB_OK);
+        MessageBoxA(HWND::default(), PCSTR("Hello".as_bytes().as_ptr()), PCSTR("World".as_bytes().as_ptr()), MB_OK);
```
kennykerr, line 7: This is a great litmus test. 😄

crates/samples/ocr/src/main.rs
```diff
@@ -13,14 +13,14 @@ async fn main_async() -> Result<()> {
     let mut message = std::env::current_dir().unwrap();
     message.push("message.png");
 
-    let file = StorageFile::GetFileFromPathAsync(message.to_str().unwrap())?.await?;
+    let file = StorageFile::GetFileFromPathAsync(&HSTRING::from(message.to_str().unwrap()))?.await?;
     let stream = file.OpenAsync(FileAccessMode::Read)?.await?;
 
-    let decode = BitmapDecoder::CreateAsync(stream)?.await?;
+    let decode = BitmapDecoder::CreateAsync(&stream)?.await?;
```
kennykerr, line 19: Ignoring strings, WinRT seems very happy with this change.

crates/libs/windows/src/core/abi.rs
```diff
@@ -4,80 +4,107 @@ use super::*;
...
-// SAFETY: raw pointers are FFI safe, `Abi` & `Self` are the same so thus they have the same in-memory representation, and
-// all representations of `Abi` are valid representations of `Self`.
-unsafe impl<T> Abi for *mut T {
-    type Abi = Self;
+    /// Whether `abi` is known to be invalid
+    ///
+    /// Note: this does not guarantee that
+    fn abi_is_possibly_valid(abi: &Self::Abi) -> bool {
+        true
+    }
```
kennykerr, line 56: AlrightGIF

crates/libs/windows/src/core/borrowed.rs
```diff
@@ -1,50 +1,77 @@
...
-    ///
-    /// For example, `IUnknown` is fine to alias as long as the reference count is guaranteed to stay above 0 
-    /// until the original `item` is dropped. On the other hand `Box<T>` is not fine to alias and thus passing a `Box` 
-    /// to this function would be UB.
-    unsafe fn new(item: &'a T) -> Self {
-        let item = unsafe { core::mem::transmute_copy(item) };
-        Self {
-            item,
-            lifetime: core::marker::PhantomData
+    pub fn new(item: Option<&'a T>) -> Self {
```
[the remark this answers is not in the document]
kennykerr: Resolved?

crates/samples/direct2d/src/main.rs
kennykerr: `type_def_is_blittable` or `type_is_blittable` should work.

crates/samples/direct2d/src/main.rs
rylev: Unfortunately not...

crates/samples/core_app/src/main.rs
```diff
@@ -59,12 +59,16 @@ fn main() -> Result<()> {
         CoInitializeEx(std::ptr::null(), COINIT_MULTITHREADED)?;
 
         if let Err(result) = Package::Current() {
-            MessageBoxW(HWND::default(), "This sample must be registered (via register.cmd) and launched from Start.", "Error", MB_ICONSTOP | MB_OK);
+            MessageBoxW(HWND::default(), pcwstr(&"This sample must be registered (via register.cmd) and launched from Start.".into()), pcwstr(&"Error".into()), MB_ICONSTOP | MB_OK);
```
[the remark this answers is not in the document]
rylev: Done!

crates/samples/direct2d/src/main.rs
rylev: Unfortunately, this is not possible as Rust's type system doesn't know that `&Option<T>` does not implement `Into<&'a T>`  and thus we get a conflicting impl error if I add an implementation to make `&Option<T>` work. It's unfortunate, but workable IMO.

crates/samples/direct2d/src/main.rs
rylev: Done!

rylev:
As far as I'm aware, the following are the things left to address before we merge this:
* [ ] WinRT interfaces that are non-exclusive currently need to be manually converted to the expected interface before being passed to WinRT methods. It would be nice if users didn't have to do this casting.
* [ ] GenericParameters must always be borrowed even if they end up being blittable. Most likely, we can't do anything about this though...
* [ ] Instead of being able to pass `None`, we now require instantiating many handle types: i.e., `None` => `HWND::default()`. 
* [ ] [TODO](https://github.com/microsoft/windows-rs/blob/b710fc2c833c47b9144358754d0862fd92847bec/crates/tests/handles/tests/win.rs#L46-L47)

crates/tests/handles/tests/win.rs
```diff
@@ -43,7 +43,8 @@ fn hfont() {
         let font: HFONT = HFONT(underlying);
         let object: HGDIOBJ = HGDIOBJ(font.0);
 
-        assert!(!DeleteObject(font).as_bool());
+        // TODO:
+        // assert!(!DeleteObject(font).as_bool());
         assert!(!DeleteObject(object).as_bool());
```
kennykerr, line 48: This tests support for the `AlsoUsableFor` attribute, which lets one handle type e.g. `HFONT` be used where another handle type .e.g. `HGDIOBJ` is expected. This was previously supported via `IntoParam` we could support this now with an explicit `Into`?

rylev: Fixed!

crates/tests/winrt/tests/composition.rs
```diff
@@ -68,8 +69,8 @@ fn composition() -> windows::core::Result<()> {
 
     // Visual.set_brush expects a CompositionBrush but CreateColorBrushWithColor returns a
     // CompositionColorBrush that logically derives from CompositionBrush.
-    let brush = compositor.CreateColorBrushWithColor(&red)?;
-    visual.SetBrush(brush)?;
+    let brush = compositor.CreateColorBrushWithColor(red)?;
+    visual.SetBrush(&CompositionBrush::from(brush))?;
```
kennykerr, line 73: This is unfortunate. We should consider what it would take to make this work for class hierarchies like this. This is a case where the parameter always expects a class (not an interface) and there is a closed set of possibilites. This is somewhat like exclusive interfaces.

crates/tests/winrt/tests/composition.rs
```diff
@@ -84,15 +85,15 @@ fn composition() -> windows::core::Result<()> {
 
     let child = compositor.CreateSpriteVisual()?;
     child.SetOffset(Vector3 { X: 1.0, Y: 0.0, Z: 0.0 })?;
-    children.InsertAtBottom(child)?;
+    children.InsertAtBottom(&Visual::from(child))?;
```
kennykerr, line 89: Same here.

crates/tests/winrt/tests/async.rs
```diff
@@ -1,17 +1,19 @@
+use std::convert::TryFrom;
+
 #[test]
 fn async_get() -> windows::core::Result<()> {
     use windows::Storage::Streams::*;
 
     let stream = &InMemoryRandomAccessStream::new()?;
 
-    let writer = DataWriter::CreateDataWriter(stream)?;
+    let writer = DataWriter::CreateDataWriter(&IOutputStream::try_from(stream)?)?;
```
kennykerr, line 9: This is an example of a parameter needing to cast from some class type to a non-exclusive interface type that should happen implicitily because we know that `InMemoryRandomAccessStream` implements `IOutputStream`.

crates/tests/map/tests/test.rs
```diff
@@ -107,16 +107,16 @@ impl IIterable_Impl<IKeyValuePair<i32, f32>> for Map {
 fn main() -> Result<()> {
     let map: IMap<i32, f32> = Map().into();
     map.Clear()?;
-    map.HasKey(0)?;
-    map.Insert(0, 0.0)?;
-    map.Lookup(0)?;
-    map.Remove(0)?;
+    map.HasKey(&0)?;
```
kennykerr, line 110: Perhaps we need a separate impl for generic interfaces depending on whether T/K/V is a primitive or blittable type...

crates/tests/winrt/tests/object.rs
```diff
@@ -3,7 +3,7 @@ use windows::Foundation::{IStringable, PropertyValue, Uri};
 
 #[test]
 fn class() -> windows::core::Result<()> {
-    let uri = Uri::CreateUri("http://kennykerr.ca")?;
+    let uri = Uri::CreateUri(&windows::core::HSTRING::from("http://kennykerr.ca"))?;
```
kennykerr, line 6: Perhaps we just need to have a universal PCWSTR type for all wide string input parameters to avoid the complication of dealing with the various different types. Then we can specialize that one to handle all the marshaling under the hood - that should simplify and harmonize the various calling patterns for strings. Not for this PR but need to land that fairly closely along with this.

crates/tests/winrt/tests/async.rs
kennykerr: Also its dependent on the version in which the non-exclusive interface was introduced. In this case, we know that the interface was introduced in v1 of the runtime class therefore a `From` will suffice and a `TryFrom` is not required.

kennykerr: If we cannot get rid of all the `from` and `try_from` requirements then perhaps we just drop the `Into<Borrowed<T>>` and just use `Borrowed<T>` everywhere. At least then it will be consistent and allow for `object.into()` to work with type inference.

rylev:
@kennykerr I believe this is ready to go. All of the outstanding issues have been fixed with the exception of how to handle strings which we agreed we should do in a follow up. 

Here are the changes since your last review: ([changes](https://github.com/microsoft/windows-rs/pull/1811/files/7c729709f9230e9e6df6713edd70230e8ba7a267..621d5352141e90f4aeba399951c3cfa827aef549))

filips123:
Are there any plans to address the situation with string parameters? Having to use `HSTRING::from` around every param is kinda annoying and I would expect the `windows` crate to take care of this automatically (as it is supposed to provide a more friendly API). Would it be possible to implement `From<&str>`, `From<String>`, etc. on `PCWSTR` and other string types or handle the conversion automatically in some other way?

rylev:
This is a tricky question: Automatically converting from `&str` or `String` to `HSTRING` is arguably too much to happen implicitly. This is because such conversions require completely new allocations since Rust string types are utf-8 while HSTRING is utf-16. I would argue that in such cases requiring the extra call to `HSTRING::from` is warranted as it makes this cost explicit. 

In fact, this was the reason that the previous `alloc` feature was present as it allowed a way for these implicit conversions to happen. We *could* bring that feature back, but I'm not sure it's a good idea. 

For string literals, the `w!` macro should be preferred. When not dealing with literals, you'll indeed have to manually convert between Rust types and the Windows types. You might want to look into using a third-party crate that allows string like functions on wide strings. This might allow a more convenient way to stay in a representation Windows expects.

PR #1812, opened 2022-06-09
Avoid string allocations in `Cfg`

kennykerr:
Code gen time continues to be painful for me. I've been meaning to remove the allocation of strings inside `Cfg` as that's a really hot bit of code so here goes.

Unfortunately, the vast majority of time running code gen is actually spent on `rustfmt` and not code gen itself. For example, the main code generator - `tool_windows` - takes about 10 seconds (under 3 seconds in release) to run without `rustfmt` and over 150 seconds with `rustfmt`.

### DIFF
--- a/crates/libs/metadata/src/reader/mod.rs
+++ b/crates/libs/metadata/src/reader/mod.rs
@@ -132,15 +132,14 @@ pub struct SignatureParam {
 }
 
 #[derive(Default, Clone)]
-pub struct Cfg {
-    // TODO: use String for now and maybe StringRef if that's too slow
-    pub types: BTreeMap<String, BTreeSet<TypeDef>>,
+pub struct Cfg<'a> {
+    pub types: BTreeMap<&'a str, BTreeSet<TypeDef>>,
     pub arches: BTreeSet<&'static str>,
 }
 
-impl Cfg {
-    pub fn add_feature(&mut self, feature: &str) {
-        self.types.entry(feature.to_string()).or_default();
+impl<'a> Cfg<'a> {
+    pub fn add_feature(&mut self, feature: &'a str) {
+        self.types.entry(feature).or_default();
     }
     pub fn union(&self, other: &Self) -> Self {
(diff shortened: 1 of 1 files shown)


MarijnS95:
> the main code generator - `tool_windows` - takes about 10 seconds (under 3 seconds in release) to run without `rustfmt` and over 150 seconds with `rustfmt`.

That's interesting: for me on Linux (with some tweaks to disable `#[link(name = "windows")]` in `-sys` and simplifying it in the core `windows` crate - issue coming up) it runs much, much faster than that:

```
cargo r --release -p tool_windows  213.75s user 24.07s system 2154% cpu 11.041 total
```

Have you optimized this in the past 12 days? Even on a 32C/64T system I didn't expect to be ±13.6 times faster :)

rylev:
Did some experimentation with `prettyplease`. Interestingly, since the `rustfmt` binary is already compiled in release mode, the use of `prettyplease` only improves things in release mode.

When running `tool_windows` in release mode on my Windows machine, I get the following results:

* `rustfmt`: ~41s
* `prettyplease`: ~16s
* no formatting: ~3s

When running in debug:
* `rustfmt`: ~58s
* `prettyplease`: ~92s
* no formatting: ~19s

It's also important to note the two added dependencies (`prettyplease` and `syn`) do increase the release build times:
* without: 13s
* with: 25s

This means that for release builds from scratch:
*  `rustfmt`: ~54s
* `prettyplease`: ~41s

Some additional tradeoffs:
* using `prettyplease` means that we don't rely on the user having `rustfmt` installed. Albeit rustfmt is pretty essential so the likelihood of this is pretty small.
* `prettyplease`'s formatting is a bit worse, but from a spot check it doesn't look *too* bad. The biggest issue is that it uses a *lot* more lines (presumably since it doesn't respect the rustfmt config where we allow for wider lines). We could use `prettyplease` in development and `rustfmt` for the final change.

This does sound promising enough to warrant thinking about. @kennykerr thoughts?

PR #1826, opened 2022-06-17
Selective `rustfmt` configuration

kennykerr:
The repo's `rustfmt.toml` applies a `max_width` of 800 to reduce the overall size of the `windows` and `windows-sys` crates, but this can be annoying for readability when it comes to the rest of the repo. Here I'm just applying it more selectively to the lib crates. We can apply it further to only the `windows` and `windows-sys` crates but I thought I'd start here first to avoid merge hell for Ryan and I. 😊

This should make the samples far more readable.

### DIFF
--- a/crates/libs/rustfmt.toml
+++ b/crates/libs/rustfmt.toml
@@ -1,1 +1,2 @@
 newline_style = "Unix"
+max_width = 800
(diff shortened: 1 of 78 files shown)


kennykerr:
Hmm, looks like the way we run `rustfmt` through piped tokens doesn't pick up the right `rustfmt.toml` - back to the drawing board.

PR #1829, opened 2022-06-18
Avoid UB in `FactoryCache`

### LINKED ISSUE #1827: Bug: Undefined Behavior Possible in windows::core::FactoryCache

windows

### Crate version

0.37 (Noticed in 0.28, confirmed still in codebase as of 0.37)

### Summary

Many WinRT constructors use a static mut windows::core::FactoryCache to instantiate their objects. For example, consider this code snippet from DataWriter:

```
impl DataWriter {
    pub fn new() -> ::windows::core::Result<Self> {
        Self::IActivationFactory(|f| f.ActivateInstance::<Self>())
    }

    fn IActivationFactory<R, F: FnOnce(&::windows::core::IGenericFactory) -> ::windows::core::Result<R>>(callback: F) -> ::windows::core::Result<R> {
        static mut SHARED: ::windows::core::FactoryCache<DataWriter, ::windows::core::IGenericFactory> = ::windows::core::FactoryCache::new();
        unsafe { SHARED.call(callback) }
    }
```

So `DataWriter::new()` calls `Self::IActivationFactory`, which calls the `call` method on a static mut instance of a `FactoryCache`. Looking at the definition of FactoryCache::call, we see:

```
pub fn call<R, F: FnOnce(&I) -> Result<R>>(&mut self, callback: F) -> Result<R> {
```

The key point is that `call` takes `&mut self`. In conjunction with the unsafe invocation of the static mut instance, this creates the potential for undefined behavior. As an example, consider the following possible scenario:

* Thread A calls `DataWriter::new()` to instantiate a` DataWriter`. This leads to invoking the `FactoryCache::call()` method on the static instance.
* During this time, Thread B also calls `DataWriter::new()`. Nothing seems to prevent this. This also leads to invoking `FactoryCache::call()` on the same static instance as Thread A.
* Both threads are now concurrently holding a mutable reference to the same object, which is undefined behavior by Rust aliasing rules. This is only possible because of the usage of unsafe to suppress the compiler error.

Interestingly, the implementation of `FactoryCache::call` does not appear to actually mutate `self`, at least in the Rust sense; there is some logical mutation in the nested AtomicPtr but that is irrelevant to the use of &mut. It does suggest a simple fix, though: `&mut self` could be replaced with just `&self`. This should fix the UB by itself. It also enables a bonus improvement, since then the static mut FactoryCache could be changed to just a static FactoryCache and the unsafe keyword could be removed, slightly reducing the unsafe exposure.

N/A: code review only

### Reproducible example

```rust
N/A: code review only
```

### Crate manifest

```toml
N/A: code review only
```

### Expected behavior

N/A: code review only

### Actual behavior

N/A: code review only

### Additional comments

Thanks for all work on this crate!

kennykerr:
Fixes #1827 

A similar issue applies to `Event` (depending on usage) but that's more complex to avoid so I'll address that separately.

### DIFF
--- a/crates/libs/windows/src/core/factory_cache.rs
+++ b/crates/libs/windows/src/core/factory_cache.rs
@@ -44,7 +44,7 @@ impl<C: RuntimeName, I: Interface> FactoryCache<C, I> {
     }
 }
 
-unsafe impl<C, I> core::marker::Send for FactoryCache<C, I> {}
+// This is safe because `FactoryCache` only holds agile factory pointers, which are safe to cache and share between threads.
 unsafe impl<C, I> ::core::marker::Sync for FactoryCache<C, I> {}
 
 /// Attempts to load the factory interface for the given WinRT class
--- a/crates/libs/windows/src/core/factory_cache.rs
+++ b/crates/libs/windows/src/core/factory_cache.rs
@@ -17,7 +17,7 @@ impl<C, I> FactoryCache<C, I> {
 }
 
 impl<C: RuntimeName, I: Interface> FactoryCache<C, I> {
-    pub fn call<R, F: FnOnce(&I) -> Result<R>>(&mut self, callback: F) -> Result<R> {
+    pub fn call<R, F: FnOnce(&I) -> Result<R>>(&self, callback: F) -> Result<R> {
         loop {
             // Attempt to load a previously cached factory pointer.
             let ptr = self.shared.load(Ordering::Relaxed);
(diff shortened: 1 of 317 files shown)


mattelkinsTC (approved):
I didn't exactly study each instance of generated code, but this looks generally correct to me.

crates/libs/windows/src/core/factory_cache.rs
```diff
@@ -44,6 +44,9 @@ impl<C: RuntimeName, I: Interface> FactoryCache<C, I> {
     }
 }
 
+unsafe impl<C, I> core::marker::Send for FactoryCache<C, I> {}
+unsafe impl<C, I> ::core::marker::Sync for FactoryCache<C, I> {}
```
rylev, line 48: Can we document why `Send` and `Sync` are safe to implement for `FactoryCache`?

rylev (approved):
🎉🎉

crates/libs/windows/src/core/factory_cache.rs
kennykerr: Done.

PR #1848, opened 2022-06-24
Strings RFC

rylev:
This is an RFC for how we'll handle the many types of Strings in Rust and Windows and how they interact with each other.

This is a complicated question, so I imagine we'll probably not get it right, but I do believe this is an improvement over the status quo not only in terms of useability but also safety.

### DIFF
--- a/rfcs/003-strings.md
+++ b/rfcs/003-strings.md
@@ -1,0 +1,325 @@
+# Strings in windows-rs
+
+## Summary
+
+Rust and Windows each have a fair amount of string types. This RFC proposes a system for making it easy and obvious how to make the string types from both environments work well together.
+
+## Motivation
+
+Making Rust and Windows string types work together is integral to ensuring correct, performant, and boiler-plate free code.
+
+### What we want to accomplish
+
+In general, we are trying to accomplish two goals:
+
+* Correctly model the Windows string types in a way that leverages Rust's type and ownership systems to prevent incorrect use.
+* Allow for easy and low cost ways to convert between Windows and Rust types.
+
+On top of this, we want these APIs to feel natural and obvious to Rust developers.
+
+## Explanation
+
+### The Different Types of Strings
+
+#### Windows
+
+Windows has the following string types:
+
+* `PSTR`: a mutable, nul-terminated string (i.e., composed of "characters" a.k.a. `u8`s) that is often but not always ANSI encoded.
+* `PCSTR`: an immutable version of `PSTR`.
+  * **QUESTION**: what are the practical differences between this and `PSTR`?
+* `PWSTR`: a mutable, nul-terminated "wide" string (i.e., composed of "wide characters" a.k.a. `u16`s) that is often but not always UTF-16 encoded.
+* `PCWSTR`: an immutable version of `PWSTR`.
+* `BSTR`: an immutable, nul-terminated "wide" string used in COM.
+* `HSTRING`: an immutable reference counted, nul-terminated "wide" string used in WinRT.
+  * Note: `HSTRING`s can sometimes be "fast pass" where the buffer and header is stack allocated. In this case the string must only be used while the stack frame is valid.
+
+Generally, callees are expected only keep reference to string types during their stake frame, and copy (or bump the reference count in the case of `HSTRING`) if they want to keep the string around for longer.
+
+#### Rust
+
+Rust's `alloc` core library and by extension it's `std` library have the following string types:
+
+* `String`: an owned pointer to UTF-8 encoded data together with a known length and capacity.
+* `&'a str`: a view into UTF-8 encoded data that lasts for the lifetime `'a`.
+* `CString`: an owned, C-compatible, nul-terminated string of bytes.
+* `&'a CStr`: a view into a `CString` data that lasts for lifetime `'a`.
+* `OsString` (on Windows): an owned sequence of 8-bit values, encoded in a less-strict variant of UTF-8.
+* `OsStr` (on Windows): a view into `OsString` data that lasts for lifetime `'a`.
+
+#### Equivalences
+
+Windows' `PCSTR` has a direct in-memory equivalent to Rust's `CString` and `&CStr` where the Rust types have clear ownership semantics while the Windows version does not.
+
+### Modeling Windows String types
+
+#### Ownership for Win32 Strings
+
+Often when dealing with Win32 strings (i.e., `PSTR`, `PCSTR`, `PWSTR`, and `PCWSTR`) who owns the data or how long the data is valid for is well known:
+
+* As "IN" params to a function, the string should only be considered valid for the lifetime of the function call.
+* As string literals, the string data is valid for the entirety of the program.
+
+However, sometimes we don't know the ownership or lifetime of a value outside of documentation:
+
+* As "OUT" params to a function, the caller is sometimes owns the data, but it may also be owned by someone else and thus only valid for some lifetime.
+  * [Example of a "borrowed" OUT param](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Globalization/struct.IOptionDescription.html#method.Id). The `id` string is owned by the `IOwnDescription` object.
+  * [Example of an "owned" OUT param](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/Com/struct.IEnumString.html#method.Next). The caller is responsible for freeing the `rgelt` param.
+* As fields in a struct, the ownership of the data is unclear. The string can sometimes be freed when the struct is no longer needed but sometimes something else owns that string.
+
+**In short**: there is no way to programmatically know the lifetime of a win32 string except when it is used as an "IN" param.
+
+#### Ownership for `BSTR`
+
+`BSTR` has the same story as a win32 string type.
+
+#### Ownership for `HSTRING`
+
+Ownership for `HSTRING` is fairly clear:
+
+* As "IN" params, the HSTRING is only being borrowed.
+* Typically in any other case, the `HSTRING` handle has had its reference counted incremented and when that handle is no longer needed, the reference count should be decremented.
+
+**In short**: `HSTRING`s are always valid when in scope, and when they are no longer used, their reference counts should be decremented **except** in the case of "IN" params where they are logically "borrowed".
+
+### Proposed API
+
+It's clear that "IN" params are the scenario where ownership differs from the status quo. Therefore, the `windows` crate will treat the types as the following.
+
+* Introduce a new type `CWString` which is the wide string equivalent to `CString`.
+* win32 strings and `BSTR`s should always be treated like raw pointers. Accessing them is unsafe since the compiler can not verify that they are still live with the exception of their use as "IN" params which we discuss below.
+* `HSTRING`s are always live and should have their reference count decremented in their `Drop` impl.
+
+### Converting between String types
+
+#### From Windows to Rust types
+
+The main distinction between `HSTRING` and the other Windows string types is that `HSTRING` is always valid while the other string types are essentially wrappers around raw pointers and thus cannot be assumed to point to valid memory. Therefore, while `HSTRING` can host a variety of safe conversions into Rust strings, the other strings cannot.
+
+The main conversions afforded to these types are to:
+
+* `&[u16]` or `&[u8]` (depending if the given Windows type is a wide string or not)
+* `String`
+
+Conversion of most types to `CString` and `OString` doesn't make much sense as all Windows types are already FFI safe and those Rust types are almost never needed outside of FFI. However, the user should be given access to the underlying buffers (including the trailing nul byte), and there should be conveniences to converting to `String` since this unlocks the entire rich `std` library functionality for string manipulation.
+
+All string types that aren't `HSTRING` can only expose unsafe functions as it is not well known that they point to valid memory. Additionally, each type should allow conversion to a raw pointer to avoid having to use `std::mem::transmute` in cases where an API expects a raw pointer.
+
+#### From Rust to Windows types
+
+It is common for Rust functionality to want to interact with some Windows APIs which require the Windows string types.
+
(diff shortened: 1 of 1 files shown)


rfcs/003-strings.md
```diff
@@ -0,0 +1,325 @@
...
+
+With the above in mind, here is what the API should look like:
+
+```rust
+impl HSTRING {
+    // String data without the trailing 0
+    fn as_wide(&self) -> &[u16] {}
+    // String data with the trailing 0
+    fn as_wide_with_nul(&self) -> &[u16] {}
+    fn to_string(&self) -> Result<String, FromUtf16Error> {}
```
mbartlett21, line 126: Note: This function collides with [`ToString::to_string`](https://doc.rust-lang.org/nightly/std/string/trait.ToString.html#tymethod.to_string), since `HSTRING` implements `Display` below.

Perhaps this could be renamed `try_to_string`?

rfcs/003-strings.md
```diff
@@ -0,0 +1,325 @@
...
+impl<T: ?Sized + AsRef<str>> From<&'_ T> for HSTRING {}
+impl<T: ?Sized + AsRef<str>> TryFrom<&'_ T> for HSTRING {
+    type Error = FromUtf16Error;
+}
+
+// --------------------------------
+
+/// The wide equivalent to std::ffi::CString;
+#[repr(transparent)]
+struct CWString(..);
```
mbartlett21, line 151: Should there be a matching borrowed `CWStr`, like `CString` and `CStr`?

rfcs/003-strings.md
```diff
@@ -0,0 +1,325 @@
...
+
+```rust
+let x: CWString = w!("hello");
+let y: CString = c!("hello");
+let z: HSTRING = h!("hello");
+```
+
+These function build null terminated string data of the appropriate width (`u8` in the case of `PCSTR` and `u16` in the other cases).
+
+**question**: for all these types built from static strings, could we special case them so that they don't use reference counting at all?
```
mbartlett21, line 248: You could just special case for a string literal, then use that with `concat!($str, "\0")`, etc. For wide strings, you'd probably need something like [`const_heap`](https://github.com/rust-lang/rust/issues/79597) (nightly api very far from stable) or compiler support.

rylev: This refers to whether we need a special version of `HSTRING` and the like that doesn't track reference counts since its backing memory is statically allocated. Actually statically allocating enough memory for a wide string is already a solved problem (e.g., [see here](https://github.com/rylev/const-utf16)).

rfcs/003-strings.md
mbartlett21: @rylev 
One reason would be so there isn't a double-indirection.
It can also be used to convert a `PCWSTR` into a safely-used reference without having to copy or own the data. (See for example `CStr::from_ptr`)

rfcs/003-strings.md
mbartlett21: Ok. I didn't realise that the const on stable had got that far.

rfcs/003-strings.md
rylev: Good catch. That's a shame. We could use `into_string` which mirrors [`CString::into_string`](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_string) directly. I originally didn't do this as it's not exactly trivial to reuse the `String`'s buffer for the `HSTRING` buffer and thus I didn't want to take `self`.

rfcs/003-strings.md
```diff
@@ -0,0 +1,325 @@
...
+* As "IN" params, the HSTRING is only being borrowed.
+* Typically in any other case, the `HSTRING` handle has had its reference counted incremented and when that handle is no longer needed, the reference count should be decremented.
+
+**In short**: `HSTRING`s are always valid when in scope, and when they are no longer used, their reference counts should be decremented **except** in the case of "IN" params where they are logically "borrowed".
+
+### Proposed API
+
+It's clear that "IN" params are the scenario where ownership differs from the status quo. Therefore, the `windows` crate will treat the types as the following.
+
+* Introduce a new type `CWString` which is the wide string equivalent to `CString`.
```
kennykerr, line 89: I'll review more carefully when I get a moment but my most obvious question is what `CWString` gives you that `HSTRING` doesn't already provide. Obviously, there are already a dizzying number of string types and I'd love to not make that dizzier. 😉

rfcs/003-strings.md
```diff
@@ -0,0 +1,325 @@
...
+
+On top of this, we want these APIs to feel natural and obvious to Rust developers.
+
+## Explanation
+
+### The Different Types of Strings
+
+#### Windows
+
+Windows has the following string types:
```
kennykerr, line 26: Another way to look at this is that Windows has a variety of string types that need to be supported for interop with various APIs but that languages can and should only use them on the ABI boundaries as much as is practical. The extreme example of this is C# where only `System.String` exists in C# and all the different string types are marshaled away behind the scenes. While I'm not suggesting we go quite that far, we should try to think of these as interop types needed only for the ABI boundary and put our energy into making it as efficient and seamless as possible to go from and get back to native and idiomatic Rust string types, and not try to make Windows replacements with all the bells and whistles of Rust's string types.

rfcs/003-strings.md
```diff
@@ -0,0 +1,325 @@
...
+
+* `PSTR`: a mutable, nul-terminated string (i.e., composed of "characters" a.k.a. `u8`s) that is often but not always ANSI encoded.
+* `PCSTR`: an immutable version of `PSTR`.
+  * **QUESTION**: what are the practical differences between this and `PSTR`?
```
mbartlett21, line 30: The difference is that `PCSTR` can be created from a `CStr` reference, and can also be converted back.

rfcs/003-strings.md
mbartlett21: For `PCSTR`, we can just use `*const c_char` directly, and `CStr` provides `from_ptr` and `as_ptr` methods to convert to and from it. For `PSTR`, we would need `CString` and `CStr` to allow mutability.

For `PWSTR` and `PCWSTR`, it would need roughly the same, except with wide characters. `PCWSTR` (I think) should be able to be created from a `BSTR` or `HSTRING`.

rfcs/003-strings.md
mbartlett21: So essentially, you'd get a wide string reference (`PCWSTR`), then pass that to [`WindowsCreateStringReference`](https://docs.microsoft.com/en-us/windows/win32/api/winstring/nf-winstring-windowscreatestringreference) ([in `windows::Win32::System::WinRT`](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/WinRT/fn.WindowsCreateStringReference.html)), then use the resultant `HSTRING` structure.

kennykerr: Using `WindowsCreateStringReference` is generally a pessimization unless the calling language natively supports null terminated wide string literals, like C# and C++. For languages like Rust, that's not worth it as the resultant `HSTRING` typically ends up being copied almost immediately.

rfcs/003-strings.md
tim-weis: That would be the ideal situation.

However, the unfortunate situation with Rust is, that Rust does not provide a zero-overhead string type that's compatible with Windows' native string encoding. `Vec<u16>`/`&[u16]` is the best Rust has to offer in this respect. I don't see any way *"to go from and get back to native and idiomatic Rust string types"* at all.

Being pragmatic here, and assuming that unmatched surrogate pairs won't appear in practice is a dangerous route to go down. Doing so will open up code based on the `windows` crate to attacks, even if only DoS attacks.

kennykerr: Indeed that's the ideal hence the "as much as is practical". So we need a null terminated wide char but we need such things as little as humanly possible with the emphasis on interop rather than string features. And since we need HSTRING anyway and HSTRING is a null terminated wide char string that should cover those needs. Even if we end up calling it CWString, for symmetry, that just happens to be implemented as an HSTRING we can at least minimize the number of implementations we're carrying around.

rylev: That makes sense. I am a bit skeptical that we'll actually save much code by using an `HSTRING` as a backing implementation detail, but if this is truly just an implementation detail then it doesn't really matter. We can keep playing with the implementation without impacting users.

ChrisDenton:
I would also add that there are some more rough edges with functions that take even more variations on strings.

There's [UNICODE_STRING](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Foundation/struct.UNICODE_STRING.html) which is roughly equivalent to the Rust types `&[u16]`, `&mut [u16]` or `Vec<u16>` depending on the function being used and the value of `MaximumLength`. This can support interior nulls.

Also there are functions such as [`CompareStringOrdinal`](https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-comparestringordinal) that can take a null terminated `LPCWCH` string but also optionally take a length so the strings need not be null terminated (and may therefore include nulls).

And some functions such as [`SetFileInformationByHandle`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfileinformationbyhandle) that take structures which require both a null terminated string and the length of the string.

While these should not be too difficult to support, I think it's worth noting there are a number of variations on the basic `ptr + length` string.

kennykerr:
Right, which is why I'm reluctant to add some new synthetic string type like `CWString` when we already have so many to contend with. If anything, something like `CWString` belongs in the standard library alongside `CString` while the `windows` crate should focus on those that are specific to Windows, like `HSTRING` and `BSTR`, as it does. 

To @ChrisDenton's point, it is also not the goal of the `window` crate to smooth over every last detail of string handling in the Windows API. There will always be APIs that are harder to call directly and possibly deserve some specialized crate to provide a more idiomatic experience if the API is that important. Many are not. 😏

rfcs/003-strings.md
```diff
@@ -0,0 +1,325 @@
...
+#### String literals
+
+Many times the user simply wants to use a string literal to build a string of a certain type.
+
+The windows crate provides the following macros for convenience:
+
+```rust
+let x: CWString = w!("hello");
+let y: CString = c!("hello");
+let z: HSTRING = h!("hello");
```
kennykerr, line 243: I don't think we need these macros to produce allocated strings. That requires runtime support. They should all be able to return `const` string literals of `PCWSTR` and `PCSTR` respectively. Then they can simply be used to call functions like `MessageBoxA/W` directly and efficiently.

kennykerr: Common Windows string types like `HSTRING` and `BSTR` can then just provide `From<&T>` implementations for `PCWSTR` for cases where a string needs to be computed at runtime and in future if/when `CString` and `CWString` are available in the std library we can just provide `From` implementations for those as well.

rfcs/003-strings.md
kennykerr: Then we shouldn't need the `Into<T>` below, unless I'm forgetting something.

rfcs/003-strings.md
kennykerr: Then we shouldn't need the `Into<T>` below, unless I'm forgetting something.

rfcs/003-strings.md
```diff
@@ -0,0 +1,325 @@
...
+### The Different Types of Strings
+
+#### Windows
+
+Windows has the following string types:
+
+* `PSTR`: a mutable, nul-terminated string (i.e., composed of "characters" a.k.a. `u8`s) that is often but not always ANSI encoded.
+* `PCSTR`: an immutable version of `PSTR`.
+  * **QUESTION**: what are the practical differences between this and `PSTR`?
+* `PWSTR`: a mutable, nul-terminated "wide" string (i.e., composed of "wide characters" a.k.a. `u16`s) that is often but not always UTF-16 encoded.
```
riverar, line 31: > [...] that is often but not always UTF-16 encoded.

Can you expand on this? I'm not aware of any non-UTF-16 encoded PWSTRs. Are you instead trying to convey that a grapheme may span 1 or 2 code 16-bit units (2-4 bytes)?

rfcs/003-strings.md
kennykerr: Here's what I have in mind: https://github.com/microsoft/windows-rs/compare/utf16?expand=1

kennykerr: And this is compatible with the existing state of the `windows` crate so would make it easier to land the `Borrowed` PR.

rfcs/003-strings.md
ChrisDenton: It's possible to create [invalid UTF-16 strings](https://www.unicode.org/faq/utf_bom.html#utf16-7) because the kernel does not check validity. This is exceedingly rare in practice but it is possible, especially if there's a malicious app causing mischief.

tim-weis: The prime candidate here is the filesystem. NTFS allows names to be just about *any* sequence of 16-bit values (with a few reserved code units). Those names will inevitably be returned from `FindFirstFile`/`FindNextFile`, so `PWSTR`/`PWCSTR` cannot make any assumptions about the validity of the UTF-16 encoding.

Likewise, named kernel objects (events, mutexes, pipes, ...) can have names that aren't valid UTF-16.

Not quite the norm, but invalid UTF-16 sequences are observed in the wild, usually as a result of an accident/bug, or, more frequently, in attacks.

kennykerr: There's no use trying to define what these are. They're just pointer `typedef`s, which is why `windows-rs` defines them as simply as possible. We can provide safe ways to create them and pass them to API calls, but we cannot make any guarantees about them or provide generally safe conversions from them.

rfcs/003-strings.md
rylev: This proposal is a subset of what I had in mind so I think it's fine to try it out. We can always expand from there to have dedicated `CWString` types and such.

kennykerr:
This PR's pretty old. Can we move it along or just close it for now?

PR #1861, opened 2022-07-01
Replace `unimplemented!` panics with conditional `#[link]` attribute for WSL/Linux support

### LINKED ISSUE #1842: Remove `unimplemented!("Unsupported target OS");` to allow _running_ on Linux and WSL
### The "problem"

Currently the `windows` crate is allowed to compile on non-`windows` systems thanks to a rather large block of code, replicated across every function:

```rust
pub unsafe fn DxcCreateInstance2(...) -> Result {
    #[cfg(windows)]
    {
        #[link(name = "windows")]
        extern "system" {
            fn DxcCreateInstance2(...) -> ::windows::core::HRESULT;
        }
        let mut result__ = ::core::option::Option::None;
        DxcCreateInstance2(...).and_some(result__)
    }
    #[cfg(not(windows))]
    unimplemented!("Unsupported target OS");
}
```

I remember discussing this long ago: it was needed for CI but seems to have never been removed, as there wasn't much reason to run on non-`windows` anyway.

Rust is supposed to omit linkage when functions aren't used: any `extern "system" { pub fn blah(); }` doesn't show up in the symbol table if it isn't called into. However, this isn't the case for `#[link]`: `-lwindows` is added to the linker invocation _even if_ no functions from the `extern "system"` block it is annotating are called.
That is evident in the `windows-sys` crate which doesn't carry any `cfg(windows)` around `#[link]` attributes: it is inconveniently used in the `metadata` crate that's subsequently used by all the tools, not allowing any of them to be ran on non-Windows (`= note: mold: fatal: library not found: windows`) even though they don't reference a single function, just some typedefs.

### Proposed solution

Now, combine those two points and we can turn the above into:

```rust
pub unsafe fn DxcCreateInstance2(...) -> Result {
    #[cfg_attr(windows, link(name = "windows"))]
    extern "system" {
        fn DxcCreateInstance2(...) -> ::windows::core::HRESULT;
    }
    let mut result__ = ::core::option::Option::None;
    DxcCreateInstance2(...).and_some(result__)
}
```

### (Dis)advantages

_Disclaimer: most of the below applies to non-`windows` only, which perhaps still isn't "in scope" of the project - but I do hope we can discuss and consider this, especially in light of [`com-rs` being officially deprecated](https://github.com/microsoft/com-rs/pull/238)!_ [^1]

[^1]: I am of course again trying to find a proper, maintained, clean Rust alternative for `DirectXShaderCompiler` COM bindings. This is perhaps one of the few - or only? - binaries that has a functioning COM API outside of Windows. The proposed Quality-of-Life improvements will shorten generated code and allow me to use `windows-rs` outside of Windows at no additional maintenance cost here.

This is much shorter and simpler (multiplied by thousands of functions), but there's an obvious catch: whenever this function is used, even in conditional code (consider for example `error.rs` using various functions for retrieving and stringifying/prettifying `Error`), linking fails.

Perhaps that's (much) better than an unsuspecting `unimplemented!()` being reached in an optional codepath (e.g. `Error` handling).

However, there's an immediate solution too. And I've alluded to / requested this in earlier issues: **define your own fallback**.
(This goes anywhere in a user's codebase, could even be a dependent crate)

```rust
#[no_mangle]
pub extern "system" DxcCreateInstance2(...) -> ::windows::core::HRESULT {
    // My custom fallback implementation
}
```
(Badly chosen example, this'd be better on e.g. `SysStringLen`.)

Then the linker error is gone again, _and_ the user has the ability to provide the necessary functions themselves _at no additional design cost to `windows-rs`_ 🎉.

### Proof of concept

You can find the end result here: https://github.com/MarijnS95/windows-rs/compare/simplify-cfg-windows (`Showing 429 changed files with 71,379 additions and 157,268 deletions`). As bonus this includes the same `cfg_attr` in `windows-sys` to allow running the tools on non-Windows.

### Alternative (even smaller) approach

Perhaps this isn't necessary (read: remove the `#[cfg(windows)]` entirely) at all if a blanket `libwindows.so` is provided _by the user_, possibly built straight from a Rust crate with aforementioned fallback implementations. This can be done directly through `cargo` by providing a:

```toml
[package]
name = "windows-fallback"

[lib]
name = "windows"
crate-type = ["cdylib"]
```

Such a crate can be specified directly as a dependency in another crate even if no Rust symbols are referenced and it's not explicitly linked against: `"windows-fallback"` here still builds and its `libwindows.so` result ends up in a path that's already part of the search path to cover the automatically-inserted `-lwindows` 🎉 (may be a bit fragile to depend on though?).

That however doesn't provide a solution to the metadata/tools currently being only compilable on Windows, and I'm _not suggesting to provide this crate within `windows-rs`_!

MarijnS95:
Fixes #1842

The `#[cfg]` structure is verbose and not needed for cross-compilation (where `target_os = "windows"` anyway).  Furthermore, the `unimplemented!("Unsupported target OS")` panic prevents using all but the most basic features windows-rs provides on non-Windows systems.  Use-cases relying on linking external functions include calling `dxcore` from WSL and `DirectXShaderCompiler` from Linux, both using a COM interface.

By removing this `#[cfg]` and `unimplemented!()` panic altogether `windows`, just like `windows-sys`, now unconditionally link using `-lwindows`.  This allows crates compiling on non-Windows targets to provide their own fallbacks for needed functions.  Any required but missing symbol becomes a linker error rather than a runtime panic.

However, such linking may not be desired or complicated to set up, and doesn't prevent the `tool_*` applications (needed to propagate `bindgen` changes in this PR to the final codebase, while working from a Linux environment) through the `metadata` -> `windows-sys` crate from linking against a `-lwindows` library (and failing).  For this reason a `cfg_attr` with `windows` predicate has been chosen: `link(name = "windows")` will be enabled on `target_os = "windows"` exclusively.

The guard is applied to `windows-sys` which previously didn't have any guarding at all (once again confirming that there's no impact on cross-compiling) to unblock `tool_*` usage on non-Windows systems.

Users are able to provide fallbacks on their own within their consumer crate through a simple:

```rust
#[no_mangle]
pub extern "system" fn SysStringLen(..) -> ...;
```

for example, without having to resort to compiling that inside a separate crate that emits a static library or shared object on the search path.  This is however still possible, though.

There are a few more issues to work out before this crate can _run_ on Linux and WSL, which have to be fleshed out before we can officially (build-)test this use-case in CI.


crates/libs/bindgen/src/functions.rs
```diff
@@ -50,7 +50,7 @@ fn gen_win_function(gen: &Gen, def: MethodDef) -> TokenStream {
         Some(link) => quote! { #[link(name = #link, kind = "static")] },
         None => {
             if gen.namespace.starts_with("Windows.") {
-                quote! { #[link(name = "windows")] }
+                quote! { #[cfg_attr(any(windows, feature = "link_windows"), link(name = "windows"))] }
```
riverar, line 53: Let's drop the feature for now and baby-step into changes here.

crates/libs/metadata/Cargo.toml
```diff
@@ -12,7 +12,8 @@ default-target = "x86_64-pc-windows-msvc"
 targets = []
 
 [dependencies.windows-sys]
-version = "0.36.1"
+path = "../sys"
+version = "0.38"
```
riverar, line 16: @kennykerr This will need your 👀. Probably can't be published like this?

riverar (changes requested):
Super excited to see those guards go away, suspect it will measurably speed up compilation. Just a relatively minor nit. 

Thanks for your overall patience/tenacity!

crates/libs/metadata/Cargo.toml
MarijnS95: I think it has to be changed back before being published - the `metadata` crate, that is. That's what happened in the `0.38` release as well.

I've previously (in https://github.com/microsoft/windows-rs/commit/4c7ae6bdc0338e2aee9b578b64306b4696816acf#r76875081) suggested to simply _not_ bump the version of `windows-sys` if it hadn't seen a release anyway. That'd solve this and more issues.
(Afaik `cargo` downloads a published crate from `crates.io` for every dependency before building and publishing the current package, instead of working on local source)

crates/libs/bindgen/src/functions.rs
MarijnS95: Okay, I can do that. No harm in doing it that way either, it simply means that the few functions one may want/need to override on non-Windows can be imported from anywhere instead of specifically (and optionally, because it's a non-default feature) residing in a `libwindows.a`/`libwindows.so` library.

MarijnS95:
> suspect it will measurably speed up compilation.

I don't think it's very useful if I try to profile this on Linux, since it'll now have a lot more code to visit and compile than the simple `unimplemented!()` :)

For profiling I'd use `hyperfine -p 'cargo clean -p windows' 'cargo b -p windows --all-features'` fwiw.

riverar:
@MarijnS95 Might be useful to stuff into a pretty chart. I can try to run that exact command here too and get you the numbers. (Haven't used hyperfine before.)

MarijnS95:
@riverar Careful with the `--all-features` though, it sat here doing nothing for a couple minutes before I canceled it.

riverar (approved):

kennykerr:
Here are a few tips to improve PR success. 

* Each PR only changes the thing it purports to change. This often just means merging with the target branch to ensure the PR is "clean". Also, don't fix something just because "you're in the neighborhood". Feel free to contribute those incidentals as unrelated PRs.
* Keep the hand-authored changes in the first commit and the generated code changes in a separate commit. That way we can easily review/diff what's changed.
* When fixing something (like Linux support), add the CI changes in a commit that "proves" it doesn't work. Then let the build fail. Then push the commit that "fixes" the build. 

Just trying to help smooth the process of adopting these changes and help you get this successfully complete. As it stands, there are a few related PRs that all seemingly need to land in some uncertain order and have a lot of overlapping changes, making review very difficult and success hard to prove.

MarijnS95:
> * Each PR only changes the thing it purports to change. This often just means merging with the target branch to ensure the PR is "clean". Also, don't fix something just because "you're in the neighborhood". Feel free to contribute those incidentals as unrelated PRs.

I don't understand where this comment comes from. That's the sole rule I always abide by. I don't think I've ever, anywhere, tried to sneak in an unrelated yet nontrivial change into an existing PR. Simply because I also hate to be on the receiving end of that as a reviewer. It makes the title+description disconnected from the contents.

So again: what's the context for writing this here? Where have I "misstepped"?

> * Keep the hand-authored changes in the first commit and the generated code changes in a separate commit. That way we can easily review/diff what's changed.

That's an actionable review comment, at least.

> * When fixing something (like Linux support), add the CI changes in a commit that "proves" it doesn't work. Then let the build fail. Then push the commit that "fixes" the build.

Thanks for sharing your opinion. Typically projects disallow merge when the CI fails, so it only makes sense to get both in at the same time (conflicts with 1., again exactly why I stated above that I'd _do that in a separate PR_) or expand CI coverage after merging the initial solution.

If you can acknowledge again that temporarily breaking `main` is _what you want_, I'll flip the order of the PRs (and obviously clearly document which depends on which).

> Just trying to help smooth the process of adopting these changes and help you get this successfully complete. As it stands, there are a few related PRs that all seemingly need to land in some uncertain order and have a lot of overlapping changes, making review very difficult and success hard to prove.

Again, a comment I don't understand. I have thee PRs open, one of which is marked WIP and states as the first sentence: `Depends on #1861 and #1862`. In other words, both PRs need to be merged for #1863 to go in. Conversely, #1861 and #1862 don't have any dependency on each-other - I didn't write `Depends on` after all, nor do they share any commits!

MarijnS95:
Force-pushed the change: first commit is manual changes, second commit is the generated result of that.

kennykerr:
Just general suggestions for anyone working on this project. I should really write it down somewhere more central but since you're one of the first outside contributors to make more substantial changes I wanted to make sure I share them. Please don't be offended. I'm not picking on you. 😉

Since you insist, I took a cursory glance at your PRs. In #1863, about testing for Linux, you made unrelated changes to the issue templates. It also includes hundreds of other changed files when I would expect only yml changes. The PRs seem to be built one on top of the other, allowing changes from one to show up in the other, making reviews impractical. While the code changes and generated code are combined in a single commit, you provide the code changes and build validation in separate PRs, so we can't validate the changes are sound. If you include them in the same PR, then there's no need to temporarily break anything other than the PR build validation. 

I realize this is not a trivial project to work on and I'm happy for the help, but the suggestions I've given are for your benefit as well as reviewers. I'm also happy to make the changes myself if it seems tedious or an inconvenience. It's my job. 😊

MarijnS95:
@kennykerr You should write those down in a global contributor guideline, they're all really good points to have. They just don't make any sense here, since some points are not at all actionable for me and left me confused again what exactly you want(ed) from me :flushed: 

We're all good! :blush: :hugs: 

---

> you made unrelated changes to the issue templates. It also includes hundreds of other changed files when I would expect only yml changes. The PRs seem to be built one on top of the other, allowing changes from one to show up in the other, making reviews impractical.

Correct. As stated, that PR is a WIP and builds on top of #1861 and #1862 to showcase that the CI succeeds. The changes live in separate commits just as you requested for regeneration, so you can look at individual changes and skip the contents that go in through other means.

You are right that this PR contains unnecessary whitespace changes for the YAML files. I can submit those separately, but they are trivial to filter out in the github review window too.

> While the code changes and generated code are combined in a single commit, you provide the code changes and build validation in separate PRs

In my eyes, while important to have, these seem like separate concerns. I suppose the CI requires a PR to contain any regeneration changes before allowing it through, so that there's no desync between the generator/metadata and checked-in generated code.

> you provide the code changes and build validation in separate PRs, so we can't validate the changes are sound. If you include them in the same PR

Secondly, more importantly, said validation depends on two disjoint PRs before succeeding.

> then there's no need to temporarily break anything other than the PR build validation.

The same CI job runs on `master`, showing that it is temporarily not building. Anyone forking and building their changes on top will face the same CI failure, until all the fixes have been merged on top.

---

To summarize, can you state exactly what contents you'd like to see in each PR, and in what order you'd like to review+merge them? It seems you want the trivial CI addition first (https://github.com/microsoft/windows-rs/pull/1863/commits/c9fd5872d21cec936d4869fbacfdfeff762a6464), have the CI fail on `master` temporarily, and then see how #1861 and #1862 resolve those errors again?

PR #1862, opened 2022-07-01
Disable all (implicit) uses of `OsStr(ing)Ext` in `HSTRING` and `PCWSTR` for WSL/Linux support

### LINKED ISSUE #1860: Guard `windows::ffi::OsStringExt` usage behind `#[cfg(windows)]` for WSL and Linux support
### Motivation

We'd like to run - _run_, **not cross-compile** - `DirectXShaderCompiler` directly on Linux. At the same time, @riverar found a case to run [`dxcore` inside WSL](https://github.com/microsoft/windows-rs/issues/1842#issuecomment-1164681869), which is also a Linux target.

The `windows` crate doesn't compile _for_ / run _on_ non-Windows targets currently, as `std::os::windows::ffi::OsStrExt` isn't available on `target_os != "windows"`.

A proposed fix guards all trait implementations containing `std::os::windows::ffi::OsStrExt` behind a `#[cfg(windows)]` flag. If needed, followup PRs can introduce Linux-specific implementations for these conversion traits (but read on for a 32-bit wchar caveat).

### Drawbacks

Extra complexity for a "purely Windows-oriented" crate. We can solve that by build-testing _for_ Linux - in addition to cross-compiling _from_ it.

Impact: Not supporting `windows-rs` on Linux and WSL while at the same time [deprecating `com-rs`](https://github.com/microsoft/com-rs/pull/238) leaves users with no alternative for [_cross-platform COM bindings_ into libraries like `DirectXShaderCompiler`](https://github.com/Traverse-Research/hassle-rs/blob/main/src/ffi.rs).

### Additional context

Prior art: https://github.com/microsoft/windows-rs/issues/1842#issuecomment-1164681869

### Caveat

Furthermore, non-Windows targets use 32-bit widechars instead of 16-bit:

https://en.cppreference.com/w/cpp/language/types#Character%20types:~:text=wchar_t,a%20distinct%20type.

> 32 bits on systems that support Unicode. A notable exception is Windows, where wchar_t is 16 bits and holds UTF-16 code units

This should be addressed in a separate issue, and seems to be trivially implementable using Rust's `char` type. This is also how https://docs.rs/widestring implements it.

MarijnS95:
Fixes #1860

On non-Windows targets - such as Linux and WSL - the `std::os::windows::ffi::OsStrExt` and `std::os::windows::ffi::OsStringExt` extension traits are not available.

To support compiling and running _on_ such non-Windows targets (not just cross-compiling _from_ them, for a Windows target) anything using these traits has to be guarded in the same way using a `#[cfg(windows)]`.

If needed fallbacks can be implemented manually, but that likely requires a bit more thought since wide-strings on these platforms typically have [32-bit `wchar_t` sizes, with Windows' 16-bit being the exception].

[32-bit `wchar_t` sizes, with Windows' 16-bit being the exception]: https://en.cppreference.com/w/cpp/language/types#Character%20types:~:text=wchar_t,a%20distinct%20type

### DIFF
--- a/crates/libs/windows/src/core/hstring.rs
+++ b/crates/libs/windows/src/core/hstring.rs
@@ -49,6 +49,7 @@ impl HSTRING {
     }
 
     /// Get the contents of this `HSTRING` as a OsString.
+    #[cfg(windows)]
     pub fn to_os_string(&self) -> std::ffi::OsString {
         std::os::windows::ffi::OsStringExt::from_wide(self.as_wide())
     }
@@ -168,18 +169,21 @@ impl core::convert::From<&alloc::string::String> for HSTRING {
     }
 }
 
+#[cfg(windows)]
 impl core::convert::From<&std::ffi::OsStr> for HSTRING {
     fn from(value: &std::ffi::OsStr) -> Self {
         unsafe { Self::from_wide_iter(std::os::windows::ffi::OsStrExt::encode_wide(value), value.len() as u32) }
     }
 }
 
+#[cfg(windows)]
 impl core::convert::From<std::ffi::OsString> for HSTRING {
     fn from(value: std::ffi::OsString) -> Self {
         value.as_os_str().into()
     }
 }
 
+#[cfg(windows)]
 impl core::convert::From<&std::ffi::OsString> for HSTRING {
     fn from(value: &std::ffi::OsString) -> Self {
         value.as_os_str().into()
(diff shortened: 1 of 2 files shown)


riverar (approved):

kennykerr:
We need a non-windows CI test to validate that this is correct.

MarijnS95:
@kennykerr I mentioned in another PR that I'd do that once this compiles, and I already pushed some preliminary revision ~20 minutes ago to my fork. I can open a PR that pulls in both #1860 and #1862 and showcase that it builds, but those PRs need to be merged in first before such a CI would succeed.

kennykerr:
Yep, just left the comment as a reminder to get the CI updated with or before this PR.

MarijnS95:
Good to know! I might as well open a draft PR once it compiles, had some fights with tool_yml and "wrong" indentation for arrays :)

And branches: master prevents it from running on a fork :)

kennykerr:
yml is not my favorite language... 😜

MarijnS95:
Draft PR open at #1863!

kennykerr:
Closing in favor of #1863

PR #1867, opened 2022-07-02
ci: Replace `diff --exit-code` with `status --porcelain` to catch untracked files

MarijnS95:
I commited this line to the repo quite some time, but discovered over time that it has one fatal flaw: `git diff` only considers _tracked_ files.  In the event that the command-at-test generates new files that weren't checked in by accident (and aren't explicitly in `.gitignore`), `git status --porcelain` will return a non-empty string and allow the test to fail appropriately.

Note that it is typically unlikely for this to happen, as e.g. a single missing Rust file generally results in a compilation failure, assuming a `mod` has likely been added in a file that was tracked already.  But better to be safe than sorry nevertheless.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ jobs:
       run: cargo run -p tool_${{ matrix.generator }}
     - name: Compare
       shell: bash
-      run: git diff --exit-code || (echo '::error::Generated `tool_${{ matrix.generator }}` are out-of-date. Please run `cargo run -p tool_${{ matrix.generator }}`'; exit 1)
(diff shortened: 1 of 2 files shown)


kennykerr:
This has worked well for some time. I'm reluctant to change it unless there's a more compelling reason to do so. It sounds like this is an unlikely issue but if you'd like to add a commit that should how this breaks and also validates that it still detects changes as before then we can consider it.

MarijnS95:
Glad to hear that it hasn't happened to this repo yet. I'm just cleaning up after my own submissions from earlier, feel free to take it or leave it. I don't think spending time to implement the sketched scenario is worth it, tbh.

PR #1888, opened 2022-07-11
Remove orphaned "alloc" feature

kennykerr:
Cleaning up a few things after #1811 - this just removes the "alloc" feature that is no longer does anything.

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -17,7 +17,6 @@ Start by adding the following to your Cargo.toml file:
 [dependencies.windows]
 version = "0.38.0"
 features = [
-    "alloc",
     "Data_Xml_Dom",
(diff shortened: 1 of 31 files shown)


docs/readme.md
```diff
@@ -17,7 +17,6 @@ Start by adding the following to your Cargo.toml file:
 [dependencies.windows]
 version = "0.38.0"
 features = [
-    "alloc",
```
riverar, line 20: Think you want to leave this one, right? Users following along in the README will fail because 0.38 doesn't represent what's in main.

kennykerr: Yes, I really should split the readme for the repo from the readme used by crates.io ...

PR #1893, opened 2022-07-12
where clauses in codegen

rylev:
This changes codegen to use `where` clauses which makes reading the generated code much easier. 

Unfortunately, this does lead to larger codegen, so we lose what was gained in #1811 

**Old**

```rust
pub fn SetPreferredDeviceKind<'a, Param0: ::std::convert::Into<LearningModelDeviceKindPreview>>(&self, value: Param0) -> ::windows::core::Result<()> {
```

**New**
```rust
pub fn SetPreferredDeviceKind<'a, P0>(&self, value: P0) -> ::windows::core::Result<()>
    where
        P0: ::std::convert::Into<LearningModelDeviceKindPreview>,
```


kennykerr:
image

Eeek. What's the difference to `cargo package` size?

Also, there have been some compatibility issues around the two syntactic approaches where Rust doesn't quite treat them the same way. I'd have to dig that up to confirm this is not a regression.

rylev:
> What's the difference to cargo package size?

Old: 13,011,416 bytes
New: 13,061,995 bytes

Surprisingly, not a big difference...

> compatibility issues around the two syntactic approaches

I'm not aware of a difference between the two approaches. Are you perhaps thinking about the `impl Trait` syntax?

kennykerr:
Yes, we're not using `impl Trait` anymore so that's probably what I was thinking of.

And the package size is the main thing so this change is probably good - and clearer for sure.

crates/libs/windows/src/core/bindings.rs
```diff
@@ -1705,31 +1726,44 @@ impl ::core::ops::Not for FORMAT_MESSAGE_OPTIONS {
     }
 }
 #[inline]
-pub unsafe fn FormatMessageW<'a, Param0: ::std::convert::Into<FORMAT_MESSAGE_OPTIONS>>(dwflags: Param0, lpsource: *const ::core::ffi::c_void, dwmessageid: u32, dwlanguageid: u32, lpbuffer: ::windows::core::PWSTR, nsize: u32, arguments: *const *const i8) -> u32 {
+pub unsafe fn FormatMessageW<'a, P0>(dwflags: P0, lpsource: *const ::core::ffi::c_void, dwmessageid: u32, dwlanguageid: u32, lpbuffer: ::windows::core::PWSTR, nsize: u32, arguments: *const *const i8) -> u32
+where
+    P0: ::std::convert::Into<FORMAT_MESSAGE_OPTIONS>,
```
kennykerr, line 1731: Enums shouldn't be convertible...

rylev: This is not an issue with this PR, but we should make an issue for this.

kennykerr: Sure, happy either way. Just noticed it here first now that the code is so much more readable. 😆

rylev: I'd rather do this as a follow up PR if you don't mind.

kennykerr (approved):

rylev:
@kennykerr just as an FYI: if I remove the rustfmt config for line width, there are a *lot* more lines, but it doesn't seem to impact package size very much. We may want to consider removing that config. It might make git diffs much larger though....

kennykerr:
Sure, we can experiment with that again. Separately. Last I checked it was still significant, but a lot has changed.

PR #1907, opened 2022-07-13
Avoid `transmute` for binding primitive argument types

kennykerr:
Now that signature types are categorized more clearly, we can remove the use of `transmute` for primitive types.


crates/libs/metadata/src/reader/mod.rs
```diff
@@ -1158,6 +1158,9 @@ impl<'a> Reader<'a> {
     pub fn signature_param_is_convertible(&self, param: &SignatureParam) -> bool {
         self.signature_param_maybe_generic(param) && self.type_is_convertible(&param.ty)
     }
+    pub fn signature_param_is_primitive(&self, param: &SignatureParam) -> bool {
+        self.signature_param_maybe_generic(param) && self.type_is_primitive(&param.ty)
```
rylev, line 1162: It feels like the name `signature_param_maybe_generic` is wrong now. I'm not sure exactly what it represents. It's an input param that's not a pointer, not a winrt array, and has no array info. How do we succinctly describe that?

kennykerr: Yep those names don't make sense to me. I'm slowly making my way to these inner helpers.

PR #1912, opened 2022-07-14
Don't suggest inlining functions

kennykerr:
The `inline` attribute was added early on to work around a linker issue with cross-compilation that somehow caused the linker to eagerly link functions that were never called. It doesn't look like this is needed anymore. 🤞


kennykerr:
Bummer. Looks like the Rust compiler still causes functions to be linked that are never called. 🙄

PR #1913, opened 2022-07-15
More accurate signature parameter parsing 

### LINKED ISSUE #1910: Borrowed types include non-borrowed types
The concept of a borrowed type defined by `type_is_borrowed` includes types like `BSTR` and `HSTRING`. The trouble is that this is used to apply `InParam` which implies convertibility and optionality, which neither of those typically are in the way that `IUnknown` typically is. 

We need to distinguish between types that are passed by reference and neither optional nor convertible and those that are borrowed, optional, and convertible.

kennykerr:
Fixes #1910

### DIFF
--- a/crates/libs/bindgen/src/gen.rs
+++ b/crates/libs/bindgen/src/gen.rs
@@ -320,11 +320,7 @@ impl<'a> Gen<'a> {
             if !self.reader.signature_param_input_value(param) {
                 continue;
             }
-            if self.reader.signature_param_is_param(param) {
-                let name: TokenStream = gen_name(position);
-                let into = self.type_name(&param.ty);
-                tokens.combine(&quote! { #name: ::std::convert::Into<::windows::core::InParam<'a, #into>>, });
-            } else if self.reader.signature_param_is_failible_param(param) {
+            if self.reader.signature_param_is_failible_param(param) {
                 let name: TokenStream = gen_name(position);
                 let error_name: TokenStream = format!("E{}", position).into();
                 let into = self.type_name(&param.ty);
--- a/crates/libs/bindgen/src/gen.rs
+++ b/crates/libs/bindgen/src/gen.rs
@@ -317,6 +317,9 @@ impl<'a> Gen<'a> {
             name
         };
         for (position, param) in self.generic_params(params) {
+            if !self.reader.signature_param_input_value(param) {
+                continue;
+            }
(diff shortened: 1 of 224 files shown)


crates/libs/bindgen/src/gen.rs
```diff
@@ -277,7 +277,7 @@ impl<'a> Gen<'a> {
     }
     /// The signature params which are generic (along with their relative index)
     pub fn generic_params<'b>(&'b self, params: &'b [SignatureParam]) -> impl Iterator<Item = (usize, &SignatureParam)> + 'b {
```
rylev, line 279: I guess this method should also change names to `convertible_params` no? Or can this be kept as `generic_params`  if we add the check for `signature_param_input_value` here?

crates/libs/windows/src/Windows/AI/MachineLearning/Preview/mod.rs
```diff
@@ -914,14 +909,11 @@ impl LearningModelBindingPreview {
...
-    pub fn HasKey<'a, P0>(&self, key: P0) -> ::windows::core::Result<bool>
-    where
-        P0: ::std::convert::Into<::windows::core::InParam<'a, ::windows::core::HSTRING>>,
-    {
+    pub fn HasKey(&self, key: &::windows::core::HSTRING) -> ::windows::core::Result<bool> {
         let this = &::windows::core::Interface::cast::<super::super::super::Foundation::Collections::IMapView<::windows::core::HSTRING, ::windows::core::IInspectable>>(self)?;
         unsafe {
             let mut result__ = ::core::mem::MaybeUninit::zeroed();
-            (::windows::core::Interface::vtable(this).HasKey)(::windows::core::Interface::as_raw(this), key.into().abi(), result__.as_mut_ptr()).from_abi::<bool>(result__)
+            (::windows::core::Interface::vtable(this).HasKey)(::windows::core::Interface::as_raw(this), ::core::mem::transmute_copy(key), result__.as_mut_ptr()).from_abi::<bool>(result__)
```
rylev, line 916: It's a shame that we had to go back to using `transmute_copy` here...

kennykerr: Yes overall, there's a ton less `transmute` but these remaining ones may still potentially be eliminated - just needs to be updated very carefully.

PR #1915, opened 2022-07-15
Ensure that `Display` impl for `HSTRING` is infallible

### LINKED ISSUE #324: Display for HString is not infallible
The Rust documentation [says](https://doc.rust-lang.org/std/fmt/index.html#formatting-traits):

> …a formatting implementation must and may only return an error if the passed-in Formatter returns an error. This is because, contrary to what the function signature might suggest, string formatting is an infallible operation.

but if an `HString` contains unpaired surrogates, its `Display::fmt` implementation will return a `std::fmt::Error` instead (and `to_string()` will panic). I think in this case it would be quite reasonable for the formatting to work like `String::from_utf16_lossy` instead.

Whatever the desired behaviour is, there should be tests to confirm the panic or replacement characters.

kennykerr:
Fixes #324


crates/tests/hstring/tests/display.rs
```diff
@@ -0,0 +1,8 @@
+use windows::core::*;
+
+#[test]
+fn display() {
```
rylev, line 4: There's a whole bunch of [tests for `HSTRING` in the `test_core` crate](https://github.com/microsoft/windows-rs/blob/d4bb98da9a6188927026ee81137216eeea8da4e1/crates/tests/core/tests/hstrings.rs) (which could just be tests in the `windows` crate itself TBH). I don't think we need separate crate for this.

crates/libs/windows/src/core/strings/hstring.rs
```diff
@@ -136,7 +136,7 @@ impl core::fmt::Display for HSTRING {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         use core::fmt::Write;
         for c in core::char::decode_utf16(self.as_wide().iter().cloned()) {
-            f.write_char(c.map_err(|_| core::fmt::Error)?)?
+            f.write_char(c.unwrap_or(core::char::REPLACEMENT_CHARACTER))?
         }
         Ok(())
```
rylev, line 141:
```rust
write!("{}", Decode(move || core::char::decode_utf16(self.as_wide().iter().cloned()))
```

PR #1925, opened 2022-07-20
Exclude MsHtml APIs

kennykerr:
These older APIs were only intended for JScript and VBScript clients, but MsHtml is by far the single largest module and as such takes a nice chunk out of the overall package size.


rylev (approved):
Wow! It would be nice if we clearly documented somewhere which APIs are being excluded since we're now not quite a 1 to 1 mapping to the metadata.

kennykerr:
Yes, I really need to spend some time and write a more complete FAQ.

kennykerr:
See #1933

kennykerr:
It can be reintroduced in a few different ways if the need arises. The `windows-bindgen` crate fulfills the promise of "any Windows API past, present, and future", is used to generate the `windows` crate, and any additional crates for supplemental APIs like this.

mwcampbell:
OK, if it's feasible to have a separate mshtml crate via windows-bindgen, I will be happy to maintain such a crate if I have the need for it.

PR #1926, opened 2022-07-21
Replace unsafe `strlen` usage in metadata library with safe Rust

Swatinem:
Removes the `unsafe` usage of C strlen and unchecked utf8 decoding
from `File::str`, replacing it with safe (but panicing) Rust code.
This also documents the method at the same time.

This pulls in the `memchr` crate, and uses almost the same code as the nightly-only `CStr::from_bytes_until_nul`.

Not sure this subcrate by itself has any tests? So I’m not 100% confident of this change due to the lack of tests?

### DIFF
--- a/crates/libs/metadata/src/reader/file.rs
+++ b/crates/libs/metadata/src/reader/file.rs
@@ -333,13 +333,21 @@ impl File {
         }
     }
 
+    /// Returns the string from the `#Strings` stream as referenced by
+    /// the (table, row, column) triple.
+    ///
+    /// # Panics
+    ///
+    /// * When any element of the (table, row, column) triple is invalid.
+    /// * When the offset in the string table is out of bounds.
+    /// * When no null terminator can be found in the string table.
+    /// * When the null-terminated string is not valid utf-8.
     pub fn str(&self, row: usize, table: usize, column: usize) -> &str {
         let offset = self.strings + self.usize(row, table, column);
 
-        unsafe {
-            let len = strlen(self.bytes.as_ptr().add(offset));
-            std::str::from_utf8_unchecked(&self.bytes[offset..offset + len])
-        }
+        let bytes = &self.bytes[offset..];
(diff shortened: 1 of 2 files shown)


crates/libs/metadata/src/reader/file.rs
```diff
@@ -333,13 +333,21 @@ impl File {
...
     pub fn str(&self, row: usize, table: usize, column: usize) -> &str {
         let offset = self.strings + self.usize(row, table, column);
 
-        unsafe {
-            let len = strlen(self.bytes.as_ptr().add(offset));
-            std::str::from_utf8_unchecked(&self.bytes[offset..offset + len])
-        }
+        let bytes = &self.bytes[offset..];
+        let nul_pos = memchr::memchr(0, bytes).unwrap();
+        std::str::from_utf8(&bytes[..nul_pos]).unwrap()
```
rylev, line 350: Perhaps we should use `expect` instead of `unwrap` so the panic message is more clear.

rylev (commented):
I'm personally in favor of most of the changes here:
* additional docs
* checking for utf-8 and panicking otherwise (using `from_utf8_unchecked` was wrong)

I'm not sure that the dependency on `memchar` is worth it though. I would say we should accept these changes without the adoption of `memchar`. We can always change towards using `memchar` in the future if we encounter issues.

crates/libs/metadata/src/reader/file.rs
tim-weis: Wouldn't

```rust
let nul_pos = bytes.iter().position(|&c| c == 0).unwrap();
```

accomplish the same without a dependency on an external crate?

Swatinem: 👍🏻 removed the `memchr` dependency and added explicit `expect` messages

kennykerr:
This dates back to when the metadata reader was primarily used from a proc macro and thus very slow. That's no longer a primary concern so this change looks good.

PR #1932, opened 2022-07-21
Provide 32-bit aliases for `SetWindowLongPtrA` and `GetWindowLongPtrA`

### LINKED ISSUE #1493: Feature request: Provide bitness-independent mappings where available
### Motivation

64-bit Windows introduced a number of new APIs to accommodate for increased pointer type widths. A prominent example being the [`[Get|Set]WindowLongPtr[A|W]`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlongptra) family of functions. They supersede the [`[Get|Set]WindowLong[A|W]`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getwindowlonga) functions, but aren't available on 32-bit Windows.

As a consequence, code that needs to be compatible with 32-bit Windows as well as 64-bit Windows must conditionally pull in the required dependencies. C and C++ developers using the Windows SDK have this solved: They can unanimously use the `*-Ptr`-variants and rely on the preprocessor to transform their code as needed.

Rust developers using the `windows` or `windows-sys` crates, on the other hand, will not find themselves in as comfortable a situation. Getting code to compile across architectures with differing pointer widths usually involves a fair amount of conditional code (see the [Direct2D sample](https://github.com/microsoft/windows-samples-rs/blob/2e6c5578a2d6199eeabd9d49389290c4ced68f3c/direct2d/src/main.rs#L638-L660) for an example).

Ideally, clients should be able to just call `GetWindowLongPtrA` and have their code compile across architectures.

### Drawbacks

I'm not aware of any adverse effects in providing this.

### Rationale

The developer experience for clients of `windows` and `windows-sys` should be on par with that of C and C++ developers using the Windows SDK. This feature request addresses an issue where this is not the case.

### Alternatives

The alternative is propagating the responsibility for writing bitness-adaptive code to clients or crate authors.

### Prior art

The [`winapi`](https://crates.io/crates/winapi) crate solves this by [re-exporting](https://github.com/retep998/winapi-rs/blob/5b1829956ef645f3c2f8236ba18bb198ca4c2468/src/um/winuser.rs#L5049-L5056) the `*-Ptr`-variants as aliases for 32-bit platforms, e.g.

```rust
#[cfg(target_pointer_width = "32")]
pub use self::GetWindowLongA as GetWindowLongPtrA;
```

### Unresolved questions

- Are there other API calls that fit this category?
- Should a solution be metadata-driven or would a hand-crafted implementation (akin to what `winapi` does) be manageable enough?

kennykerr:
I was hoping for a more generalized solution, but this will suffice until I think of something better. As long as there aren't any more of these.

Fixes #1493


crates/libs/bindgen/src/lib.rs
```diff
@@ -109,12 +109,23 @@ pub fn namespace(gen: &Gen, tree: &Tree) -> String {
...
     };
 
+    if tree.namespace == "Windows.Win32.UI.WindowsAndMessaging" {
+        tokens.combine(&quote! {
+            #[cfg(target_pointer_width = "32")]
+            #[cfg(feature = "Win32_Foundation")]
+            pub use SetWindowLongA as SetWindowLongPtrA;
+            #[cfg(target_pointer_width = "32")]
+            #[cfg(feature = "Win32_Foundation")]
+            pub use GetWindowLongA as GetWindowLongPtrA;
```
tim-weis, line 125: This seems to be missing the Unicode versions (`SetWindowLongPtrW` and `GetWindowLongPtrW`). Is this intentional or an oversight?

There's also a slight discrepancy between types: For 32-bit targets the functions accept/return `i32`, for 64-bit targets `isize`. I wonder if that has any practical implications for clients. Current stable Rust (1.62.1) doesn't warn about superfluous casts (such as `0_isize as isize`), so I'm guessing this is ok.

crates/tests/window_long/tests/sys.rs
```diff
@@ -0,0 +1,10 @@
+use windows_sys::Win32::UI::WindowsAndMessaging::*;
+
+// Validates that these possibly aliased functions can be called on all supported targets.
+#[test]
+fn test() {
+    unsafe {
+        SetWindowLongPtrA(0, GWLP_USERDATA, 0);
+        GetWindowLongPtrA(0, GWLP_USERDATA);
```
tim-weis, line 8: Should include tests for `SetWindowLongPtrW` and `GetWindowLongPtrW`.

crates/tests/window_long/tests/win.rs
```diff
@@ -0,0 +1,10 @@
+use windows::Win32::UI::WindowsAndMessaging::*;
+
+// Validates that these possibly aliased functions can be called on all supported targets.
+#[test]
+fn test() {
+    unsafe {
+        SetWindowLongPtrA(None, GWLP_USERDATA, 0);
+        GetWindowLongPtrA(None, GWLP_USERDATA);
```
tim-weis, line 8: Same as above.

tim-weis (commented):
Looks good overall and works for `windows` as well as `windows-sys` without code duplication. :tada:

crates/libs/bindgen/src/lib.rs
kennykerr: I left out the W versions since the A and W versions are identical under the hood - there's really no reason to have both functions. Another less confusing approach might be to consolidate them by removing the W versions and dropping the suffix from the remaining version. Then you'd effectively get the post-macro names from C++ e.g. `SetWindowLong` and `GetWindowLong`. 

Yes, I wondered about the return type as well but all the samples I ported handled it naturally so this seems to work fine in practice.

crates/tests/window_long/tests/sys.rs
kennykerr: See other comment.

crates/tests/window_long/tests/win.rs
kennykerr: See other comment.

crates/libs/bindgen/src/lib.rs
tim-weis: I don't feel that this type of consolidation should be done here. That's really something the Platform/SDK Team would need to initiate.

Whether the implementations are identical or not is not something us developers can observe. We have to go by the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlongptra) and as long as that documentation explicitly says: "Don't mix encodings" then that's what those of us that program against a documented contract will have to do.

I'm not opposed to keeping things as simple as possible. I just feel that attempting to fix this here is going to lead to a more confusing situation.

kennykerr: That's fair. In that case, I'm inclined to do nothing and just let this be handled downstream, perhaps by a dedicated crate.

riverar: I hear ya but getting the platform to fix this will be impossible. And creating/consuming a separate crate to fix this special case is not ideal. I think we should keep these as ugly as they are.

tim-weis: > I think we need to add the W variants tho.

Agreed. As far as I understand, the ANSI and Unicode versions are distinct exports, with distinct ordinals and distinct addresses. Leaving either one out is harmful to clients that need the function addresses for reasons other than calling them. Not an issue with `windows` (which doesn't expose the raw entry points), but for `windows-sys` it is, and it would just feel awkward to not have those around for reasons we cannot reasonably verify.

kennykerr: I'll refrain from discussing the implementation details, but agree that it would be more consistent to provide both. Coming up.

kennykerr:
Added both A and W variants as well as tests for both.

PR #1936, opened 2022-07-22
Adding documentation for factory

### LINKED ISSUE #1733: Need to document `factory` function
sorry to bump this closed issue, but i think something that is also tricky is that the factory function isn't present in the documentation for the core module (as far as i could find) 

I would expect the documentation to be here: 
https://microsoft.github.io/windows-docs-rs/doc/windows/core/index.html

_Originally posted by @denniswjeong in https://github.com/microsoft/windows-rs/issues/1565#issuecomment-1116983843_

BenJKuhn:
This pull request updates the factory function to be included in the documentation set, since it's useful for accessing WinRT interop interfaces implemented on activation factory objects.

Fixes #1733


kennykerr (approved):
Thanks!

Issue #1890, opened 2022-07-11
The `windows` and `windows-sys` crates need their own `readme` files


kennykerr:
These need to be distinct from each other, since the crates are quite different, and distinct from that of the repo to avoid confusion.

@riverar https://github.com/microsoft/windows-rs/pull/1888#discussion_r918412546

kennykerr:
I streamlined the readme a bit and don't think this is urgent, so I'll close this for now.

PR #1940, opened 2022-07-27
Used std::ptr::null() instead of None in readme

blwarren:
The current example in the readme produces a mismatched type error in which CreateEventW is expecting either a *const SECURITY_ATTRIBUTES or null. The error is corrected by replacing None in the example with std::ptr::null().

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -42,7 +42,7 @@ fn main() -> Result<()> {
     assert!(root.InnerText()? == "hello world");
 
     unsafe {
-        let event = CreateEventW(None, true, false, None)?;
+        let event = CreateEventW(std::ptr::null(), true, false, None)?;
         SetEvent(event).ok()?;
         WaitForSingleObject(event, 0);
         CloseHandle(event).ok()?;


kennykerr:
The repo's samples reflect the latest updates to the `windows` crate rather than what is available on crates.io - in other words, it is correct for the *next* version of the `windows` crate. Related: #1890

blwarren:
Ah, I understand. I made the mistake of working from the repo readme instead of the crates.io docs. Lesson learned! Thanks!

msyyces8x95:
@kennykerr I checked using the last version from git, it works as intended : `None` instead of `std::ptr::null()` works fine.
A genuine question, next version will break backward compatibility for 'windows' crate ? compiling old code created with master branch of the crate produces some error. 

Example : 
code : `let conf_event = CreateEventW(std::ptr::null(), true, false, None)?;`
error : `Error : expected enum 'Option', found *-ptr`

rylev:
Yes, the windows crate does not yet guarantee backwards compatibility between versions. The rate of change is slowing, and hopefully we'll get there soon, but we're not there yet.

PR #1941, opened 2022-07-27
Support fixed-size array parameters

### LINKED ISSUE #1938: Add slice support for `MemorySize` attribute

kennykerr:
Following on  #1939 this update adds support for Win32 parameters that are declared as fixed-size arrays. This avoid any confusion around treating these arrays as references. It also more consistently applies `Option` to optional parameters rather than trying to transform empty slices into null pointers under the hood. 

Unfortunately, it also highlighted a few more metadata bugs:

https://github.com/microsoft/win32metadata/issues/1008
https://github.com/microsoft/win32metadata/issues/1010
https://github.com/microsoft/win32metadata/issues/1011
https://github.com/microsoft/win32metadata/issues/1012
https://github.com/microsoft/win32metadata/issues/1014

Fixes #1938


kennykerr:
I used BCrypt to validate a lot of this work. 

https://github.com/microsoft/windows-rs/blob/fixed-arrays/crates/tests/bcrypt/tests/tests.rs

It's now a lot easier to reason about use such APIs from Rust thanks to the awareness of optional and array parameters.

crates/libs/bindgen/src/gen.rs
```diff
@@ -922,23 +922,23 @@ impl<'a> Gen<'a> {
...
-                        }
-                        ArrayInfo::RelativeLen(_) => {
-                            if self.reader.param_flags(param.def).output() {
-                                quote! { ::core::mem::transmute(::windows::core::as_mut_ptr_or_null(#name)), }
-                            } else {
-                                quote! { ::core::mem::transmute(::windows::core::as_ptr_or_null(#name)), }
-                            }
+                                quote! { #name.as_ptr() }
+                            };
+                            quote! { ::core::mem::transmute(#map), }
```
rylev, line 932: Since we know this is a pointer does `as _` work instead? Would be nice to avoid another use of transmute.

crates/tests/bcrypt/tests/tests.rs
```diff
@@ -0,0 +1,64 @@
...
+pub const fn trim_null_end(mut bytes: &[u8]) -> &[u8] {
+    while let [rest @ .., last] = bytes {
+        if *last == 0 {
+            bytes = rest;
+        } else {
+            break;
+        }
+    }
+    bytes
+}
```
rylev, line 64: Is there a reason this is const?

Could be rewritten as:
```rust
fn trim_null_end(bytes: &[u8]) -> &[u8] {
    let end = bytes
        .iter()
        .rposition(|&s| s != 0)
        .map(|i| i + 1)
        .unwrap_or(bytes.len());
    &bytes[..end]
}
```

rylev (approved):
This looks like a nice change!

crates/libs/bindgen/src/gen.rs
kennykerr: Not just yet - it still requires a bit more untangling before that will work. There are a few cases where the underlying types aren't simply the pointer equivalent of the slice. But I'd love to get there.

crates/tests/bcrypt/tests/tests.rs
kennykerr: It doesn't but the iterator approach seems overly complicated in comparison.

PR #1956, opened 2022-08-08
feat: Add clearer Debug implementation for enum/define wrappers

poliorcetics:
Note: this will not affect type defined in windows-rs `core` since their `Debug`
impl is not aware of the constants linked to the type, so `HRESULT` and
`WIN32_ERROR` won't be changed. It would be nice if we found a way to automate
this for them too.

### DIFF
--- a/crates/libs/bindgen/src/enums.rs
+++ b/crates/libs/bindgen/src/enums.rs
@@ -107,6 +107,37 @@ pub fn gen(gen: &Gen, def: TypeDef) -> TokenStream {
         });
     }
 
+    if is_scoped || !gen.sys {
+        let debug_details = fields.iter().map(|(field_name, value)| {
+            let field_name = field_name.as_str();
+            let value = value.as_str();
+            if is_scoped {
+                quote! {
+                    Self::#field_name => (#field_name, #value),
+                }
+            } else {
+                // This branch is `!gen.sys`
+                quote! {
+                    #field_name => (#field_name, #value),
+                }
+            }
+        });
+
+        let type_name = type_name.name;
+        tokens.combine(&quote! {
+            #features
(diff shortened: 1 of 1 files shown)


poliorcetics:
I haven't regenerated the code because I don't know how to do it from a macOS machine. If you tell me how, I can do it, else you (the reviewer) can do it on your side and push a commit here, that's okay too.

riverar:
Can you summarize the issue this resolves and maybe show the before/after? Or link this to an existing issue?

poliorcetics:
Ah sorry, no existing issue, this comes from my usage of the crate but I can explain it here:

When debugging/seeing values in errors, the output is `Type(value)`, like `NSTATUS(3221227524)` for `STATUS_CONTENT_BLOCKED`, which is defined as `0xC0000804`.

With this change, I want to have a debug output of `NSTATUS::STATUS_CONTENT_BLOCKED(0xC0000804)` when possible, that way it's much easier to search for error codes and enum variants in general.

This can only be done if the enum is generated via the  `bindgen` crate in this repo, so `WIN32_ERROR` and `HRESULT` are not affected I think, though I would like to make it so for them too.

kennykerr:
Thanks for the contribution! Sorry I've been away for a few weeks. While this would be nice in some ways, I'm a little reluctant due to the excessive amount of extra string literals it would have to inject into the pre-generated crates. I'll close this PR for now, but feel free to open an issue to discuss further. We'd have to analyze the impact in various ways before considering and we can more easily discuss all of that in an issue.

poliorcetics:
> We'd have to analyze the impact in various ways before considering and we can more easily discuss all of that in an issue.

Opened #1981

PR #1961, opened 2022-08-14
Use the msvc target crates for gnullvm targets

### LINKED ISSUE #1964: Reducing the size of the import library crates
(This issue is focused on import libraries, but I'll also open issues about other size aspects for in the coming days.)

Vendoring crates with large amounts of data is costly. More so when semver incompatible versions are released often and different crates in the ecosystem use different versions.

The first and most obvious issue is that the version for the import libraries should be separate from that of the other crates. There hasn't been changes to the `crates/targets/*/lib` files since 0.36.0, yet, there have been 0.37.0, 0.38.0, 0.39.0 since then. Even if there had been changes, semver compatible versions would have been fine. I would suggest decoupling the versions and to publish 0.36.2, 0.37.1 and 0.38.1 "dummy" versions that depend on 0.39.0.

The second issue is the multiplication of crates for new targets, when they are not strictly necessary (especially in the context of what's coming below). This is addressed in #1961.

A third issue is that depending on the tooling used, the import library size can vary (for instance, if I run the tool_gnu crate right now, I somewhat end up with smaller import libraries). Relatedly, even if the same size is produced, the content is different because of timestamps within the libraries.

The remaining issue is that of the size of the import libraries themselves.

My proposed plan of action is the following (I have working patches for those, although they need some cleanup and proper commit messages):
- Remove the separate crates for gnullvm targets (#1961)
- Change tool_msvc to use the LLVM tools to produce the import libraries: it's faster, the resulting import libs are 25% smaller and deterministic. Bonus: it allows to build the import libs on any OS. I don't think there is any concern about using "non-native" import libraries with MSVC because that's effectively what `raw-dylib` support is doing already.
- Change tool_gnu to generate more deterministic import libraries (working around https://sourceware.org/bugzilla/show_bug.cgi?id=29489 and some other sources of non-determinism).
- Change tool_gnu to not require two different MINGw environments.
- Tweak tool_gnu to reduce the size of the import libraries it produces (saving ~388K per importlib)
- Merge tool_gnu and tool_msvc because at that point, they're 95% identical, and it is easier to run just one command to refresh all the import libraries.

It's possible to make the LLVM variants something like <100K lighter (IIRC), but that makes MSVC not work with them, so I don't think it's worth bothering having separate crates for that small of a difference.

BTW, while digging in the import libs, I found that the opposite problem of #1447 exists for libraries like windows.data.pdf (all libraries with a period in their name). Fixing it ends up being easier after the work above.

Further down the road, I'm planning to submit code to LLVM to produce MINGw-style import libraries (i.e. not using the short import form), incidentally, it should produce files about ~3MB smaller each if my math is right. (This will also allow rustc's raw-dylib support to stop requiring MINGw binutils).

Eventually, I assume the goal would be for windows/windows-sys to use raw-dylib, removing the need for the import library crates entirely, but the crates will still be desirable to have around for compatibility with older (now current) versions of the rust compiler for some while. I think at that point it would be desirable to replace those crates with code that generates the import libraries on demand when the compiler version is older than the one that stabilizes raw-dylibs (which seems it might be 1.65.0?). I have some sketch code that produces MINGw-style objects already, producing the full import libraries would not be much more effort code-wise, and would be fast.

glandium:
The import libraries in crates/target/*_gnullvm only differ from the
ones in crates/target/*_msvc the following ways:
- their file name
- they don't contain the second "/" linker member with symbol names in
  lexical order.
- the members are not in the same order.

That's about it. Both binutils ld and lld will happily find $lib.lib
import libraries as well as lib$lib.a ones, so the file name doesn't
matter. The other differences also don't have any practical consequence.

This means it is not necessary to add extra crates for the purpose of
the gnullvm target, especially when e.g. `cargo vendor` will want to
vendor them all, and we can reuse the msvc ones, although their name
could cause some confusion.

Fixes #1964


riverar:
Hey @glandium, we explicitly asked for the separate crate https://github.com/microsoft/windows-rs/pull/1846#pullrequestreview-1026401722 as
* we don't think making LLVM ingest a _non native_ lib format is a good idea long term
* having target specific libs lets us leverage optimizations, flags, tricks, etc. that may be specific to that toolset

But we're open minded. Is there a problem with what we're doing today that we should consider? Please create an issue to discuss!

glandium:
I will file an issue shortly for various changes I have pending wrt the import libraries. I wanted to get this one out of the way first, before the extra crates end up published.

kennykerr:
Thanks for the contribution! Sorry I've been gone for a few weeks. I'll close this PR for now but let's continue the discussion #1964 as there appears to be a lot of open questions.

PR #1967, opened 2022-08-17
Produce the *_gnu import libraries more deterministically

glandium:
This addresses item 3 in #1964.


kennykerr:
> This addresses item 3 in #1964.

Thanks for the fix! Sorry to be a stickler, but could you open a new issue for this PR? It's not clear to me what "item 3" refers to. 

@riverar is this related to the consistent application of full file names in the win32 metadata that you've been working on?

riverar:
This PR stabilizes dlltool temporary file output (which, by default, uses the process ID in its naming scheme) for reproducible build purposes. Isn't related to our suffix issue.

riverar (approved):

PR #1968, opened 2022-08-17
Remove unnecessary symbols and sections from *_gnu import libs

glandium:
This cuts their size by more than 20% by removing what is essentially
dead-weight:
 crates/targets/i686_gnu/lib/libwindows.a   | Bin 14009016 -> 10924886 bytes
 crates/targets/x86_64_gnu/lib/libwindows.a | Bin 13879932 -> 10487912 bytes

Note this PR includes the commits from PR #1967 because github doesn't do PR dependencies.

### DIFF
--- a/crates/tools/gnu/src/main.rs
+++ b/crates/tools/gnu/src/main.rs
@@ -91,10 +91,22 @@ EXPORTS
     }
     cmd.output().unwrap();
 
-    // Work around lack of determinism in dlltool output.
+    // Work around lack of determinism in dlltool output, and at the same time remove
+    // unnecessary sections and symbols.
     std::fs::rename(output.join(format!("lib{}.a", library)), output.join("tmp.a")).unwrap();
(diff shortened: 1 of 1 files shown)


kennykerr:
> This cuts their size by more than 20% 

This sounds great! @riverar is this a no-brainer or should we discuss further with an issue? 

As I mentioned elsewhere, I'd like to avoid unnecessary churn in the targets if `raw-dylib` is close to stabilization.

riverar (approved):
Looks good; let's merge https://github.com/microsoft/windows-rs/pull/1967 first.

kennykerr:
Merged #1967 - this PR now has merge conflicts - once that's cleared we can get it in - thanks!

riverar (approved):

mati865:
This is just the legacy format of import libs. Short import libraries created by LLVM for `gnullvm` target are by far the smallest at around 3 MiB, while short import libraries produced by MSVC are around 4 MiB.

Issue #1969, opened 2022-08-17
Test (cross-compilation) actions fail on this repository


glandium:
The error message is:
> actions/cache@v3 is not allowed to be used in microsoft/windows-rs. Actions in this workflow must be: within a repository that belongs to your Enterprise account or matching the following: actions/checkout@v2.

riverar:
Where are you seeing this? Can you link to the particular test run? I don't see any failures in recent commits.

glandium:
https://github.com/microsoft/windows-rs/actions

riverar:
I walked right into that one, thanks.

kennykerr:
Need to figure out why its not failing the CI validation for PRs - and then we should be able to add the necessary action.

kennykerr:
I allowed that cache action but it still seems to fail:

https://github.com/microsoft/windows-rs/actions/runs/2873464088

Might need to revert https://github.com/microsoft/windows-rs/pull/1883 until we can figure this out.

mati865:
Oh, that's unfortunate but probably there is no other choice.
It'd be nice if https://github.com/microsoft/windows-rs/pull/1961 could be reconsidered instead before next release.

riverar:
Last build seemed to pass https://github.com/microsoft/windows-rs/actions/runs/2905093873, looks okay now.

kennykerr:
The build passed but it still skipped the cross compilation workflow.

kennykerr:
OK, I was looking at an older build that should also have succeeded but at least the latest build seems green. 👍

kennykerr:
So, I'll close this now - the fix was to enable the cache action (which I did) and the CI build finally seems to have caught up to that fact. 😜

PR #1971, opened 2022-08-18
Downgrade cache action to comply with organizational policy

riverar:
Fixes: #1969


riverar:
Ah it's worse than I thought. Will chat offline with @kennykerr.

riverar:
Closing, won't need this. We're working on approving v3.

Issue #1977, opened 2022-08-22
Bug: /crates/samples/create_window/src/main.rs Line 40 expected *-ptr, found enum `Option`


bmariuszb:
### Which crate is this about?

windows

### Crate version

0.39.0

### Crate manifest

```toml
[package]
name = "winows"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies.windows]
version = "0.39.0"
features = [
    "Win32_Foundation",
    "Win32_Graphics_Gdi",
    "Win32_System_LibraryLoader",
    "Win32_UI_WindowsAndMessaging",
]
```

### Expected behavior

Sample runs without errors.

### Actual behavior

error[E0308]: mismatched types
     --> src\main.rs:40:38
      |
40    |                 ValidateRect(window, None);
      |                 ------------         ^^^^ expected *-ptr, found enum `Option`
      |                 |
      |                 arguments to this function are incorrect
      |
      = note: expected raw pointer `*const RECT`
                        found enum `Option<_>`

bmariuszb:
yes

kennykerr:
The samples in the repo compile with the latest, usually pre-release, version of the `windows` or `windows-sys` crate. To find the samples for a particular release you can use a specific release tag. For example:

https://github.com/microsoft/windows-rs/tree/0.39.0

That will give you the samples that compile with version 0.39 of the `windows` crate.

bmariuszb:
Maybe you could add that note to readme in samples directory?

kennykerr:
So yes, I'd love to but Cargo 1.46 doesn't appear to support that. #1980

Perhaps in future when we no longer need to validate against 1.46 for `windows-sys`.

kennykerr:
Fixed! #1991

PR #1980, opened 2022-08-23
Add samples readme

kennykerr:
This adds a brief explanation for how to find the correct samples for a given release of the `windows` and `windows-sys` crates. 

Fixes: #1977

### DIFF
--- a/crates/samples/readme.md
+++ b/crates/samples/readme.md
@@ -1,0 +1,6 @@
+The samples in the repo compile with the latest, usually pre-release, version of the `windows` or `windows-sys` crate. 
+To find the samples for a particular release you can use a specific release tag. For example:
+
+https://github.com/microsoft/windows-rs/tree/0.39.0
+
+That will give you the samples that compile with version 0.39 of the windows crate.


rylev:
@kennykerr what was the error you encountered? Seems strange that a markdown file would trip up cargo.

ChrisDenton:
You may need to add an exclude to [windows-rs/Cargo.toml](https://github.com/microsoft/windows-rs/blob/f06b0bb3e88bffc1ce18cd1ea65f1bcbe587afa7/Cargo.toml) otherwise it's matched by the `crates/samples/*` member. See [the Cargo docs](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspace-section).

kennykerr:
https://github.com/microsoft/windows-rs/runs/7976243489?check_suite_focus=true

rylev:
@ChrisDenton's suggestion is exactly what you need to do:

```toml

[workspace]
members = [
  "samples/*",
  # ...
]
exclude = ["samples/readme.md"]
```

kennykerr:
Its interesting that it works with newer versions of cargo, but that's a reasonable workaround - thanks!

kennykerr:
This issue was also addressed some time between 1.46 and 1.49 so I think I'll just update the MSRV of `windows-sys` and avoid complicating the workspace. #1987

riverar:
This particular issue was fixed in Cargo 1.47 (https://github.com/rust-lang/cargo/pull/8511)

PR #1987, opened 2022-08-26
Update min rust version for `windows-sys` to 1.49

kennykerr:
This update just bumps the minimum supported Rust version for the `windows-sys` crate from 1.46 to 1.49 to pick up some Cargo fixes needed to simplify the build, like the issue encountered in #1980. I chose 1.49 as that version has already been picked as the min version for the prominent `parking_lot` crate that depends on the `windows-sys` crate so should be broadly adopted.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
     name: Check windows-sys
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


rylev (approved):
Seems reasonable as 1.49 is over a year and a half old. However, it would be nice if we had a bit more formal of policy on MSRV

Issue #1989, opened 2022-08-28
`windows-sys` without CRT?


Zymlex:
As I see, `windows-sys` uses `no_std`, but still try to link with CRT.
Is it possible to use `windows-sys` without CRT?

upd: https://github.com/microsoft/windows-rs/issues/1989#issuecomment-1230421401

tim-weis:
Could you provide a repro?

I'm not sure I'm seeing the same as you are, however, the (failed) imports I see aren't *actually* from the CRT, but rather the SEH infrastructure (`__CxxFrameHandler3`, `_CxxThrowException`). I don't know where those are getting pulled in (or the `memset`s further down).

This is the repro I'm using:

*Cargo.toml*

```toml
[package]
name = "nostd_windows-sys"
version = "0.0.0"
edition = "2021"

[dependencies.windows-sys]
version = "0.36.1"
features = [
    "Win32_Foundation",
    "Win32_System_Threading",
]
```

*src/main.rs*

```rust
#![no_std]
#![no_main]

use core::panic::PanicInfo;

use windows_sys::Win32::System::Threading::{GetCurrentProcess, TerminateProcess};

#[panic_handler]
fn panic_abort(_info: &PanicInfo) -> ! {
    unsafe {
        let h = GetCurrentProcess();
        TerminateProcess(h, 1);
    };
    unreachable!();
}

#[no_mangle]
pub extern "C" fn EntryPoint() -> u32 {
    0
}
```

*.cargo/config.toml*

```toml
[build]
rustflags = [
    "-C", "link-arg=/ENTRY:EntryPoint"
]
```

(Note: I don't know why the linker cannot find the entry point.)

riverar:
I'm seeing similar CRT functions getting referenced (__CxxFrameHandler3) as @Zymlex reported using @tim-weis's sample (after adding an empty eh_personality function and setting the subsystem):

```
...
 libcore-798bc7c2d986ac97.rlib(core-798bc7c2d986ac97.core.db87af78-cgu.0.rcgu.o) : error LNK2019: unresolved external symbol memcpy referenced in function _ZN4core3num7flt2dec8strategy6dragon15format_shortest17h09419dd069410d1fE
[...]

PR #1992, opened 2022-08-29
Update the faq to call out `no_std` support

kennykerr:
Clarifies that the `windows-sys` crate offers no_std support. 

Related: #1989

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -7,6 +7,7 @@
 | Reason | `windows` | `windows-sys`|
 | --- | --- | --- |
 | Fast compile times are one of your top concerns | | ✅ |
+| You need `no_std` support | | ✅ |
 | You need COM or WinRT support | ✅ | |
 | You would prefer to use APIs that feel idiomatic to Rust | ✅ | |
 


riverar (changes requested):
There's no data to suggest this is accurate. The issue this fixes, for example, demonstrates no_std doesn't work.

kennykerr:
> There's no data to suggest this is accurate. The issue this fixes, for example, demonstrates no_std doesn't work.

This PR just clarifies that the `windows-sys` crate explicitly supports `no_std` compilation. This is compiler-enforced but if that's not helpful to include in the FAQ then I'm happy to ditch it. 

https://github.com/microsoft/windows-rs/blob/9627a203d99f90fb36c63fcb4120866a6686b18d/crates/libs/sys/src/lib.rs#L5

https://doc.rust-lang.org/reference/names/preludes.html#the-no_std-attribute

Again, this has nothing to do with whether or not the CRT is linked into a given downstream binary. The `windows-sys` crate has no control over that. Anyway, I'll close this for now.

ChrisDenton:
IMHO, the `no_std` support for windows-sys is one of the biggest motivating factor for using it instead of `windows` (the others being compile times and stability).

As kenny says, this has nothing to do with no-CRT support (which is a very very niche requirement) and in either case that has nothing to do with the `windows-sys` crate itself. It will support whatever the toolchain supports and has no requirements beyond `core`.

riverar (approved):

PR #1994, opened 2022-08-31
Remove unused `bindgen` options

kennykerr:
These options were used by the internal bindings as well as Xaml code generation and are no longer needed.

### DIFF
--- a/crates/libs/bindgen/src/gen.rs
+++ b/crates/libs/bindgen/src/gen.rs
@@ -4,31 +4,14 @@ pub struct Gen<'a> {
     pub reader: &'a Reader<'a>,
     pub namespace: &'a str,
     pub sys: bool,
-    pub flatten: bool,
     pub cfg: bool,
     pub doc: bool,
-    pub min_enum: bool,
-    pub min_inherit: bool,
-    pub min_xaml: bool,
-    pub windows_extern: bool,
     pub component: bool,
 }
 
 impl<'a> Gen<'a> {
     pub fn new(reader: &'a Reader) -> Self {
-        Self {
-            reader,
-            namespace: "",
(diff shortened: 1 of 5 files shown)


crates/libs/bindgen/src/gen.rs
```diff
@@ -4,31 +4,14 @@ pub struct Gen<'a> {
     pub reader: &'a Reader<'a>,
     pub namespace: &'a str,
     pub sys: bool,
-    pub flatten: bool,
     pub cfg: bool,
     pub doc: bool,
-    pub min_enum: bool,
-    pub min_inherit: bool,
-    pub min_xaml: bool,
-    pub windows_extern: bool,
```
riverar, line 13: So `windows_extern` is critical for anyone that wants to build on top of this crate and expose features dependent on upstream features (e.g. [windows-app-rs](https://github.com/microsoft/windows-app-rs/blob/master/Cargo.toml)). I don't think we have any active crates using this at this very moment, but I'd vote for keeping this as it has minimal impacts to `bindgen` and removes a significant crate adoption roadblock, especially for those looking to transform additional Microsoft SDKs.

riverar (approved):
Looks good, though I made a plea for `windows_extern`.

crates/libs/bindgen/src/gen.rs
kennykerr: I'm just not sure how common that is or whether that was something you'll only hit when you're doing some crazy like Xaml or WinAppSDK. The DIA example doesn't appear to need that.

riverar: Fair. We can re-add easy enough when a new case presents itself.

PR #1996, opened 2022-09-01
Update Win32 metadata version from 24.0.1 to 30.0.12

kennykerr:
This update includes a rollup of fixes and improvements to the Win32 metadata. 

More information about what's changed: https://github.com/microsoft/win32metadata/releases


kennykerr:
@riverar & @glandium I will follow up with updates to support new calling convention metadata and publish new libs.

PR #1997, opened 2022-09-01
Remove workaround for `GUID` constants

kennykerr:
Now that https://github.com/microsoft/win32metadata/pull/1204 has been fixed, we no longer need this workaround to harmonize `GUID` constants.


rylev (approved):

kennykerr:
Sigh. Turns out we need to keep the workaround. 

https://github.com/microsoft/win32metadata/issues/1212

Issue #1975, opened 2022-08-22
How to cast C++ `IUnknown` pointer to Rust `IUnknown` struct?


seungha-yang:
I'm currently trying to cast C++ IUnknown pointer (`*mut c_void`) to rust IUnknown via `std::mem::transmute_copy` 
but not successful so far.

Is it supposed to be working or is there any guide for the scenario?

kennykerr:
It's not particularly easy because it's not particularly safe. There's also the question of whether the C++ pointer is owned or dangling. Do you want the Rust `IUnknown` to assume ownership - and thus call `Release` when it is dropped - or only reference the C++ pointer that is potentially owned by someone else? 

If we assume you're only borrowing someone else's pointer, then it would look something like this:

```rust
let raw: *mut std::ffi::c_void = ptr; // someone else will call `ptr->Release()`
let unknown: &windows::core::IUnknown = std::mem::transmute(&raw);
```

You can now use `unknown` for as long as `raw` is valid...

If we assume you're taking ownership of the pointer, then it would look something like this:

```rust
let raw: *mut std::ffi::c_void = ptr; // nobody else will touch `ptr` again
let unknown: windows::core::IUnknown = std::mem::transmute(raw);
```

You can now use `unknown` for as long as it is valid as it owns a reference and will call `Release` when it is dropped.

seungha-yang:
Thanks for the feedback. My case is borrowing C++ pointer and it was indeed ownership issue. I worked around it via `core::mem::forget` or something else so that rust `IUnknown` can be stay in the scope without `IUnknown::Release`.

I understood that this kind of stuff is hard (or almost impossible) to be safe but I'm wondering if there can be a higher level and a bit safer helper method. Particularly I'm working on rust binding for [GStreamer D3D11 library](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/tree/main/subprojects/gst-plugins-bad/gst-libs/gst/d3d11) and [plugin](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/tree/main/subprojects/gst-plugins-bad/sys/d3d11)

sdroege:
Such helper API would certainly be useful, even if it has to be `unsafe`. It's not obvious which memory representations are equivalent and having to call `transmute()` with all it's pre-conditions is a lot less safe than having e.g.

```rust
fn from_raw_owned(ptr: *mut c_void) -> IUnknown;
[...]

Issue #1885, opened 2022-07-11
implement struct add lifetime 


gityoog:
```rust
#[implement(DWebBrowserEvents2)]
pub struct BrowserEvent<'a> {
    parent: &'a Browser,
    on_complete: Box<dyn Fn() + 'a>,
}
``` 

i want to add a lifetime for BrowserEvent, but
![QQ截图20220711175510](https://user-images.githubusercontent.com/36879029/178239103-36013293-0195-42c8-9bf1-7ef7f0a95f1e.png)

kennykerr:
That isn't supported and I'm not sure whether we can or should support it for objects with COM lifetime, but @rylev should have a better idea of that.

kennykerr:
Unless someone can offer a compelling reason for this feature, I'm going to close this issue as unsupported. Freel free to chime in with suggestions.

gityoog:
> Unless someone can offer a compelling reason for this feature, I'm going to close this issue as unsupported. Freel free to chime in with suggestions.

I want to do something in `Invoke`, such as triggering callback or using ref of other structs, which all need to declare the life cycle

expamle:
```rust
// when the life cycle is not supported, i can only use SendMessage or raw pointer
pub struct BrowserEvent<'a> {
    browser: &'a Browser
    js_callback:  RefCell<HashMap<String, Box<dyn Fn(*const DISPPARAMS) -> Option<VARIANT> + 'a>>>
}
impl IDispatch_Impl for BrowserEvent {
    // ...
    fn Invoke(
        &self,
        dispidmember: i32,
        _riid: *const windows::core::GUID,
        _lcid: u32,
        _wflags: u16,
        pdispparams: *const DISPPARAMS,
        _pvarresult: *mut VARIANT,
        _pexcepinfo: *mut EXCEPINFO,
        _puargerr: *mut u32,
    ) -> windows::core::Result<()> {
        match dispidmember as u32 {
            DISPID_BEFORENAVIGATE2 => unsafe {
                // (*pdispparams).rgvarg.write(val);
                let url = (*(*(*pdispparams).rgvarg.add(5))
                    .Anonymous
                    .Anonymous
                    .Anonymous
                    .pvarVal)
                    .Anonymous
                    .Anonymous
                    .Anonymous
                    .bstrVal
                    .to_string();
                // need to get allow list from &browser
                if !url.eq("about:blank") {
                    (*(*(*pdispparams).rgvarg).Anonymous.Anonymous)
[...]

PR #2010, opened 2022-09-07
Adds `unsafe` helpers to cast from C++ or raw `IUnknown` pointers to Rust `IUnknown` values

kennykerr:
Fixes: #1975

### DIFF
--- a/crates/libs/windows/src/core/unknown.rs
+++ b/crates/libs/windows/src/core/unknown.rs
@@ -7,6 +7,18 @@ use super::*;
 #[repr(transparent)]
 pub struct IUnknown(core::ptr::NonNull<core::ffi::c_void>);
 
+impl IUnknown {
+    /// Creates an `IUnknown` value by taking ownership of the `raw` pointer.
+    pub unsafe fn from_raw_owned(raw: *mut core::ffi::c_void) -> Self {
+        std::mem::transmute(raw)
+    }
+
+    /// Creates a borrowed `IUnknown` that is valid so long as the `raw` pointer is valid.
+    pub unsafe fn from_raw_borrowed<'a>(raw: &'a *mut core::ffi::c_void) -> &'a Self {
+        std::mem::transmute(raw)
+    }
+}
+
 #[doc(hidden)]
 #[repr(C)]
 pub struct IUnknownVtbl {
--- a/crates/libs/windows/src/core/unknown.rs
+++ b/crates/libs/windows/src/core/unknown.rs
@@ -8,12 +8,22 @@ use super::*;
 pub struct IUnknown(core::ptr::NonNull<core::ffi::c_void>);
 
 impl IUnknown {
-    /// Creates an `IUnknown` value by taking ownership of the `raw` pointer.
-    pub unsafe fn from_raw_owned(raw: *mut core::ffi::c_void) -> Self {
+    /// Creates an `IUnknown` value by taking ownership of the `raw` COM interface pointer.
+    ///
+    /// # Safety
+    ///
+    /// The `raw` pointer must be owned by the caller and represent a valid COM interface pointer. In other words,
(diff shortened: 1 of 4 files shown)


crates/libs/windows/src/core/unknown.rs
```diff
@@ -7,6 +7,18 @@ use super::*;
 #[repr(transparent)]
 pub struct IUnknown(core::ptr::NonNull<core::ffi::c_void>);
 
+impl IUnknown {
+    /// Creates an `IUnknown` value by taking ownership of the `raw` pointer.
```
rylev, line 11: It would be good to point out what the `raw` pointer is expected to be (i.e., an owned pointer to a vtable pointer).  You should also note that when the returned value is dopped, the reference count will be decremented. 

It's convention to put the requirements for using this function safely in a section labeled `# Safety`.

crates/libs/windows/src/core/unknown.rs
```diff
@@ -7,6 +7,18 @@ use super::*;
...
+    pub unsafe fn from_raw_owned(raw: *mut core::ffi::c_void) -> Self {
```
rylev, line 12: You might consider calling this just `from_raw` as usually the assumption is that `from` functions return owned values. While this doesn't have symmetry with the `from_raw_borrowed` function it is symmetrical with other `from_raw` functions for other types.

crates/libs/windows/src/core/unknown.rs
```diff
@@ -7,6 +7,18 @@ use super::*;
...
+        std::mem::transmute(raw)
+    }
+
+    /// Creates a borrowed `IUnknown` that is valid so long as the `raw` pointer is valid.
```
rylev, line 16: Same with this. It's very much not obvious what `raw` is supposed to be a pointer to.

crates/tests/core/tests/unknown.rs
```diff
@@ -0,0 +1,54 @@
...
+use windows::core::*;
+
+#[interface("f7ea748b-8121-41c1-aaee-406ba6f148a9")]
+unsafe trait ITest: IUnknown {
+    unsafe fn Test(&self) -> u32;
+}
+
+#[implement(ITest)]
+struct Test {
+    drop: *mut u32,
```
rylev, line 12: Just out of curiousity. Can this be made safe or does the `implement` macro not handle lifetimes?

```rust
#[implement(ITest)]
struct Test<'a> {
    drop: &mut u32,
}
```

crates/tests/core/tests/unknown.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        {
+            let raw_borrowed: *mut std::ffi::c_void = test.as_raw();
+            let unknown_borrowed: &IUnknown = IUnknown::from_raw_borrowed(&raw_borrowed);
+            assert_eq!(unknown_borrowed.as_raw(), test.as_raw());
+            let test_query: ITest = unknown_borrowed.cast().unwrap();
+            assert_eq!(test_query.Test(), 0);
+        }
+        {
+            let raw_owned: *mut std::ffi::c_void = std::mem::transmute(test.clone());
```
rylev, line 43: It would be nice if this could be replaced with:

```rust
let raw_owned = test.clone().into_raw();

```

crates/tests/core/tests/unknown.rs
kennykerr: Yep it doesn't handle lifetimes. #1885

crates/tests/core/tests/unknown.rs
```diff
@@ -52,3 +52,28 @@ fn test() {
         assert_eq!(dropped, 1);
     }
 }
+
+#[test]
+fn test_test() {
```
rylev, line 57:
```suggestion
fn test_pointer_conversion_functions() {
```

rylev (approved):

Issue #1985, opened 2022-08-25
Discrepancy in symbol names between i686*/lib/*windows.* and .lib files in windows SDK


glandium:
I think I missed some libraries from the SDK, but I dumped symbol lists from windows SDK/MSVC for x86 and compared to what's in i686_msvc, and there are tons of discrepancies (this is after #1979).

Discrepancies where the numbers are different (windows.lib on the left, SDK on the right):
__imp__CfUpdateSyncProviderStatus@8 __imp__CfUpdateSyncProviderStatus@12
__imp__CfReportProviderProgress@28 __imp__CfReportProviderProgress@32
__imp__CfReportProviderProgress2@40 __imp__CfReportProviderProgress2@44
__imp__CfQuerySyncProviderStatus@8 __imp__CfQuerySyncProviderStatus@12
__imp__CfDisconnectSyncRoot@4 __imp__CfDisconnectSyncRoot@8
__imp__RegisterClusterNotifyV2@24 __imp__RegisterClusterNotifyV2@28
__imp__EapHostPeerQueryUserBlobFromCredentialInputFields@44 __imp__EapHostPeerQueryUserBlobFromCredentialInputFields@48
__imp__EapHostPeerQueryCredentialInputFields@36 __imp__EapHostPeerQueryCredentialInputFields@40
__imp__EapHostPeerInvokeIdentityUI@60 __imp__EapHostPeerInvokeIdentityUI@64
__imp__EapHostPeerInvokeConfigUI@40 __imp__EapHostPeerInvokeConfigUI@44
__imp__EapHostPeerGetMethodProperties@48 __imp__EapHostPeerGetMethodProperties@52
__imp__EapHostPeerConfigBlob2Xml@32 __imp__EapHostPeerConfigBlob2Xml@36
__imp__EapHostPeerGetIdentity@64 __imp__EapHostPeerGetIdentity@68
__imp__EapHostPeerBeginSession@64 __imp__EapHostPeerBeginSession@68
__imp__alljoyn_interfacedescription_property_getannotationscount@12 __imp__alljoyn_interfacedescription_property_getannotationscount@16
__imp__alljoyn_interfacedescription_property_getannotationatindex@32 __imp__alljoyn_interfacedescription_property_getannotationatindex@36
__imp__alljoyn_interfacedescription_property_getannotation@24 __imp__alljoyn_interfacedescription_property_getannotation@28
[...]

PR #2015, opened 2022-09-09
Update the win32 metadata to version 32.0.17

kennykerr:
Update to the latest version of the win32 metadata that includes the fixes for the blocking calling convention and stack size issues. 

More info here: https://github.com/microsoft/win32metadata/releases/tag/v32.0.17-preview

Fixes: #1985


kennykerr:
Note that I have updated all the msvc libs but none of the gnu/llvm libs.

PR #2016, opened 2022-09-09
Misc. changes to import library generation tooling

glandium:
This covers the first 3 points in https://github.com/microsoft/windows-rs/issues/1964#issuecomment-1227046538 and the first commit mentioned in the 4th point, that reduces differences from updating the msvc import libraries variants.


kennykerr:
> Handle both *_gnu and *_gnullvm import libraries from the same command

If you do this, you need to rerun `cargo run -p tool_yml` to rebuild the build scripts.

crates/tools/gnu/src/main.rs
```diff
@@ -1,28 +1,38 @@
...
         return;
     };
 
+    for platform in platforms {
+        let tools = if platform.ends_with("_gnu") { &["dlltool", "ar", "objcopy"][..] } else { &["llvm-dlltool", "llvm-ar"][..] };
+        for tool in tools {
+            if which::which(tool).is_err() {
+                eprintln!("Could not find {}. Is it in your $PATH?", tool);
+                return;
+            }
```
riverar, line 28: We can remove the `which` dependency by just trying to invoke the tool. Alternatively, we could do nothing at all and just fail too, given this is mostly maintainer tooling. I think either approach is preferable over the dependency hit.

```suggestion
            if Command::new(tool).stdout(Stdio::null()).spawn().is_err() {
                eprintln!("Could not find {}. Is it in your $PATH?", tool);
                return;
            }
```

crates/tools/msvc/src/main.rs
```diff
@@ -40,6 +42,31 @@ fn main() {
...
+
+    // Clear out timestamps in the resulting library.
+    let mut archive = std::fs::File::options().read(true).write(true).open(output.join("windows.lib")).unwrap();
+    let len = archive.metadata().unwrap().len();
+    let mut header = [0u8; 8];
+    archive.read_exact(&mut header).unwrap();
+    assert_eq!(&header, b"!<arch>\n");
+    for num in 1.. {
+        archive.seek(SeekFrom::Current(16)).unwrap(); // identifier
+        assert_eq!(archive.write(b"0           ").unwrap(), 12); // replace archive timestamp
```
riverar, line 54: To align with the Windows SDK (e.g. kernel32.lib), let's set this to `-1`.

```suggestion
        assert_eq!(archive.write(b"-1          ").unwrap(), 12); // replace archive timestamp
```

The below timestamp replacement will also need a tweak.

riverar (changes requested):
Looks great, some minor changes requested.

crates/tools/msvc/src/main.rs
glandium: Interestingly kernel32.lib has random timestamps for the member themselves. That said, does it really matter to set this to -1? This introduces a new difference with the files that llvm-dlltool produces for no benefit.

riverar: I see `-1` for members in kernel32.lib from 22621.0. Looks like this is the sentinel value chosen by msvc and friends in the SDK, so we don't want to stray and cause problems. 

`dumpbin`, for example, already behaves differently when using `-1` and `0`:

```
// Timestamp set to 0
Archive member name at 25EBE: KERNEL32.dll/
       0 time/date Wed Dec 31 16:00:00 1969
         uid
         gid
       0 mode
     1F2 size
correct header end

// Timestamp set to -1
Archive member name at 25EBE: KERNEL32.dll/
FFFFFFFF time/date
         uid
         gid
       0 mode
     1F2 size
correct header end
```

glandium: That's the timestamp for from the ar headers, and it doesn't matter. The timestamps in the members themselves also don't matter, and they are random.
```
Archive member name at 25EBE: KERNEL32.dll/
FFFFFFFF time/date
         uid
         gid
       0 mode
     1F2 size
correct header end

FILE HEADER VALUES
            8664 machine (x64)
               3 number of sections
        FAC6008A time date stamp
             10E file pointer to symbol table
               8 number of symbols
               0 size of optional header
               0 characteristics

Archive member name at 260EC: KERNEL32.dll/   
FFFFFFFF time/date
         uid
         gid
       0 mode
      FB size
correct header end

FILE HEADER VALUES
            8664 machine (x64)
               2 number of sections
        D33470E4 time date stamp
              BA file pointer to symbol table
               2 number of symbols
               0 size of optional header
               0 characteristics
(...)

Archive member name at 26380: KERNEL32.dll/
FFFFFFFF time/date
         uid
         gid
       0 mode
      39 size
correct header end

  Version      : 0
  Machine      : 8664 (x64)
  TimeDateStamp: D1A71C8A
  SizeOfData   : 00000025
  DLL name     : KERNEL32.dll
  Symbol name  : AcquireSRWLockExclusive
  Type         : code
  Name type    : name
  Hint         : 0
  Name         : AcquireSRWLockExclusive
```
(that's from SDK 22621)

riverar: Ah yup, I see that now too thanks!

Given the demonstrable impact to `dumpbin`, I'm still leaning towards `-1` for the archive headers.

For the members though, we'll have to double-check with the msvc folks. We don't have enough information right now to know what "doesn't matter" or is "random". (Microsoft, for example, relies on random-looking PE TimeDateStamp values for symbol/image lookup in its symbol server.) I'm happy with whatever there for this PR then we go back and fix it later if needed.

kennykerr:
What is the purpose of post processing the libs and is that really necessary?

glandium:
> What is the purpose of post processing the libs

Assuming you use the same MSVC version, it makes the libs reproducible.

crates/tools/msvc/src/main.rs
tim-weis: Probably related to the random-looking timestamps: [Why are the module timestamps in Windows 10 so nonsensical?](https://devblogs.microsoft.com/oldnewthing/20180103-00/?p=97705).

glandium:
What's random is what's in the actual SDK. What's in the import libs provided by windows-rs right now is an actual timestamp of when the tool_msvc tool is run. What's in the import libs generated by raw-dylibs by rust is an zeroed timestamp. Also, these "timestamps" appear nowhere in the binaries linked against those import libraries. I think you're overthinking this. The timestamps that matter more as per @oldnewthing's blog post are those on .dlls and .exes, not import libs.

kennykerr:
> What's random is what's in the actual SDK. 

Well, that's not at all surprising. 🤣 I have no further objections. Thanks for your patience.

riverar:
We need to zoom out a bit and not assume these timestamps "random" and/or unnecessary. We don't have the data to support those assertions just yet. It's more likely these are internal implementation details (read: hacks) and are actually useful/used in some niche scenario.

I think this latest commit is a great _best effort_ alignment of our generated libs to the official SDK libs, down to the `-1`!

@glandium Appreciate your patience and compromise.

riverar (approved):

glandium:
Can this be merged?

kennykerr:
It looks like the gnullvm targets are still there. Should they be removed now?

glandium:
I'd rather get to that in a followup PR, especially since the modalities haven't been worked out yet.

glandium:
Thank you.

glandium:
I opened #2018 for discussion.

kennykerr:
Great, thanks for the continued contributions!

PR #2017, opened 2022-09-12
Remove win32metadata workarounds

kennykerr:
Removes various workarounds for bugs recently fixed by the win32 metadata project.


riverar (commented):
Looks good. Delayed due to work schedule this week.

Issue #2019, opened 2022-09-13
Remove dlltool prefix hack


glandium:
#1967 added a dlltool option (`-t`) to make the *_gnu import library member names more consistent across different versions of binutils. That required a sort of a hack to use a different prefix based on what the library name looked like, because of inconsistencies in the win32metadata. Those have since been fixed, so the hack can now be removed, which essentially looks like this:
```diff
diff --git a/crates/tools/gnu/src/main.rs b/crates/tools/gnu/src/main.rs
index aef4b96f9..e4bb99aba 100644
--- a/crates/tools/gnu/src/main.rs
+++ b/crates/tools/gnu/src/main.rs
@@ -109,11 +109,7 @@ EXPORTS
         });
         // Ensure consistency in the prefixes used by dlltool.
         cmd.arg("-t");
-        if library.contains('.') {
-            cmd.arg(format!("{}_", library).replace('.', "_").replace('-', "_"));
-        } else {
-            cmd.arg(format!("{}_dll_", library).replace('-', "_"));
-        }
+        cmd.arg(format!("{}_", library).replace('.', "_").replace('-', "_"));
     }
     cmd.output().unwrap();
 
```

Unfortunately, as is, this also generates different import libraries than what's in the repo because of what's now fixed in https://github.com/microsoft/win32metadata/pull/1247.
So once there's a new metadata release and it's pulled in here, the above patch can be applied.

kennykerr:
I believe this can now be fixed since #2050 has arrived.

PR #2050, opened 2022-09-21
Update the win32 metadata to version 33.0.18

kennykerr:
Update to the latest version of the win32 metadata that includes the lib fixes for the Windows 8 kernel synchronization functions. 

Fixes: #2027

PR #2052, opened 2022-09-21
Remove dlltool prefix hack

glandium:
and refresh the *_gnu{,llvm} libraries after #2050.

What's this all about?

Fixes: #2019


kennykerr (approved):
Thanks!

Issue #2056, opened 2022-09-22
`hstring::Header::duplicate` does not uphold aliasing rules


ChrisDenton:
Rust's aliasing rules mean that an `&mut` must ensure unique access to the memory. Not even `unsafe` code is allowed to subvert that. Therefore the following code is UB because the duplicate pointer is derived from `&mut self`:

https://github.com/microsoft/windows-rs/blob/49317c3a78f7b644f7fd2522ceb6265c698ff89f/crates/libs/windows/src/core/strings/hstring.rs#L411-L416

PR #2057, opened 2022-09-22
Fix `HSTRING` to conform to Rust's aliasing rules

ChrisDenton:
Pointers derived from `&mut` references must still uphold the uniqueness guarantee. Fortunately the fix is simple here:  we just use a shared reference when duplicating `HSTRING`s. Casting between pointer types is fine, it's effectively just a lint.

Fixes: #2056

### DIFF
--- a/crates/libs/windows/src/core/strings/hstring.rs
+++ b/crates/libs/windows/src/core/strings/hstring.rs
@@ -408,11 +408,11 @@ impl Header {
         header
     }
 
-    fn duplicate(&mut self) -> *mut Header {
+    fn duplicate(&self) -> *mut Header {
         if self.flags & REFERENCE_FLAG == 0 {
             // If this is not a "fast pass" string then simply increment the reference count.
             self.count.add_ref();
-            self
+            self as *const Header as *mut Header
         } else {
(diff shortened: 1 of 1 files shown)


kennykerr (approved):
Thanks!

Muon:
This is wrong. You can have as many `*mut T` pointing to the same `T` as you like, you just can't have multiple `&mut T` at once (otherwise [this entirely safe example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fc002006a0f3736a1a2522cef1834366) should not compile). Furthermore, this PR potentially *introduces* UB, since you cannot legally mutate through a `*mut T` that was obtained from an `&T`. Here's a playground example you can run with MIRI to see that it picks up UB: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=11397c0ce7a6705b02c7d78518d90383

ChrisDenton:
The first safe playground example you linked is safe because the pointers are never used. It's always safe to not use pointers.

Indeed, mutating a pointer obtained through an `&T` would not be sound. However, in this case it's being used in a `HSTRING` which is immutable.

PR #2076, opened 2022-09-28
Fix Natvis for HSTRING after latest changes.

ridwanabdillahi:
The underlying data structure for HSTRING changed, this caused the debugger visualizer tests to fail.

This change updates the Natvis to depend on the memory layout of the HSTRING type as opposed to relying on internal fields.

### DIFF
--- a/crates/libs/windows/windows.natvis
+++ b/crates/libs/windows/windows.natvis
@@ -27,18 +27,20 @@
   </Type>
 
   <Type Name="windows::core::strings::hstring::HSTRING">
-    <DisplayString Condition="__0 == nullptr">""</DisplayString>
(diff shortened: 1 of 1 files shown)


ridwanabdillahi:
Moving this PR directly to the master branch.

ridwanabdillahi:
@kennykerr would you like this PR to go directly into the master branch? This change can be made against master as of now and wouldn't break any of the debugger visualizer tests.

kennykerr:
Sounds good thanks!

ridwanabdillahi:
> Sounds good thanks!

Okay, I'll close this PR and make the change directly against master. Thanks!

PR #2077, opened 2022-09-28
Update the Natvis definition for `HSTRING`

ridwanabdillahi:
This change updates the Natvis definition of `HSTRING` to rely on its memory layout as opposed to the internal structure.

This allows for changes to the `HSTRING` data structure without breaking the Natvis.


kennykerr (approved):
Sweet - thanks!

Issue #2049, opened 2022-09-21
Feature request: Provide more `const` support for string types


seritools:
### Motivation

`w!` allows you to create a `&'static HSTRING`, but `.as_ptr()` and other conversions don't seem to be `const`, so it doesn't seem possible to get a const `PCWSTR` from it, in const context. Currently I'm using `wchar::wchz!` as alternative: `PCWSTR::from_raw(wchz!("my string").as_ptr())`. It would be nice to have a built-in way to do this.

Another tiny one is to create a const null-`PCWSTR`, but `PCWSTR::null` is not marked as const. `PCWSTR(std::ptr::null())` is the slightly uglier workaround :)
The same should be the case for the other string types (`PWSTR`, and the 8bit string types). `null()`, `as_ptr()` and `is_null()` all can be marked as `const fn` as far as I can tell.

kennykerr:
I'd love to get there. I think we're still waiting for things to stabilize. For example, if I try to make `HSTRING::is_empty` a `const` function I get the following error:

```
D:\git\windows-rs>cargo check -p windows
    Checking windows v0.40.0 (D:\git\windows-rs\crates\libs\windows)
    Finished dev [unoptimized + debuginfo] target(s) in 0.55s

D:\git\windows-rs>cargo check -p windows
    Checking windows v0.40.0 (D:\git\windows-rs\crates\libs\windows)
error: `ptr::mut_ptr::<impl *mut T>::is_null` is not yet stable as a const fn
  --> crates\libs\windows\src\core\strings\hstring.rs:19:9
   |
19 |         self.0.is_null()
   |         ^^^^^^^^^^^^^^^^
   |
   = help: add `#![feature(const_ptr_is_null)]` to the crate attributes to enable

error: could not compile `windows` due to previous error
```

I'm not sure when the `const_ptr_is_null` feature will be stabilized. It doesn't seem to have made much progress of late:

https://github.com/rust-lang/rust/issues/74939

ChrisDenton:
Hm, if you don't mind a bit of type verbosity you maybe could use `Option<NonNull<Header>>` as the inner type. Then it's just `None` for the null case.

seritools:
Testing a bit on `HSTRING` (https://github.com/seritools/windows-rs/tree/hstring-tests) it seems that it's possible.

There's some ugliness around NonNull::as_ptr() returning a `*mut`, which rustc currently doesn't allow to deref in const context, triggering:

```rs
[...]

PR #2078, opened 2022-09-29
Make more of the `HSTRING` methods `const`

kennykerr:
Based on a suggestion by @ChrisDenton, I've made more of the `HSTRING` methods `const`. In theory I could apply the same technique to `BSTR` if desired. I'm reluctant to mess with `PCWSTR` and friends as they're used in more unorthodox ways. 

Fixes: #2049

### DIFF
--- a/crates/libs/windows/src/core/strings/hstring.rs
+++ b/crates/libs/windows/src/core/strings/hstring.rs
@@ -3,44 +3,43 @@ use super::*;
 /// A WinRT string ([HSTRING](https://docs.microsoft.com/en-us/windows/win32/winrt/hstring))
 /// is reference-counted and immutable.
 #[repr(transparent)]
-pub struct HSTRING(*mut Header);
+pub struct HSTRING(Option<std::ptr::NonNull<Header>>);
 
 impl HSTRING {
     /// Create an empty `HSTRING`.
     ///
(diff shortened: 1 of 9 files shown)


kennykerr:
Looks like I'll have to update the MSRV to 1.64 for the `windows` crate:

https://github.com/microsoft/windows-rs/actions/runs/3148068177/jobs/5118211480

https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html

crates/libs/windows/src/core/strings/hstring.rs
```diff
@@ -119,11 +127,13 @@ impl Drop for HSTRING {
         }
 
         unsafe {
-            let header = std::mem::replace(&mut self.0, std::ptr::null_mut());
-            // REFERENCE_FLAG indicates a string backed by static or stack memory that is
-            // thus not reference-counted and does not need to be freed.
-            if (*header).flags & REFERENCE_FLAG == 0 && (*header).count.release() == 0 {
-                heap_free(header as *mut std::ffi::c_void);
+            if let Some(header) = self.0.take() {
```
rylev, line 130: Just a nit that the `unsafe` block can be moved inside of the `if let` block.

kennykerr:
done - thanks!

PR #2079, opened 2022-09-30
Replace boilerplate COM interface code gen with `macro_rules`

kennykerr:
The `windows` and `windows-sys` crates avoid `derive` macros as much as possible as they tend to be very slow at scale. There is however a lot of boilerplate code that could be avoided with selective use of macros, so I am going to start making conservative use of `macro_rules` to replace large chunks of boilerplate code. I tested the change in this update and could not measure any difference in build time, so I think this is a good first step in reducing code gen and overall crate size. 

The new `interface_hierarchy` macro stamps out most of the `From` implementations for COM and WinRT types that can often have large interface hierarchies. For example, `ID3D12Device9` derives from 11 other interfaces leading to 33 `From` implementations that can be generated by the `interface_hierarchy` macro far more concisely. 

https://github.com/microsoft/windows-rs/blob/6ed9da4b44d06fa451a14bfc0cf8c6e7b7524ce8/crates/libs/windows/src/core/mod.rs#L101-L123

### DIFF
--- a/crates/libs/windows/src/core/mod.rs
+++ b/crates/libs/windows/src/core/mod.rs
@@ -95,3 +95,32 @@ fn wide_trim_end(mut wide: &[u16]) -> &[u16] {
     }
     wide
 }
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! interface_hierarchy {
+    ($child:ty, $parent:ty) => {
+        impl ::core::convert::From<$child> for $parent {
+            fn from(value: $child) -> Self {
+                unsafe { ::core::mem::transmute(value) }
+            }
+        }
+        impl ::core::convert::From<&$child> for &$parent {
+            fn from(value: &$child) -> Self {
+                unsafe { ::core::mem::transmute(value) }
+            }
+        }
+        impl ::core::convert::From<&$child> for $parent {
+            fn from(value: &$child) -> Self {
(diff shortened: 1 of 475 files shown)


crates/libs/windows/src/core/mod.rs
```diff
@@ -95,3 +95,32 @@ fn wide_trim_end(mut wide: &[u16]) -> &[u16] {
...
+        }
+    };
+    ($child:ty, $first:ty, $($rest:ty),+) => {
+        $crate::core::interface_hierarchy!($child, $first);
+        $crate::core::interface_hierarchy!($child, $($rest),+);
+    };
+}
+
+#[doc(hidden)]
+pub use interface_hierarchy;
```
rylev, line 126: I don't believe this is necessary given the `#[macro_export]`.

crates/libs/windows/src/core/mod.rs
```diff
@@ -95,3 +95,32 @@ fn wide_trim_end(mut wide: &[u16]) -> &[u16] {
     }
     wide
 }
+
+#[doc(hidden)]
```
rylev, line 99: Despite this being not shown to end users, it could be good to have some documentation here for those working on the internals of windows-rs.

rylev (commented):
Looks great! ❤️

kennykerr:
Will do - thanks @rylev!

Issue #2082, opened 2022-10-02
`PCWSTR`'s `From<&HSTRING>` implementation may lead to dangerous code


num3ric:
### Which crate is this about?

windows

### Crate version

0.42.0

### Summary

Failure on (debug) launch in visual studio code with the automatically generated configuration:

```
         {
            "type": "lldb",
            "request": "launch",
            "name": "Debug executable 'sample_direct3d12'",
            "cargo": {
                "args": [
                    "build",
                    "--bin=sample_direct3d12",
                    "--package=sample_direct3d12"
                ],
                "filter": {
                    "name": "sample_direct3d12",
                    "kind": "bin"
                }
            },
            "args": [],
            "cwd": "${workspaceFolder}"
        },
```

Resulting error:
`Error: Error { code: HRESULT(0x80004005), message: "Unspecified error" }`

In stepping through the code, it looks like the failure stems from the first `D3DCompileFromFile` call in `create_pipeline_state`.

### Toolchain version/configuration

Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\num3ric\.rustup

stable-x86_64-pc-windows-msvc (default)
rustc 1.64.0 (a55dd71d5 2022-09-19)

### Expected behavior

Successful launch.

### Actual behavior

`Error: Error { code: HRESULT(0x80004005), message: "Unspecified error" }`

riverar:
```
#[test]
fn test()
{
    let s = "abcd";
    let pcwstr = PCWSTR::from(&s.into());
    unsafe { assert_eq!(pcwstr.as_wide(), [97, 98, 99, 100, 0]); }
}
```

@kennykerr  Not clear if the sample should be null terminating or if `&str -> &HSTRING` should be producing a null-terminated string.

riverar:
There also appears to be an issue with a temporary value getting dropped here as we end up in `d3dcompiler_47!D3DCompileFromFile` with `rcx` pointing to nothing useful.

https://github.com/microsoft/windows-rs/blob/6d45cc6924209746a2a12b2df9bd760eef016265/crates/samples/direct3d12/src/main.rs#L429-L432

kennykerr:
The problem is here:

https://github.com/microsoft/windows-rs/blob/3054d4065dbe11345608fdf2787fe614f96ec2f9/crates/libs/windows/src/core/strings/pcwstr.rs#L62-L66

The `PCWSTR` is not limited to the lifetime of the `HSTRING`.

[...]

PR #2084, opened 2022-10-03
Fix D3D12 sample

kennykerr:
This fixes the D3D sample mentioned here: #2082

The issue was detected using App Verifier and stems from an unsafe `From` implementation, which I will fix separately as it is rather involved.

### DIFF
--- a/crates/samples/direct3d12/src/main.rs
+++ b/crates/samples/direct3d12/src/main.rs
@@ -426,17 +426,17 @@ mod d3d12_hello_triangle {
         let asset_path = exe_path.parent().unwrap();
         let shaders_hlsl_path = asset_path.join("shaders.hlsl");
         let shaders_hlsl = shaders_hlsl_path.to_str().unwrap();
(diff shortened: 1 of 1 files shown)


crates/samples/direct3d12/src/main.rs
```diff
@@ -426,17 +426,17 @@ mod d3d12_hello_triangle {
         let asset_path = exe_path.parent().unwrap();
         let shaders_hlsl_path = asset_path.join("shaders.hlsl");
         let shaders_hlsl = shaders_hlsl_path.to_str().unwrap();
-        let shaders_hlsl = PCWSTR::from(&shaders_hlsl.into());
+        let shaders_hlsl: HSTRING = shaders_hlsl.into();
```
kennykerr, line 429: A temporary `HSTRING` was being created and then used in the subsequent calls to `D3DCompileFromFile`.

riverar (approved):
Thanks!

PR #2090, opened 2022-10-09
Fix a few new `clippy` warnings

kaivol:
removed some unnecessary casts (`clippy::unnecessary-cast`) and ignored another warning (`clippy::manual_clamp`)

### DIFF
--- a/crates/samples/direct2d/src/main.rs
+++ b/crates/samples/direct2d/src/main.rs
@@ -179,6 +179,7 @@ impl Window {
 
         let size = unsafe { target.GetSize() };
 
+        #[allow(clippy::manual_clamp)]
         let radius = size.width.min(size.height).max(200.0) / 2.0 - 50.0;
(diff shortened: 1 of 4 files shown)


ChrisDenton:
Is there a reason not to use clamp here?

riverar:
Sounds reasonable. Can we also change the title of the PR to reflect what was changed?

kennykerr:
> Is there a reason not to use clamp here?

The trouble is that it will panic if min > max.

crates/samples/direct2d/src/main.rs
```diff
@@ -179,6 +179,7 @@ impl Window {
 
         let size = unsafe { target.GetSize() };
 
+        #[allow(clippy::manual_clamp)]
```
kennykerr, line 182: Seems like a `clippy` bug that this warning is not appropriate and must be suppressed.

PR #2110, opened 2022-10-21
Fix example

FlusherDock1:
Fixes error inside readme example: 

```
error: cannot find macro `h` in this scope
 --> src\main.rs:8:17
  |
8 |     doc.LoadXml(h!("<html>hello world</html>"))?;
  |                 ^ help: a macro with a similar name exists: `s`
  |
 ::: windows-0.42.0\src\core\strings\literals.rs:3:1
  |
3 | macro_rules! s {
  | -------------- similarly named macro `s` defined here
```

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -33,7 +33,7 @@ use windows::{
 
 fn main() -> Result<()> {
     let doc = XmlDocument::new()?;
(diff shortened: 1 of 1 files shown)


kennykerr:
That example isn't broken. 😏 It represents the *next* version of the `windows` crate...

PR #2114, opened 2022-10-23
fix DEVMODEA/W struct error.

aosemp:
Fixes: DEVMODEA/W  from VS SDK " C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\um\wingdi.h"

example code:
```
  let mut dev_mode = DEVMODEA::default();
    dev_mode.dmSize = mem::size_of::<DEVMODEA>() as u16;

    unsafe {
        if !EnumDisplaySettingsA(None, ENUM_CURRENT_SETTINGS, &mut dev_mode).as_bool() {
            return Err(Error::from_win32());
        }
    }

    println!(
        "now {} {}x{} change to {}x{}",
        dev_mode.dmSize, dev_mode.dmPelsWidth, dev_mode.dmPelsHeight, width, height
    );
```

### DIFF
--- a/crates/libs/windows/src/Windows/Win32/Graphics/Gdi/mod.rs
+++ b/crates/libs/windows/src/Windows/Win32/Graphics/Gdi/mod.rs
@@ -7082,11 +7082,11 @@ impl ::core::ops::Not for DC_LAYOUT {
 #[doc = "*Required features: `\"Win32_Graphics_Gdi\"`*"]
 #[repr(transparent)]
 #[derive(::core::cmp::PartialEq, ::core::cmp::Eq)]
-pub struct DEVMODE_COLLATE(pub u32);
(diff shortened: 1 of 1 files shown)


aosemp:
@microsoft-github-policy-service agree [company="Guangdong Yizhisec Technology Co., Ltd"]

aosemp:
@microsoft-github-policy-service agree company="Guangdong Yizhisec Technology Co., Ltd"

tim-weis:
This PR attempts to modify *generated* code. This won't persist into the next release.

If there is an issue with the code generator, please file an issue here. If the issue is rather due to incorrect metadata, please file an issue [here](https://github.com/microsoft/win32metadata/issues). Either way, analyzing the root cause should be the first step in solving this issue.

PR #2120, opened 2022-10-25
Add performance counter samples

kennykerr:
Just adds a basic Performance Counter API sample for both `windows` and `windows-sys`.

### DIFF
--- a/crates/samples/counter/src/main.rs
+++ b/crates/samples/counter/src/main.rs
@@ -1,0 +1,21 @@
+use windows::{core::*, Win32::System::Performance::*};
+
+fn main() {
+    unsafe {
+        let mut query = 0;
+        PdhOpenQueryW(None, 0, &mut query);
+
+        let mut counter = 0;
+        PdhAddCounterW(query, w!("\\Processor(0)\\% Processor Time"), 0, &mut counter);
+
+        loop {
+            std::thread::sleep(std::time::Duration::new(1, 0));
+            PdhCollectQueryData(query);
+
+            let mut value = Default::default();
+            if 0 == PdhGetFormattedCounterValue(counter, PDH_FMT_DOUBLE, None, &mut value) {
+                println!("{:.2}", value.Anonymous.doubleValue);
+            }
+        }
+    }
+}
--- a/crates/samples/counter_sys/src/main.rs
+++ b/crates/samples/counter_sys/src/main.rs
@@ -1,0 +1,21 @@
+use windows_sys::{core::*, Win32::System::Performance::*};
+
+fn main() {
+    unsafe {
+        let mut query = 0;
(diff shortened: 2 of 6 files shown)


crates/samples/counter/src/main.rs
```diff
@@ -0,0 +1,21 @@
+use windows::{core::*, Win32::System::Performance::*};
+
+fn main() {
+    unsafe {
+        let mut query = 0;
+        PdhOpenQueryW(None, 0, &mut query);
```
riverar, line 6:
```suggestion
        let handle = PdhOpenQueryW(None, 0, &mut query);
```

crates/samples/counter_sys/src/main.rs
```diff
@@ -0,0 +1,21 @@
+use windows_sys::{core::*, Win32::System::Performance::*};
+
+fn main() {
+    unsafe {
+        let mut query = 0;
+        PdhOpenQueryW(std::ptr::null(), 0, &mut query);
```
riverar, line 6:
```suggestion
        let handle = PdhOpenQueryW(std::ptr::null(), 0, &mut query);
```

crates/samples/counter_sys/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+        loop {
+            std::thread::sleep(std::time::Duration::new(1, 0));
+            PdhCollectQueryData(query);
+
+            let mut value = std::mem::zeroed();
+            if 0 == PdhGetFormattedCounterValue(counter, PDH_FMT_DOUBLE, std::ptr::null_mut(), &mut value) {
+                println!("{:.2}", value.Anonymous.doubleValue);
+            }
+        }
+    }
```
riverar, line 20:
```suggestion
        }
        
        PdhCloseQuery(handle);
    }
```

crates/samples/counter/src/main.rs
```diff
@@ -0,0 +1,21 @@
...
+        loop {
+            std::thread::sleep(std::time::Duration::new(1, 0));
+            PdhCollectQueryData(query);
+
+            let mut value = Default::default();
+            if 0 == PdhGetFormattedCounterValue(counter, PDH_FMT_DOUBLE, None, &mut value) {
+                println!("{:.2}", value.Anonymous.doubleValue);
+            }
+        }
+    }
```
riverar, line 20:
```suggestion
        }
        
        PdhCloseQuery(handle);
    }
```

kennykerr:
@riverar the win32 metadata really should provide native type defs for these.

riverar:
@kennykerr Agreed, was thinking that too. Feel free to ignore the suggestions (as they'll be rendered moot when we fix that up)!

PR #2124, opened 2022-10-27
Use more readable formatting for samples

kennykerr:
This repo uses a very long line width to improve crate compression, but it does make the code a little harder to read. This update just tweaks the line width for samples to improve readability.

### DIFF
--- a/crates/samples/rustfmt.toml
+++ b/crates/samples/rustfmt.toml
@@ -1,0 +1,1 @@
+newline_style = "Unix"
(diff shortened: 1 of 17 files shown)


crates/samples/rustfmt.toml
```diff
@@ -0,0 +1 @@
+newline_style = "Unix"
```
riverar, line 1: This may clash with some folks that have Git configured to checkout in non-LF. There doesn't appear to be any real value in explicitly setting this.

kennykerr: All of the generated code uses "Unix" newline formatting, so if this is not set it marks a ton of code as "changed" for the purpose of git differentiation.

riverar: Ah we have `*       text eol=lf` in gitattributes, so this is OK. (Ideally we'd delete both of these but not going to fight that battle today.)

kennykerr: I gave up trying to fight git on line endings. Life's too short. 😜

PR #2129, opened 2022-10-31
Add DirectComposition sample

kennykerr:
This sample illustrates various concepts related to UI development for Windows including:

* DirectComposition
* Direct2D
* DirectWrite
* Windows Animation Manager
* Windows Imaging Component

This sample first appeared in my [DirectComposition course on Pluralsight](https://www.pluralsight.com/courses/directcomposition-in-action). 

dcomp


crates/samples/dcomp/src/main.rs
```diff
@@ -0,0 +1,781 @@
...
+        Ok(image)
+    }
+}
+
+fn create_device_3d() -> Result<ID3D11Device> {
+    let mut flags = D3D11_CREATE_DEVICE_BGRA_SUPPORT;
+
+    if cfg!(debug_assertions) {
+        flags |= D3D11_CREATE_DEVICE_DEBUG;
+    }
```
riverar, line 552: I'm not sure where you get the debug layers these days; this doesn't work on my retail Windows 11 machines.

```
Error { code: HRESULT(0x887A002D), message: "The application requested an operation that depends on an SDK component that is missing or mismatched."
```

crates/samples/dcomp/src/main.rs
```diff
@@ -0,0 +1,781 @@
...
+        D3D11CreateDevice(
+            None,
+            D3D_DRIVER_TYPE_HARDWARE,
+            None,
+            flags,
+            None,
+            D3D11_SDK_VERSION,
+            Some(&mut device),
+            None,
+            None,
```
riverar, line 566: Optional: Consider using `D3D_DRIVER_TYPE_WARP` instead for better perf over RDP.

riverar (changes requested):
`image.jpg` appears to be corrupted, leading to panic.

```
code: HRESULT(0x88982F60), message: "The image is unrecognized."
```

crates/samples/dcomp/src/main.rs
riverar: Update: We can add this to the `README.md`:

To run this sample, you will need the Direct3D 11 debug layer installed. To install the layer, go to Settings > Apps > Optional features and add the optional feature `Graphics Tools`.

riverar (approved):
Tested on aarch64, works well.

crates/samples/dcomp/src/main.rs
```diff
@@ -0,0 +1,769 @@
...
+
+            AdjustWindowRect(
+                &mut rect,
+                WINDOW_STYLE(GetWindowLongW(self.handle, GWL_STYLE) as _),
+                false,
+            )
+            .ok()?;
+
+            Ok((rect.right - rect.left, rect.bottom - rect.top))
+        }
```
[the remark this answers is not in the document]
kennykerr: In that case yes, but in general most of the functions make so many unsafe calls it becomes very noisy. Not sure how to think about that.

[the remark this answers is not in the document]
kennykerr: Sure, feel free to contribute something like that. I find it highly subjective and would really like to reduce the underlying *need* for `unsafe` in the first place. Many of these calls are perfectly safe. I just need to think some more about how to define that.

Godnoken:
Thank you for this! I read your article from 2014 about DirectComposition and transparency and tried to implement it myself in Rust before finding out you already had done it yourself.

I just have one question. Is there a simple way in this example to make the transparent areas click-through to the windows behind? Like using `HTTRANSPARENT` on `NCHITTEST`, but instead propagating down to windows that are not on the same thread.

If you have any quick solution or perhaps a link to information for me to dig through, that would be very much appreciated, thank you.

kennykerr:
2014 was a long time ago. 😜 I don't recall exactly, but I thought `WS_EX_NOREDIRECTIONBITMAP` along with a 0% alpha channel was all it took. @robmikh may know for sure.

Godnoken:
Haha, indeed! Hmm. But is it possible to set a 0% alpha channel without `WS_EX_LAYERED`, and if not, wouldn't that put me back to square one performance-wise?

kennykerr:
Dug into this a bit. Unfortunately, if you need per-pixel hit testing then you need a layered window e.g. `WS_EX_LAYERED` with the implied performance penalty. Nothing has changed from what I described in that 2014 article.

Godnoken:
Ah, that's too bad, but I think I'll be able to come up with a solution that fits my purposes anyways.
Thank you very much!

PR #2131, opened 2022-11-02
Don't require files for reading/writing

Jake-Shadle:
The reader/writer implementations in `windows-metadata` only supported reading/writing directly to files, which makes them cumbersome to use if you want to do anything with buffers before/after reading or writing (eg compression), this just adds a `windows_metdata::reader::Reader::from_buffer` and `windows_metadata::writer::file::write_to_buffer` method/function to not require a file on disk.


Jake-Shadle:
@microsoft-github-policy-service agree [company="Embark Studios AB"]

Jake-Shadle:
@microsoft-github-policy-service agree company="Embark Studios AB"

kennykerr (approved):

PR #2132, opened 2022-11-02
Validate baseline for lib files

kennykerr:
As the target libs are binary files and cannot be easily compared, this update stores a text baseline so that we can easily tell if any lib changes have snuck in unawares. The CI build should also validate that no such changes have occurred and fail the build if that's the case. 

This just validates the function signatures and associated library names. Ideally, the build can also rebuild all the libs and validate that the lib binaries are identical but that will require more yml trickery.


kennykerr:
Hmm for some reason, my intentional break of the baseline (https://github.com/microsoft/windows-rs/pull/2132/commits/479946ea287526b72abd6a6376e1d15aab306edb) isn't breaking the build.

kennykerr:
Yay, build now fails if target lib baseline has changed.

Issue #2126, opened 2022-10-31
Feature request: Apply standard Rust naming conventions for types that implement ownership semantics


tim-weis:
### Motivation

The motivation behind this feature request is to make client code more comprehensible. It's currently difficult to tell ABI(-like) *"all-management-on-the-client"* types and Rust's *"ownership-is-a-solved-problem"* types apart.

Since the [question](https://github.com/microsoft/windows-rs/issues/2123) recently came up as to what the ownership semantics of the various types of the windows-rs crate are, I'm proposing to encode this information by following standard Rust [naming](https://rust-lang.github.io/api-guidelines/naming.html) conventions.

The only types affected by this proposal (to my knowledge) are [`BSTR`](https://github.com/microsoft/windows-rs/blob/master/crates/libs/windows/src/core/strings/bstr.rs) and [`HSTRING`](https://github.com/microsoft/windows-rs/blob/master/crates/libs/windows/src/core/strings/hstring.rs) (to be renamed to `Bstr` (`BStr`?) and `HString`, respectively). COM and WinRT interfaces are already following this convention, signaling to clients that ownership is handled. `P[C][W]STR` may need to be considered, still.

### Drawbacks

This is a breaking change. Like any breaking change, it's going to cause noise, downstream.

I believe the perturbations to be manageable. From personal experience, both *rustc* and *clippy* do a fabulous job at driving authors to make the right adjustments.

### Rationale and alternatives

As an alternative to renaming the types, type aliases could be provided for easier migration. I'm not sure this is strictly necessary pre-v1.0, and type aliases tend to increase complexity.

### Additional context

This consideration is certainly going to come up again when [support for `VARIANT`s](https://github.com/microsoft/windows-rs/issues/539) (or [`SAFEARRAY`s](https://github.com/microsoft/windows-rs/issues/2123#issuecomment-1293733992), maybe) is approached. Having distinct casing for the resource management wrappers and raw ABI types makes it easier to tell them apart.

kennykerr:
I think this has potential but is not very clearcut. Simple examples like `BSTR` and `HSTRING` have potential but there are many Win32 structs that contain types with a `Drop` implementation and I can't imagine renaming all of these. This is also further complicated by #1896 and where that ends up landing - I'm not yet sure.

tim-weis:
> not very clearcut

Hm, true, I hadn't considered generated types...

[...]

Issue #1896, opened 2022-07-12
`Borrowed` needs to removed/combined with `InParam`


kennykerr:
Following #1811, Borrowed needs to be applied to Win32 struct fields.

kennykerr:
Borrowed really doesn't work for struct fields at all, as was hoped. Win32 fields should probably all just be `ManuallyDrop` as their ownership is not defined. I'll probably also just fold `Borrowed` and `InParam` together as its only useful for input parameters but the current implementation is unnecessarily complicated as it considers, but does not support, field usage.

kennykerr:
Fixed in #2237.

Issue #907, opened 2021-06-23
Structs that hold pointers to arrays of structs should have lifetime parameters


damyanp:
For example:

```c++
typedef struct D3D12_ROOT_DESCRIPTOR_TABLE
    {
    UINT NumDescriptorRanges;
    _Field_size_full_(NumDescriptorRanges)  const D3D12_DESCRIPTOR_RANGE *pDescriptorRanges;
    } 	D3D12_ROOT_DESCRIPTOR_TABLE;
```

pDescriptorRanges points to the start of an array of NumDescriptorRanges D3D12_DESCRIPTOR_RANGE instances.  The generated rust type should have an appropriate lifetime parameter indicate that this array must be valid for as long as the D3D12_ROOT_DESCRIPTOR table instance is valid.

This pattern exists across many D3D12 structs, and possibly in other Win32 ones as well.

damyanp:
Currently, win32metadata does not include anything associating these two fields together, so it is likely some work will be required there. I think we need to agree that including these lifetime parameters is the right thing to do, figure out what metadata is required and get the appropriate issue filed.  Once that's done we'll be able to do the windows-rs work to resolve this issue.

kennykerr:
Sweet, thanks for opening the issue. I have other issues open for supporting arrays in general, which I currently only support minimally, but this will be useful to track this specific pattern. 

We'll need another issue on the win32metadata repo for adding metadata to account for the `_Field_size_full_(NumDescriptorRanges)` SAL annotation. The metadata already parses SAL annotations so that shouldn't be a problem - we just need to get them to add this one somehow.

kennykerr:
Other array-related issues: #212 #443 #479 #790

I hope to address all of these together when I get a moment.

kennykerr:
Doesn't look like there's been any progress metadata issue. https://github.com/microsoft/win32metadata/issues/547

I'll close this issue for now but keep an eye on that metadata issue. If that is resolved, we can revisit this.

stormshield-gt:
Hi, sorry for commenting on old issue, the meta issue has been closed https://github.com/microsoft/win32metadata/issues/547 so maybe we can revisit this one ?

PR #2136, opened 2022-11-04
Remove unsafe conversion for PCWSTR

rylev:
As @michaelwoerister pointed out this conversion, while safe, makes it *very* easy to write incorrect code. I think we should avoid such code.

Here's an example of some code that uses safe Rust but creates a dangling pointer:

```rust
fn create_dangling_pcwstr_pointer() -> PCWSTR {
    let s = HSTRING::from("abc");
    PCWSTR::from(&s)
}
```

### DIFF
--- a/crates/libs/windows/src/core/strings/pcwstr.rs
+++ b/crates/libs/windows/src/core/strings/pcwstr.rs
@@ -59,12 +59,6 @@ unsafe impl Abi for PCWSTR {
     type Abi = Self;
 }
 
-impl From<&HSTRING> for PCWSTR {
-    fn from(hstring: &HSTRING) -> Self {
-        Self(hstring.as_ptr())
-    }
-}
-
 // This just ensures that `None` can be used for optional PCWSTR parameters, which can be quite common
 // with some Windows APIs.
 impl From<Option<PCWSTR>> for PCWSTR {


michaelwoerister:
Yes, this does indeed look like a potential source of errors to me.

Overall, I don't know of a good way of dealing with `PCWSTR` and similar string types that are essentially raw pointers with no ownership or lifetime guarantees attached.

kennykerr:
This has come up before (#2084). In this case, we caught it with [App Verifier](https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/application-verifier) which does a great job of detecting such errors. 

I'm just not sure simply removing the impl is sufficient since we still need a way to perform such a conversion.

kennykerr:
This is also related to #907 and #1896 where we really need to track lifetime even in cases where ownership is ambiguous.

kennykerr:
In this case, perhaps its preferable to be explicit and require something like this (from the build break)?

```rust
        let vertex_shader = unsafe {
            D3DCompileFromFile(
                PCWSTR(shaders_hlsl.as_ptr()),
                None,
                None,
```

michaelwoerister:
> In this case, perhaps its preferable to be explicit and require something like this (from the build break)?
> 
> ```rust
>         let vertex_shader = unsafe {
>             D3DCompileFromFile(
>                 PCWSTR(shaders_hlsl.as_ptr()),
>                 None,
>                 None,
> ```

That does look better to me, yes.

An explicit `to_pcwstr()` method might also make sense. It could be safe, like e.g. [Vec::as_ptr()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_ptr) is safe, but would serve as a place where documentation about safety/lifetime can be added -- but I hesitate to suggest adding ad-hoc convenience methods like this without an overall design strategy for strings and/or win32 raw-pointers.

kennykerr:
> but I hesitate to suggest adding ad-hoc convenience methods like this without an overall design strategy for strings and/or win32 raw-pointers

Right, once the overall ownership story has been sorted out, see also https://github.com/microsoft/windows-rs/issues/2126#issuecomment-1304686765, this should be a lot clearer. Until then, I'm happy to require `PCWSTR(from.as_ptr())`.

PR #2161, opened 2022-11-14
Experimental support for delay-loading API calls 

kennykerr:
Using the `windows_delay_load` flag, you can enable support for delay-loading API calls. This can be handy if you know that some of the APIs may not be available on all of your supported target operating systems and you would like to delay loading imports, preventing the operating system loader from resolving them all at load time. This has the same effect as the [MSVC delayload linker](https://learn.microsoft.com/en-us/cpp/build/reference/linker-support-for-delay-loaded-dlls?view=msvc-170) option but implemented entirely in Rust. 

The main drawback is that there's no way to limit delay loading to certain libraries and since it is enabled via `rustflags` it will presumably impact all dependent crates being compiled. Suggestions welcome. 😊


kennykerr:
Looks like gnu nightly on ubuntu is busted outside of my control.

crates/libs/sys/src/core/mod.rs
```diff
@@ -46,5 +49,26 @@ macro_rules! link {
...
+#[cfg(windows_delay_load)]
+#[macro_export]
+macro_rules! link {
+    ($library:literal $abi:literal $(#[$($doc:tt)*])* fn $name:ident($($arg:ident: $argty:ty),*)->$ret:ty) => (
+        #[cfg(target_arch = "x86")]
+        type $name = ::core::option::Option<unsafe extern $abi fn($($argty),*) -> $ret>;
+        #[cfg(not(target_arch = "x86"))]
+        type $name = ::core::option::Option<unsafe extern "system" fn($($argty),*) -> $ret>;
+        pub unsafe fn $name($($arg: $argty),*) -> $ret {
+            static mut CACHE: $name = None;
```
wesleywiser, line 61: This is almost a nitpick but `static mut` is very difficult to use correctly (to the point where there is a strong possibility it will be deprecated and removed in a future edition). I think it would be better to use a `static AtomicPtr<{fn ptr type}>` instead. Acquire/Release are nearly free on x86 IIRC so there shouldn't be too much of a performance difference.

PR #2164, opened 2022-11-15
Add optional support for `raw-dylib` 

kennykerr:
Following on from #2149, this update adds optional support for `raw-dylib` in the `windows` and `windows-sys` crates. This is currently only supported on nightly but is on track for stabilization. The `link` macro is "overloaded" to adjust the way the Rust `link` attribute is configured depending on whether the `windows_raw_dylib` flag is specified. 

The `link` macro will generate the traditional form if the `windows_raw_dylib` flag is not specified. In that mode, it continues to rely on the target libs provided by the `windows-targets` crate. However, the `windows_raw_dylib` flag ensures that no libs are needed or downloaded. The name of the DLL is then used to instruct the Rust compiler to inject the appropriate import table entry directly and without the use of an import lib. 

I'll probably switch the `windows` crate over to `raw-dylib` unconditionally once the feature is stabilized but keep the `windows-sys` crate optional to retain support for an older MSRV.


kennykerr:
In addition to waiting on stabilization, testing uncovered an issue blocking support on i686-pc-windows-gnu: 

https://github.com/rust-lang/rust/issues/104453

kennykerr:
cc @udoprog @dpaoliello @bjorn3 @wesleywiser

kennykerr:
You're welcome! This certainly makes it easier to kick the tires with `raw-dylib`. 😊

crates/libs/sys/src/core/mod.rs
```diff
@@ -35,6 +35,29 @@ impl GUID {
...
+        }
+    )
+}
+
+#[cfg(all(windows_raw_dylib, not(target_arch = "x86")))]
+#[macro_export]
+macro_rules! link {
+    ($library:literal $abi:literal $(#[$($doc:tt)*])* fn $name:ident($($arg:ident: $argty:ty),*)->$ret:ty) => (
+        #[link(name = $library, kind = "raw-dylib", modifiers = "+verbatim")]
+        extern "system" {
```
glandium, line 54: Is there a specific reason this was made to ignore $abi? (it's now "C" rather than "system", as a workaround for variadic functions).

kennykerr: Yes, it's a workaround for this: https://github.com/rust-lang/rust/issues/110505

glandium: Yes, that's why it's C now, but why was it "system" rather than $abi?

kennykerr: Because the ABI is generally only relevant for x86 and "system" should work for all ARM64 and x64 Windows APIs.

PR #2165, opened 2022-11-15
Remove deprecated WinRT namespaces

kennykerr:
The WinRT metadata includes a `deprecated` attribute that signals that the type or method has been replaced, but some APIs are deprecated in their entirety, leaving whole modules deprecated and effectively unusable. As these are very obscure APIs and not widely supported, I've just gone ahead and added these to the exclusion list for the `windows` crate to ditch the extra weight. APIs with individual deprecated APIs will continue to be available behind the "deprecated" feature. 

The Win32 metadata does not have a similar deprecated attribute and the `windows-sys` crate doesn't include WinRT APIs, so this change does not affect the `windows-sys` crate.


riverar (approved):
Wow, some of these "deprecated" APIs flat out don't work anymore. Good catch!

PR #2183, opened 2022-11-25
Downgrade target crates versions to 0.42.1

glandium:
New imports were added, but that didn't necessitate the semver incompatible version bump. Fortunately, version 0.43.0 of those crates or crates that depend on them have never been published, so we can pretend the bump never happened.

What's this all about?

Fixes: #2384 ⬅️ Be sure to refer to an existing issue here!

### DIFF
--- a/crates/libs/sys/Cargo.toml
+++ b/crates/libs/sys/Cargo.toml
@@ -16,7 +16,7 @@ targets = []
 all-features = true
 
 [target.'cfg(not(windows_raw_dylib))'.dependencies]
-windows-targets = { path = "../targets",  version = "0.43.0" }
+windows-targets = { path = "../targets",  version = "0.42.1" }
 
 [features]
 default = []
--- a/crates/libs/targets/Cargo.toml
+++ b/crates/libs/targets/Cargo.toml
@@ -1,7 +1,7 @@
 
 [package]
 name = "windows-targets"
-version = "0.43.0"
+version = "0.42.1"
 authors = ["Microsoft"]
 edition = "2018"
 license = "MIT OR Apache-2.0"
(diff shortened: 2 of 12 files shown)


riverar:
Not sure why we keep throwing around the word "incompatible". Any change we make in 0.x is semver compliant; the crate is in pre-release.

ChrisDenton:
It's a strong convention in the Rust community. See [SemVer Compatibility](https://doc.rust-lang.org/cargo/reference/semver.html#semver-compatibility)

> Initial development releases starting with "0.y.z" can treat changes in "y" as a major release, and "z" as a minor release. "0.0.z" releases are always major changes. This is because Cargo uses the convention that only changes in the left-most non-zero component are considered incompatible.

riverar:
I see thanks! I haven't heard of this until now. So really this is more about being _Cargo-flavored semver_ incompatible. Not sure we follow that but if the ecosystem is rallying around this, it's probably looking into.

glandium:
When a crate (like parking-lot and many others) has the following dependency in its Cargo.toml (omitting features):
```
[dependencies]
windows-sys = "0.42.0"
```
what cargo reads is "the crate can use any 0.42.* version".
When windows-sys goes from 0.42.0 to 0.43.0, it means those crates need updated dependencies to pick the new version, while the new version has no change that would break any of the crates that depend on 0.42.0.

glandium:
> if the ecosystem is rallying around this

that's how the cargo ecosystem has worked for something like 7 years, FWIW.

riverar:
I don't think it makes sense for crate owners to have to care about this until 1.x, but that's just my opinion not the crate's. 🙃

crates/libs/sys/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 
 [package]
 name = "windows-sys"
-version = "0.43.0"
+version = "0.42.1"
```
kennykerr, line 4: #2166 is almost certainly a breaking change, so while the targets are compatible this won't be.

kennykerr:
Please limit the change to the crates in the `targets` folder. That should suffice for this PR. Thanks!

kennykerr (approved):
Thanks!

Issue #2185, opened 2022-11-25
windows_sys::Win32::Graphics::Direct2D contains APIs that need Foundation_Numerics


glandium:
It contains the following functions:
- `D2D1ComputeMaximumScaleFactor`
- `D2D1InvertMatrix`
- `D2D1IsMatrixInvertible`
- `D2D1MakeRotateMatrix`
- `D2D1MakeSkewMatrix`

and types:
- `D2D1_BRUSH_PROPERTIES`
- `D2D1_DRAWING_STATE_DESCRIPTION`
- `D2D1_DRAWING_STATE_DESCRIPTION1`
- `D2D1_INK_STYLE_PROPERTIES`
- `D2D1_LAYER_PARAMETERS`
- `D2D1_LAYER_PARAMETERS1`

They are all behind the `Foundation_Numerics` feature, which doesn't exist, and the `Foundation::Numerics` types they rely on are not exposed in the crate.

glandium:
There's something similar in `windows_sys::Win32::System::WinRt`, which has `CreateDispatcherQueueController` behind the non-existent `System` feature, using a `::System::DispatcherQueueController` type that doesn't exist in windows-sys.

kennykerr:
Yep, there are a few such instances where the corresponding types in the `windows` crate are excluded from the `windows-sys` crate (notably all WinRT APIs) since the `windows-sys` crate lacks support for COM and WinRT. I've been meaning to exclude more of the COM-heavy namespaces from `windows-sys`, which should also help reduce the overall size and potential for breaking changes. 

https://github.com/microsoft/windows-rs/blob/773f19970b225668d4e314c35e5cc76d0220d6af/crates/tools/sys/src/main.rs#L6

PR #2194, opened 2022-11-28
Exclude major COM APIs from the `windows-sys` crate

kennykerr:
The `windows-sys` crate already excludes all WinRT APIs. This just removes some of the major families of APIs that are all but unusable without COM support and thus unusable from `windows-sys` anyway. 

This should also avoid some unnecessary breaking changes in the `windows-sys` crate. 

Fixes: #2185


riverar (commented):
Additional possible candidates for removal:
* `Win32_Media_PictureAcquisition` (https://learn.microsoft.com/en-us/previous-versions/windows/desktop/acquisition/programming-reference?redirectedfrom=MSDN)
* `Win32_System_SideShow` (https://learn.microsoft.com/en-us/previous-versions/windows/desktop/cc982216(v=vs.85))
* `Win32_System_TransactionServer`

kennykerr:
Added those.

crates/tools/sys/src/main.rs
```diff
@@ -3,7 +3,35 @@ use std::collections::*;
...
+    "Windows.Win32.Graphics.Direct2D",
+    "Windows.Win32.Graphics.Direct3D",
+    "Windows.Win32.Graphics.Direct3D10",
+    "Windows.Win32.Graphics.Direct3D11",
+    "Windows.Win32.Graphics.Direct3D11on12",
+    "Windows.Win32.Graphics.Direct3D12",
+    "Windows.Win32.Graphics.Direct3D9",
+    "Windows.Win32.Graphics.Direct3D9on12",
+    "Windows.Win32.Graphics.DirectComposition",
+    "Windows.Win32.Graphics.DirectDraw",
```
glandium, line 18: Wouldn't things like wgpu want those in windows-sys? (cc @Elabajaba)

Elabajaba: I don't think so? As far as I know `wgpu` needs COM support (and like Kenny said, these are all probably unusable without COM), which means `wgpu` needs to use `windows` instead of `windows-sys` anyways.

Issue #2180, opened 2022-11-23
Bug: In D3D demo


zhouhang95:
### Which crate is this about?

other (please share in the comments)

### Summary

```
    // Copy the triangle data to the vertex buffer.
    unsafe {
        let mut data = std::ptr::null_mut();
        vertex_buffer.Map(0, None, Some(&mut data))?;
        std::ptr::copy_nonoverlapping(
            vertices.as_ptr(),
            data as *mut Vertex,
            std::mem::size_of_val(&vertices),
        );
        vertex_buffer.Unmap(0, None);
    }
```
->
```
    // Copy the triangle data to the vertex buffer.
    unsafe {
        let mut data = std::ptr::null_mut();
        vertex_buffer.Map(0, None, Some(&mut data))?;
        std::ptr::copy_nonoverlapping(
            vertices.as_ptr(),
            data as *mut Vertex,
            vertices.len(),
        );
        vertex_buffer.Unmap(0, None);
    }
```
pub const unsafe fn copy_nonoverlapping<T>(src: *const T, dst: *mut T, count: usize)
Copies count * size_of::<T>() bytes from src to dst.

zhouhang95:
std::mem::size_of_val(&vertices) == 84
vertices.len() == 3

kennykerr:
@damyanp

riverar:
@zhouhang95 It's a holiday here in the US starting tomorrow, so you may not receive a response until Monday. Can you also provide a small complete sample that I can run here? Thank you! / 从明天开始，这是美国的假期，所以你可能要到周一才能收到回复。您能否提供我可以在这里运行的小型完整示例？谢谢！

zhouhang95:
> @zhouhang95 It's a holiday here in the US starting tomorrow, so you may not receive a response until Monday. Can you also provide a small complete sample that I can run here? Thank you! / 从明天开始，这是美国的假期，所以你可能要到周一才能收到回复。您能否提供我可以在这里运行的小型完整示例？谢谢！

Hello,
the total code is in crates/samples/direct3d12 demo.

kennykerr:
Is the sample misbehaving? It seems to work fine for me. Unfortunately, I have no idea what that code is trying to do as I'm not familiar with Direct3D, but perhaps @damyanp can confirm that it is correct.

zhouhang95:
Yes.
It copy more data than it need.
memcpy thrid param is count of byte.
[...]

PR #2195, opened 2022-11-28
Direct3D12 sample - fix copy_nonoverlapping size

damyanp:
copy_nonoverlapping expects size to be the number of elements in the array, but the previous code was providing the size in bytes.

Fixes: #2180.


kennykerr (approved):
Thanks!

PR #2206, opened 2022-11-30
Add convert methods for `VARIANT_BOOL`

aquacash5:
Adds bool conversions for `VARIANT_BOOL` to improve ergonomics.

### DIFF
--- a/crates/libs/bindgen/src/extensions/variant_bool.rs
+++ b/crates/libs/bindgen/src/extensions/variant_bool.rs
@@ -5,8 +5,7 @@ pub fn gen() -> TokenStream {
         impl VARIANT_BOOL {
             #[inline]
             pub fn as_bool(self) -> bool {
-                debug_assert!(self == VARIANT_TRUE || self == VARIANT_FALSE, "{:?} is in an invalid state", self);
-                self == VARIANT_TRUE
+                self.0 != 0
             }
             #[inline]
             pub fn ok(self) -> ::windows::core::Result<()> {
--- a/crates/libs/bindgen/src/extensions/variant_bool.rs
+++ b/crates/libs/bindgen/src/extensions/variant_bool.rs
@@ -26,11 +26,6 @@ pub fn gen() -> TokenStream {
                 self.ok().expect(msg);
             }
         }
-        impl ::core::convert::From<::core::option::Option<VARIANT_BOOL>> for VARIANT_BOOL {
-            fn from(optional: ::core::option::Option<Self>) -> Self {
-                optional.unwrap_or_default()
-            }
-        }
         impl ::core::convert::From<VARIANT_BOOL> for bool {
             fn from(value: VARIANT_BOOL) -> Self {
                 value.as_bool()
--- a/crates/libs/bindgen/src/extensions/variant_bool.rs
+++ b/crates/libs/bindgen/src/extensions/variant_bool.rs
@@ -27,6 +27,11 @@ pub fn gen() -> TokenStream {
                 self.ok().expect(msg);
             }
         }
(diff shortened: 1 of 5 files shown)


aquacash5:

kennykerr:
Thanks for the contribution, however the code is generated from metadata so you cannot simply add to this source file. Have a look at how this is done for `BOOL` here:

https://github.com/microsoft/windows-rs/blob/master/crates/libs/bindgen/src/extensions/bool32.rs

aquacash5:
Thank you, I am starting to understand how this project is put together.

I am trying to figure out where this code is generated from?

https://github.com/microsoft/windows-rs/blob/2ddb06e51568dc2f401f10245fd6158b3312107b/crates/libs/windows/src/Windows/Win32/Foundation/mod.rs#L20129-L20133

```rust
impl ::core::convert::From<::core::option::Option<BOOLEAN>> for BOOLEAN {
    fn from(optional: ::core::option::Option<BOOLEAN>) -> BOOLEAN {
        optional.unwrap_or_default()
    }
}
```

I see that it is implemented on both `BOOL` and `BOOLEAN`, but it isn't mentioned in the bindgen package.

kennykerr:
The `windows-bindgen` takes care of generating the basic definitions and the extensions are only there to add whatever cannot be derived from metadata.

crates/libs/bindgen/src/extensions/variant_bool.rs
```diff
@@ -0,0 +1,75 @@
+use super::*;
+
+pub fn gen() -> TokenStream {
+    quote! {
+        impl VARIANT_BOOL {
+            #[inline]
+            pub fn as_bool(self) -> bool {
+                debug_assert!(self == VARIANT_TRUE || self == VARIANT_FALSE, "{:?} is in an invalid state", self);
+                self == VARIANT_TRUE
```
kennykerr, line 9:
```suggestion
                self.0 != 0
```

I think we should avoid assuming too much about how VARIANT_BOOL will be used in practice. Since its a 16-bit value I'm sure there is some ancient piece of code somewhere that breaks this assumed invariant, much like `BOOL`.

crates/libs/bindgen/src/extensions/variant_bool.rs
```diff
@@ -0,0 +1,79 @@
...
+            #[track_caller]
+            pub fn expect(self, msg: &str) {
+                self.ok().expect(msg);
+            }
+        }
+        impl ::core::convert::From<::core::option::Option<VARIANT_BOOL>> for VARIANT_BOOL {
+            fn from(optional: ::core::option::Option<Self>) -> Self {
+                optional.unwrap_or_default()
+            }
+        }
```
kennykerr, line 33: Hmm, why was this added? I'm just comparing against `BOOL` (in `bool32.rs`) which does not have this conversion.

aquacash5: I added this because it is boolean functionality that was derived for `BOOL` and `BOOLEAN`, but not derived for `VARIANT_BOOL`. Even though this functionality is valid functionality for `VARIANT_BOOL`.

aquacash5: This is why I asked my question earlier. https://github.com/microsoft/windows-rs/pull/2206#issuecomment-1332352847

kennykerr: Oh, I see. That's because (for unknown reasons) the `BOOL` type is defined as a `struct` by the Win32 metadata whereas the `VARIANT_BOOL` type is defined as an `enum`. We can raise this discrepancy on the https://github.com/microsoft/win32metadata/ repo, but for now let's just make the extension code consistent.

kennykerr (approved):
Thanks!

PR #2216, opened 2022-12-01
Remove any `.dll` suffix from non-Windows link import file names

kennykerr:
This allows non-Windows imports to fallback to using the traditional Rust link attribute that expects a lib rather than a dll. 

Unfortunately, I don't have an easy way to test this.


damyanp (approved):
Do you have a way to manually generate metadata without having to read a winmd file?  If so then maybe could test some stable API like D3D12CreateDevice that has the library set to D3D12.DLL and then verify that the dll is stripped from the linker line.

BTW - are the filenames already lower case or does the trim function ignore case?

damyanp:
Thank you!

kennykerr:
> are the filenames already lower case or does the trim function ignore case?

Yep, on line 60 you can see that `to_lowercase` is called. 

I'm working on metadata generation right now - not quite ready yet tho.

PR #2235, opened 2022-12-07
Remove any implied ownership semantics for Win32 structs

kennykerr:
WinRT structs expect any non-blittable fields to be owned and dropped as necessary. Win32 structs don't provide any ownership semantics that can be gleaned from metadata and for the most part require the caller to know whether any resources are owned by the caller or callee of any particular API. 

This update solidifies this by ensure that all non-blittable field types are stored as `ManuallyDrop` to make that clear. A variant of `ManuallyDrop` is also provided and used for fields to simplify programming with the Win32 API where such fields must be accessible as references without necessarily taking ownership. 

I tested the Direct3D 12 sample and the warnings that @polymonster mentioned are no longer reported when running under a debugger. 

Fixes: #1410

### DIFF
--- a/crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
+++ b/crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
@@ -1209,14 +1209,14 @@ impl ID3D12Device {
         T: ::windows::core::Interface,
     {
         let mut result__ = ::core::mem::MaybeUninit::zeroed();
-        (::windows::core::Vtable::vtable(self).CreateGraphicsPipelineState)(::windows::core::Vtable::as_raw(self), ::core::mem::transmute(pdesc), &<T as ::windows::core::Interface>::IID, result__.as_mut_ptr()).from_abi(result__)
+        (::windows::core::Vtable::vtable(self).CreateGraphicsPipelineState)(::windows::core::Vtable::as_raw(self), pdesc, &<T as ::windows::core::Interface>::IID, result__.as_mut_ptr()).from_abi(result__)
     }
     pub unsafe fn CreateComputePipelineState<T>(&self, pdesc: *const D3D12_COMPUTE_PIPELINE_STATE_DESC) -> ::windows::core::Result<T>
     where
         T: ::windows::core::Interface,
     {
         let mut result__ = ::core::mem::MaybeUninit::zeroed();
-        (::windows::core::Vtable::vtable(self).CreateComputePipelineState)(::windows::core::Vtable::as_raw(self), ::core::mem::transmute(pdesc), &<T as ::windows::core::Interface>::IID, result__.as_mut_ptr()).from_abi(result__)
+        (::windows::core::Vtable::vtable(self).CreateComputePipelineState)(::windows::core::Vtable::as_raw(self), pdesc, &<T as ::windows::core::Interface>::IID, result__.as_mut_ptr()).from_abi(result__)
     }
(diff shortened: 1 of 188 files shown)


crates/libs/windows/src/Windows/Win32/Graphics/Direct3D12/mod.rs
```diff
@@ -25710,12 +25644,12 @@ impl ::core::default::Default for D3D12_RENDER_TARGET_VIEW_DESC_0 {
 #[repr(C)]
 #[doc = "*Required features: `\"Win32_Graphics_Direct3D12\"`*"]
 pub struct D3D12_RESOURCE_ALIASING_BARRIER {
-    pub pResourceBefore: ::core::option::Option<ID3D12Resource>,
-    pub pResourceAfter: ::core::option::Option<ID3D12Resource>,
+    pub pResourceBefore: ::windows::core::ManuallyDrop<ID3D12Resource>,
```
damyanp, line 25647: Is it possible to store the equivalent of `None`  in a ManuallyDrop - so these show up as NULL when passed to the D3D12 runtime?

kennykerr: Yes, `ManuallyDrop::none()` will give you that.

kennykerr: I added this test to illustrate: https://github.com/microsoft/windows-rs/pull/2235/commits/d9be39dc2f3808d15ca32800545d1967908a31bc

damyanp: Cool.  Is this something specific about windows-rs interfaces?  I can't find any documentation for none on https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html.

kennykerr: Yes. The `ManuallyDrop` provided by the standard library basically forces ownership, so the `windows` crate now provides a simple `ManuallyDrop` alternative for use in such cases reference semantics are desired. See here: 

https://github.com/microsoft/windows-rs/pull/2235/commits/18546d2361abeac538b1e86ea03599d86761ea5a#diff-594f4615fa893560c293b924a5380f92ad1c39e74b45a643b979c14c1a957c88

damyanp: Ah, I see. Any concerns about possible confusion to the same named struct in std::mem?

damyanp (approved):

crates/libs/windows/src/core/manually_drop.rs
```diff
@@ -0,0 +1,46 @@
...
+    abi: T::Abi,
+}
+
+impl<T: super::Abi> ManuallyDrop<T> {
+    pub fn new(value: &T) -> Self {
+        Self { abi: value.abi() }
+    }
+
+    pub fn abi(&self) -> T::Abi {
+        unsafe { std::mem::transmute_copy(&self.abi) }
```
rylev, line 17: Is `Abi` not constrained to be `Copy`? I guess we sometimes set `Abi` to `std::mem::ManuallyDrop` which would not be `Copy` even though semantically it usually should be.

kennykerr: Sadly, it can't be `Copy`. That would make it a lot simple for callers. Rust requires `Copy` for fields of unions but won't let me make a `Copy` type without generating considerably more code by essentially unrolling `ManuallyDrop` types and duplicating their memory layout.

PR #2237, opened 2022-12-08
Simplify parameter bindings

kennykerr:
Following on #2235, the input parameter bindings can be simplified further.


rylev:
I don't believe that it is really sound to remove the lifetime param from `InParam`, but I'm not sure I have the energy to explain why. The fact this was removed with no real explanation is certainly concerning.

kennykerr:
Happy to add the lifetime back if it serves a purpose, but it wasn't constraining anything so seemed harmless enough. The `InParam` very clearly is only used for input parameters and should thus always live as long the function call, but I could be missing something.

PR #2248, opened 2022-12-12
Infallible return values and more efficient out parameters 

### LINKED ISSUE #2246: Support infallible return values
Now that we're starting to get more accurate trailing optional parameters thanks to https://github.com/microsoft/win32metadata/issues/1005 we can start to more accurately represent infallible return values like `ID2D1BitmapBrush.GetBitmap`.

kennykerr:
Now that we're starting to get more accurate optional parameter metadata thanks to https://github.com/microsoft/win32metadata/issues/1005 we can more accurately represent infallible return values. These are functions and COM interface methods that have a `void` return type in metadata but also have a trailing out parameter that may be treated as a logical return value. For example, `GetLocalTime` goes from:

```rust
let mut time = Default::default();
GetLocalTime(&mut time);
```

To this:

```rust
let time = GetLocalTime();
```

And `ID2D1DeviceContext`'s GetTarget method goes from:

```rust
let mut previous = None;
target.GetTarget(&mut previous);
```

To this:

```rust
let previous = target.GetTarget()?;
```

A `Result` is used to allow detection of null COM interface pointer values.

I now also more carefully detect large structs and avoid transforming those to avoid large stack usage. For example, `IDCompositionDevice2`'s `GetFrameStatistics` method now returns the `DCOMPOSITION_FRAME_STATISTICS` value as an output parameter. Any structure larger than 128 bits (the size of `GUID`) is passed by reference rather than by return value. This is especially useful for DirectX matrix structures. 

Fixes #2246

### DIFF
--- a/crates/libs/bindgen/src/com_methods.rs
+++ b/crates/libs/bindgen/src/com_methods.rs
@@ -79,6 +79,35 @@ pub fn gen(gen: &Gen, def: TypeDef, kind: InterfaceKind, method: MethodDef, meth
                 }
             }
         }
+        SignatureKind::ReturnValue => {
+            let args = gen.win32_args(&signature.params, kind);
+            let params = gen.win32_params(&signature.params, kind);
+            let return_type = signature.params[signature.params.len() - 1].ty.deref();
+            let is_nullable = gen.reader.type_is_nullable(&return_type);
+            let return_type = gen.type_name(&return_type);
+
+            if is_nullable {
+                quote! {
+                    #doc
+                    #features
+                    pub unsafe fn #name<#generics>(&self, #params) -> ::windows::core::Result<#return_type> #where_clause {
+                        let mut result__ = ::core::mem::MaybeUninit::zeroed();
+                        (::windows::core::Vtable::vtable(self)#bases.#vname)(::windows::core::Vtable::as_raw(self), #args);
+                        <#return_type as ::windows::core::Abi>::from_abi(result__.assume_init())
+                    }
+                }
+            } else {
+                quote! {
+                    #doc
+                    #features
+                    pub unsafe fn #name<#generics>(&self, #params) -> #return_type #where_clause {
+                        let mut result__ = ::core::mem::MaybeUninit::zeroed();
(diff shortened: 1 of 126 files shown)


MarijnS95:
@kennykerr we just noticed that `IDXGIAdapter4::GetDesc3` went from `GetDesc3()` to `GetDesc3(&mut desc)` in `windows 0.44.0`. Now I'm not sure what the metadata contains or if the metadata is off, but https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_6/nf-dxgi1_6-idxgiadapter4-getdesc3 states, for `[out] pDesc`:

> This parameter must not be **NULL**

Any idea why [this changed](https://github.com/Traverse-Research/gpu-allocator/pull/150/files#diff-0905e6011ee596acd9d557cd8d5a97cf8fcbcc0c0a399e2f4ef2e337f914b275L36-R37)?

kennykerr:
I mentioned this in the PR description above. Testing this transformation revealed that it negatively affects certain methods that logically return very large structs, like DirectX matrixes, so I put a cap on the size of the structs that would be candidates for this transformation. Unfortunately, the metadata for return values is not very helpful at this point so I have to use a heuristic. The heuristic was just a little too generous before. 

https://github.com/microsoft/windows-rs/blob/f5373bb681f2c3f8ed1401744406f337dc3ca6f3/crates/libs/metadata/src/reader/mod.rs#L1324-L1349

MarijnS95:
Ack, `DXGI_ADAPTER_DESC3` seems to be bigger than 128 bytes. Didn't notice this when originally scrubbing the description, thanks for pointing this out.

MarijnS95:
@kennykerr in hindsight I'm not sure what kind of RVO or other optimizations Rust would perform here (also considering lack of `#[inline]`), but I hope the move semantics language feature doesn't come at the detriment of performance. The previous approach is definitely easier to write.

If so, the only place this might be useful is when updating fields already in a struct, though I suspect most would be constructed by moving `DESC3` (and other struct values) out of a function, and into a struct constructor.

PR #2274, opened 2023-01-09
Reduce `windows` crate build time by moving non-essential functions behind the `default` feature

kennykerr:
The `windows` crate's build time continues to be a challenge as it depends on many small functions that make it substantially simpler to use the Windows API. Unfortunately, the Rust compiler spends a lot more time compiling functions than your typical C++ compiler, and this ends up hurting the `windows` crate. 

While many of these functions are essential, a large enough number are avoidable. This includes the functions used to implement `Default`, `PartialEq`, and `Debug` traits. COM interfaces also support inheritance hierarchies that can end up duplicating numerous inherited methods in Rust. While convenient, none of this is necessary to use the Windows API. 

In order to give developers more control, this update moves these non-essential functions behind the `default` feature. That means that nothing should change for existing code, but a dependent crate can opt to exclude this code as follows:

```
[dependencies.windows]
default-features = false
...
```

This update also includes a copy of the Direct3D 12 sample that does not use the default features. It required relatively minor changes and reduced build time from 9-10 seconds down to 6-7 seconds. Naturally, your mileage may vary and depend on which modules you rely on. Without default features, the `Win32::Graphics::Direct3D12` module avoids parsing roughly 13,700 lines of code. That's over 2,400 functions in that module alone. 

Of course, if you're not using COM or WinRT APIs then you can substantially reduce build time simply by opting to use the `windows-sys` crate rather than the `windows` crate. More info here: 

https://kennykerr.ca/rust-getting-started/windows-or-windows-sys.html


damyanp:
This sounds cool.  Is there something about Direct3D12 or the sample that made it stand out as one that needs looking at, maybe something that we should consider in future API designs?  Or was it picked as a semi-random example of the impact of this change?

kennykerr:
> This sounds cool. Is there something about Direct3D12 or the sample that made it stand out as one that needs looking at, maybe something that we should consider in future API designs? Or was it picked as a semi-random example of the impact of this change?

I was just looking for a relatively large sample that relied on a number of larger modules with both traits and COM inheritance.

riverar:
Can you explicitly list a few _non-essential functions_ that would be tucked away?

kennykerr:
> Can you explicitly list a few non-essential functions that would be tucked away?

All of the functions that now appear in `default.rs` rather than `mod.rs`. For example, here's what moves from `mod.rs` to `default.rs` for the `ID2D1Bitmap` interface: 

https://github.com/microsoft/windows-rs/blob/d64f0df34e16d32352430ed4bfe16b45d7d8ec97/crates/libs/windows/src/Windows/Win32/Graphics/Direct2D/default.rs#L3059-L3076

The traits are non-essential and the `GetFactory function is inherited by `ID2D1Bitmap` from the `ID2D1Resource` interface.

kennykerr:
Was hoping to get more feedback on this but as it stands, I'm not super impressed with the wins. I'm going to incorporate some of these tweaks separately and keep experimenting with ways to reduce build time.

PR #2284, opened 2023-01-14
Only show visible windows and include x/y in output

jessemcdowell:
I'm not sure if contributes to the samples are welcome... I wanted an example that showed passing a struct to a Win32 function and working with bit masks, so I added one.

### DIFF
--- a/crates/samples/windows/enum_windows/src/main.rs
+++ b/crates/samples/windows/enum_windows/src/main.rs
@@ -1,6 +1,8 @@
 use windows::{
     Win32::Foundation::{BOOL, HWND, LPARAM},
-    Win32::UI::WindowsAndMessaging::{EnumWindows, GetWindowTextW},
+    Win32::UI::WindowsAndMessaging::{
+        EnumWindows, GetWindowInfo, GetWindowTextW, WINDOWINFO, WS_VISIBLE,
+    },
 };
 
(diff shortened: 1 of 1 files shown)


crates/samples/windows/enum_windows/src/main.rs
```diff
@@ -13,8 +15,14 @@ extern "system" fn enum_window(window: HWND, _: LPARAM) -> BOOL {
...
 
-        if !text.is_empty() {
-            println!("{text}");
+        let mut info = WINDOWINFO {
+            cbSize: core::mem::size_of::<WINDOWINFO>() as u32,
+            ..Default::default()
+        };
+        GetWindowInfo(window, &mut info).unwrap();
+
+        if !text.is_empty() && info.dwStyle & WS_VISIBLE.0 != 0 {
```
kennykerr, line 24: Looks like `WINDOWINFO`'s `dwStyle` and `dwExStyle` fields need to be adjusted to use the `WINDOW_STYLE` and `WINDOW_EX_STYLE` types. Could you open an issue for that? https://github.com/microsoft/win32metadata

jessemcdowell: No problem. https://github.com/microsoft/win32metadata/issues/1430

kennykerr (approved):
Thanks!

Issue #2286, opened 2023-01-16
Bug: D3D11 `XXSet*` functions incorrectly project optional array of optional objects as optional array of required objects


chyyran:
I'll use `PSSetShaderResources` as an example but this applies to most `XXSet*` functions.

[`PSSetShaderResources`](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-pssetshaderresources) has the following signature

```cpp
void PSSetShaderResources(
  [in]           UINT                     StartSlot,
  [in]           UINT                     NumViews,
  [in, optional] ID3D11ShaderResourceView * const *ppShaderResourceViews
);
```

In 0.43 this was projected as 

```rust
unsafe fn PSSetShaderResources(&self, startslot: u32, ppshaderresourceviews: Option<&[Option<ID3D11ShaderResourceView>]>)
```

This signatured allowed Rust to express the following C++ pattern

```c++
auto shaderResources = {
   resource1,
   nullptr,
   resource2
};

context.PSSetShaderResources(0, &shaderResources);
```

```rust
let shader_resources: &[Option<ID3D11ShaderResourceView>; 3] = &[
   Some(resource1),
   None,
   Some(resource2)
];

context.PSSetShaderResources(0, Some(shader_resources));
```

This pattern is inexpressible in 0.44 without transmute because `PSSetShaderResources` is projected as

```rust
unsafe fn PSSetShaderResources(&self, startslot: u32, ppshaderresourceviews: ::core::option::Option<&[ID3D11ShaderResourceView]>)
```

i.e. every `ID3D11ShaderResourceView` must be valid.

The workaround is to use the following transmute

```rust

let shader_resources: &[Option<ID3D11ShaderResourceView>; 3] = &[
   Some(resource1),
   None,
   Some(resource2)
];

// windows-rs does not guarantee that the layout of IUnknown will remain NonNull<c_void>. If it changes, the following 
// transmute goes from UB that "works on my machine", to instant nasal demons. This const assert should give us the 
// confidence that the transmute "does what it should".

// This also assumes that all D3D11 COM pointer types wrap an IUnknown. I feel like this is a stronger guarantee than the
// Rust layout of IUnknown given COM lite pointers are guaranteed castable to IUnknown.

const _: () = assert!(std::mem::size_of::<Option<windows::core::IUnknown>>() == std::mem::size_of::<windows::core::IUnknown>());

[...]

PR #2293, opened 2023-01-18
Slices should use default (optional) type

kennykerr:
This effectively reverts #2233 - it turns out that some DirectX APIs require arrays of pointers where some or all must be null.

Fixes: #2286

### DIFF
--- a/crates/libs/bindgen/src/gen.rs
+++ b/crates/libs/bindgen/src/gen.rs
@@ -980,11 +980,10 @@ impl<'a> Gen<'a> {
                 }
                 SignatureParamKind::ArrayRelativeLen(_) => {
                     let ty = param.ty.deref();
+                    let ty = self.type_default_name(&ty);
                     let ty = if self.reader.param_flags(param.def).output() {
-                        let ty = self.type_default_name(&ty);
                         quote! { &mut [#ty] }
                     } else {
-                        let ty = self.type_name(&ty);
                         quote! { &[#ty] }
                     };
                     if self.reader.param_flags(param.def).optional() {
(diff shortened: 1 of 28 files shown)


damyanp (approved):
I can see that it's painful to force Optional on all arrays here.  I guess the proper way to do this would be to get some metadata to describe that the array elements themselves are optional (I don't think that the SAL annotation in the D3D headers cover this already, so it'd be a game of whack-a-mole to try and find all the right places to add it).

kennykerr:
Yep, I don't see a good way to do this retroactively.

ChrisDenton:
Isn't a recurring problem that DirectX APIs are COM-like rather than actual COM? I mean they look like COM but can differ in various ways. If so then I'm of the tentative opinion that DirectX APIs should be handled differently from normal Windows API COM.

kennykerr:
The `windows` crate does not distinguish, nor care, whether a given API depends on the COM runtime services. It only cares that a given API inherits from `IUnknown` and thus supports dynamic discovery and lifetime management. That is why it can transparently support both WinRT and `IUnknown`-based APIs as they share this in common. Short of additional metadata, I believe the change here is correct however inconvenient it may be for some APIs.

damyanp:
> I'd encourage DirectX to introduce additional metadata that we can then light up in Rust. 😊

ID3D11DeviceContext's idl has `pointer_default( unique )`, which I think would be a good indicator that any pointer value in the interface could be NULL.  I don't think that this attribute makes it through the `.h` -> `.winmd` translations.

> DirectX APIs should be handled differently from normal Windows API COM.

I think we already have a bunch of general "do things differently" set in win32metadata for the d3d headers.  Maybe this could be another one of them?

Issue #2296, opened 2023-01-20
Bug: README no longer links to API docs


Ben-Lichtman:
### Which crate is this about?

windows, windows-sys

### Crate version

master

### Summary

The README used to link to the (full) API docs so they were relatively easy to reach (ie https://microsoft.github.io/windows-docs-rs/doc/windows/), now they are nowhere.

Please could these links be restored to the readme? Thanks

### Expected behavior

See the links to the docs

### Actual behavior

No links to the docs

kennykerr:
The readme is used to build the readme for the crates. For example, here's the [windows-sys](https://crates.io/crates/windows-sys) crate where you can see both the readme content as well as the doc link on the right. I find this a little simpler and less confusing as I don't have to worry about the doc links pointing at the wrong version of the docs. In general, the most reliable way to find the docs for a given crate is to use the documentation link on https://crates.io/.

Another approach would be to have dedicated readme pages for the respective crates, but that's a bit more work that I've not got around to yet.

Ben-Lichtman:
I do like that. My usual flow for looking up docs is to go to docs.rs/cratename.

For the windows crate I would go from the docs.rs -> click the link to the git repo -> click the docs link on the repo.

Maybe there is some way to link to the docs from docs.rs also to avoid this problem?

ChrisDenton:
It seems like this could be changed to point to the full docs https://github.com/microsoft/windows-rs/blob/e06dc97e00f15fddbf9685c02b2b28d17fa466f7/crates/libs/windows/src/lib.rs#L1-L3

But the problem is the same wording will be used for both the stub docs and the full docs. This could perhaps be finessed with the right wording. Failing that a `#[cfg(docsrs)]` (see https://docs.rs/about/builds) could be used to change the docs depending on who is building it.

riverar:
The documentation Microsoft hosts is typically only applicable to the latest version so linking it to any crate directly may cause more confusion.

Perhaps until the docs.rs/windows connundrum is solved, you could just add a browser favorite. And maybe we can add a link to the Getting Started page.

kennykerr:
https://kennykerr.ca/rust-getting-started/how-to-find-api.html

kennykerr:
@ChrisDenton I've run out of steam trying to get docs.rs to handle the windows crate. Feel free to see what you can come up with.

ChrisDenton:
[...]

PR #2297, opened 2023-01-20
Add link from docs.rs to full API documentation

ChrisDenton:
This attempts to help with discovery for people who find Rust documentation directly via docs.rs and not the crates.io readme. The wording chosen hopefully makes it clear that it's not necessarily the same version as the doc.rs page (note that old versions on crates.io have the same issue). I guess we could have something about building old docs locally for that case but that's probably better left to the FAQ if it's desired.

The docs.rs documentation can be generated locally by setting `$env:RUSTDOCFLAGS = "--cfg docsrs"`. Though obviously testing properly would require actually publishing it, unfortunately.

Fixes: #2296

### DIFF
--- a/crates/libs/windows/Cargo.toml
+++ b/crates/libs/windows/Cargo.toml
@@ -14,6 +14,7 @@ rust-version = "1.64"
 [package.metadata.docs.rs]
 default-target = "x86_64-pc-windows-msvc"
 targets = []
+rustc-args = ["--cfg", "docsrs"]
 
 [target.'cfg(not(windows_raw_dylib))'.dependencies]
 windows-targets = { path = "../targets",  version = "0.42.1" }
--- a/crates/libs/windows/src/lib.rs
+++ b/crates/libs/windows/src/lib.rs
@@ -1,3 +1,5 @@
+#![cfg_attr(docsrs, doc = "This is a stub. The latest API documentation is here: <https://microsoft.github.io/windows-docs-rs/>")]
+#![cfg_attr(docsrs, doc = "")]
 /*!
 Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs>
 */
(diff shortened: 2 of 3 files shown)


crates/libs/windows/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(docsrs, doc = "This is a stub. The latest API documentation is here: <https://microsoft.github.io/windows-docs-rs/>")]
+#![cfg_attr(docsrs, doc = "")]
 /*!
 Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs>
 */
```
kennykerr, line 5:
```suggestion
/*!
This is a stub. The latest API documentation is here: <https://microsoft.github.io/windows-docs-rs/>

Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs>
*/
```

Thanks Chris, I think the same effect can be achieved without messing with docsrs flags/attributes. No?

ChrisDenton: Sure. But wouldn't that also affect the docs generated for microsoft.github.io? Which is not a stub.

kennykerr: Oh, got it. 😊

kennykerr (approved):

crates/libs/windows/Cargo.toml
```diff
@@ -14,6 +14,7 @@ rust-version = "1.64"
 [package.metadata.docs.rs]
 default-target = "x86_64-pc-windows-msvc"
 targets = []
+rustc-args = ["--cfg", "docsrs"]
```
Nemo157, line 17: `rustc-args` don't apply to the doc-generation step but to the check/proc-macro builds of all the dependencies. This should be `rustdoc-args`.

ChrisDenton: Yeah, I just figured that. Btw, maybe the [cross-compiling](https://docs.rs/about/builds#cross-compiling) section should mention this?

Nemo157: Yep https://github.com/rust-lang/docs.rs/issues/2284

ChrisDenton: Thanks!

PR #2303, opened 2023-01-24
Fix test component re-build reliability

kennykerr:
The build script wasn't reliably detecting when the MIDL compiler failed, leading cargo to cache the last successful result and seemingly ignoring IDL changes.


riverar (approved):
Great to see midlrt errors in stdout/err now too.

PR #2308, opened 2023-01-25
Update Win32 metadata to version 41.0.25

kennykerr:
The changes in the `windows-sys` crate are due to reserved parameters now being treated as input parameters, thus not requiring mutability. The remaining changes (to both crates) stem from improvements and additions to the Win32 metadata. 

https://github.com/microsoft/win32metadata/releases/tag/v41.0.25-preview


riverar:
If this isn't urgent, we might want to wait for https://github.com/microsoft/win32metadata/issues/1288 to get added. This will create a large amount of additional reserved parameter metadata causing you to do this again.

kennykerr:
Happy to wait if that will be landing relatively soon.

PR #2312, opened 2023-01-27
Add `raw-dylib` testing with `nightly-i686-pc-windows-gnu`

kennykerr:
This should now work thanks to https://github.com/rust-lang/rust/issues/104453 being closed.

### DIFF
--- a/.github/workflows/raw_dylib.yml
+++ b/.github/workflows/raw_dylib.yml
@@ -20,6 +20,7 @@ jobs:
           - target: x86_64-pc-windows-msvc
           - target: i686-pc-windows-msvc
           - target: x86_64-pc-windows-gnu
+          - target: i686-pc-windows-gnu
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


kennykerr:
Hmm, seems something is still not working with `raw-dylib` on `nightly-i686-pc-windows-gnu`.

wesleywiser:
@kennykerr You might want to retry this with the latest nightly build now that https://github.com/rust-lang/rust/pull/107360 has landed.

kennykerr:
@wesleywiser I reran but it still fails with the same error.

kennykerr:
@dpaoliello suggested it might be https://github.com/rust-lang/rust/issues/99534

mati865:
DLL mismatch wouldn't cause `error adding symbols: File format not recognized` error.

riverar:
I have never been successful with the self-contained GNU tools that ships with the tier 1 GNU targets. Switched to using the external tooling.

kennykerr:
Thanks @riverar!

PR #2322, opened 2023-02-03
Update Win32 metadata

kennykerr:
More info here: https://github.com/microsoft/win32metadata/releases/tag/v41.0.25-preview


matteomeli:
How would an implementation make use of the AgilitySDK for DirectX12 using windows-rs since this merge? 

Does the user need to anything else, in particular are any of the steps mentioned in https://devblogs.microsoft.com/directx/gettingstarted-dx12agility/ needed on the client application side to make sure the AgilitySDK is used?

kennykerr:
You may want to ask the folks on this PR: https://github.com/microsoft/win32metadata/pull/1445

Issue #2327, opened 2023-02-05
Bug: cross compilation with the GNU toolchain doesn't work for static libs with custom StaticLibraryAttribute


wravery:
### Which crate is this about?

other (please share in the comments)

### Crate version

0.44.0

### Summary

If you use a custom winmd generated with the `StaticLibrary` attribute, indicating a preferred static library along with a DLL, `windows-bindgen` can generate a static `#[link]` attribute pointing to that static library. However, this only works when building on Windows, with the MSVC toolchain. If you try to cross-compile from another host or opt-in to the GNU toolchain from Windows, it doesn't know how to link the static library. It can only handle linking against DLLs.

### Toolchain version/configuration

Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\wrave\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc (default)
beta-x86_64-pc-windows-msvc
nightly-x86_64-pc-windows-msvc

installed targets for active toolchain
--------------------------------------

aarch64-pc-windows-msvc
i686-pc-windows-msvc
wasm32-unknown-unknown
x86_64-pc-windows-gnu
x86_64-pc-windows-msvc

active toolchain
----------------

stable-x86_64-pc-windows-msvc (default)
rustc 1.67.0 (fc594f156 2023-01-24)

### Reproducible example

```rust
I worked around this in https://github.com/wravery/webview2-rs by patching the output of windows-bindgen to support cross-compilation: [webview2-com-sys](https://github.com/wravery/webview2-rs/blob/1861c1583c0d9f606484b17b4b1d397a8f8957d4/crates/bindings/build.rs#L390).

To reproduce this issue, you could turn it into a no-op that just returns Ok(bindings).
```

### Expected behavior

The output of `windows-bindgen` should cross-compile with the GNU toolchain.

### Actual behavior

Without the patching function in build.rs, the `#[link]` attributes all look like this:

```rust
#[link(name = "WebView2LoaderStatic", kind = "static")]
```

This works perfectly for MSVC, but it breaks with the GNU toolchain.

### Additional comments

I'm about to publish a PR with a fix, I'm just filling in background information here.

riverar:
Hey @wravery! Can you provide a bit more detail on what you're seeing in terms of errors when it comes to static libs?

I created a repro crate that compiles successfully and demonstrates
* creating metadata w/ static lib attributes
* generating bindings using metadata
* creating a static lib (using gnu tooling)
[...]

PR #2328, opened 2023-02-05
Fallback to dynamic linking for GNU toolchain

wravery:
If you use a custom winmd generated with the `StaticLibrary` attribute, indicating a preferred static library along with a DLL, `windows-bindgen` can generate a static `#[link]` attribute pointing to that static library. However, this only works when building on Windows, with the MSVC toolchain. If you try to cross-compile from another host or opt-in to the GNU toolchain from Windows, it doesn't know how to link the static library. It can only handle linking against DLLs.

This change is adapted from a patch method I used in [webview2-com-sys](https://github.com/wravery/webview2-rs/blob/1861c1583c0d9f606484b17b4b1d397a8f8957d4/crates/bindings/build.rs#L390) to support cross-compilation. It uses `#[cfg_attr]` to conditionally link against either the static library or the DLL depending on the target toolchain.

The output of the patch in `webview2-com-sys` looks like this:

```rust
    #[cfg_attr(
        target_env = "msvc",
        link(name = "WebView2LoaderStatic", kind = "static")
    )]
    #[cfg_attr(not(target_env = "msvc"), link(name = "WebView2Loader"))]
```

With this change in `windows-bindgen`, it looks like this:

```rust
    #[cfg_attr(
        target_env = "msvc",
        link(name = "WebView2LoaderStatic", kind = "static")
    )]
    #[cfg_attr(not(target_env = "msvc"), link(name = "webview2loader"))]
```

The only difference between my custom patch and the updated `windows-bindgen` output is the case of the DLL name.

Fixes: #2327

### DIFF
--- a/crates/libs/bindgen/src/functions.rs
+++ b/crates/libs/bindgen/src/functions.rs
@@ -55,14 +55,7 @@ fn gen_win_function(gen: &Gen, def: MethodDef) -> TokenStream {
     let extern_abi = gen.reader.method_def_extern_abi(def);
     let abi_return_type = gen.return_sig(&signature);
 
-    let link = if let Some(link) = gen.reader.method_def_static_lib(def) {
-        quote! {
-            #[link(name = #link, kind = "static")]
-            extern #extern_abi {
-                fn #name(#(#abi_params),*) #abi_return_type;
-            }
-        }
-    } else {
+    let link = {
         let impl_map = gen.reader.method_def_impl_map(def).expect("ImplMap not found");
         let scope = gen.reader.impl_map_scope(impl_map);
         let link = gen.reader.module_ref_name(scope).to_lowercase();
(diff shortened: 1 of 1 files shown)


wravery:
I did get a test failure, but I don't think it's related to this change. It seems to fail consistently on my machine against the `master` branch as well:

```
failures:

---- agile_send stdout ----
Error: Error { code: HRESULT(0x80010002), message: "Call was canceled by the message filter." }


failures:
    agile_send

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 10.03s

error: test failed, to rerun pass `-p test_agile_reference --test tests`
```

kennykerr:
Closing as per https://github.com/microsoft/windows-rs/issues/2327#issuecomment-1419217669

PR #2329, opened 2023-02-06
metadata `Reader::attribute_args` enhancements

ChrisDenton:
I've been playing with reading attribute metadata and wanted to help fix a couple of issues I encountered:

* `ComVisibleAttribute` has a bool argument, so it's useful to be able to read it. I also added a panic if the blob bool value is not either 0 or 1 (as per the spec). Admittedly this may be overkill but I figured it'd help detect if something has gone wrong somewhere.
* `UnmanagedFunctionPointerAttribute` requires `System.Runtime.InteropServices.CallingConvention` and `AttributeUsageAttribute` requires `System.AttributeTargets`. Neither of which are part of the default metadata files. I've added the type name to the panic message to make this easier to diagnose but I guess this is a win32metadata issue?

### DIFF
--- a/crates/libs/metadata/src/reader/blob.rs
+++ b/crates/libs/metadata/src/reader/blob.rs
@@ -70,6 +70,14 @@ impl<'a> Blob<'a> {
             String::from_utf16_lossy(slice)
         }
     }
+    pub fn read_bool(&mut self) -> bool {
+        // A bool is specified as "a single byte with value 0 (false) or 1 (true)".
+        match self.read_u8() {
+            0 => false,
+            1 => true,
+            _ => panic!("Illegal bool value"),
+        }
+    }
     pub fn read_i8(&mut self) -> i8 {
         let value = i8::from_le_bytes(self[..1].try_into().unwrap());
         self.offset(1);
--- a/crates/libs/metadata/src/reader/mod.rs
+++ b/crates/libs/metadata/src/reader/mod.rs
@@ -1591,7 +1591,11 @@ impl<'a> Reader<'a> {
             }
         }
 
-        Type::TypeDef((self.get(full_name).next().expect("Type not found"), Vec::new()))
+        if let Some(ty) = self.get(full_name).next() {
+            Type::TypeDef((ty, Vec::new()))
+        } else {
+            panic!("Type not found: {}", full_name);
+        }
     }
     fn type_from_blob(&self, blob: &mut Blob, enclosing: Option<TypeDef>, generics: &[Type]) -> Option<Type> {
         let is_winrt_const_ref = blob.read_modifiers().iter().any(|def| self.type_def_or_ref(*def) == TypeName::IsConst);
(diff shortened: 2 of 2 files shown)


crates/libs/metadata/src/reader/blob.rs
```diff
@@ -70,6 +70,14 @@ impl<'a> Blob<'a> {
             String::from_utf16_lossy(slice)
         }
     }
+    pub fn read_bool(&mut self) -> bool {
+        // A bool is specified as "a single byte with value 0 (false) or 1 (true)".
```
riverar, line 74: ECMA-335, II.7.4 Native data types ([PDF](https://www.ecma-international.org/wp-content/uploads/ECMA-335_6th_edition_june_2012.pdf) page 125), states:
> Boolean. 4-byte integer value where any non-zero value represents TRUE, and 0 represents 
FALSE.

riverar:
> * `UnmanagedFunctionPointerAttribute` requires `System.Runtime.InteropServices.CallingConvention` and `AttributeUsageAttribute` requires `System.AttributeTargets`. Neither of which are part of the default metadata files. I've added the type name to the panic message to make this easier to diagnose but I guess this is a win32metadata issue?

Don't believe metadata is using these attributes. Are you by any chance using ILSpy? (It automatically/annoyingly loads in unrelated .NET reference assemblies.)

crates/libs/metadata/src/reader/blob.rs
ChrisDenton: Sorry, I should have been clear, this is a custom attribute, ECMA-335, II.23.3 (page 268).

> If the parameter kind is simple (first line in the above diagram) ( bool, char, float32, float64, int8, int16, int32, int64, unsigned int8, unsigned int16, unsigned int32 or unsigned int64) then the 'blob' contains its binary value (Val). (A bool is a single byte with value 0 (false) or 1 (true); char is a two-byte Unicode character; and the others have their obvious meaning.) This pattern is also used if the parameter kind is an enum -- simply store the value of the enum's underlying integer type.

ChrisDenton:
No, just the metadata reader. A simplified example:

```rust
let files = metadata::reader::File::with_default(&[]).unwrap();
    let reader = &metadata::reader::Reader::new(&files);
    let tree = reader.tree("Windows", &[]).unwrap();
    let root = tree.flatten();

    let mut attributes_table = std::collections::BTreeMap::new();
    for tree in root {
        if !tree.namespace.starts_with("Windows.Win32.") {
            continue;
        }
        for ty in reader.namespace_types(tree.namespace) {
            for attr in reader.type_def_attributes(ty) {
                let name = reader.attribute_name(attr);
                let args = reader.attribute_args(attr);
                attributes_table.insert(name, args);
            }
        }
    }
```

crates/libs/metadata/src/reader/blob.rs
riverar: Ah, I should've paid closer attention--this is a blob. 👍

riverar:
I think I'm having a moment. Are you saying the reader is pulling in default metadata with references to `UnmanagedFunctionPointerAttribute` or `System.Runtime.InteropServices.CallingConvention`? I'm not seeing that here, so I think I'm missing something.

ChrisDenton:
I mean, they do appear to exist in the generated C# as well. E.g. https://github.com/microsoft/win32metadata/search?q=UnmanagedFunctionPointer

kennykerr (approved):
Looks good - thanks!

Issue #2330, opened 2023-02-06
metadata reader: inspect enum type in attribute arguments


ChrisDenton:
It'd be cool if I could inspect the type of enums in attribute arguments. Currently this type information is discarded and only the raw value is kept:

https://github.com/microsoft/windows-rs/blob/721027379419bc58a9293adc51ea04a55e87e993/crates/libs/metadata/src/reader/mod.rs#L292-L302

My thinking is to create a new `Value::Enum` variant for this. Something like:

```rust
pub enum Value {
    ...
    // I suspect that, in practice, all attribute enums are i32 (or at least fit into it).
    // I could be wrong though.
    Enum(TypeDef, i32),
}
```

I can submit a PR if this is ok.

kennykerr:
I'm busy working on the metadata writer and that might be useful for testing metadata generation more carefully. The precise type should only matter when it comes to reading constant values, so a single variant seems reasonable here.

PR #2331, opened 2023-02-07
Metadata reader: Preserve type when reading enum values in attributes

ChrisDenton:
Also implement parsing enums in named arguments.

I've gone back on forth on whether to use an `i32`, `u64` or an `Integer` type to store the value. I eventually landed on the latter but please stop me if I'm over-engineering this.

Closes: #2330

### DIFF
--- a/crates/libs/metadata/src/reader/mod.rs
+++ b/crates/libs/metadata/src/reader/mod.rs
@@ -127,6 +127,18 @@ pub enum Value {
     F64(f64),
     String(String),
     TypeDef(TypeDef),
+    Enum(TypeDef, Integer),
+}
+
+pub enum Integer {
+    U8(u8),
+    I8(i8),
+    U16(u16),
+    I16(i16),
+    U32(u32),
+    I32(i32),
+    U64(u64),
+    I64(i64),
 }
 
 pub struct Signature {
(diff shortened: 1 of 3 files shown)


crates/libs/metadata/src/reader/mod.rs
```diff
@@ -127,6 +127,18 @@ pub enum Value {
     F64(f64),
     String(String),
     TypeDef(TypeDef),
+    Enum(TypeDef, Integer),
+}
+
+pub enum Integer {
```
kennykerr, line 133: Looks good.

crates/libs/metadata/src/reader/mod.rs
```diff
@@ -312,14 +314,19 @@ impl<'a> Reader<'a> {
...
-            let name = values.read_str().to_string();
+            let mut name = values.read_str().to_string();
             let arg = match arg_type {
                 0x02 => Value::Bool(values.read_bool()),
                 0x06 => Value::I16(values.read_i16()),
                 0x08 => Value::I32(values.read_i32()),
                 0x09 => Value::U32(values.read_u32()),
                 0x0E => Value::String(values.read_str().to_string()),
                 0x50 => Value::TypeDef(self.get(TypeName::parse(values.read_str())).next().expect("Type not found")),
+                0x55 => {
```
kennykerr, line 325: Is this being tested anywhere? Perhaps add a test to `crates/tests/metadata` that reads the `SupportedArchitecture` attributes on `SetWindowLongPtrA`.

ChrisDenton: Oh good point, I'll add that. It is being tested in my own code at least ;)

ChrisDenton: So the only attribute that actually hits this code path is `GCPressure` (which I guess is useful to GCs). I've added a test for it but I don't know how reliable that attribute is.

kennykerr: As soon as I have the metadata generation completed, we can stop digging around for test cases and just write discrete tests from scratch.

kennykerr (approved):
Thanks!

Issue #91, opened 2020-04-28
Support for creating collections


kennykerr:
Many APIs assume the caller will provide a collection of values. This should integrate with `winrt::Param` to allow regular Rust collections and iterators to bind to WinRT input parameters of WinRT collection types.

rookboom:
> Many APIs assume the caller will provide a collection of values. This should integrate with `winrt::Param` to allow regular Rust collections and iterators to bind to WinRT input parameters of WinRT collection types.

Is there any workaround for this? I am trying to create a `windows::ai::machine_learning::TensorFloat`. It  has a `create2` constructor function which requires an `Param<IIterable<i64>>` as input .

Ideally I would like to call `TensorFloat::create2([1,2,3])`.
If I could create an `IVector` `v` somehow, I could maybe do something like `TensorFloat::create2(v.get_view().into())` but the `single_threaded_vector` function does not seem to be available in Rust.

kennykerr:
As a workaround, you can create a WinRT component with C++/WinRT that creates and returns the collection and then call that from Rust. You can even just have a component that returns an empty collection that you can then populate using Rust/WinRT.

riverar:
Hey @kennykerr, is this still a current workaround? Just hit this myself, needing a `IVector<IRandomAccessStream>`.

kennykerr:
Yes, this is still blocked on #81 where I need to add support to implement generic interfaces. Right now, the `implement` macro will only let you implement non-generic interfaces. I hope to get to this soon.

alexjgreig:
Hey @kennykerr , I am trying to create a collection that implements IntoParam<'a, IIterable<HSTRING>>. Is the workaround above still the only solution to this problem? Thanks for the development.

GPeye:
Hey @kennykerr , is this still blocked? #81 is now closed. Granted, it did spin off a few other issues. I just need some way to pass an IIterable of HSTRINGs from Rust to wintrt. I'm hoping a year of progress has yielded at least a work around at this point that doesn't involve C++.

tangp3:
[...]

PR #2339, opened 2023-02-20
Simplify namespace extensions in the `windows-bindgen` crate

kennykerr:
This mostly just changes bindgen to use a per-namespace extension model rather than a per-type model. This aligns more with the way extensions are injected in cppwinrt, which has stood the test of time, and simplifies the upcoming extensions for implementing collections (#91) as part of the larger component authoring support. 

This PR also changes the bindgen project to use the default Rust formatting, which added some noise to the PR, but makes the extension code much more readable and manageable.


riverar:
Does this impact external `gen` users at all? Doesn't look like it but thought I'd double check.

kennykerr:
> Does this impact external `gen` users at all? Doesn't look like it but thought I'd double check.

I don't expect so as it only impacts namespace covered by the `windows`/`windows-sys` crates.

PR #2346, opened 2023-02-23
Stock collection support for `IIterable`

kennykerr:
As part of #91, this update provides the groundwork to support stock collections and introduces `TryFrom` for `IIterable<T>` as a start. Stock implementations of the remaining collection types will follow. In its simplest form, you can write the following:

```rust
let collection = IIterable::<i32>::try_from(vec![1, 2, 3])?;
```

Along with building on the type trait support and `implement` macro improvements in #2343, this update fixes a few things that have become increasingly difficult to work around. Notably, most of the internal implementation details of the `windows` crate are now moved to the `windows::imp` module rather than the `windows::core` module that provides public supporting types. Previously these types and helpers were simply hidden from documentation, but this change makes it more obvious what's an implementation detail to be avoided as opposed to simply missing documentation. Conversely, it will make it easier to figure out what needs documenting.


crates/libs/windows/src/lib.rs
```diff
@@ -10,6 +10,12 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
...
-pub mod core;
+
 pub use Windows::*;
+pub mod core;
+
+#[doc(hidden)]
+mod Windows;
+
+#[doc(hidden)]
+pub mod imp;
```
tim-weis, line 21: Is there a technical reason why this cannot be

```rust
pub(crate) mod imp;
```

?

kennykerr: Yes, there are implementation details that are nevertheless used outside of the `windows` crate because the `windows-bindgen` crate will generate them for non-Windows metadata.

Jamzy01:
Hi, do I need to enable a feature flag in the Cargo.toml file or something to use this feature? Because the rust compiler won't recognize my conversion.

```rs
let collection = IIterable::<i32>::try_from(vec![1, 2, 3])?;
```

returns this error

```rs
error[E0277]: the trait bound `IIterable<i32>: std::convert::From<Vec<{integer}>>` is not satisfied
   --> src\file_system_object\clipboard_handler\mod.rs:280:26
    |
280 |         let collection = IIterable::<i32>::try_from(vec![1, 2, 3])?;
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::convert::From<Vec<{integer}>>` is not implemented for `IIterable<i32>`
    |
    = note: required for `Vec<{integer}>` to implement `Into<IIterable<i32>>`
    = note: required for `IIterable<i32>` to implement `TryFrom<Vec<{integer}>>`
```

kennykerr:
You need the `implement` feature.

Jamzy01:
Hi there, sorry but I'm still struggling to understand how this works, so I'm trying to set the clipboard contents like this:

```rs
let new_clipboard_data = DataPackage::new().unwrap();

let mut clipboard_items = Vec::new();

clipboard_items.push(Some(FileSystemLocation::load_from_string("J:\\Test").unwrap()));

new_clipboard_data
    .SetStorageItems(
        &IIterable::<IStorageItem>
            ::try_from(clipboard_items)
            .expect("Failed to convert clipboard items to IIterable"),
        false
    )
    .expect("Failed to set clipboard storage items");

Clipboard::SetContent(&new_clipboard_data).unwrap_or_else(|err|
    eprintln!("Failed to set clipboard data {err}")
);
```

And ```FileSystemLocation``` implements [IStorageItem_Impl](https://microsoft.github.io/windows-docs-rs/doc/windows/Storage/trait.IStorageItem_Impl.html), except it still gives me this error:

```rs
error[E0277]: the trait bound `windows::Foundation::Collections::IIterable<windows::Storage::IStorageItem>: std::convert::TryFrom<std::vec::Vec<std::option::Option<file_system_location::FileSystemLocation>>>` is not satisfied
   --> src\file_system_object\clipboard_handler\mod.rs:291:18
    |
291 |                   &IIterable::<IStorageItem>
    |  __________________^
292 | |                     ::try_from(clipboard_items)
    | |______________________________^ the trait `std::convert::TryFrom<std::vec::Vec<std::option::Option<file_system_location::FileSystemLocation>>>` is not implemented for `windows::Foundation::Collections::IIterable<windows::Storage::IStorageItem>`
    |
    = help: the trait `std::convert::TryFrom<std::vec::Vec<<T as windows::core::Type<T>>::Default>>` is implemented for `windows::Foundation::Collections::IIterable<T>`
```

I assume passing [IStorageItem](https://microsoft.github.io/windows-docs-rs/doc/windows/Storage/struct.IStorageItem.html)s themselves would work, except from what I can gather, I cannot create these objects myself, besides there must be a reason why IStorageItem_Impl exists anyway. One more thing, are there even any examples of something like this?

Issue #2347, opened 2023-02-23
Bug: CONTEXT structure should be aligned to 16 bytes


TimMisiak:
### Which crate is this about?

windows-sys

### Crate version

0.45.0

### Summary

The following code will fail randomly on x86/x64 because CONTEXT is defined as DECLSPEC_ALIGN(16) in winnt.h, but does not have any alignment annotation in windows-rs.

```
        let hthread = unsafe { OpenThread(THREAD_ALL_ACCESS, FALSE, debug_event.dwThreadId) };

        let mut ctx: CONTEXT = unsafe { std::mem::zeroed() };
        ctx.ContextFlags = CONTEXT_ALL;
        let ret = unsafe { GetThreadContext(hthread, &mut ctx) };
```

Winnt.h:

```
typedef struct DECLSPEC_ALIGN(16) DECLSPEC_NOINITALL _CONTEXT {
```

### Toolchain version/configuration

Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\timmi\.rustup

stable-x86_64-pc-windows-msvc (default)
rustc 1.67.0 (fc594f156 2023-01-24)

### Crate manifest

```toml
[dependencies.windows-sys]
version = "0.45.0"
features = [
    "Win32_Foundation",
    "Win32_Security",
    "Win32_System_Kernel",
    "Win32_System_Threading",
    "Win32_UI_WindowsAndMessaging",
    "Win32_System_Diagnostics_Debug",
    "Win32_System_Environment",
    "Win32_System_WindowsProgramming",
]
```

### Expected behavior

Alignment should be 16 bytes

### Actual behavior

Structure has no explicit alignment, causing random failures when using it with APIs that expect alignment.

kennykerr:
Hi Tim, this is a known issue: https://github.com/microsoft/win32metadata/issues/1044

I hope to bribe someone to fix this very soon. @riverar donuts are on the way.

kennykerr:
I tried working around the issues in #2352 but that didn't work out. Transferring to the Win32 metadata repo, as I don't think there's anything I can do until the metadata has been fixed.

PR #2352, opened 2023-02-24
Improve handling of packing vs alignment

kennykerr:
The `CONTEXT` issue (#2347) has come up numerous times and is still not addressed in the Win32 metadata where this definition originates (https://github.com/microsoft/win32metadata/issues/1044). Rather than waiting indefinitely for that to get fixed, I'm providing a workaround directly inside `windows-rs`. 

Addressing this issue requires a pair of workarounds:

* Add workaround for https://github.com/microsoft/win32metadata/issues/380 where unnecessary packing metadata is present. This is important because [Rust will not allow](https://doc.rust-lang.org/reference/type-layout.html) you to mix packing and alignment information on a struct. 

* Inject alignment information manually gathered from MSVC to properly set the alignment for the `CONTEXT` struct. 

The latter is obviously not scalable and hopefully the Win32 metadata will eventually be able to capture this information correctly as this no doubt affects other structs. 

Fixes: #2347

### DIFF
--- a/crates/libs/bindgen/src/structs.rs
+++ b/crates/libs/bindgen/src/structs.rs
@@ -25,9 +25,16 @@ fn gen_struct_with_name(gen: &Gen, def: TypeDef, struct_name: &str, cfg: &Cfg) -
     let flags = gen.reader.type_def_flags(def);
     let cfg = cfg.union(&gen.reader.type_def_cfg(def, &[]));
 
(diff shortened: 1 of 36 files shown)


kennykerr:
Hmm that's not quite it. The Win32 metadata is just too difficult to work around in its current state since I have to patch the packing and alignment information for certain structs but only for certain architectures, but the metadata incorrectly packs some of these definitions together (e.g. `MINIDUMP_THREAD_CALLBACK`).

PR #2359, opened 2023-03-02
Simplify parameter bindings

kennykerr:
This update builds on #2343 and dramatically simplifies input parameter bindings, somethings that's always been rather complicated as we've attempted to model convertibility in Rust. Rather than depend on thousands of `From` and `TryFrom` implementations, types now implement the marker traits called `CanInto` and `CanTryInto` that do nothing but indicate to the `windows` crate that such types are convertible. The various Windows APIs can then rely on these traits to transform input arguments in the most efficient way possible, either casting directly to a parent interface (in a COM hierarchy) or calling `QueryInterface` when a conversion is expected (in a WinRT hierarchy). 

The end result is a simpler abstraction with far less code. Specifically, it avoids generating around 7,000 `From` and `TryFrom` implementations. That amounts to a lot of functions that the Rust compiler doesn't have to parse, saving on build time.


kennykerr:
Forgot to merge - I'll be back!

Issue #2294, opened 2023-01-19
How to use core::ManuallyDrop


Flakebi:
I just updated to 0.44.0, which introduces `windows::core::ManuallyDrop` and I’m slightly confused on how this is supposed to be used.

The `std::mem::ManuallyDrop` is a container that does not call drop of the inner value, so if I understand it correctly,
```rust
// incorrect use:
{
  let nodrop = std::mem::ManuallyDrop::new(vec![0; 10]);
}
// nodrop is dropped but the contained vector is not, so memory is leaked

// correct use:
{
  let nodrop = std::mem::ManuallyDrop::new(vec![0; 10]);
  let vec: Vec<_> = std::mem::ManuallyDrop::into_inner(nodrop);
}
// vec is dropped so the memory is freed
```

So, using it for unions in the windows api makes a lot of sense for me, if the unions have memory that needs to be dropped, because the compiler has no chance to figure out which variant of the union to drop.

However, 0.44.0 introduce `core::ManuallyDrop`, which – as I currently understand it – has nothing to do with the reason and meaning behind `std::mem::ManuallyDrop`. Instead, it is just a reference (`*mut void` for things that implement `Vtable`).

I was searching for how to call drop on the contained value for a while, like one needs to do on the std ManuallyDrop, but I wasn’t able to find anything like that :)

So, two questions:
1. Why is it called `ManuallyDrop` isn’t `Borrowed` a better name?
2. If a union contains only pointers with no drop functions, can we remove the `std::mem::ManuallyDrop` wrapper?
  (E.g. for `D3D12_RESOURCE_BARRIER_0`, the example never calls `std::mem::ManuallyDrop::into_inner` anyway, which – as far as I understand – one would have to do for a correct use case)

For cross-reference, #1410 is AFAIK the reason for `core::ManuallyDrop` and #2235 is the PR that introduces ManuallyDrop.

kennykerr:
Sorry for the late response. The answer's a bit complicated but I hope to explain the issue here soon. It's definitely something I'd like to improve but Rust is a bit unwieldy in this regard. Anyway, I'll try to summarize the issue here asap.

kennykerr:
Looked at this some more this morning. I might be able to untangle some of the problems that necessitate a custom `ManuallyDrop`. It's all somewhat tied up in the parameter binding support that I have slowly been simplifying but I think more could be done to make this "just work".

Lokathor:
[...]

PR #2361, opened 2023-03-02
Drop `ManuallyDrop`

kennykerr:
Thanks to #2360, I can now get rid of the weird custom `ManuallyDrop` in favor of the standard type. Previously, the parameter binding support depended on a custom `ManuallyDrop` implementation. 

Fixes: #2294


crates/samples/windows/direct3d12/src/main.rs
```diff
@@ -492,7 +492,7 @@ mod d3d12_hello_triangle {
             Flags: D3D12_RESOURCE_BARRIER_FLAG_NONE,
             Anonymous: D3D12_RESOURCE_BARRIER_0 {
                 Transition: std::mem::ManuallyDrop::new(D3D12_RESOURCE_TRANSITION_BARRIER {
-                    pResource: ManuallyDrop::new(resource),
+                    pResource: unsafe { std::mem::transmute_copy(resource) },
```
MarijnS95, line 495: @kennykerr what's the guidance for this? Seems the typical user might go for a safe `ManuallyDrop::new(Some(resource.clone()))`, not immediately realizing that this keeps the refcount on `resource` up, leaking it.

IMO the previous `ManuallyDrop` implementation was clearer in terms of sneakily holding a borrow for which the refcount wasn't increased.

Perhaps a much more invasive change would be having lifetimes on structs: `&'a X` has the same layout as `*const X`/`*mut X`.

kennykerr: Oh, I just saw your example was cloning - yes that would be bad. 😅

I don't know what to say - hopefully developers using Direct3D know enough about COM to realize that would be wrong. Direct3D also has leak detection support so that should help catch bugs like that.

MarijnS95: > So here at least we're acknowledging you need to pay close attention.

You already seem to have figured it out, and I hope that most Rust programmers' alarm bells ring upon seeing `ManuallyDrop::new(x.clone())` (in whatever context), but it is still rather easy to make this mistake as it's the only safe way to write this. Even worse when already having ownership of `x` and not needing to clone at all making it rather invisible (apart needing to call `ManuallyDrop::new()`) :grimacing: 

In my own code I didn't see the leak detector triggering (on resources), only resource overlapping further down the line - might have to dig a bit further.

> I addressed that briefly here: https://github.com/microsoft/windows-rs/issues/2386

Great to have that, even better if it could be part of some obvious/up-front documentation? Or an API that nudges towards this altogether? Not sure how it would look like, or if borrows would be more beneficial in the end.

PR #2365, opened 2023-03-07
Enable hosted lib generation

kennykerr:
The lib workflow now builds and uploads the libs for all targets. This can be used to update those targets without having to build them locally. 

I’ve also added a GitHub action to centralize the environment adjustments needed to get Rust to work properly with the targets inside the GitHub VMs.


kennykerr:
https://github.com/microsoft/windows-rs/pull/2365/commits/01ce49fef4dd6b0d3d0faba590c3a33b6ec6d825 just calls an API that's recently been added and isn't covered by the stale GNU libs. Once that failure is confirmed I can push the libs generated by the hosted build.

kennykerr:
It's not finding the `llvm-dlltool` in the path. Not sure where to find that. 🤔

riverar:
@kennykerr That's in `C:\msys64\mingw64\bin` on the runners. The environment steps put that into `GITHUB_PATH` but I'm not sure if that works across the new workflows you created.

riverar:
Looks like updating path via composite actions is indeed not really supported https://github.com/actions/runner/issues/789.

This aligns with my experience with all of GitHub's shiny new workflow tools/modularity efforts. They are so enticing on paper but when you try to actually use them... 😊

PR #2377, opened 2023-03-12
Check diff of all targets, not just baseline

kennykerr:
Building on #2365, this should now ensure we have updated libs at all times, much like the gen workflow ensures code gen is always updated.


kennykerr:
Scrapping this for now as lib generation is still not reproducible.

PR #2402, opened 2023-03-28
Restore reproducible libs

riverar:


crates/tools/msvc/src/main.rs
```diff
@@ -67,7 +67,7 @@ fn main() {
             break;
         }
     }
-
+    archive.flush();
```
kennykerr, line 70: May want to call `drop(archive)` right after this line to ensure the file is closed after flushing and before copying.

kennykerr: I think the gnu tool does call `drop` but it fails to call `flush`.

kennykerr (dismissed):

riverar:
Investigating gnu lib reproducibility issue, see if we can get two birds with one PR here.

riverar:
OK @kennykerr looking good. Manually verified the msvc/gnu libs against the workflow ones and they match now. We'll revisit and hopefully clean all this up soon.

kennykerr (dismissed):

kennykerr:
A MacOS concurrency limit is holding things up - since everything else has passed I'm going to cancel the run and complete the PR to relieve some pressure.

PR #2404, opened 2023-03-29
Check diff for all targets

kennykerr:
Now that #2402 has been fixed, here's hoping we can enable diff checks for all targets.


riverar:
So msvc is failing because unbeknownst to me, the linker is embedding its version number into the output. `14.35.32215` becomes `0E 00 | 23 00 | D7 7D`. Let me go spelunking in the linker.

image

riverar:
Looks like it emits the version string in `link.exe!CImplib::EmitImportDescriptor`. Doesn't look like it's tied into the existing undocumented `/Brepro`; that would have been too easy.

PR #2405, opened 2023-03-29
Use safe DLL loading

kennykerr:
Restricts DLL search directories for safety.

This update corresponds to the following change: https://github.com/microsoft/cppwinrt/pull/1293


crates/libs/windows/src/imp/delay_load.rs
```diff
@@ -8,7 +8,7 @@ use super::*;
 ///
 /// * Both the library and function names must be valid null-terminated strings.
 pub unsafe fn delay_load<T>(library: core::PCSTR, function: core::PCSTR) -> Option<T> {
-    let library = LoadLibraryA(library);
+    let library = LoadLibraryExA(library, 0, LOAD_LIBRARY_SEARCH_DEFAULT_DIRS);
```
riverar, line 11: Would it be better to use `LOAD_LIBRARY_SEARCH_SYSTEM32` here? Seems all internal use of delay_load only uses modules from \system32 context.

kennykerr: No, both cppwinrt and the windows crate support xcopy activation for WinRT.

riverar (dismissed):

kennykerr:
This test fails because Cargo doesn't provide a reliable way to determine where output binaries are placed.

crates/libs/windows/src/imp/delay_load.rs
Eyryu: @kennykerr

Issue #2409, opened 2023-03-30
Feature request: `windows_bindgen::standalone` should allow specifying specific enum constants


Jake-Shadle:
### Motivation

Currently, `windows_bindgen::standalone` will emit _every_ constant in an enum with no way to filter to only the ones that are needed by the crate using the bindings. As a consequence `#![allow(dead_code)]` is always emitted in the generated bindings so the user doesn't get warnings for the unused constants.

### Drawbacks

In a _vast majority_ of use cases, only a small fraction of enum values are used from any particular enum, from the relatively tame [`PAGE_`](https://learn.microsoft.com/en-us/windows/win32/memory/memory-protection-constants) flags enum to the frankly insane (at the time of this writing) 3240 item [`WIN32_ERROR`](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes). While it's not the end of the world, refining the emitted items to only the ones requested by the user means less generated code and thus less code to compile.

### Rationale and alternatives

The impact of not doing this is that crates which generate enum bindings with `windows_bindgen::standalone` will almost always be compiling code they don't use. This cost will be negligible in the grand scheme of things, but there's also no reason to give the compiler more work to do if it can be avoided. It also means that if the user manually removes the items they don't use, regenerating the bindings will need manual cleanup every time rather than just having a feature to always emit bindings without a need for manual fixup.

kennykerr:
This one's a little trickier but agreed it would be nice if this were possible.

kennykerr:
And here's the fix for this issue: #2430

PR #2396, opened 2023-03-28
Support standalone code generation

kennykerr:
Even with a choice between the `windows` and `windows-sys` crates, some developers may prefer to use completely standalone bindings. This update introduces the `standalone` function to the `windows-bindgen` crate that provides the ability to generate entirely standalone bindings for Windows APIs with a single function call that you can run from a test to automate the generation of bindings. This can help to reduce your dependencies while continuing to provide a sustainable path forward for any future API requirements you might have, or just to refresh your bindings from time to time to pick up any bug fixes automatically from Microsoft.

Here's an example of how you might use it. Here's what your project's Cargo.toml file might look like:

```
[dependencies.windows-targets]
version = "0.42"

[dev-dependencies]
windows-bindgen = "vnext"
```

The `windows-targets` crate is a lib-only crate that provides import libs for supported targets. This isn't strictly necessary but simplifies importing Windows API functions. `raw-dylib` may also be used to avoid this dependency. The `windows-bindgen` crate is only needed for generating bindings and is thus a dev dependency only.

Given this dev dependency, we can now write a test to generate the bindings as follows:

```rust
#[test]
fn gen_bindings() {
    use windows_bindgen::standalone;

    let apis = [
        "Windows.Win32.System.SystemInformation.GetTickCount",
    ];

    let bindings = standalone(&apis);
    std::fs::write("src/bindings.rs", bindings).unwrap();
}
```

This will generate bindings for the `GetTickCount` function. Running the test will result in the following being generated and written to the `src/bindings.rs` file:

```rust
#![allow(
    non_snake_case,
    non_upper_case_globals,
    non_camel_case_types,
    dead_code,
    clippy::all
)]
pub type PSTR = *mut u8;
pub type PWSTR = *mut u16;
pub type PCSTR = *const u8;
pub type PCWSTR = *const u16;
#[link(name = "windows")]
extern "system" {
    pub fn GetTickCount() -> u32;
}
```

The crate can now make use of these standalone bindings as follows:

```rust
mod bindings;
use bindings::*;

fn main() {
    unsafe {
        println!("{}", GetTickCount());
    }
}
```

Let's now add a call to the `MessageBoxA` function, as a slightly more interesting example. Update the `apis`  array in the test as follows:

```rust
    let apis = [
        "Windows.Win32.System.SystemInformation.GetTickCount",

        "Windows.Win32.UI.WindowsAndMessaging.MessageBoxA",
        "Windows.Win32.Foundation.HWND",
[...]

PR #2430, opened 2023-04-05
Produce conservative enum definitions in standalone code generation

kennykerr:
Unscoped enumeration types are now generated without generating all of the enumerator constants. Practically, this means that if you directly or indirectly refer to a type like `WIN32_ERROR` - with its many constants - that you will only get the `WIN32_ERROR` type itself and will need to request any constants directly. This can greatly reduce the overall size of the code produced with standalone code generation (#2396).

Fixes: #2409


crates/libs/bindgen/src/standalone.rs
```diff
@@ -19,6 +19,7 @@ pub fn standalone(names: &[&str]) -> String {
 
     let mut type_names = BTreeSet::new();
     let mut core_types = BTreeSet::new();
+    let mut enums = BTreeMap::new();
```
ChrisDenton, line 22: @kennykerr I think you might have meant this to be a `Vec`? Otherwise this will overwrite the entry instead of adding to it:
```rust
enums.insert(enum_name, type_name.name);
```

kennykerr: Ah, good catch - will fix asap.

kennykerr: Here's the fix: #2434 - I use sorted containers to ensure that the generated code is reproducible.

ChrisDenton: Great, thanks!

Issue #1447, opened 2022-01-23
Bug: Application is linked to winspool.dll instead of winspool.drv


ancwrd1:
### Which crate is this about?

windows

### Crate version

master branch

### Summary

Attempt to use any winspool functions (e.g. `SetDefaultPrinterW`) results in application linking to `winspool.dll` instead of `winspool.drv` and it fails to start.

### Expected behavior

The `winspool.drv` DLL name must be used for imported functions.

### Actual behavior

The `winspool.dll` DLL name is used for imported functions which is missing.

kennykerr:
@riverar I thought you already opened a bug for this on win32metadata but I can't find it.

riverar:
Related to https://github.com/microsoft/win32metadata/issues/714

PR #2431, opened 2023-04-05
Strip `.dll` extension from raw-dylib link names

Jake-Shadle:
As noted in the comment, currently `windows-bindgen` emits the full name of the dll including the extension, which will cause runtime failures since somewhere (linker? dynamic loader?) the extension is automatically added, causing the dll not to be found and abort.

I think the confusion may come from the [RFC](https://rust-lang.github.io/rfcs/2627-raw-dylib-kind.html) containing example code which _does_ include the extension.

```rust
#[cfg(windows)]
#[link(name = "kernel32.dll", kind = "raw-dylib")]
#[allow(non_snake_case)]
extern "system" {
    fn GetStdHandle(nStdHandle: u32) -> *mut u8;
}
```

Note this doesn't impact `bthprops.cpl` which is the 1? case I could find of a dynamic lib having an extension _other_ than .dll.

I was opening an issue and realized it was easier to just fix it.

### DIFF
--- a/crates/libs/bindgen/src/functions.rs
+++ b/crates/libs/bindgen/src/functions.rs
@@ -58,6 +58,10 @@ fn gen_link<P: IntoIterator<Item = TokenStream>>(
     for param in params {
         tokens.push_str(&format!("{}, ", param.as_str()));
     }
+
+    // Strip the dll suffix, including it will cause runtime failures as the
+    // dynamic loader will look for eg. "kernel32.dll.dll"
+    let link = link.strip_suffix(".dll").unwrap_or(link);
     let tokens = tokens.trim_end_matches(", ");
     format!(
         "::windows_targets::link!(\"{link}\" \"{abi}\"{doc} fn {name}({tokens}) {return_type});"


riverar:
> Note this doesn't impact bthprops.cpl which is the 1? case I could find of a dynamic lib having an extension other than .dll.

There are plenty of other cases, see:
* https://github.com/microsoft/win32metadata/issues/714
* https://github.com/microsoft/windows-rs/issues/1447
* https://github.com/microsoft/win32metadata/issues/1021
* etc.

It's not optimal or really possible to guess the correct answer so we make it explicit in metadata.

Do you have a repro that demonstrates a failure? If so, please open an issue. Be aware raw-dylib is in use by rustc today.

ChrisDenton:
I think the issue here is that it needs the `+verbatim` modifier.

```diff
#[cfg(windows)]
-#[link(name = "kernel32.dll", kind = "raw-dylib")]
+#[link(name = "kernel32.dll", kind = "raw-dylib", modifiers = "+verbatim")]
#[allow(non_snake_case)]
extern "system" {
    fn GetStdHandle(nStdHandle: u32) -> *mut u8;
}
```

kennykerr:
This is a complicated issue, but it has been discussed at length here: 

https://github.com/rust-lang/rust/issues/58713

The `windows-targets` crate support for `raw-dylib` can be found here: 

https://github.com/microsoft/windows-rs/blob/master/crates/libs/targets/src/lib.rs

And it is tested here:

https://github.com/microsoft/windows-rs/blob/master/.github/workflows/raw_dylib.yml

PR #2446, opened 2023-04-14
Update Win32 v49 and WDK v4

kennykerr:
This updates both Win32 and WDK metadata, bringing in a large number of additional WDK APIs.


kennykerr:
Still a remaining issue blocking this update: https://github.com/microsoft/wdkmetadata/issues/31

Issue #2443, opened 2023-04-13
Why does windows-targets have compiled static libraries?


rodrigocfd:
This issue is just a question.

The windows crate depends on [windows-targets](https://crates.io/crates/windows-targets) crate for [linker support](https://github.com/Amanieu/parking_lot/pull/378#issue-1660949224). And windows-targets depends on [internal crates](https://github.com/microsoft/windows-rs/tree/master/crates/targets) which vary according to the platform. These crates have compiled static libraries like [this one](https://github.com/microsoft/windows-rs/tree/master/crates/targets/x86_64_msvc/lib), which are passed to the linker via `cargo:rustc-link-search=native={LIB_DIR}`.

Now I'm wondering:

1. Why are those static libraries necessary?
2. Do they contain only the compiled code from [baseline](https://github.com/microsoft/windows-rs/tree/master/crates/targets/baseline) folder?

I'm curious because the program below, with zero dependencies, works fine:

```rust
#[link(name = "user32")]
extern "system" {
	fn MessageBoxW(
		_: *mut std::ffi::c_void,
		_: *const u16,
		_: *const u16,
		_: u32,
	) -> i32;
}

fn main() {
	unsafe {
		MessageBoxW(
			std::ptr::null_mut(),
			"Text".encode_utf16()
				.chain(std::iter::once(0))
				.collect::<Vec<_>>()
				.as_ptr(),
			"Caption".encode_utf16()
				.chain(std::iter::once(0))
				.collect::<Vec<_>>()
				.as_ptr(),
			0,
		);
	}
}
```

riverar:
Hi @rodrigocfd, those are [import libraries](https://learn.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-creation#creating-an-import-library) not static libraries.

The baseline folder is used to detect import library differences--it's less cumbersome than trying to diff a binary blob. See also https://github.com/microsoft/windows-rs/pull/2132.

Closing as there's nothing actionable here but feel free to keep the discussion going/ask any follow up questions.

ChrisDenton:
To put it another way, your example has a dependency:

```rust
#[link(name = "user32")]
```

You're depending on the target specific import library "user32.lib". These libraries are usually managed by the Visual Studio or the Windows SDK installer. Rust is somewhat "smart" and will attempt to locate these kinds of dependencies for you.

[...]

PR #2463, opened 2023-04-20
Allows custom windows-targets function visibility.

rodrigocfd:
Currently the `link!` macro only produces functions with `pub` visibility. I'm looking forward to add `windows_targets` as a dependency to [`winsafe`](https://crates.io/crates/winsafe) – as proposed [here](https://github.com/rodrigocfd/winsafe/pull/75) and clarified [here](https://github.com/microsoft/windows-rs/issues/2443) – and being able to specify the function visibility would be handy.

### DIFF
--- a/crates/libs/targets/src/lib.rs
+++ b/crates/libs/targets/src/lib.rs
@@ -8,12 +8,12 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 #[macro_export]
 #[doc(hidden)]
 macro_rules! link {
-    ($library:literal $abi:literal $($link_name:literal)? $(#[$doc:meta])? fn $($function:tt)*) => (
+    ($library:literal $abi:literal $($link_name:literal)? $(#[$doc:meta])? $vis:vis fn $($function:tt)*) => (
         #[link(name = $library, kind = "raw-dylib", modifiers = "+verbatim", import_name_type = "undecorated")]
         extern $abi {
             $(#[$doc])?
             $(#[link_name=$link_name])?
-            pub fn $($function)*;
+            $vis fn $($function)*;
         }
     )
 }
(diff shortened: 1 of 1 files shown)


rodrigocfd:

rodrigocfd:
> Please open an issue. This change alone will not work as the other crates that depend on `windows-targets` would need to agree on how they use the `link` macro.

The `vis` attribute [can be empty](https://doc.rust-lang.org/reference/macros-by-example.html), so no changes would be needed on the user's side.

kennykerr:
Fair enough, I'll let the build validate.

kennykerr:
> The `vis` attribute [can be empty](https://doc.rust-lang.org/reference/macros-by-example.html), so no changes would be needed 

I think the problem is going to be that existing code expects it to be public.

kennykerr:
You should be able to use the `link` macro within a module, typically called `bindings`, and then control the visibility from the outer module. That should give you what you need without changes to the `link` macro.

rodrigocfd:
> You should be able to use the `link` macro within a module, typically called `bindings`, and then control the visibility from the outer module. That should give you what you need without changes to the `link` macro.

That's exactly how I'm using it today, I confined all declarations whithin a module, so they won't escape to public visibility. I saw the change as a good opportunity to make it more strictly correct.

However we have a problem now. If "no visibility" is informed (empty `$vis`), Rust sees as "module private" visibility. However, the current `link!` implementation interprets "no visibility" as `pub`, because `$vis` is never informed today (it simply doesn't exist).

So adding `$vis` to the `link!` macro would fundamentally change the `link!` signature, and all users would be forced to update its usage, because everything would be private by default. It would be fairly easy to catch though – the compiler would immediately inform that the funcions are private.

But is such a change acceptable?

kennykerr:
I'm not opposed to it in theory, but it would be a breaking change and there just doesn't seem to be enough reason to warrant that right now.

rodrigocfd:
I agree. Since the visibility can me mitigated by wrapping the declarations inside a module, a breaking change doesn't seem reasonable by now.

If needed in the future, however, the seed has been planted.

kennykerr:
Yep, I wish I thought of it sooner. 😜 Thanks for the suggestion!

PR #2466, opened 2023-04-21
Windows Firewall Com Sample

Skwerl23:
added samples for the firewall, 
This took some effort to figure out, so hopefully it helps some one else

What's this all about?

Fixes: #0000 ⬅️ Be sure to refer to an existing issue here!
- no fixes, only a sample of the firewall api/com object. It's not straight forward and might help some one out!


Skwerl23:

kennykerr:
Thanks for sharing! We already have a number of COM-related samples, so I'd prefer to leave it as is.

WofWca:
Thank you!
I'm not familiar with Windows API or this crate, and I was looking for exactly this.

Issue #2467, opened 2023-04-21
Feature request: Win32 BOOL+GetLastError() mapping to Result


talagrand:
### Motivation

It's very helpful that both HRESULT/NTSTATUS/WIN32_ERROR-returning functions are mapped to returning a Result so standard error-handling conventions can be used against them. Request the same treatment for Win32  BOOL + GetLastError()-convention functions as well.

### Drawbacks

Performance? Though there's windows-sys and HRESULTs call the error info APIs.

kennykerr:
Last time I checked there were a few issues with the Win32 metadata that prevented this transformation, but it's been a while so I'll check again and hopefully we can turn it on.

PR #2476, opened 2023-04-25
Transform `BOOL` and `Win32` return values

kennykerr:
Currently, the `windows` crate will transform API functions that return `HRESULT` or `NTSTATUS` into `Result<T>` to provide a more idiomatic developer experience. 

This update extends that transformation to functions that return `WIN32_ERROR`, like `RegOpenKeyExW`, as well as functions that return `BOOL` and also have metadata that indicates they call `SetLastError` on failure, like `LookupPrivilegeNameW`. 

Fixes: #2467

### DIFF
--- a/crates/libs/metadata/src/reader/mod.rs
+++ b/crates/libs/metadata/src/reader/mod.rs
@@ -602,6 +602,13 @@ impl<'a> Reader<'a> {
         };
         self.module_ref_name(self.impl_map_scope(impl_map)).to_lowercase()
     }
+    pub fn method_def_last_error(&self, row: MethodDef) -> bool {
+        if let Some(map) = self.method_def_impl_map(row) {
+            self.impl_map_flags(map).contains(PInvokeAttributes::LAST_ERROR)
+        } else {
+            false
+        }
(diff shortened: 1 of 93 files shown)


kennykerr:
CC @talagrand and @riverar if you'd like to spot check.

talagrand:
Thank you @kennykerr! I spot checked all the APIs I was using and they're all covered by this PR.

Now, I can open a separate issue for this but it would be ideal if APIs that use the pattern *sentinel* + GetLastError() were also covered. Things like ExpandEnvironmentStringsW (returns 0 on failure) or ResumeThread (returns -1).

riverar:
> Now, I can open a separate issue for this but it would be ideal if APIs that use the pattern sentinel + GetLastError() were also covered. Things like ExpandEnvironmentStringsW (returns 0 on failure) or ResumeThread (returns -1) or GlobalLock (returns NULL).

This will require additional metadata support; do file an issue at https://github.com/microsoft/win32metadata/issues thanks!

riverar (approved):

Issue #2479, opened 2023-04-27
Bug: Nested types not crawled when gather types to emit


Jake-Shadle:
### Which crate is this about?

other (please share in the comments)

### Crate version

windows-bindgen

### Summary

When crawling through types to emit based on the initial set requested by the user, nested records were not being crawled, which would mean any types unique to that nested record would not get emitted resulting in compile errors.

### Expected behavior

I expected the full definitions of every type needed for any root type requested by the user.

### Actual behavior

Missing type definitions for fields in nested records.

Issue #2478, opened 2023-04-27
Bug: Type aliases for interfaces not publicly exposed


Jake-Shadle:
### Which crate is this about?

other (please share in the comments)

### Crate version

bindgen 0.49.0

### Summary

When using `windows_bindgen::standalone_sys` any COM interfaces encountered are emitted as `type ...` instead of being `pub`.

### Expected behavior

The items should be pub like everything else emitted by `standalone_sys`

### Actual behavior

The items are not pub.

PR #2480, opened 2023-04-27
Include nested dependencies in standalone code generation

Jake-Shadle:
This PR fixes 2 issues:

- When emitting COM interfaces with `standalone_sys` the type aliases were not pub like everything else.
- Nested types were not being crawled when gathering the types, which would result in fields in nested records that used types that were unique to them to not be emitted

Fixes: #2478
Fixes: #2479


kennykerr:
Thanks for the fix! Just need to rerun `cargo run -p tool_core` to regenerate the internal bindings. 

> When emitting COM interfaces with standalone_sys the type aliases were not pub like everything else.

The trouble here is that this may introduce name collisions. I'm not opposed to changing it, but we'd need to be a bit more explicit in `windows-core` about how we refer to those bindings.

Jake-Shadle:
Oof, yah, that broke everything.

kennykerr:
I can fix that up in a separate PR if you'd prefer. Let me know.

kennykerr:
Sounds good - I can find a prettier solution separately.

PR #2485, opened 2023-04-28
Split out type crawling

Jake-Shadle:
This PR splits out type crawling into a separate function and makes `standalone_imp` public so that the user can do the type crawling in a separate pass before emitting them via `standalone_imp`.

Initially I only did this so that the user can query the set of types that will be emitted, but I then encountered the bug described in microsoft/win32metadata#1554 and rolled in a fix that takes advantage of the split out type crawling, while preserving the current behavior of the other `standalone_` functions.

Since I found this one `VK_F` case, I have to assume there are others given the scope of the Win32 API, but I wasn't sure if there were cases where classes or interfaces could clash with functions/constants thus I only have the 3 different variants.

Another possible option would be more narrowly focused on enum constants, where the enum's name could be followed by `::<enum variant>` rather than just `.<enum variant>` as it is now, but I wasn't sure if there might be other cases where a clash could occur between eg. a struct and a function which that fix would not address.

Fixes: microsoft/win32metadata#1554

### DIFF
--- a/crates/libs/bindgen/src/standalone.rs
+++ b/crates/libs/bindgen/src/standalone.rs
@@ -5,7 +5,9 @@ pub fn standalone_win(names: &[&str]) -> String {
     let files = &File::with_default(&[]).unwrap();
     let reader = &Reader::new(files);
     let gen = &mut Gen::new(reader);
-    standalone_imp(gen, names)
+
+    let items = gather_items(gen, names.iter().map(|s| (*s, None)));
+    standalone_imp(gen, items)
 }
 
 /// Generates standalone Windows bindings based on the `windows-sys` crate's bindings.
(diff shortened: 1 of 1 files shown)


kennykerr:
Thanks, but I'd prefer not to complicate the standalone code generation's interface if we can help it. Hopefully https://github.com/microsoft/win32metadata/issues/1554 can be addressed and make this unecessary. Otherwise, I'd like to find an automatic way to handle such collisions so that it "just works".

kennykerr:
Since this should be a rare issue, I simple workaround could be just to emit all overloads. Rust will be happy with that. What do you think?

Jake-Shadle:
That'd be fine I suppose, they can just be manually edited to remove the unneeded items in those rare cases.

kennykerr:
Here you go: #2486

PR #2495, opened 2023-05-09
Collect all typedefs with same name

Jake-Shadle:
This fixes an issue introduced in #2475 where all typedefs with the same name are not collected if they aren't explicitly requested by the user.

Let's take `Windows.Win32.System.Diagnostics.Debug.RtlCaptureContext` as an example.

If we request standalone generation of this function, the code in `standalone_imp` will find the function, iterate through its parameters, and collect the `CONTEXT` typedef, however, it will only collect the _first_ one that it finds, which is the `aarch64` version in this particular case, causing the `x86` and `x86_64` `CONTEXT` typedefs (and their unique nested/referenced typedefs) to not be collected. Previously this worked in 0.49.0 because the top level type collection code used `signature_cfg` to gather all of the parameter/return types.

Resolves: #2496


kennykerr:
Thanks for the contribution - good catch. As a reminder, in future please open an issue with a minimal repro before creating a PR. That just makes it easier for me to isolate and identify the necessary change and track a fix.

crates/tests/standalone/build.rs
```diff
@@ -44,6 +44,7 @@ fn main() {
         &[
             "Windows.Win32.Networking.WinSock.WSADATA",
             "Windows.Win32.UI.WindowsAndMessaging.GetWindowLongPtrW",
+            "Windows.Win32.System.Diagnostics.Debug.RtlCaptureContext",
```
kennykerr, line 47: Let put this in a new test module, say `b_arch_dependency`. That way we can easily validate that the change doesn't perturb anything else.

crates/tests/standalone/build.rs
```diff
@@ -47,6 +47,14 @@ fn main() {
         ],
     );
 
+    // Ensures that root items that reference architecture independent types
+    // of the same name correctly collect all of those types rather than just
+    // the first one
+    write_sys(
+        "src/b_arch_dependencies.rs",
```
kennykerr, line 54: Be sure to include module in compilation e.g. `mod b_arch_dependencies;` in `lib.rs`.

kennykerr:
Looks like you're hitting some new rustup changes. I'll quickly get those addressed in a separate PR to get this unblocked.

kennykerr:
Apologies, I had to step away for 2 days - I have now completed #2497 so you should be able to merge master and get this PR building.

kennykerr (approved):

Issue #2499, opened 2023-05-10
`standalone_sys` should support string constants


Jake-Shadle:
### Which crate is this about?

windows-bindgen

### Crate version

0.49.0

### Summary

When binding string constants there are 2 issues.

1. String constants are `Type::String`, which gets resolved to `HSTRING`, and while it's unlikely, if you don't bind another item that pulls in `PCSTR` or `PCWSTR` (the string types that `Type::String` constants actually get resolved to) then there will be a missing typedef.
2. String constants unconditionally assume that the `s!` and `w!` macros are available.

### Expected behavior

The emitted string constants should compile correctly.

### Actual behavior

The emitted constants will fail to compile due to the unknown `s!` or `w!` macro(s), and possibly due to missing the type aliases for `PCSTR` and/or `PCWSTR`.

kennykerr:
Much has changed in `windows-bindgen` since this issue was opened so I'm not sure whether its still relevant. A major update was recently published that you can read about here: #3359

Take it for a spin and let me know if you have any feedback.

PR #2500, opened 2023-05-10
Fix string constants for standalone sys

Jake-Shadle:
This change fixes two issues described in #2499.

1. String constants are (currently) specially handled to resolve to `PCSTR` or `PCWSTR` based on the constant field, but this was not reflected in type collection. The standalone type collection now ignores string constants since it knows they are special, so `HSTRING` is no longer added if not used elsewhere.
2. String constant emission unconditionally emits `s!` or `w!`  to create the string literal, but in standalone + sys mode this means the macros are not prefixed with the crate, the only special case in the current bindgen for the crate name. Now if in standalone + sys mode it assumes `windows_core` is used and emits that.

The other way to fix issue 2 is to get rid of the macros altogether in favor of just emitting either a byte string (with null terminator) for ansi, or a precomputed u16 array (with null terminator) directly in the bindings. The byte string would be as readable as the macro is today, and it would be trivial to add a comment on top of the u16 array to say the human readable string that it represents. But I did the simplest fix for now.

Resolves: #2499


crates/libs/bindgen/src/constants.rs
```diff
@@ -12,7 +12,11 @@ pub fn gen(gen: &Gen, def: Field) -> TokenStream {
 
         if ty == constant_type {
             if ty == Type::String {
-                let crate_name = gen.crate_name();
+                let crate_name: TokenStream = if gen.sys && !gen.standalone {
```
kennykerr, line 15: This is a bit problematic since this assumes that sys-style code gen should assume a dependency on `windows-core` which is designed for non-sys style code gen. I'd like to avoid adding a "core" crate for `windows-sys` but this conflates the two.

crates/libs/bindgen/src/standalone.rs
```diff
@@ -67,8 +67,15 @@ fn standalone_imp(gen: &mut Gen, names: &[&str]) -> String {
...
             constants.insert(field);
-            gen.reader
-                .type_collect_standalone(&gen.reader.field_type(field, None), &mut types);
+
+            let ft = gen.reader.field_type(field, None);
+
+            // String normally gets resolved to HSTRING, but since macros
+            // from windows_core/windows_sys::core are used to create the
+            // string literals we can skip the type collection
+            if ft != Type::String {
```
kennykerr, line 76: This seems like a hack. String constants should not be depending on `HSTRING`.

kennykerr:
I think the main issue is support for string literals in standalone code generation. I'll think some more on the best way to handle those in sys-style code generation. It should work fine for non-sys code generation.

crates/libs/bindgen/src/constants.rs
kennykerr: Yep, it's not a simple fix but I'd prefer to find a good solution rather than fix a problem by introducing another problem. 😉

Jake-Shadle: ok

PR #2502, opened 2023-05-11
Update Win32 v51 and WDK v5

kennykerr:
This updates both Win32 and WDK metadata.


kennykerr:
Hmm, looks like I need to handle the new `MetadataTypedefAttribute` attribute before I can support this metadata update.

kennykerr:
For now, I'll just treat `MetadataTypedefAttribute` as synonymous with `NativeTypedefAttribute` as there's no other way to handle this. 

https://github.com/microsoft/win32metadata/issues/1575

kennykerr:
More issues with metadata - leaving the update for now.

Issue #1093, opened 2021-08-26
Support metadata generation


kennykerr:
Dedicating an issue to this topic. Originally part of #81 (huge thread).

kennykerr:
Getting back to authoring (#1094) and metadata generation as originally described in #81 I'd like to support a `define` macro that would take the place of using MIDL and instead let you define your metadata directly in Rust as follows:

```rust
fn main() {
    define! {
        mod Microsoft {
            mod Windows {
                struct StructType { x: i32, y: i32 }

                interface IInterfaceType {
                    fn Method(&self) -> StructType;
                }

                class ClassType : IInterfaceType;
            }
        }
    }
}
```

The `define` macro would typically be called by a build script to generate the component or project's .`winmd` file to `OUT_DIR/project.winmd` and a corresponding set of Rust bindings for the implementation to use to `OUT_DIR/project.rs`. The project can then use that along with the `implement` macro to provide the implementation.

Alovchin91:
MIDL 3.0 is pretty simple, very well documented and is rather powerful. Do you think it would make sense to support both ways of defining runtime classes (so also in an .idl file)?

kennykerr:
MIDL also has a lot of problems. Of course, nothing prevents you from using MIDL if you prefer. Fundamentally, the winmd format is the canonical representation for metadata. How you get to that is up to you. I'm just working on a way to do it organically in Rust so that MIDL would not be required, but first I'm working out a few minor issues to make sure that developers can start using MIDL in the short term. Basically, the same issues you run into in your sample project and just making sure is streamlined.

hmyan90:
Thanks, Kenny for putting this together and super helpful to provide a validated short-term solution. For MIDL define, personally I also don't mind defining it, it can particularly serve as a document purpose that can share with the team who will consume it, we can add notes/comments on it. But if we define in Rust macro, it's hard to share it with another team, I think. But sure we can choose the option!

mominshaikhdevs:
@kennykerr is the work on the "complete metadata generator" aka "the riddle tool" complete yet? or is it in backlog?

PR #2510, opened 2023-05-18
Update to riddle and metadata generation

kennykerr:
Here's another update as I continue to work toward #1093. 

Metadata generation requires a stricter metadata reader that more accurately captures type information in a lossless manner rather than just capturing what I happen to need for the `windows` crate. To that end, this update brings another batch of small improvements to the reader. 

I have also switched the `windows-metadata` crate to use standalone code generation for its internal definitions to ease maintenance and improve clarity. Many numeric literals have been replaced by generated constants. 

The standalone code generation functions will now assert if they can't find the requested type, constant, or function.

Finally, riddle and metadata generation still have a long way to go but this update is the first to support pretty rich metadata generation of both WinRT and Win32 metadata. It's not quite ready to kick the tires but it's getting close! 

This update also deals with a few new warnings raised by the latest Rust toolchain updates.


crates/libs/metadata/src/error.rs
```diff
@@ -0,0 +1,84 @@
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Default, Debug)]
+pub struct Error {
```
MarijnS95, line 4: `std::error::Error` is not implemented for this type, are we supposed to convert it to `std::io::Error` (and loose the span info)?

This for example prevents using the `?` operator to bubble the error up in an `anyhow::Result` chain.

Issue #2521, opened 2023-05-27
Error in sample


MartinAskestad:
The EnumWindows sample contains an error. It doesn't return an Ok(()).

kennykerr:
See https://github.com/microsoft/windows-rs/pull/2522#pullrequestreview-1447304831

PR #2522, opened 2023-05-27
Fix sample

MartinAskestad:
Fix the missing Ok(())

Fixes: #2521⬅️ Be sure to refer to an existing issue here!

### DIFF
--- a/crates/samples/windows/enum_windows/src/main.rs
+++ b/crates/samples/windows/enum_windows/src/main.rs
@@ -1,7 +1,8 @@
 use windows::{Win32::Foundation::*, Win32::UI::WindowsAndMessaging::*};
 
 fn main() -> windows::core::Result<()> {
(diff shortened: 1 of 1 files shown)


tim-weis (commented):
Samples are implemented to compile with the latest (potentially not yet released) versions of the crates. The signature for [`EnumWindows`](https://github.com/microsoft/windows-rs/blob/72e9bf4b6790fd68ab0aba7eb4482ddd238a8d02/crates/libs/windows/src/Windows/Win32/UI/WindowsAndMessaging/mod.rs#L1075) has changed from returning a `BOOL` to returning a `windows_core::Result`, meaning that the code is correct already.

The proposed changes aren't required, and indeed cause a regression: The result of calling `EnumWindows` is now ignored (which hopefully triggers a clippy warning).

PR #2539, opened 2023-06-14
Implement `type_name()` for `TypeRef`

MarijnS95:
When running `tool_windows` on the latest `.winmd` generated with https://github.com/microsoft/win32metadata/pull/1591 to test the results of that metadata change, no implementation exists for `TypeRef`.  This implementation is rather trivial/straightforward based on `TypeDef`.

Going forward, also print the failed value when a `match` statement ends up in `_ => unimplemented!()` so that it is quicker to immediately debug what enum variant needs to be implemented.

### DIFF
--- a/crates/libs/bindgen/src/gen.rs
+++ b/crates/libs/bindgen/src/gen.rs
@@ -60,6 +60,29 @@ impl<'a> Gen<'a> {
     }
 
     //
+    // TypeRef
+    //
+
+    pub fn type_ref_name(&self, ref_: TypeRef) -> TokenStream {
+        self.type_ref_name_imp(ref_, "")
+    }
+    pub fn type_ref_vtbl_name(&self, ref_: TypeRef) -> TokenStream {
+        self.type_ref_name_imp(ref_, "_Vtbl")
+    }
+    pub fn type_ref_name_imp(&self, ref_: TypeRef, suffix: &str) -> TokenStream {
+        let type_name = self.reader.type_ref_type_name(ref_);
+
+        assert!(!type_name.namespace.is_empty());
+
+        let mut namespace = self.namespace(type_name.namespace);
+        let mut name = to_ident(type_name.name);
+        name.push_str(suffix);
+
+        namespace.combine(&name);
+        namespace
+    }
+
+    //
     // Type
     //
 
@@ -156,13 +179,14 @@ impl<'a> Gen<'a> {
             Type::WinrtArray(ty) => self.type_name(ty),
             Type::WinrtArrayRef(ty) => self.type_name(ty),
             Type::ConstRef(ty) => self.type_name(ty),
-            _ => unimplemented!(),
+            Type::TypeRef(TypeDefOrRef::TypeRef(ref_)) => self.type_ref_name(*ref_),
+            x => unimplemented!("{:?}", x),
         }
     }
     pub fn type_vtbl_name(&self, ty: &Type) -> TokenStream {
         match ty {
             Type::TypeDef((def, generics)) => self.type_def_vtbl_name(*def, generics),
-            _ => unimplemented!(),
+            x => unimplemented!("{:?}", x),
         }
     }
     pub fn type_abi_name(&self, ty: &Type) -> TokenStream {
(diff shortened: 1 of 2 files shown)


MarijnS95:
Note that while `tool_windows` succeeds, all `Foundation` things are gone, and running `tool_core` still runs into:

```
thread 'main' panicked at 'Windows.Foundation.IReference not found', crates\libs\bindgen\src\standalone.rs:97:9
```

These seem to come from `Windows.winmd`. I've downloaded the latest `10.0.22621.755` version and regenerated, and now it is fine. Did the format for `mdmerge` change? After all, I downloaded the same version that `Microsoft.Windows.SDK.Contracts` was supposedly downloaded from previously.

kennykerr:
I haven't taken the last few releases of the Win32 metadata as there are a number of problematic changes that block Rust adoption. I also have some major changes coming to metadata and code generation and I'd like to land those before dealing with the metadata breaking changes.

MarijnS95:
@kennykerr ah that's unfortunate, I'd love to have the updated metadata in windows-rs at some point for these small improvements, but makes sense to delay that until getting those major changes in. Will sit tight, thanks!

MarijnS95:
@kennykerr that doesn't make this PR invalid though? It is still very much necessary to compile with the latest win32metadata and it still rebases nicely on top of `master` even after `riddle` made its way in.

Since that's in it sounds like we're unblocked here now to look into "dealing with the metadata breaking changes"?

kennykerr:
There are just a bunch of incompatibilities that were introduced by the Win32/Wdk metadata in the last few releases that need to be dealt with. I've worked on, among other things, making the error reporting through `riddle` far more helpful for this sort of thing. Anyway, I'll take a look at the latest metadata when I have a sec and see if it's something we can now easily handle or whether it needs a bit more thought and engineering.

crates/libs/bindgen/src/gen.rs
```diff
@@ -156,13 +179,14 @@ impl<'a> Gen<'a> {
             Type::WinrtArray(ty) => self.type_name(ty),
             Type::WinrtArrayRef(ty) => self.type_name(ty),
             Type::ConstRef(ty) => self.type_name(ty),
-            _ => unimplemented!(),
+            Type::TypeRef(TypeDefOrRef::TypeRef(ref_)) => self.type_ref_name(*ref_),
```
kennykerr, line 182: Note that code gen from `TypeRef` is not a good idea as it implies the `TypeDef` is not available and thus you have no idea what kind of type you're actually generating code for. Usually this implies you're missing metadata. I should have better diagnostics for this in `riddle` soon.

riverar: Throwing in my experiences from last night smoke testing riddle. I ran into this too; I was missing base Win32 Metadata winmd in the riddle input folder.

MarijnS95: @kennykerr ah, thanks, so you're basically saying that I fixed this in the wrong way. Which might match the [observation of `Foundation` having vanished](https://github.com/microsoft/windows-rs/pull/2539#issuecomment-1591977671).

Improved diagnostics (even something simple as a format-string in an `unimplemented!()` call) are definitely welcome!

@riverar note that this PR was from before `riddle` and might be caused by the Win32metadata updates instead.

MarijnS95: > Note that code gen from `TypeRef` is not a good idea as it implies the `TypeDef` is not available and thus you have no idea what kind of type you're actually generating code for. Usually this implies you're missing metadata. I should have better diagnostics for this in `riddle` soon.

@kennykerr are you still planning on improving the diagnostics here? A similar issue is appearing again because of incomplete metadata: https://github.com/microsoft/win32metadata/issues/1760

kennykerr: Please open a new issue and include a repro. Thanks.

MarijnS95: Reported at #2742. The repro is generating new metadata with `win32metadata` `main` (relevant commit hash is linked in case this is fixed upstream), followed by running `tool_windows`.

Issue #2492, opened 2023-05-05
Generalize the `component`/`standalone` functions to support namespace/glob generation


MarijnS95:
### Which crate is this about?

windows-bindgen

### Crate version

0.49

### Summary

One of my custom crates based on `windows-bindgen` sets `gen.cfg = false`, resulting in no `#[cfg]`s on `windows-bindgen 0.48` (and `windows-metadata`):

https://github.com/MarijnS95/DirectX-Headers/blob/9767b2895cdfc10d4aa72df6efd4ef24d3945017/.rust/api_gen/src/main.rs#L38

After upgrading to `0.49` however, without changing anything else, the API is riddled with `#[cfg]` guards, forcing me to default-define features within my own crate to use the bindings, even if `gen.cfg = false` remains in place:

https://github.com/MarijnS95/DirectX-Headers/commit/9767b2895cdfc10d4aa72df6efd4ef24d3945017

### Toolchain version/configuration

N/A

### Reproducible example

```rust
git clone https://github.com/MarijnS95/DirectX-Headers
cd DirectX-Headers
dotnet build .metadata
cargo r -p api_gen
```

### Crate manifest

```toml
[dependencies]
windows-bindgen = "0.49"
windows-metadata = "0.49"
```

### Expected behavior

No `#[cfg]` in autogenerated code, like in `windows-bindgen 0.48`.

### Actual behavior

`#[cfg]` guards show up in autogenerated code.

kennykerr:
Thanks, but there's a reason that `Gen` isn't documented. 😉 

https://docs.rs/windows-bindgen/

It isn't tested directly as it is a rather complex beast that is used under the hood by some of the other public APIs, which are tested. I'm trying to provide higher level abstractions that are simpler to use and support, hence the `component` and `standalone` functions. 

Anyway, I'm happy to take a quick look at this issue if you can share a minimal repro (no external dependencies) and perhaps provide more of the underlying functionality through simpler functions.

kennykerr:
Actually, the `cfg` field is no longer used. I should just remove it.

MarijnS95:
> Thanks, but there's a reason that `Gen` isn't documented. wink
> 
> https://docs.rs/windows-bindgen/

Is `component()` the way to go then?

> It isn't tested directly as it is a rather complex beast that is used under the hood by some of the other public APIs, which are tested. I'm trying to provide higher level abstractions that are simpler to use and support, hence the `component` and `standalone` functions.

[...]

PR #2544, opened 2023-06-19
Consolidate code generation

kennykerr:
This update moves all of the code generation features into the `riddle` tool. This includes:

* `windows` and `windows-sys` crate generation
* third-party crate generation
* component development (DLLs)
* standalone code generation

This should greatly simplify the way developers consume bindings and provides a more declarative tool for specifying which bindings to generate and how.

In addition, `riddle` is the tool that can both parse and generate .idl and .winmd files so developers have a single tool for all metadata processing. This part is still very much in development, but this consolidation means I can now focus on a single metadata engine and more quickly bring this to life. 

Note that a few changes will impact existing developers:

* The `windows-bindgen` crate is no more. You can migrate to the `riddle` tool which does all of the same things without requiring you to write code. Whether its standalone code generation or component development, `riddle` provides a simpler way to generate those bindings. It will be published to crates.io so that you can install it using `cargo install riddle`.
* The `windows-metadata` crate has been trimmed down and represents a low-level reader. Generally, you shouldn't have to interact with this crate directly and can just use `riddle`. In particular, the crate no longer embeds the .winmd files for the Windows API. This has two main benefits. The first is that the crate itself is much smaller and the second is that it means you can more easily target whatever metadata your project is built and tested for rather than whatever version windows-rs happens to contain at any given time.


MarijnS95:
> * It will be published to crates.io so that you can install it using `cargo install riddle`.

(Without looking at the code) curious if this still allows users to consume `riddle` in their `build.rs` as a library crate?

kennykerr:
You can call `riddle` from `build.rs` via `std::process::Command`.

MarijnS95:
And that requires a `build.rs` to call `cargo install riddle` first and clobber the users' environment?

kennykerr:
That depends on how you've chosen to configure the build and prepare the crate for publication.

kennykerr:
I'll write more about the available options when I'm closer to publishing riddle itself.

MarijnS95:
As expected I've added a `cargo install riddle --git my-fork` to my CI scripts now.

Marerc:
Hi, there, thank you for the contributions, your works really help. And is there any toturial for the riddle?

kennykerr:
I'm just polishing it up some more before I publish the version at which point I'll be sure to add a tutorial here:

https://kennykerr.ca/rust-getting-started

djc:
Is there a particular reason there is no library interface for riddle, just a CLI interface?

kennykerr:
That was briefly discussed in #2492 - I'm not strongly opposed but there wasn't a very strong argument for it, so I haven't prioritized that work. Feel free to open a new issue if you have a good case for it.

PR #2549, opened 2023-06-21
Riddle fixes and improvements

kennykerr:
* Adds support for type references in winmd generation
* Speeds up Rust code formatting
* Adds more test coverage
* Introduces a new `MINIMAL` config option for omitting enumerators


crates/tools/riddle/src/rust/mod.rs
```diff
@@ -30,6 +30,7 @@ pub fn from_reader(
     gen.std = config.remove("STD").is_some();
     gen.sys = gen.std || config.remove("SYS").is_some();
     gen.implement = config.remove("IMPLEMENT").is_some();
+    gen.minimal = config.remove("MINIMAL").is_some();
```
riverar, line 33: Nit: I'd prefer to see `--minimal` etc. instead of `-config MINIMAL`

riverar (approved):

crates/tools/riddle/src/rust/mod.rs
kennykerr: Yeah, I'm still mulling this over. The `-config` options are just passed onto the respective generators (idl/rust/winmd) and will tend to be very specific to one or the other, which is why I didn't make them first-class options.

PR #2551, opened 2023-06-21
Improve panic diagnostics

kennykerr:
This update attempts to provide some basic debug information for unimplemented code paths and standardizes on `unimplemented` rather than `todo` for consistency.


MarijnS95:
Thanks for copying this over!

PR #2554, opened 2023-06-26
Add initial riddle support for methods and parameters 

kennykerr:
This includes reading/writing .idl as well as generating .winmd files. This update also renames the internal `Gen` type to `Writer` for consistency, hence the noisy update.


riverar:
image

Lib test failure is a false negative, resulting from compiler servicing 11 hours ago https://github.com/actions/runner-images/pull/7760. This results in a new `@comp.id` symbol getting emitted; our [lib post-processing doesn't handle that today](https://github.com/microsoft/windows-rs/blob/master/crates/tools/msvc/src/main.rs#L56-L94).

PR #2563, opened 2023-07-01
Tweak riddle arguments

riverar:
Aligns `riddle` long form arguments with experiences found in other command line tooling.

Some prior art on short/long arg syntax:
* https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
* https://docs.rs/clap/latest/clap/

### DIFF
--- a/crates/tests/component/build.rs
+++ b/crates/tests/component/build.rs
@@ -31,14 +31,14 @@ fn main() {
         "--target-dir",
         "../../../target/test_component", // TODO: workaround for https://github.com/rust-lang/cargo/issues/6412
         "--",
-        "-in",
+        "--in",
         "component.winmd",
         &metadata_dir,
-        "-out",
+        "--out",
         "src/bindings.rs",
-        "-filter",
+        "--filter",
         "test_component",
-        "-config",
+        "--config",
         "IMPLEMENT",
     ]);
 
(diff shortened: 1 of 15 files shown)


ChrisDenton:
Hm, unless I'm missing it there doesn't seem to be any precedent for `-@`? That seems like the worst of both worlds, no? It's sigil soup and non-standard.

riverar:
That's fair, thoughts on a replacement? `-etc` doesn't seem right given it's not _et cetera_ (and other things).

riverar:
Dropped the short forms and restored `--etc`. Kept the long forms (with double hyphens).

ChrisDenton:
I admit I don't otherwise have any strong preferences, so long as it's not too weird. Consistency would be great but unfortunately Windows command line tools are all over the place. Even relatively recent first party, developer focused tools can't quite seem to agree on a convention (e.g. vswhere and winget).

kennykerr:
Yes I've not been overwhelmed by the consistency in command line tools either, so I've always just tried to find the simplest and most intuitive approach for my tools. 

By the looks of it, this (final) commit only changes from a single `-` to a `--` and otherwise leaves things as they were. I don't love it, but I recognize it is at least a little more in line with the Rust family of command line tools, so I think this is probably fine.

kennykerr (approved):

PR #2568, opened 2023-07-06
Move to Rust edition 2021 and version 1.56

kennykerr:
Now that the major crate dependencies like `syn` and `tokio` have moved to 1.56 or later, `windows-rs` can finally make the move to the 2021 edition of Rust and MSRV for `windows` and `windows-sys` will be 1.56 from here on.


kennykerr:
Looks like this is failing due to a change in 2021: https://doc.rust-lang.org/stable/edition-guide/rust-2021/reserving-syntax.html

Thomasdezeeuw:
@TheBlueMatt I don't this that's true, Mio only depends on v0.48, see https://github.com/tokio-rs/mio/blob/8dd37c5e14bf7031365321c5aad5dc85788b9a09/Cargo.toml#L52, Cargo isn't allowed to use v0.49 or v0.50 because of that.

kennykerr:
Hopefully #2613 will help.

kennykerr:
That comment was purely about the fact that Rust 1.48 doesn't support the `version` key.

PR #2599, opened 2023-08-10
metadata: Parse new `Char` value type

MarijnS95:
The latest metadata now contains `char` constants.

Note that `quote! { #value }` generates the actual character code in the source file, whereas we prefer the escaped literals just like in `Value::String()`.

```rust
pub const DEVPROP_FALSE: char = '\u{0}';
pub const DEVPROP_TRUE: char = '\u{ffff}';
````

### DIFF
--- a/crates/libs/metadata/default/readme.md
+++ b/crates/libs/metadata/default/readme.md
@@ -1,22 +1,25 @@
 These `.winmd` files provide the default metadata for the Windows API. This is used to
 generate the `windows` and `windows-sys` crates. To view the metadata, use a tool
-like [ILSpy](https://github.com/icsharpcode/ILSpy). 
+like [ILSpy](https://github.com/icsharpcode/ILSpy).
 
-## Windows.Win32.winmd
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by riverar** [comment]
I don't think the metadata impl. is correct; this should probably be a `System.Byte` or `Int8` rather than a `char`.

**Comment by MarijnS95** [comment]
`ECMA-335` says that `char` is UTF16:

https://www.ecma-international.org/wp-content/uploads/ECMA-335_5th_edition_december_2010.pdf

image

**Comment by ChrisDenton** [comment]
A Rust `char` is 32 bits, not 16. Further, these true and false values don't look like they're semantically characters either.

**Comment by riverar** [comment]
DEVPROP booleans are 8-bit making them incompatible 

```cpp
typedef char CHAR;
typedef CHAR DEVPROP_BOOLEAN, *PDEVPROP_BOOLEAN;
#define DEVPROP_TRUE  ((DEVPROP_BOOLEAN)-1)
#define DEVPROP_FALSE ((DEVPROP_BOOLEAN) 0)
```

Issue #2603, opened 2023-08-12
Release request for windows-rs 0.49.0 for WDK additions.


chyyran:
Requesting a new release of windows-rs for WDK additions, particularly ntifs.h headers in `Wdk_Storage_FileSystem`.

kennykerr:
Which crate are you using?

chyyran:
I’m using windows-rs

kennykerr:
Thanks, have you tested against github? Happy to release an update if you've validated that the APIs you need work as expected.

chyyran:
Tested against GitHub and it seems to be working fine with one caveat. WDK functions convert their NTSTATUS errors to HRESULT so they can return `Result<T, windows::core::Error>`. My particular use case requires the base NTSTATUS in case of errors. Could WDK functions be made to return `Result<T, NTSTATUS>` or a wrapped version such that the original NTSTATUS is recoverable where relevant? (Should I file a new issue for this?)

tim-weis:
@chyyran [`NTSTATUS`](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-erref/87fba13e-bf06-450e-83b1-9241dc81e781) values can be losslessly represented as `HRESULT` values. To recover the corresponding `NTSTATUS` value from a [`windows::core::Error`](https://microsoft.github.io/windows-docs-rs/doc/windows/core/struct.Error.html) you would call [`code()`](https://microsoft.github.io/windows-docs-rs/doc/windows/core/struct.Error.html#method.code) to get the `HRESULT` value, and then unset bit 4 (N).

chyyran:
Thanks @tim-weis, in which case everything works as expected.

riverar:
Heads up, the documentation linked above numbers the most significant bit as bit 0. Bit 4 in this case is really bit 28.

kennykerr:
Thanks for testing. Will publish today.

PR #2605, opened 2023-08-14
Version 0.51.0

kennykerr:
As requested by #2603, this update includes the first published update to the [windows](https://crates.io/crates/windows) crate in 5 months. As a reminder, updates are only published by request. 

* This includes a semver-compatible update to the [windows-targets](https://crates.io/crates/windows-targets) crate.
* The `windows` crate now depends on the [windows-core](https://crates.io/crates/windows-core) crate.
* The `windows`, `windows-core`, `windows-implement`, and `windows-interface` crates have been bumped to version 0.51.0 as there have been breaking changes. 
* The `windows-targets` crate has been bumped to version 0.48.2 as it remains compatible with previous versions. 
* Notably, an expanded set of WDK APIs are now available. 
* Other crates will not be updated at this time.

Fixes: #2603

### DIFF
--- a/crates/libs/core/Cargo.toml
+++ b/crates/libs/core/Cargo.toml
@@ -1,7 +1,7 @@
 
 [package]
 name = "windows-core"
-version = "0.50.0"
+version = "0.51.0"
 authors = ["Microsoft"]
 edition = "2021"
 license = "MIT OR Apache-2.0"
@@ -16,7 +16,7 @@ default-target = "x86_64-pc-windows-msvc"
 targets = []
 
 [dependencies.windows-targets]
-version = "0.48.1"
+version = "0.48.2"
 path = "../targets"
 
 [features]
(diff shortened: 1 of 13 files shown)


MarijnS95:
@kennykerr are you sure `windows-bindgen`/`windows-metadata` don't need an update? They generate calls to `::windows::imp::interface_hierarchy!()`, but these have moved into the core.

It cannot be `::windows::core::imp::interface_hierarchy!()` either because the macro internally references `::windows_core` and requires the `windows-core` crate to be in the dependencies list too.

kennykerr:
I only publish crates upon request and only the `windows` crate was requested. I'm also actively working on the extensibility provided by those other crates so I'm eager to continue refining that before I publish another update. 

And yes, it is an unfortunate side-effect of the `windows-core` crate that this additional dependency is necessary. I'm still mulling over how to address that most elegantly.

MRahilly:
Great to see 0.51.0

The HMODULE/HINSTANCE tango resumes and need let _ = is needed to consume return types.

Issue #2607, opened 2023-08-15
Release request: `windows-bindgen` (etc?) for compatibility with `0.51` release


MarijnS95:
Code generated by the current `windows-bindgen`/`windows-metadata` `0.49` crates is not compatible with the latest `windows(-core) 0.51` release.

> @kennykerr are you sure `windows-bindgen`/`windows-metadata` don't need an update? They generate calls to `::windows::imp::interface_hierarchy!()`, but these have moved into the core.
>
> `::windows::core::imp::interface_hierarchy!()` would also be inconvenient because the macro internally references `::windows_core` and requires the `windows-core` crate to be in the dependencies list too.

_Originally posted by @MarijnS95 in https://github.com/microsoft/windows-rs/issues/2605#issuecomment-1678185817_

---

Apparently compatible releases need to be explicitly requested, so here goes.

MarijnS95:
Thanks so much @kennykerr, it's all working now!

Only `riddle` isn't published yet, but I'm using that as a nice excuse to just have a one-liner `api_gen` crate linking to `windows-bindgen`. After all, as mentioned in https://github.com/Traverse-Research/amd-ext-d3d-rs/pull/3:


> Still on the fence whether to go the `cargo install riddle; riddle --etc bindings.txt` route, or whether to _keep_ the `cargo r -p api_gen` crate now that windows-rs decided to **keep** a `windows-bindgen` library crate that we can instead give these arguments to [^1].
> 
> [^1]: Meaning that the `riddle` utility does not need to be separately installed, because running the `api_gen` crate through `cargo` automatically downloads and compiles dependencies.

Issue #2593, opened 2023-07-31
Feature request: riddle lib interface


Yuri6037:
### Motivation

I have a low-level crate which uses currently windows-sys but imports a great set of features and uses only about 1 function for each of these features, I'm searching to optimize the bindings to only exactly what I'm using based on what feature the user enabled on the lib crate itself.

I've looked into the riddle tool which looks like it could help me but there's one big drawback of using it: it's absolute pain to setup. What I need is a way to install the riddle tool as part of build.rs script in order to generate the bindings in dev only. Forcing other people to install globally this tool is unacceptable. So the current setup will force me to make some hacky code to download (or use submodule), build and run the tool manually from build.rs.

Could you provide a simple lib interface so that I can import this tool in dev and use it from build.rs?

kennykerr:
I've had a few suggestions for a lib. That should be doable.

riverar:
Some dev discussion notes to refresh our memory later:
* Look at `windows-bindgen` crate as possible deployment vehicle
* Shipping _without_ metadata (add winmds to consuming crate) won't satisfy folks asking for turnkey solution
  * Dependency on winmd files needs to be added (e.g., windows-metadata default function)
  * Metadata itself is still very unstable; breakage is almost guaranteed between releases
  * May be easier to punt and explore riddle bootstrapping first

kennykerr:
We could hide the default winmd files behind a crate feature, so they're not included if not desired. Mainly, it doesn't make sense for `riddle.exe` to include them but lib dependents may appreciate that it's all included.

PR #2609, opened 2023-08-15
Make `riddle` available as a library

kennykerr:
This update makes the `riddle` tool available as a library crate. 

* I've repurposed the `windows-bindgen` crate rather than minting yet another crate. 
* The "default" metadata is now included by the `windows-bindgen` crate directly. This can be turned off using `default-features = false` as the `riddle` tool does. 
* The `tool_core` crate used internally by `windows-rs` illustrates how to use `windows-bindgen` as a library (https://github.com/microsoft/windows-rs/commit/9b17c65883332844fa718b9e563f61f562cc7ffa).
* The interface is the same between `riddle` and `windows-bindgen` namely string arguments. The choice is thus purely whether you need a library or prefer a command line tool. 

@ChrisDenton does this improve your developer experience? 

The rest of the PR has a lot of noise but that's largely just generated comments getting updated/removed. 

Fixes: #2593
Fixes: #2607


kennykerr:
Ideally you can just pass `library/std/src/sys/windows/c/windows_sys.lst` directly to `windows-bindgen`. 😊

ChrisDenton:
I had a brief look and, yep, this looks easier to integrate. I'll do a draft PR rust-lang tomorrow and see if anything shakes out (iirc there have also been quite a few changes since the last metadata update).

> Ideally you can just pass `library/std/src/sys/windows/c/windows_sys.lst` directly to `windows-bindgen`. 😊

Slightly tricky atm because we have comments in the `windows_sys.lst` file. And we add a few bits to the output file too (mostly just to make arm32 not be a compile error),

kennykerr:
Sounds good. Let me know if you need me to tweak anything. I can also skip comments if that's the only obstacle.

ChrisDenton:
Yeah, I think the comments are the only issue with input.

I did notice that `FindFirstFileW` is returning a `HANDLE` type instead of a `FindFileHandle`. Was that intentional?

Also minor thing but I think it'd be good to remove the std special case for `SOCKET`. It should now be easy to use the windows-sys type internally.

kennykerr:
The `FindFirstFileW` thing probably flipped around in the Win32 metadata. I can't point to a commit or PR as they don't have a good way of tracking changes. I'm planning on stabilizing that as soon as I have a few more pieces in place for `riddle`. Sorry about that. For what it's worth, I think `HANDLE` is correct. 

I'll have a PR up tomorrow with the changes for stripping comments and removing the specializations. You can then double check that for me. Be great to have less differences.

ChrisDenton:
> The `FindFirstFileW` thing probably flipped around in the Win32 metadata. I can't point to a commit or PR as they don't have a good way of tracking changes. I'm planning on stabilizing that as soon as I have a few more pieces in place for `riddle`. Sorry about that. For what it's worth, I think `HANDLE` is correct.

Tbh it ultimately doesn't matter for sys style bindings, other than making the diff nicer, I was more curious.

> Be great to have less differences.

Yep! I hope to continue pecking away at the differences.

crates/libs/bindgen/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "windows-bindgen"
+version = "0.49.0"
```
MarijnS95, line 3: > I've repurposed the windows-bindgen crate rather than minting yet another crate.

This is still the same version as that from 4 months ago:

https://crates.io/crates/windows-bindgen/versions

Looks like a publish didn't make its way through?

kennykerr: Getting that ready now: #2611

PR #2610, opened 2023-08-16
Improve `windows-bindgen` for `std` usage 

kennykerr:
As discussed in #2609, this update makes the following changes:

* Strips out comments from response files.
* Removes the special case for `SOCKET`


kennykerr:
Sweet, thanks for checking!

kennykerr:
@ChrisDenton I'm preparing a release.

ChrisDenton:
Great! My PR should be ready to go once that's done (give or take some minor tweaks).

klensy:
Wait a second, maybe fix for https://github.com/microsoft/win32metadata/issues/1680 can be here too?

kennykerr:
I'd need another release of the Win32 metadata. I don't think there is one yet with that change. That would also mean a breaking change to the API surface so unless that's really urgent I'd prefer to hold off on that. Thoughts?

PR #2611, opened 2023-08-16
Version 0.51.1

kennykerr:
This is mainly for publishing an update of the `windows-bindgen` crate but also fixes a few minor readme and version issues to avoid confusion. 

* The `windows` and `windows-core` crates are updated to 0.51.1 to update the readme but is otherwise unchanged.
* The `windows-targets` crate and arch-specific crates are updated to 0.48.3 to update the readme and bring the lib file names in line with the crate version.

Those are all semver-compatible updates. 

* The `windows-bindgen` crate is updated to 0.51.1 and includes the new riddle-based metadata processing capabilities and is compatible with the `windows` and `windows-core` crates. 
* The `windows-metadata` crate is updated to 0.51.1 and includes many refinements but is largely for internal use. 

These are major breaking changes. 

Following the completion of this PR I'll republish these crates. The `windows-sys` crate will not be updated at this time.


kennykerr:
@riverar & @ChrisDenton if you could take a quick look I'd appreciate it.

crates/tests/targets/tests/files.rs
```diff
@@ -0,0 +1,13 @@
+// This test just makes it easier to ensure that the lib file names match the crate versions
+// so we don't accidentally forget to bump the version numbers embedded in the lib file names.
```
ChrisDenton, line 2: Nice idea 👍

ChrisDenton (approved):
Looks right to me

ChrisDenton:
Oh wait, just noticed https://github.com/microsoft/windows-rs/blob/0.51.1/crates/samples/readme.md

There versions of windows and windows-sys have now diverged which isn't reflected there. I'm not sure how much that matters though,

docs/readme.md
```diff
@@ -3,14 +3,14 @@
 The [windows](https://crates.io/crates/windows) and [windows-sys](https://crates.io/crates/windows-sys) crates let you call any Windows API past, present, and future using code generated on the fly directly from the [metadata describing the API](https://github.com/microsoft/windows-rs/tree/master/crates/libs/bindgen/default) and right into your Rust package where you can call them as if they were just another Rust module. The Rust language projection follows in the tradition established by [C++/WinRT](https://github.com/microsoft/cppwinrt) of building language projections for Windows using standard languages and compilers, providing a natural and idiomatic way for Rust developers to call Windows APIs.
 
 * [Getting started](https://kennykerr.ca/rust-getting-started/)
-* [Samples](https://github.com/microsoft/windows-rs/tree/0.48.0/crates/samples)
+* [Samples](https://github.com/microsoft/windows-rs/tree/0.51.1/crates/samples)
```
kennykerr, line 6: @ChrisDenton I think it's fine since this link here will point to the "current" release on crates.io for each respective version and crate. The samples readme is mainly trying to describe how to get to the right link for any particular version.

ChrisDenton: Agreed. I think that's fine then.

PR #2620, opened 2023-08-18
Switch all crates to Rust edition 2021

kennykerr:
As discussed in #2613, all crates are switching to use edition 2021 and MSRV 1.56 or later. 

Version 0.52 will be the next released version of the `windows-rs` crates, but there are no further release plans at this time.


MarijnS95:
> Version 0.52 will be the next released version of the windows-rs crates, but there are no further release plans at this time.

The downside of bumping this early is that we can now no longer `[patch.crates-io]` to git commits on this crate :)

Issue #2630, opened 2023-08-25
Add warning/error if a particular `bindgen` filter has no effect


kennykerr:
> Maybe it could warn/error if a binding is not found?

Yes I'd like to report a warning/error if a particular filter has no effect. 

_Originally posted by @kennykerr in https://github.com/microsoft/windows-rs/issues/2625#issuecomment-1686918152_

PR #2634, opened 2023-08-28
Detect unused `bindgen`/`riddle` filters

kennykerr:
As an example:

```
> riddle.exe -i windows.winmd -o output.rs -f Windows.Foundation.Rect WIndows !Windows.Fondation
error: unused filters
  Windows.Fondation
  WIndows
```

Fixes: #2630

### DIFF
--- a/crates/libs/bindgen/src/winmd/verify.rs
+++ b/crates/libs/bindgen/src/winmd/verify.rs
@@ -2,6 +2,18 @@ use super::*;
 use metadata::RowReader;
 
 pub fn verify(reader: &metadata::Reader, filter: &metadata::Filter) -> crate::Result<()> {
+    let unused: Vec<&str> = filter.unused(reader).collect();
+
+    if !unused.is_empty() {
+        let mut message = "unused filters".to_string();
+
+        for unused in unused {
+            message.push_str(&format!("\n  {unused}"));
+        }
+
+        return Err(crate::Error::new(&message));
+    }
+
     for item in reader.items(filter) {
         // TODO: cover all variants
         let metadata::Item::Type(def) = item else {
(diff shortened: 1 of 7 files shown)


ChrisDenton (approved):
Looks great to me! I also tried it on the rustc riddle PR and it showed up things that had been moved or removed. I'll update that when I have a mo.

This PR also has a few misc cleanups? Not complaining mind you.

crates/libs/bindgen/src/winmd/verify.rs
```diff
@@ -2,6 +2,18 @@ use super::*;
...
+    let unused: Vec<&str> = filter.unused(reader).collect();
+
+    if !unused.is_empty() {
+        let mut message = "unused filters".to_string();
+
+        for unused in unused {
+            message.push_str(&format!("\n  {unused}"));
+        }
+
+        return Err(crate::Error::new(&message));
```
riverar, line 14: Surfacing these are errors versus warnings may make crate third party crate maintenance harder than it needs to be. I can see a scenario where folks will want to add a preemptive filter for, say, X.Y.Z.Preview namespaces when they appear/don't appear during a dev cycle. Or simply don't care if they over-specify the filter list.

No change needed right now though, we can take the "lets see what happens" approach.

riverar (approved):

crates/libs/bindgen/src/winmd/verify.rs
ChrisDenton: Personally, I do prefer the error because I'd like the filter list to reflect what goes into the generated bindings (no more, no less). But I accept that other people may have other preferences.

kennykerr:
> This PR also has a few misc cleanups? 

Only some dead/commented code that I happened upon as well as hardening the tests to ensure that the filtering was working correctly.

kennykerr:
Thanks for checking! I'll leave it like this for now and we can reevaluate in future as needed.

PR #2644, opened 2023-09-04
Update readme.md - fix link

Spirarel:
What's this all about?

Fix broken link

Fixes: #0000 ⬅️ Be sure to refer to an existing issue here!


docs/readme.md
```diff
@@ -3,7 +3,7 @@
 The [windows](https://crates.io/crates/windows) and [windows-sys](https://crates.io/crates/windows-sys) crates let you call any Windows API past, present, and future using code generated on the fly directly from the [metadata describing the API](https://github.com/microsoft/windows-rs/tree/master/crates/libs/bindgen/default) and right into your Rust package where you can call them as if they were just another Rust module. The Rust language projection follows in the tradition established by [C++/WinRT](https://github.com/microsoft/cppwinrt) of building language projections for Windows using standard languages and compilers, providing a natural and idiomatic way for Rust developers to call Windows APIs.
 
 * [Getting started](https://kennykerr.ca/rust-getting-started/)
-* [Samples](https://github.com/microsoft/windows-rs/tree/0.52.0/crates/samples)
+* [Samples](https://github.com/microsoft/windows-rs/tree/master/crates/samples)
```
tim-weis, line 6: This was changed in #2226 so that the `Samples` link on crates.io always refers to the corresponding sample code for any given version. This means the link in this repository's readme is usually ahead of its time and 404's.

The proposed change, however, makes matters worse: On crates.io, the `Samples` link will now practically always be out of sync with *any* published version.

kennykerr: Yes, I think I just need to split the repo readme files so that each crate carries its own readme and the repo itself has a single non-published readme as we just keep having this same issue pop up repeatedly. 🤷‍♂️

kennykerr:
Closing in favor of #2645.

PR #651, opened 2021-03-28
Update readme.md

Immortalin:
For Rust 1.50 and latest version of crate.

PR #2645, opened 2023-09-05
Provide individual crate readme files

kennykerr:
Up to now we've just provided a single readme shared by all crates. This has proven to be a simple and manageable approach but has also caused quite a bit of confusion for developers poking around the repo directly. In this update I've simply copied the readme into each individual library crate so they can evolve separately and removed most of the content from the repo's readme which can now also focus more on supporting developers visiting the repo itself. Hopefully this helps to avoid some of the confusion.

Fixes #2644 #2110 #1940 #780 #651 etc.

@tim-weis @ChrisDenton @riverar for suggestions.


riverar:
I can't post comments on the files directly, seems GitHub is slightly broken right now.

Was just going to suggest maybe adding comments `<!-- ... --->` in tricky areas to help folks / discourage PRs. It shouldn't get rendered on crates.io/GitHub/etc.

riverar:
@kennykerr Looks fine to me, we can adjust per feedback if any further PRs mistakenly come in.

ChrisDenton (approved):
This makes a lot of sense to me, especially as it gives the freedom to diverge in the future. It'd be great to some text on the repo's readme that explains there's a difference between master and the latest published version of crates.

Issue #2639, opened 2023-08-30
Wrong HRESULT error code if the call was made from a different facility


dergecko:
### Summary

The `GetLastError` calls internally the `HRESULT::from_win32`  which set the facility bits of the `HRESULT` to 7 aka _win32_.
However the facility is not always _win32_.

In this example I call the `GetLastError` however the same holds true for functions that already return the `windows::core::Result` like `Cryptography::CryptAcquireCertificatePrivateKey`
I provided a simple example that should look for a fingerprint in the CurrentUser/Personal Cert store if the fingerprint exists.
The is maybe not the simplest version of the bug but hopefully verbose enough.

Easy solutions would be a way to specify the facility on the `GetLastError` call or provide different `windows::core::Error::from_` implementations for each facility. E.g
`windows::core::Error::from_sspi`

### Crate manifest

```TOML
[package]
name = "hresult_example"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
hex = "0.4"
windows = { version = "0.51", features = [
    "Win32_Security_Cryptography",
    "Win32_Foundation",
] }
```

### Crate code

```Rust
use hex::FromHex;
use windows::Win32::Security::Cryptography;

// looks for a fingerprint in CurrentUser\My (Personal)
fn main() {
    let path_as_wstr = "My"
        .encode_utf16()
        .chain(Some(0))
        .collect::<Vec<_>>();
    let dwflags = Cryptography::CERT_OPEN_STORE_FLAGS(Cryptography::CERT_SYSTEM_STORE_CURRENT_USER_ID << Cryptography::CERT_SYSTEM_STORE_LOCATION_SHIFT);
    let handle = unsafe {
        Cryptography::CertOpenStore(
            Cryptography::CERT_STORE_PROV_SYSTEM_W,
            Cryptography::CERT_QUERY_ENCODING_TYPE::default(),
            Cryptography::HCRYPTPROV_LEGACY::default(),
            dwflags,
            Some(path_as_wstr.as_ptr() as _),
        )
    }.unwrap();

    let hash = <[u8; 20]>::from_hex(&"ABF777B90E01AFFDC21BCCA4BB98D7D884DF702E").unwrap();
    let fingerprint_blob = Cryptography::CRYPT_INTEGER_BLOB {
        cbData: hash.len() as _,
        pbData: hash.as_ptr() as _,
    };

[...]

PR #2646, opened 2023-09-06
Tweak Win32 error code conversion to handle `HRESULT` input

kennykerr:
In some cases, the result of `GetLastError` will be an `HRESULT` rather than a Win32 error code despite the fact that the Win32 metadata incorrectly indicates that `GetLastError` returns a Win32 error code. That is a separate issue we'll have to deal with.

For now, this just hardens the `windows-core` crate to deal with error codes consistent with how `HRESULT_FROM_WIN32` behaves in C++.

Fixes: #2639


ChrisDenton (approved):
Looks good to me!

PR #2651, opened 2023-09-11
Derive `PartialEq`, `Eq`, `Debug`, `Clone` for interfaces

kennykerr:
The `derive` attribute has historically been problematic, mostly because it can be very slow, but improvements over the last few years have improved its performance to the point where we can begin to use it more liberally in the `windows` crate without impacting build time. This update limits use of the `derive` attribute to COM interfaces as a start. Applying it to structs continues to be a little problematic, but it seems worthwhile to at least simplify things in degrees.


riverar (approved):
lgtm

ChrisDenton (approved):
More use of `derive` is great, if only because it can substantially reduce the lines of code (something e.g. IDEs complain about).

Issue #2656, opened 2023-09-14
NTSTATUS values are erroneously transformed into `Ok(())`


chyyran:
### Summary

NTSTATUS values that are not `STATUS_SUCCESS` (`0x0`) are transformed into `Ok` rather than `Err`. 

### Crate manifest

```TOML
windows = { version = "0.51.0", features = ["Win32_Foundation"] }
```

### Crate code

This test should pass, but it fails instead.

```Rust
#[cfg(test)]
mod test {
    #[test]
    pub fn status_pending() {
        assert!(matches!(windows::Win32::Foundation::STATUS_PENDING.ok(), Err(_)))
    }
}
```

```
assertion failed: matches!(windows :: Win32 :: Foundation :: STATUS_PENDING.ok(), Err(_))
```

chyyran:
This is a tangentially related request, but I think it's more than reasonable to turn off the transformation of `NTSTATUS` to `windows::core::Error` for APIs that return `NTSTATUS`. It's a lot more work to unwrap `Result<()>` than to simply check `STATUS_SUCCESS` (which should be the _only_ status code that converts to `Ok`), than to call `.ok()?` on an `NTSTATUS`.

ChrisDenton:
I don't think `STATUS_PENDING` should be an error. I think we should stick to the C++ semantics (albeit encoded in Rust types) rather than inventing our own. Otherwise we're going to get issues in places where the semantics don't quite match.

I prefer the suggestion to just return `NTSTATUS`. The only alternative that preserves the full semantics would be `Result<NTSTATUS, NTSTATUS>` which seems less ideal to me, though having a method on `NTSTATUS` that does this would be fine, imho. In any case, I think we would still want a method that performs the transformation into `core::Error` so that it can bubble up into more general code.

kennykerr:
Note that the Win32 metadata has a way to deal with this through the `CanReturnMultipleSuccessValues` attribute but this is applied in a whack-a-mole fashion so I'm not too fond of it. If alternative success codes like `STATUS_PENDING` are more common with functions returning `NTSTATUS` then I'm fine with disabling the transformation for such signatures. 

Note also that `NTSTATUS` provides the `ok()` method so that you can easily transform it into a `Result` if needed.

PR #2658, opened 2023-09-14
Disable signature transformation for `NTSTATUS`

kennykerr:
Fixes: #2656

### DIFF
--- a/crates/libs/bindgen/src/metadata.rs
+++ b/crates/libs/bindgen/src/metadata.rs
@@ -293,7 +293,6 @@ pub fn signature_kind(reader: &Reader, signature: &Signature) -> SignatureKind {
                 SignatureKind::ResultVoid
             }
         }
-        Type::TypeDef(def, _) if reader.type_def_type_name(*def) == TypeName::NTSTATUS => SignatureKind::ResultVoid,
         Type::TypeDef(def, _) if reader.type_def_type_name(*def) == TypeName::WIN32_ERROR => SignatureKind::ResultVoid,
         Type::TypeDef(def, _) if reader.type_def_type_name(*def) == TypeName::BOOL && method_def_last_error(reader, signature.def) => SignatureKind::ResultVoid,
         _ if type_is_struct(reader, &signature.return_type) => SignatureKind::ReturnStruct,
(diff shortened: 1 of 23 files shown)


riverar:
I don't think there are enough `STATUS_PENDING` returning functions to warrant degrading the developer experience here. Is there something preventing us from specializing the return transformation for `NTSTATUS`?

kennykerr:
I think the issue is that there are many many more success codes when it comes to `NTSTATUS` compared with say `HRESULT`. 

The alternative would be to return `Result<NTSTATUS, Error>` but that would be very different to the way we handle error transformations today and just seems like it would cause more confusion than anything else.

ChrisDenton:
The status quo relies on `CanReturnMultipleSuccessValues` being accurately applied (or not) to all NTSTATUS returning functions. Otherwise the experience is not just degraded, it's unusable for certain scenarios. But I think we simply don't have the people to ensure that the metadata 100% accurately applies the attribute. Maybe that'll come in time but we're not there yet.

So, all else aside, I think something has to change.

riverar:
Agree we'd just be applying `CanReturnMultipleSuccessValues` to functions as we find them, but in the absence of more data demonstrating a large swath of functions that will be better off, we may want to consider it over regressing the experience across the board. I recognize, though, getting that data is probably not possible in the short term and concede as a result.

ChrisDenton:
I guess there are alternatives but they'll require more work to get started on. So maybe the opposite attribute for `NTSTATUS` would be more useful (e.g. `CanReturnSingleSuccessValues`)? Have the least destructive behaviour by default and only apply `Result` ergonomics for functions that have been reviewed as only returning one success value.

kennykerr:
I just don't think any of these attribute-based approaches are scalable.

kennykerr:
Unrelated, the build is failing on the diff check for libs. That's probably because GitHub updated their MSVC toolchain and now the libs have a different version stamp again. I really need to get that version stamp wiped out.

ChrisDenton (dismissed):

riverar:
> Unrelated, the build is failing on the diff check for libs. That's probably because GitHub updated their MSVC toolchain and now the libs have a different version stamp again. I really need to get that version stamp wiped out.

Yup, will pick this up (removing the compiler ID per last discussion).

kennykerr:
Manually updated the libs to unblock this PR.

Issue #2657, opened 2023-09-14
Allow to manually call QueryInterface from `windows`


pie-flavor:
### Suggestion

Consider the function `IActiveScript::GetScriptSite`:

```c
HRESULT GetScriptSite(  
    REFIID iid,           // interface identifier  
    void **ppvSiteObject  // address of host site interface  
);  
```

The natural way to implement this function is, given some variable `site`, `site.QueryInterface(iid, ppvSiteObject)`. And the function is designed to make that the natural way, as the `IActiveScriptSite` is provided via a parallel `SetScriptSite` function and the implementor of `GetScriptSite` doesn't necessarily know all the interfaces it implements, and therefore all the interfaces the caller of `GetScriptSite` might be wanting.

However `windows` does not allow you to call `QueryInterface` directly. This means I can't implement it that way, and can only cast to interfaces I *do* know about. If the caller wants `IActiveScriptSiteTraceInfo`, and I forgot to include it, then the function fails, where a C implementation would not have to fail.

The generic wrappers around `QueryInterface` to encapsulate its quirks are great, but it would be good to be able to call it directly when you need to.

kennykerr:
Oh, I forgot to mention that this does actually exist on the `ComInterface` trait although the `query` method is hidden from docs. It is however used for just this purpose. 

https://github.com/microsoft/windows-rs/blob/58eaa528e13b4b5bcfdba18bcebedeaa06c3f9c0/crates/libs/core/src/com_interface.rs#L48-L50

Take that for a spin and if that works for you, I can polish it up and unhide it.

pie-flavor:
It's callable, though it does mean that I have to nullcheck the `REFIID` myself and cast the `*mut *mut c_void` to `*mut *const c_void`.

kennykerr:
That is what I addressed in #2659.

PR #2659, opened 2023-09-14
Unhide `query` method on `ComInterface` trait

kennykerr:
While I'm at it, I fixed some usability issues with the `query` methods. Briefly, I originally wrote the `IUnknown` vtable and trait by hand and it doesn't quite line up with how the Win32 metadata's version of `GUID*, void**` is translated into Rust. This just makes the signatures consistent so that it is easier to call `query` and simply forward the `iid` and `interface` parameters without any coercion. 

Fixes: #2657


kennykerr:
https://github.com/microsoft/windows-rs/pull/2658#issuecomment-1719713440

PR #2661, opened 2023-09-18
Mask non-reproducible linker artifacts in libs

riverar:
The Microsoft Visual C++ (MSVC) linker (as of 14.37.32822.0) does not create reproducible import libs. It supports a `/Brepro` flag to stabilize archive member header timestamps but misses other elements, such as import header time/date stamps, coff file header time/date stamps, and the compiler version symbols emitted by itself into the COFF symbol table.

The tool_msvc crate has been updated to take advantage of the `/Brepro` flag and mask additional non-reproducible elements, including `S_COMPILE2_ST` symbols in the `.debug$S` symbol section. (See also: `link!CImplib::EmitNullImportDescriptor`)

Related: https://github.com/microsoft/windows-rs/pull/2016


kennykerr (dismissed):
Looks good - thanks!

riverar:
Strangely the workflow didn't run after two attempts. Let me submit another commit to kick it.

kennykerr (approved):

PR #2664, opened 2023-09-21
Update SDK and WDK metadata

riverar:
Windows.Win32.winmd
- Version: 55.0.45

Windows.Wdk.winmd
- Version: 0.8.2


kennykerr:
I would have expected the following to catch the stale baseline:

https://github.com/microsoft/windows-rs/blob/126755f60bce63c3c7b0c8edc3de5eef26c37082/.github/workflows/lib.yml#L55-L59

kennykerr:
#2673 confirms the baseline validation continues to work. I also just noticed that the validation was just GitHub failing to report correctly that it didn't actually run the workflows:

image

riverar:
https://github.com/microsoft/win32metadata/issues/1566 is still open, so `../Windows.Win32.49.winmd` stays in this round. Will get that fixed up in metadata for next round.

kennykerr (approved):
Thanks!

Issue #2665, opened 2023-09-21
Hide `windows-sys`'s documentation behind a feature flag?


LikeLakers2:
### Suggestion

Hi! I've been building a program that uses the `bevy` game engine crate, and in the midst of this, I've found it useful to render documentation for everything via `cargo doc`.

However, I noticed that the `windows-sys` crate generates quite a lot of documentation. Unfortunately, `cargo doc` has no way to exclude certain dependencies without excluding all dependencies (the `--exclude` flag only seems to work on packages that are within the current workspace).

I don't believe this is an issue with the `windows-sys` crate, per se - it's doing its job just fine. However, unless a user is using `windows-sys` directly, they may not need to see `windows-sys`'s documentation anyways - meaning `windows-sys`'s documentation is simply wasting disk space, and worse, making `cargo clean` take a lot longer due to the mass amount of files generated.

Still, some users may need `windows-sys`'s docs, so hiding them behind a blanket `#[doc(hidden)]` may not be a good idea. Perhaps `windows-sys` could attach `#[doc(hidden)]` to most stuff, *unless* a certain feature flag (let's say `windows-sys/render_docs`) is specified?

I just think it may be useful to avoid generating so much documentation unless a user needs it.

riverar:
So you'd like to generate docs for your crate and all dependencies except `windows-sys`, right? (Just making sure I understand, thanks!)

LikeLakers2:
Yes. `windows-sys` generates a lot of documentation, and because I'm using the `bevy` game engine crate, I have no need to look at the `windows-sys` documentation ever. So I'd like a way to avoid generating `windows-sys` documentation.

There is https://github.com/rust-lang/cargo/issues/4049, however that was opened in 2017 with no plans to add it yet. Thus, as a stopgap, I'm hoping `windows-sys` can provide a method to skip generating so much documentation.

kennykerr:
Hey, sorry for the delay. Been off the grid camping. 🏕️

Just catching up here but this is what I do. Is that sufficient?

```cargo doc -p package --all-features --no-deps```

LikeLakers2:
@kennykerr It could work, but given the large number of `bevy_*` crates that I'd actually want to look at in the resulting documentation, plus the fact that I have other dependencies Id want the documentation for, specifying each wanted package isn't really something I'd be keen on doing.

I'd much rather just ask for a way to exclude certain crates (i.e. `windows-sys`) than need to include every crate I want as a command line parameter.

[...]

PR #2671, opened 2023-09-28
Slim doc generation

kennykerr:
While looking into #2665, I found some areas where the doc generation could be simplified. Mainly this removes a bunch of redundant doc comments. And shaves a few hundred thousand lines of code. 😏


tim-weis:
The *"Required features"* documentation items seem to have dropped the quotation marks around the features. That's unfortunate as it requires extra work to copy them into a *Cargo.toml* file now. It would be nice to have them back unless there is a compelling reason for this change.

kennykerr:
Right, I originally removed them as the italics rendering was causing some problems where the quotes were not plain quotes supported by toml but it seems now that it's not italic that it would work fine so I can add them back.

tim-weis:
Thanks!

For reference, the non-ASCII quotation marks transformation was fixed<sup>*\[citation needed\]*</sup> by wrapping the quoted feature names in backticks (\`). This has worked reliably with and without italics. I liked the italics better, but I can probably be trained to like the new layout no less than the old, familiar formatting, given enough training data.

kennykerr:
Great, the whole docs things has been frustrating and tedious as rustdoc is so problematic that I've forgotten some of the twists and turns. I'm still considering scrapping it in favor of generating docs directly.

PR #2676, opened 2023-10-03
Add `docs` feature

kennykerr:
This avoids documentation generation when undesired. 

Fixes: #2665

### DIFF
--- a/crates/libs/sys/Cargo.toml
+++ b/crates/libs/sys/Cargo.toml
@@ -21,6 +21,7 @@ path = "../targets"
 
 [features]
 default = []
+doc = []
 # generated features
 Wdk = []
 Wdk_Foundation = ["Wdk"]
--- a/crates/libs/sys/Cargo.toml
+++ b/crates/libs/sys/Cargo.toml
@@ -21,7 +21,7 @@ path = "../targets"
 
 [features]
 default = []
-doc = []
+docs = []
 # generated features
 Wdk = []
 Wdk_Foundation = ["Wdk"]
--- a/crates/libs/sys/src/lib.rs
+++ b/crates/libs/sys/src/lib.rs
@@ -5,7 +5,7 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 #![no_std]
 #![doc(html_no_source)]
 #![allow(non_snake_case, non_upper_case_globals, non_camel_case_types, clippy::all)]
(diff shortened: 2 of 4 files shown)


crates/libs/sys/Cargo.toml
```diff
@@ -21,6 +21,7 @@ path = "../targets"
 
 [features]
 default = []
+doc = []
```
riverar, line 24: How about docs, plural? I recognize this is temporary so not a blocker. But I may grumble something in the hallway when we walk past each other.

riverar (dismissed):

crates/libs/sys/Cargo.toml
kennykerr: Wouldn't want that! It could go either way. I based it on the `doc` attribute. 🤷‍♂️

kennykerr: Switched to `docs`.

tim-weis:
This took me a moment to comprehend, so I'm leaving a few notes here in case someone else gets confused in the future.

# What

This PR gates documentation generation for the `windows` and `windows-sys` crates behind a feature named `"docs"`. This feature is disabled by default. Other crates (notably `windows-core`) are unaffected by this change.

# How To

The `"docs"` feature is intended to be used with `cargo doc`, exclusively (see [feature selection](https://doc.rust-lang.org/cargo/commands/cargo-doc.html#feature-selection) in The Cargo Book). By default, no documentation is generated. To enable documentation generation, e.g., for the `windows` crate run the following command line:

> ```none
> cargo doc --features windows/docs
> ```

Similarly for the `windows-sys` crate.

# How Not To

The `"docs"` feature is a regular Cargo feature that *can* be enabled from a *Cargo.toml* file. Clients of `windows` and `windows-sys`, and crate authors in particular, should refrain from doing so, as it would effectively invalidate the ability to generate documentation for `windows` and `windows-sys` on client request.

kennykerr:
Thanks Tim, that's helpful.

tim-weis:
Glad to have you back, Kenny!

LikeLakers2:
Thanks, Kenny. :)

It appears that my comments at #2665 have restarted the discussion of https://github.com/rust-lang/cargo/issues/4049, a suggestion that allows skipping certain crates from `cargo doc`. See https://github.com/rust-lang/cargo/issues/4049#issuecomment-1740248035.

Maybe `windows-sys` and `windows` won't actually need a `docs` feature for that long?

PR #2682, opened 2023-10-17
Simplify metadata reader

kennykerr:
While working on metadata processing and generation it became clear I needed to further simplify metadata parsing to battle the growing complexity. The metadata reader now works much more like the one I wrote for cppwinrt in the sense that every table column can be read with a simple method call and does not require an indirection through the reader. 

Before:

```rust
for def in reader.get_type_def(TypeName::new("Windows.Foundation", "IAsyncInfo")) {
    println!("{}", reader.type_def_name(def));

    for method in reader.type_def_methods(def) {
        println!("{}", reader.method_def_name(method));
    }
}
```

After:

```rust
for def in reader.get_type_def("Windows.Foundation", "IAsyncInfo") {
    println!("{}", def.name());

    for method in def.methods() {
        println!("{}", method.name());
    }
}
```

Under the hood this requires that the reader is pinned, or leaked, for the duration of the process to allow circular referencing. This seems like a reasonable trade given that this is only ever used by `windows-bindgen` for code generation.


riverar:
Did some testing, looks like a rust bug, at least in the gnu toolchain. Updating unblocks us, doing more research.

❌
```
stable-x86_64-unknown-linux-gnu (default)
rustc 1.72.1 (d5c2e9c34 2023-09-13)
```

✔️
```
stable-x86_64-unknown-linux-gnu (default)
rustc 1.73.0 (cc66ad468 2023-10-03)
```

kennykerr:
Nice catch, I can repro the failure if I use 1.72.1 locally. Fixed with 1.73. Weird.

PR #2683, opened 2023-10-17
Add bindgen config option to disable generating inner attributes

dpaoliello:
Issue Details:
It can be useful to include the files generated by `riddle` or the `bindgen` crate in an existing package: either by generating them into the package's directory or using the `include!` macro. For this to work, the generator needs to create a single .rs file AND not have any inner attributes (e.g., `#![allow(...)]`) at the top of the file, which is currently not an option.

Fix Details:
Adds a new config option, `no-inner-attributes` which instructs the `bindgen` crate not to emit the inner attributes at the top of the file.


kennykerr (approved):

ChrisDenton:
Does this need a test case to avoid accidental regression?

kennykerr:
You could update tool_core to use this option.

kennykerr (approved):
Looks great - thanks for the test!

PR #2685, opened 2023-10-19
Added RPC_S_OK constant

rthidfrev:
Added the constant `pub const RPC_S_OK: RPC_STATUS = RPC_STATUS(0i32);`


tim-weis:
@kennykerr The final link [here](https://kennykerr.ca/rust-getting-started/how-are-crates-built.html) is no longer correct. Could you update it to point to its [new home](https://github.com/microsoft/windows-rs/tree/master/crates/libs/bindgen/default)?

PR #2687, opened 2023-10-20
Enable all features for docs.rs generation

GuillaumeGomez:
The current docs.rs documentation for the `windows-rs` crate is useless currently: https://docs.rs/windows/latest/windows/

This enables all features so docs for all items will be available.

### DIFF
--- a/crates/libs/windows/Cargo.toml
+++ b/crates/libs/windows/Cargo.toml
@@ -16,6 +16,7 @@ categories = ["os::windows-apis"]
 default-target = "x86_64-pc-windows-msvc"
 targets = []
 rustc-args = ["--cfg", "docsrs"]
+all-features = true
 
 [dependencies]
 windows-core = { path = "../core", version = "0.52.0" }


GuillaumeGomez:
Once merged, could you make a new release so we can have access to the documentation? Thanks in advance!

tim-weis:
> This enables all features so docs for all items will be available.

Is `docs.rs` ready to deliver? Documentation for the `windows` crate is hosted [here](https://microsoft.github.io/windows-docs-rs/doc/windows/) because `docs.rs` wasn't quite up for the task. Has the situation changed since?

GuillaumeGomez:
What was the issue? It took too long to generate the docs and it timed out?

tim-weis:
I do not recall the precise issues, but I believe it was either one of *"too long"* or *"too much memory"*.

GuillaumeGomez:
Let me check with my docs.rs teammates to see if we can give it another try.

GuillaumeGomez:
So the current blocker is the maximum memory from what I could infer. For that, we'll have to wait until the problem is solved in docs.rs.

In the meantime, would it be possible to add a links to the documentation of your crate so people (like me) actually can discover where it's used?

kennykerr:
I think that's what #2297 by @ChrisDenton was meant to solve but I don't see the "stub" message here: 

https://docs.rs/windows/latest/windows/

So perhaps something's not quite working.

GuillaumeGomez:
No I really meant in your documentation directly. I wasn't even aware of this link like one minute ago (and I'm in the docs.rs team XD).

ChrisDenton:
The intent was to display a link in the documentation but only when published on docs.rs. I think I screwed up [#2297](https://github.com/microsoft/windows-rs/pull/2297/files) by using `rustc-args` instead of `rustdoc-args`.

GuillaumeGomez:
Ah yes indeed. Good catch! :)

PR #2688, opened 2023-10-20
Fix link from docs.rs to full API documentation

ChrisDenton:
Fixes https://github.com/microsoft/windows-rs/pull/2687#issuecomment-1772864560


kennykerr (approved):
Thanks!

kennykerr:
Looks like that worked! https://docs.rs/windows/latest/windows/

Issue #2694, opened 2023-11-04
windows-core crate should enable `implement` feature when building for docs.rs


MarijnS95:
### Suggestion

I tried very hard to link someone the `#[implement]` docs (there are none to begin with it seems: https://docs.rs/windows-implement/latest/windows_implement/attr.implement.html) via `windows-core` after deducing from `crates/samples/windows/bits/src/main.rs` that that is where this macro must reside / be reexported from... But couldn't find it.

As it turns out the `docs.rs` metadata doesn't turn on this feature (same for the `windows::core` docs generated at https://microsoft.github.io/windows-docs-rs/doc/windows/core/), can this be added so that the documentation is more complete? It could possibly use `doc_cfg`, and `doc_auto_cfg` _or_ a manual `#[doc(cfg(feature = "implement"))]` to make it clear that this is not provided by default.

---

Perhaps there are more features that are currently undocumented on docs.rs? I haven't checked.

kennykerr:
It's not really the `windows-core` crate. The `windows` crate documentation lacks this because it builds with `no-deps`.

https://github.com/microsoft/windows-docs-rs

Rust docs are generally miserable for large crates. Something I still need to deal with more fully.

MarijnS95:
`--no-deps` is mostly a local thing to not **build** documentation for dependencies in place, but it can (and should!) still link to "external" documentation.

For the `windows-core` (title of this issue) crate specifically, https://crates.io/crates/windows-core says that the official docs are at https://docs.rs/windows-core/latest/windows_core/, those could have been built _with_ the feature and link to https://docs.rs/windows-implement. Both crates are small.

For https://github.com/microsoft/windows-docs-rs you can specify multiple crates, e.g. `-p windows -p windows-core -p windows-implement` should get everything, together with --all-features`?

MarijnS95:
Note that I'm mostly filing this issue on request, otherwise I would have created a PR immediately that adds:

```toml
[package.metadata.docs.rs]
features = ["implement"]
rustdoc-args = ["--cfg", "docsrs"] # Optional, in case we do the below
```

To `core`'s `Cargo.toml`. And then see if the following works in its' `lib.rs`:

```rust
#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
```

kennykerr:
Not to worry. It's a little deeper than that as the macros are re-exported through the `windows` crate and those are hidden. I'll get it fixed up asap.

[...]

PR #2696, opened 2023-11-06
Document `implement` and `interface` macros

kennykerr:
Still plan to provide more comprehensive docs here (https://kennykerr.ca/rust-getting-started) but this should at least help folks get started. 

Fixes: #2694

### DIFF
--- a/crates/libs/windows/src/lib.rs
+++ b/crates/libs/windows/src/lib.rs
@@ -13,11 +13,9 @@ extern crate self as windows;
 pub mod core {
     pub use windows_core::*;
 
-    #[doc(hidden)]
     #[cfg(feature = "implement")]
     pub use windows_implement::implement;
 
-    #[doc(hidden)]
     #[cfg(feature = "implement")]
     pub use windows_interface::interface;
 }
--- a/crates/libs/implement/src/lib.rs
+++ b/crates/libs/implement/src/lib.rs
@@ -1,5 +1,29 @@
 use quote::{quote, ToTokens};
 
+/// Implements one or more COM interfaces.
+///
+/// # Example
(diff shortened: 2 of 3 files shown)


crates/libs/windows/src/lib.rs
```diff
@@ -13,11 +13,9 @@ extern crate self as windows;
...
     pub use windows_core::*;
 
-    #[doc(hidden)]
     #[cfg(feature = "implement")]
     pub use windows_implement::implement;
 
-    #[doc(hidden)]
     #[cfg(feature = "implement")]
     pub use windows_interface::interface;
 }
```
MarijnS95, line 21: Now that is cheeky :grin: and explains why I was trying to find an `implement` reexport in the `windows-core` crate.

Not to be a nuisance, but I wouldn't have done it this way in favour of not tricking people that `implement`/`interface` exist in the `windows-core` crate, whenever they might want to migrate from this `windows` crate. Instead, how about:

```rust
pub use windows_core as core;
pub use windows_implement::implement;
pub use windows_implement::interface;
```

?

kennykerr: These were originally in the `windows::core` module so this was just done for compatibility.

MarijnS95: Are there plans to break that backwards-compat at some point?

alexiokay:
guys i still dont know how to use it... i get 

```
#[implement(IAudioEndpointVolumeCallback)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ could not find `windows_core` in the list of imported crates
```

Issue #2388, opened 2023-03-16
Design: Arithmetic Overflow in the wrapper, how do we deal with it?


a4lg:
Some wrappers inside the `windows` crate takes one or more slices as arguments.

What I concern is, on current `windows` crate, `slice.len()` is converted to another type without checking validity against the target type.  As a result, it sometimes causes arithmetic overflow.

For instance, `MultiByteToWideChar` Win32 API uses `int` as the length type.

Quoting `crates/libs/windows/src/Windows/Win32/Globalization/mod.rs`:
```rust
#[inline]
pub unsafe fn MultiByteToWideChar(codepage: u32, dwflags: MULTI_BYTE_TO_WIDE_CHAR_FLAGS, lpmultibytestr: &[u8], lpwidecharstr: ::core::option::Option<&mut [u16]>) -> i32 {
    ::windows::imp::link ! ( "kernel32.dll""system" fn MultiByteToWideChar ( codepage : u32 , dwflags : MULTI_BYTE_TO_WIDE_CHAR_FLAGS , lpmultibytestr : :: windows::core::PCSTR , cbmultibyte : i32 , lpwidecharstr : :: windows::core::PWSTR , cchwidechar : i32 ) -> i32 );
    MultiByteToWideChar(codepage, dwflags, ::core::mem::transmute(lpmultibytestr.as_ptr()), lpmultibytestr.len() as _, ::core::mem::transmute(lpwidecharstr.as_deref().map_or(::core::ptr::null(), |slice| slice.as_ptr())), lpwidecharstr.as_deref().map_or(0, |slice| slice.len() as _))
}
```

OK, if we overflow `i32`, that would be a problem.

Following test code intentionally exploits this and try to pass `0x1_0000_0003`-byte sized buffer to `MultiByteToWideChar`.  It is handled as `3`-byte sized buffer in the wrapper and the result is corrupted.

For `windows-sys` crate, I don't find any issues.  The user must handle such conditions to use the API safely.  However, in `windows` crate, slice length type is hidden inside the wrapper and there's no safety measures.

I'm not sure what to do but I can say that *some* decision making will be required.

```rust
// Run this code on the 64-bit environment!

use windows::Win32::Globalization::*;
use windows::Win32::Foundation::*;

fn main() {
    // Japanese Shift_JIS (Microsoft variant)
    const CODE_PAGE: u32 = 932;
    // Your current ANSI code page (if 932 does not work)
[...]

PR #2699, opened 2023-11-10
Perform checked integral type conversions for WinRT APIs

kennykerr:
This update partially addresses #2388 by checking integer conversions in WinRT APIs. Passing a slice whose size exceeds `u32::MAX` will now return an error containing `E_INVALIDARG`. 

This doesn't completely cover #2388 where the example refers to a Win32 API. This fix is relatively simple to roll out for WinRT APIs since they all return a `Result`. Win32 APIs do not uniformly do so, so I'll address Win32 APIs separately, but this update at least proves out that general approach. 

I also don't have a good way to test this.


kennykerr:
Chatting with @riverar I think that returning an `HRESULT` may be confusing as it will be assumed to come from the API rather than being the caller's issue. I think I'll switch this to panic and then I can apply it to WinRT and Win32 APIs uniformly.

PR #2702, opened 2023-11-14
Add `windows-version` crate

kennykerr:
Getting operating system version information simply and reliably is a challenge given the sordid history of versioning APIs on Windows. A simple mechanism that does not require application manifest files is desirable in many instances, particularly for Rust apps that don't necessarily build with Visual Studio and all of its build and packaging conveniences. Here's a simple example:

```rust,no_run
use windows_version::*;

fn main() {
    println!("Current version: {:?}", OsVersion::current());

    if is_server() {
        println!("Running on a Windows Server release.");
    }

    if OsVersion::current() >= OsVersion::new(10, 0, 0, 12345) {
        println!("Can use a feature available on this version or later.")
    }
}
```


kennykerr:
Mainly unsure whether we need something a little less wordy than `if Version::new(10, 0, 0, 12345).ge() {` ... 

Perhaps a helper function like this: `if current_ge(10, 0, 0, 12345) {`

ChrisDenton:
Looks cool!

I suspect the two most common use cases for this would be

- Diagnostic purposes (e.g. reporting the version in bug reports)
- Testing if some feature/behaviour can be replied up

I think the former is sufficiently covered by `Version::current`.

Is there any reason not to implement `Ord` here? `Version::current() > Version::new(10, 0, 0, 12345)` would seem nice to have. Especially with some nice const aliases in place of `Version::new` (not necessarily defined in this crate).

riverar:
I think I would prefer more verbose names, favoring the clarity that brings over keystroke count.

```rust
OSVersion::is_greater_than(...)
OSVersion::is_less_than(...)
OSVersion::is_at_least(...)
```

kennykerr:
The derived implementation of `Ord` matches what I need:

https://doc.rust-lang.org/stable/std/cmp/trait.Ord.html#derivable

So that simplifies things.

In terms of verbosity, you can use the `Ord` and `PartialOrd` trait-provided methods explicitly to make things explicit if needed.

riverar:
> Windows usually discourages developers from using version checks to determine feature availability. Is there a reason why the Rust APIs should be any treated any differently from other platform APIs?

The lack of robust (or any) feature detection capabilities in Windows makes that advice impractical.

crates/libs/version/src/lib.rs
```diff
@@ -0,0 +1,132 @@
+#![cfg_attr(not(test), no_std)]
+
+mod bindings;
+use bindings::*;
+
+/// Operating system version information.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Version {
```
ChrisDenton, line 8: I think calling this `OsVersion` might be better? At least, it makes more sense to me when taken out of the context of this crate.

riverar (dismissed):

crates/libs/version/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    /// Hook used for testing `ge`.
+    #[cfg(test)]
+    fn current() -> Self {
+        test::test_current()
+    }
+}
+
+/// Determines if the currently running operating system is a Windows Server release.
+pub fn is_server() -> bool {
```
ChrisDenton, line 61: It feels like this should be a method so you can do `current().is_server()`. However, after thinking about it I'm not sure the extra complexity would be worth it for a single function. We'd need another struct, say `OsInfo`, with a method for getting the version.

ChrisDenton (commented):
This looks good to me. I have a few thoughts but nothing that I feel should block merging this.

crates/libs/version/src/lib.rs
kennykerr: Switched to `OsVersion`.

crates/libs/version/src/lib.rs
kennykerr: It's a bit of an odd pairing but I can't think of something better than a free function without turning this into a bigger "info" crate.

ChrisDenton: Yeah, I talked myself round to that point of view.

ChrisDenton (approved):

riverar (approved):

tim-weis:
> The lack of robust (or any) feature detection capabilities in Windows makes that advice impractical.

We have

* [Run-time dynamic linking](https://learn.microsoft.com/en-us/windows/win32/dlls/run-time-dynamic-linking) (alongside [delay-loaded DLLs](https://learn.microsoft.com/en-us/cpp/build/reference/linker-support-for-delay-loaded-dlls))
* [`IUnknown::QueryInterface`](https://learn.microsoft.com/en-us/windows/win32/api/unknwn/nf-unknwn-iunknown-queryinterface(refiid_void)) for COM and the WinRT
* [`ApiInformation`](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.metadata.apiinformation) for the WinRT

I'd used either of the above for feature detection, and things worked out as advertised.

@riverar Do you have an example where none of the above can be used or report unreliable results? I'm genuinely curious.

riverar:
@tim-weis Sure!

* An example where I workaround a bug across supported operating systems https://github.com/File-New-Project/EarTrumpet/blob/d6d9f6130178eec3d9d5a5fbcd19a91881966629/EarTrumpet/UI/Themes/Manager.cs#L24-L42
* An example where I check if I can use rounded corner features https://github.com/File-New-Project/EarTrumpet/blob/d6d9f6130178eec3d9d5a5fbcd19a91881966629/EarTrumpet/Extensions/WindowExtensions.cs#L32-L36
* An example where I light up the ability to move audio sessions between devices, introduced in RS4+ https://github.com/File-New-Project/EarTrumpet/blob/d6d9f6130178eec3d9d5a5fbcd19a91881966629/EarTrumpet/UI/ViewModels/AppItemViewModel.cs#L39C2-L40
* An example where I have to nudge our flyout window to align with UI changes in Windows 11 https://github.com/File-New-Project/EarTrumpet/blob/d6d9f6130178eec3d9d5a5fbcd19a91881966629/EarTrumpet/UI/Views/FlyoutWindow.xaml.cs#L169-L173

PR #2709, opened 2023-11-18
Add link to the generated documentation to often-seen readmes

seritools:
Just a tiny bit of friction I hit sometimes - there's actually no link to the generated documentation inside the readmes, and thus no easy way to get to the docs from GitHub either. The only link I can find is in the sidebar on crates.io. Usually googling (or hoping it's in the browser history) seems to be faster than finding it that way.

This PR just adds it to the main readme file, and to the readme for the windows crate itself.


ChrisDenton:
Adding a link in the readme of the `windows` crate makes some sense to me as it's at an unusual address. I'm less sure about adding it to the docs readme as that's meant to be an overview of all crates.

kennykerr:
There's also a link on `crates.io` for each crate, so this doesn't seem all that valuable. I try to keep the readmes as uncluttered as possible, so I'm reluctant to add convenience links like this.

seritools:
sounds good

Issue #2713, opened 2023-11-25
Add ability to override def impl of IInspectable::GetTrustLevel


riverar:
### Suggestion

Without this, in certain environments, WinRT component instantiation will result in the following error:

> Object's implementation of IInspectable::GetTrustLevel returned a value that is different from the trust level configured for the class in the catalog

PR #2714, opened 2023-11-27
Adds ability to override `GetTrustLevel` return value

kennykerr:
In rare cases, the return value should be changed to allow partial/full trust semantics for WinRT activation. This update introduces a generalized syntax for expressing this so that future implementation customization may be supported as needed. For example:

```rust
#[implement(IStringable, TrustLevel = Full)]
struct FullTrust;
```

Fixes: #2713

### DIFF
--- a/crates/libs/implement/src/lib.rs
+++ b/crates/libs/implement/src/lib.rs
@@ -74,6 +74,8 @@ pub fn implement(attributes: proc_macro::TokenStream, original_type: proc_macro:
         }
     });
 
+    let trust_level = proc_macro2::Literal::usize_unsuffixed(attributes.trust_level);
+
     let conversions = attributes.implement.iter().enumerate().map(|(enumerate, implement)| {
         let interface_ident = implement.to_ident();
         let offset = proc_macro2::Literal::usize_unsuffixed(enumerate);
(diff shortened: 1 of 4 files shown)


crates/libs/implement/src/lib.rs
```diff
@@ -336,6 +349,18 @@ impl syn::parse::Parse for UseTree2 {
...
+                    return Err(syn::parse::Error::new(ident.span(), "Unrecognized key-value pair"));
+                }
+                input.parse::<syn::Token![=]>()?;
+                let span = input.span();
+                let value = input.call(syn::Ident::parse_any)?;
+                match value.to_string().as_str() {
+                    "Partial" => Ok(UseTree2::TrustLevel(1)),
+                    "Full" => Ok(UseTree2::TrustLevel(2)),
+                    _ => Err(syn::parse::Error::new(span, "`TrustLevel` must be `Partial` or `Full`")),
+                }
```
riverar, line 363: Should we support the explicit specification of `BaseLevel` for folks that want to make it explicitly clear in their code?

```suggestion
                match value.to_string().as_str() {
                    "BaseLevel" => Ok(UseTree2::TrustLevel(0)),
                    "Partial" => Ok(UseTree2::TrustLevel(1)),
                    "Full" => Ok(UseTree2::TrustLevel(2)),
                    _ => Err(syn::parse::Error::new(span, "`TrustLevel` must be `BaseLevel`, `Partial`, or `Full`")),
                }
```

kennykerr: That just seems redundant and may confuse developers into thinking that maybe that's required. 🤷

riverar (approved):

PR #2727, opened 2023-12-06
Remove `invalid_mut` when `writer.std`

ChrisDenton:
This removes one more difference between `std` and `sys`.

The remaining two differences are in linking and handle type. The former should more or less solve itself once everyone is happy using raw-dylib. The latter requires getting either win32metadata or rust-lang/rust to shift their position on what type a `HANDLE` should be.


kennykerr (approved):
I'm not opposed to changing the definition of `HANDLE` in windows-rs...

ChrisDenton:
Ah, fair enough! Someone has to blink first at least 😉

riverar:
I am opposed to that, naturally.

Issue #2739, opened 2023-12-12
Deny `rust_2018_idioms` and `unused_qualifications` in crate roots


MarijnS95:
### Suggestion

When using a local `path` reference to the `windows-rs` crate, from a project that denies a bunch of lints globally (via `rustflags` in `.cargo/config.toml` until workspace lints get implicit/forced inheritance), there are quite a few warnings from the `windows` crate which would be trivial and neat to resolve.

Most/all originate from the `rust_2018_idioms` lint (and a few more like `unused_qualifications`, I think https://github.com/ash-rs/ash/blob/e6d80badc389d94e2a747f442e5ed4189b66d7d3/ash/src/lib.rs#L1-L8 might be a good starting point). This lint isn't enabled by default despite `windows` already setting `edition = "2021"`, and certain things like `extern crate` might be nice to clear out.

Some samples:

```rust
warning: unused extern crate
  --> windows-rs\crates\libs\windows\src\lib.rs:11:1
   |
11 | extern crate self as windows;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove it
   |
   = note: `-W unused-extern-crates` implied by `-W rust-2018-idioms`
   = help: to override `-W rust-2018-idioms` add `#[allow(unused_extern_crates)]`
```

```rust
warning: unnecessary qualification
  --> windows-rs\crates\libs\core\src\strings\bstr.rs:87:21
   |
87 |     fn from(value: &std::string::String) -> Self {
   |                     ^^^^^^^^^^^^^^^^^^^
   |
help: remove the unnecessary path segments
   |
87 -     fn from(value: &std::string::String) -> Self {
87 +     fn from(value: &String) -> Self {
   |
```

```rust
warning: hidden lifetime parameters in types are deprecated
   --> windows-rs\crates\libs\implement\src\lib.rs:241:34
    |
241 |     fn parse(cursor: syn::parse::ParseStream) -> syn::parse::Result<Self> {
    |                      ------------^^^^^^^^^^^
    |                      |
    |                      expected lifetime parameter
    |
    = note: `-W elided-lifetimes-in-paths` implied by `-W rust-2018-idioms`
    = help: to override `-W rust-2018-idioms` add `#[allow(elided_lifetimes_in_paths)]`
help: indicate the anonymous lifetime
    |
241 |     fn parse(cursor: syn::parse::ParseStream<'_>) -> syn::parse::Result<Self> {
    |                                             ++++
```

[...]

PR #2740, opened 2023-12-12
Fixes some non-default warnings when running clippy

kennykerr:
The following lint produces some additional suggestions fixed by this update. As mentioned in #2739 it would be nice to have a way to tighten up these checks automatically via yml workflows, but for now at least this takes care of the majority of the noise coming from this lint. 

`cargo clippy -- --warn rust_2018_idioms`

Fixes: #2739

### DIFF
--- a/crates/libs/bindgen/src/rdl/mod.rs
+++ b/crates/libs/bindgen/src/rdl/mod.rs
@@ -136,7 +136,7 @@ pub struct Interface {
 syn::custom_keyword!(interface);
 syn::custom_keyword!(class);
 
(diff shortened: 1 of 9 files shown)


MarijnS95:
Is this enough to not emit the warnings at all (for some crates)? If so, it might be time to add `#![warn(rust_2018_idioms)]` to those crates so that the same issues won't slip in again.

Don't forget about `unused_qualifications`!

kennykerr:
Will open a new PR for workspace lints.

PR #2741, opened 2023-12-12
Enable workspace lints

kennykerr:
As suggested in #2739, I'm enabling workspace lints to more easily enable non-default Rust and Clippy lints.

I've started with only `elided-lifetimes-in-paths` as some of the others, like `rust_2018_idioms`, has some false positives that prevent them from being turned on for this workspace.

Fixes: #2739

### DIFF
--- a/crates/libs/sys/src/lib.rs
+++ b/crates/libs/sys/src/lib.rs
@@ -7,7 +7,6 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 #![allow(non_snake_case, non_upper_case_globals, non_camel_case_types, clippy::all)]
 #![cfg_attr(not(feature = "docs"), doc(hidden))]
 
-extern crate self as windows_sys;
 pub mod core;
 
 include!("Windows/mod.rs");
--- a/crates/libs/sys/src/lib.rs
+++ b/crates/libs/sys/src/lib.rs
@@ -7,6 +7,9 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 #![allow(non_snake_case, non_upper_case_globals, non_camel_case_types, clippy::all)]
 #![cfg_attr(not(feature = "docs"), doc(hidden))]
 
+#[allow(unused_extern_crates)]
+extern crate self as windows_sys;
+
 pub mod core;
 
 include!("Windows/mod.rs");
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,4 +13,4 @@ exclude = [
 ]
 
 [workspace.lints.rust]
-elided-lifetimes-in-paths = "warn"
+rust_2018_idioms = "warn"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,3 +11,6 @@ members = [
 exclude = [
     "crates/targets/baseline",
 ]
+
+[workspace.lints.rust]
+elided-lifetimes-in-paths = "warn"
--- a/crates/libs/bindgen/Cargo.toml
+++ b/crates/libs/bindgen/Cargo.toml
@@ -9,6 +9,9 @@ description = "Windows metadata compiler"
 repository = "https://github.com/microsoft/windows-rs"
 readme = "readme.md"
 
+[lints]
+workspace = true
+
 [package.metadata.docs.rs]
 default-target = "x86_64-pc-windows-msvc"
 targets = []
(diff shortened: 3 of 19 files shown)


ChrisDenton:
If it's only a single case, would adding an `#[allow(unused_extern_crates)] // false positive` be worth it so you still get everything else linted?

kennykerr:
Seems to work!

ChrisDenton:
Is the reason for the false positive that CI doesn't always use features that use it? E.g. clippy is run with no features at all https://github.com/microsoft/windows-rs/blob/831b99b240505c8a6b02f1cd8fba5ecc0a32c351/.github/workflows/clippy.yml#L146

ChrisDenton (approved):
In any case, this PR lgtm.

kennykerr:
> Is the reason for the false positive that CI doesn't always use features that use it? E.g. clippy is run with no features at all
> 
> https://github.com/microsoft/windows-rs/blob/831b99b240505c8a6b02f1cd8fba5ecc0a32c351/.github/workflows/clippy.yml#L146

Yes, it isn't used in all cases. It is used by some features and by the optional `implement` and `interface` macros.

crates/libs/sys/src/lib.rs
```diff
@@ -7,7 +7,9 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 #![allow(non_snake_case, non_upper_case_globals, non_camel_case_types, clippy::all)]
 #![cfg_attr(not(feature = "docs"), doc(hidden))]
 
+#[allow(unused_extern_crates)]
 extern crate self as windows_sys;
```
MarijnS95, line 11: It does not seem that this is needed for anything; the lint is valid. There is nothing inside `sys` / `windows` referring to `(::)windows_sys::` or `(::)windows::` respectively. Could these lines just be removed altogether?

MarijnS95: Err, I meant to say that:

1. On `windows_sys` I don't seem to get a false-positive here, because it's actually used;
2. On `windows` I only get a few errors originating from the `windows_implement` macro, which expects the `windows` crate to be available as an external crate (which is interesting for a macro that generates code to be used both inside and outside of a crate).

ChrisDenton:
```
> cargo clippy -p windows-sys
    Checking windows-sys v0.52.0 (R:\windows-rs\crates\libs\sys)
warning: unused extern crate
  --> crates\libs\sys\src\lib.rs:11:1
   |
11 | extern crate self as windows_sys;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove it
   |
   = note: `-W unused-extern-crates` implied by `-W rust-2018-idioms`
   = help: to override `-W rust-2018-idioms` add `#[allow(unused_extern_crates)]`

warning: `windows-sys` (lib) generated 1 warning (run `cargo clippy --fix --lib -p windows-sys` to apply 1 suggestion)
    Finished dev [unoptimized + debuginfo] target(s) in 2m 49s
```

MarijnS95: See the `console` highlight:

> ```console
> > cargo clippy -p windows-sys
>     Checking windows-sys v0.52.0 (R:\windows-rs\crates\libs\sys)
> warning: unused extern crate
>   --> crates\libs\sys\src\lib.rs:11:1
>    |
> 11 | extern crate self as windows_sys;
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove it
>    |
>    = note: `-W unused-extern-crates` implied by `-W rust-2018-idioms`
>    = help: to override `-W rust-2018-idioms` add `#[allow(unused_extern_crates)]`
> 
> warning: `windows-sys` (lib) generated 1 warning (run `cargo clippy --fix --lib -p windows-sys` to apply 1 suggestion)
>     Finished dev [unoptimized + debuginfo] target(s) in 2m 49s
> 
> > cargo clippy -p windows-sys --features Win32
>     Checking windows-sys v0.52.0 (R:\windows-rs\crates\libs\sys)
>     Finished dev [unoptimized + debuginfo] target(s) in 1.41s
> ```

Indeed, it is conditionally used :(

kennykerr:
> On windows I only get a few errors originating from the windows_implement macro, which expects the windows crate to be available as an external crate (which is interesting for a macro that generates code to be used both inside and outside of a crate).
EDIT: It's actually including ::windows::core::. Wasn't there a request before to use ::windows_core:: here so that the macro can also be used without the windows crate (at the downside that windows+implement users now need both crates)?

If there's a simpler solution, I'd love to hear it. I'd like to make the macros depend only on the `windows_core` crate but that makes it less convenient for what I'd consider is the majority of developers.

PR #2743, opened 2023-12-13
Enable workspace lint `unused_qualifications`

kennykerr:
Following on from #2741, this was a bit painful as I was relying on a lot of wildcard `use` declarations. A bit wordier now but also less ambiguous. Clippy was completely unable to auto fix most of this. 

Fixes: #2739


Cargo.toml
```diff
@@ -14,3 +14,4 @@ exclude = [
 
 [workspace.lints.rust]
 rust_2018_idioms = "warn"
+unused_qualifications = "warn"
```
ChrisDenton, line 17: Github is complaining about the lack of trailing new line. Not that it particularly matters, unless you're a POSIX purist 😆.

kennykerr: Fixed! 🤷

Issue #2748, opened 2023-12-17
Please improve docs for `HSTRING::len()`


Enyium:
### Suggestion

Currently, [it says](https://microsoft.github.io/windows-docs-rs/doc/windows/core/struct.HSTRING.html#method.len):

> `pub fn len(&self) -> usize`
> 
> Returns the length of the string.

Does this mean the number of wide chars (excluding the terminating null)? So, Unicode codepoints above U+FFFF are counted as two? When reading the docs, you should know exactly what's up.

tim-weis:
Would you be fine with

> Returns the length (in code units) of the string.

?

kennykerr:
The length is the number of UTF-16 characters, not including the terminating null character.

ChrisDenton:
Rust's [`String::len`](https://doc.rust-lang.org/std/string/struct.String.html#method.len) is documented as so:

> Returns the length of this `String`, in bytes, not [`char`](https://doc.rust-lang.org/std/primitive.char.html)s or graphemes. In other words, it might not be what a human considers the length of the string.

Change "bytes" for "`u16`s" and it's pretty much the same.

Enyium:
> The length is the number of UTF-16 characters

"characters" doesn't seem to be quite correct and not helpful when trying to understand more than the convention "We just think of all of these units as characters, so they're conveniently the same size."

> Would you be fine with
>
> > Returns the length (in code units) of the string.

Better.

> Rust's [`String::len`](https://doc.rust-lang.org/std/string/struct.String.html#method.len) is documented as so:
> 
> > Returns the length of this `String`, in bytes, not [`char`](https://doc.rust-lang.org/std/primitive.char.html)s or graphemes. In other words, it might not be what a human considers the length of the string.
> 
> Change "bytes" for "`u16`s" and it's pretty much the same.

The best approach.

Combined:

```
Returns the length of the string in `u16`s, i.e., UTF-16 code points, not `char`s or graphemes. It might not be what a human considers the length of the string.
```

kennykerr:
Feel free to contribute something along those lines.

kennykerr:
Agree with Tim. Here's a minor clarification: #2749

PR #2749, opened 2023-12-19
Clarify `HSTRING::len` documentation

kennykerr:
Fixes: #2748

### DIFF
--- a/crates/libs/core/src/strings/hstring.rs
+++ b/crates/libs/core/src/strings/hstring.rs
@@ -19,7 +19,7 @@ impl HSTRING {
         self.0.is_none()
     }
 
-    /// Returns the length of the string.
+    /// Returns the length of the string. The length is measured in `u16`s, not including the terminating null character.
     pub fn len(&self) -> usize {
         if let Some(header) = self.get_header() {
             header.len as usize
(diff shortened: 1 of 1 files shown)


crates/libs/core/src/strings/hstring.rs
```diff
@@ -19,7 +19,7 @@ impl HSTRING {
         self.0.is_none()
     }
 
-    /// Returns the length of the string.
+    /// Returns the length of the string. The length is measured in `u16`s, not including the terminating null character.
```
riverar, line 22: I think it's important to additionally indicate these are UTF-16 code units, otherwise looks good to me.


```suggestion
    /// Returns the length of the string. The length is measured in `u16`s (UTF-16 code units), not including the terminating null character.
```

crates/libs/core/src/strings/hstring.rs
```diff
@@ -19,7 +19,7 @@ impl HSTRING {
         self.0.is_none()
     }
 
-    /// Returns the length of the string.
+    /// Returns the length of the string. The length is measured in `u16`s (UTF-16 characters), not including the terminating null character.
```
riverar, line 22:
```suggestion
    /// Returns the length of the string. The length is measured in `u16`s (UTF-16 code units), not including the terminating null character.
```

Example:
```cpp
// Character 𐐏 is represented with two UTF-16 code units 0xD801 0xDC0F
winrt::hstring(L"𐐏").size() == 2
```

riverar (approved):

PR #2757, opened 2023-12-26
Remove unnecessary std usage from bindgen in writer.rs

ajw221:
Recently ran into some issues while attempting to use `windows-bindgen` in a `no_std` crate. The generated code contains almost all `::core` usage except for a few areas where `::std` is used instead.

### DIFF
--- a/crates/libs/bindgen/src/rust/writer.rs
+++ b/crates/libs/bindgen/src/rust/writer.rs
@@ -193,10 +193,10 @@ impl Writer {
                 quote! { *mut ::core::ffi::c_void }
             }
             metadata::Type::String => {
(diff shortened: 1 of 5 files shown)


ajw221:
> @microsoft-github-policy-service agree

kennykerr:
Maybe we need a dedicated test crate that "proves" this works for `no_std` to avoid future regressions, unless all of `test_standalone` can be marked `no_std`.

kennykerr:
This would also depend on the `windows-core` being `no_std`. While I think this is a good idea in general, this will require a bit more of a systematic approach. Why don't you start by creating an issue and we can discuss.

PR #2774, opened 2024-01-04
Relax the build target checks in windows-targets sub-crates

seritools:
I've done this change for Rust9x initially, but realized that the [recently-added tier3 `win7` targets](https://github.com/rust-lang/rust/pull/118150) also fail to link with the strict check in the target crates:

```plain
> cargo +nightly build --target x86_64-win7-windows-msvc -Zbuild-std
[...]

error: linking with `link.exe` failed: exit code: 1181
  |
  = note: "C:\\Program Files\\Microsoft Visual Studio\\2022\\Preview\\VC\\Tools\\MSVC\\14.39.33321\\bin\\HostX64\\x64\\link.exe" "/NOLOGO" [...]
  = note: LINK : fatal error LNK1181: cannot open input file 'windows.0.52.0.lib'
```

This change aligns the checks for the msvc targets with the checks in `crates\libs\targets\Cargo.toml`.
I didn't touch the gnu/gnullvm ones since I'm unclear about `target_abi` actually being stable/checkable yet. LMK if I should add the respective checks.

Also please feel free to close if this doesn't align with windows-rs. For example, technically it could be interpreted as support for unsupported target vendors.

CC @roblabla

### DIFF
--- a/crates/targets/aarch64_msvc/build.rs
+++ b/crates/targets/aarch64_msvc/build.rs
@@ -1,6 +1,8 @@
 fn main() {
-    let target = std::env::var("TARGET").unwrap();
-    if target != "aarch64-pc-windows-msvc" && target != "aarch64-uwp-windows-msvc" {
+    let family = std::env::var("CARGO_CFG_TARGET_FAMILY").unwrap();
+    let arch = std::env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let env = std::env::var("CARGO_CFG_TARGET_ENV").unwrap();
+    if family != "windows" || arch != "aarch64" || env != "msvc" {
         return;
     }
 
--- a/crates/targets/aarch64_gnullvm/build.rs
+++ b/crates/targets/aarch64_gnullvm/build.rs
@@ -1,6 +1,9 @@
 fn main() {
+    let family = std::env::var("CARGO_CFG_TARGET_FAMILY").unwrap();
+    let arch = std::env::var("CARGO_CFG_TARGET_ARCH").unwrap();
+    let env = std::env::var("CARGO_CFG_TARGET_ENV").unwrap();
     let target = std::env::var("TARGET").unwrap();
-    if target != "aarch64-pc-windows-gnullvm" {
+    if family != "windows" || arch != "aarch64" || env != "gnu" || !target.ends_with("-gnullvm") {
         return;
     }
 
--- a/crates/targets/i686_gnu/build.rs
+++ b/crates/targets/i686_gnu/build.rs
@@ -1,6 +1,8 @@
 fn main() {
-    let target = std::env::var("TARGET").unwrap();
-    if target != "i686-pc-windows-gnu" && target != "i686-uwp-windows-gnu" {
+    let family = std::env::var("CARGO_CFG_TARGET_FAMILY").unwrap();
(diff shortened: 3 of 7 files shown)


crates/targets/aarch64_msvc/build.rs
```diff
@@ -1,6 +1,8 @@
 fn main() {
-    let target = std::env::var("TARGET").unwrap();
-    if target != "aarch64-pc-windows-msvc" && target != "aarch64-uwp-windows-msvc" {
```
ChrisDenton, line 3: Your patch seems to be missing the UWP special case?

ChrisDenton (commented):
Being less sensitive to the precise target name seems like it might lessen churn in the future. Obviously win9x support is off-label use of this crate but if it works then fair enough.

crates/targets/aarch64_msvc/build.rs
ChrisDenton: Oh, wait `UWP` is a "vendor" so it'll work. Never mind then!

seritools: 🤔 I'm not seeing it, it's not checking for `uwp` because it doesn't check the vendor at all now, so all of `(pc, win7, uwp, rust9x)` would pass there.

ChrisDenton: Ha, I got there eventually!

ChrisDenton:
My only other comment is that if this is worth doing then it's worth doing for all the target crates, not just the msvc ones.

seritools:
Agreed, just wasn't sure about the target_abi checks. If someone could chime in there I'll update it asap!

ChrisDenton:
Ah, unfortunately `cfg_target_abi` is not stable yet https://github.com/rust-lang/rust/issues/80970

So it'd need to be parsed from the target string.

seritools:
That's what I thought as well, but wasn't sure anymore after seeing https://github.com/microsoft/windows-rs/blob/a6bdfc873e9e9b7a3018830e1c243ec399555853/crates/libs/targets/Cargo.toml#L28

Anyways, would a check for `.ends_width("-gnullvm")` be good enough for now, then?

seritools:
I went ahead and added those changes as a separate commit 👍

kennykerr:
It's quite a mess since (a) `cfg_target_abi` isn't stable and (b) gnu/llvm in general seems to be quite a mess. (#2515).

ChrisDenton:
gnu/llvm is potentially getting promoted to a tier 2 target soon so I'll try to push for stabilization of `target_abi`.

riverar:
Yikes, these conditionals are hard to review without loading up a bunch of toolchain context. If the tests pass, that's great. But it would be helpful to see table of supported targets and the attributes being used to identify them to understand the composition. (Not a blocker here.)

riverar (approved):

kennykerr:
Documented here: https://doc.rust-lang.org/cargo/reference/environment-variables.html

Be good to have some targeted yml workflows that verify each target.

kennykerr (approved):

evmar:
Another positive consequence of this change is that you can now use windows-sys with `i586-pc-windows-msvc`, where before it would filter specifically on `i686`.

(I discovered this PR because I was trying to figure out why the released windows-sys didn't work for me even though the code at Git head did...)

PR #2775, opened 2024-01-05
Simplified parameter bindings and type projection

### LINKED ISSUE #2759: Automatic conversion from WinRT class to implemented interface
### Use case
Currently, the following code does not compile:
```rust
let vector: Collections::IVector<IStringable> = ...;
let uri = Uri::CreateUri(h!("https://github.com"))?;
vector.Append(&uri)?;
       ------  ^^^ the trait `CanInto<IStringable>` is not implemented for `Uri`
       |
       required by a bound introduced by this call
= note: required for `&windows::Foundation::Uri` to implement `IntoParam<windows::Foundation::IStringable, ReferenceType>`
```
Instead, one has to manually cast the `Uri` to a `IStringable`. This should not be necessary, because  it is guaranteed that [`Uri` implements `IStringable`](https://learn.microsoft.com/en-us/uwp/api/windows.foundation.uri?view=winrt-22621).

I understand that `Uri` implements `CanTryInto<IStringable>`, which makes it compatible with `TryIntoParam<IStringable>` arguments, but `TryIntoParam` arguments only seem to be used for a few methods. 

The same problem also exists for WinRT interfaces which require another interface. 

### Suggestion
I'm not sure if I'm missing something, but I would suggest getting rid of the distinction between `CanInto`/`IntoParam` and `CanTryInto`/`TryIntoParam`, replacing the latter with the former. 
This would require changes to `CanInto` to support conversions using `QueryInterface`, but that should not be a problem, as the conversions are guaranteed to succeed. 

---
I would be happy to hear any opinion on this; whether the proposal makes sense or if there are reasons for not doing this and keeping the current solution.

kennykerr:
This update simplifies a few notable areas of code generation and type projection:

* The distinction between compile- and run-time convertible parameters has been removed from the generated parameter bindings and pushed down into the type traits themselves. Practically that means that there is no longer the confusing difference between the `IntoParam` and `TryIntoParam` constraints. 

* The `ComInterface` and `Interface` traits have been combined. The distinction only matters for a small number of non-`IUnknown` interfaces, but the divide caused a great deal of complexity. The single `Interface` trait now handles both. The `cast` method will however `panic` if called from a non-`IUnknown` interface for obvious reasons. The branching is a compile-time decision so the unused branch should be compiled away.

* A lot more of the boilerplate code around COM interface declarations is now hidden behind simplified macros which makes the code a lot easier to read and substantially more concise.  

This also fixes #2759


crates/libs/core/src/interface.rs
```diff
@@ -1,11 +1,15 @@
...
 ///
 /// This trait is automatically implemented by the generated bindings and should not be
 /// implemented manually.
 ///
 /// # Safety
-pub unsafe trait Interface: Sized {
+pub unsafe trait Interface: Sized + Clone {
     type Vtable;
+    const IID: GUID;
+    const UNKNOWN: bool = true;
```
kaivol, line 12: Would it make sense to combine these into an `Option<GUID>`?

crates/libs/core/src/param.rs
```diff
@@ -17,68 +17,10 @@ impl<T: Type<T>> Param<T> {
...
-    }
-}
-
-#[doc(hidden)]
-pub trait CanTryInto<T>: ComInterface
-where
-    T: ComInterface,
-{
-    const CAN_INTO: bool = false;
+pub trait CanInto<T>: Sized {
```
kaivol, line 20: If I understand correctly, the `CanInto` trait is used for conversions between Interfaces (through `QueryInterface` or direct) and also for reusable handles.
To avoid confusion, I think we could use different traits for these purposes, with possibly more expressive names (something like `ImplementsInterface` and `AlsoUsableForHandle` ?).

crates/libs/core/src/interface.rs
```diff
@@ -63,6 +67,36 @@ pub unsafe trait Interface: Sized {
...
+    /// Call `QueryInterface` on this interface
+    ///
+    /// # Safety
+    ///
+    /// `interface` must be a non-null, valid pointer for writing an interface pointer.
+    unsafe fn query(&self, iid: *const GUID, interface: *mut *mut std::ffi::c_void) -> HRESULT {
+        if Self::UNKNOWN {
+            (self.assume_vtable::<IUnknown>().QueryInterface)(self.as_raw(), iid, interface)
+        } else {
+            panic!()
```
kaivol, line 97: Should probably add a meaningful message here

crates/libs/core/src/param.rs
kennykerr: Feel free to elaborate on a new issue. I'm not sure I want to add more traits than needed, but I'm open to ideas.

crates/libs/core/src/interface.rs
kennykerr: It could be made to work, but it would again complicate the nominal case of COM interfaces.

kaivol: Sorry, didn't knew that. I was always under the impression that both (deriving from `IUnknown` and having a unique IID) would be required for a COM interface. 

I hope you don't mind me asking, but do you happen to know if such interfaces exist in the Windows API? I am curious in which situations this would be useful.

kennykerr: Sure, there are probably a few dozen in the Windows SDK. They're generally used as callback interfaces where some caller may provide the implementation, but it doesn't necessarily represent an interface on a COM object that you'd query for.

kaivol: Just to make sure I'm not misunderstanding, do any of these actually implement `IUnknown`?
I know about `IXAudio2VoiceCallback` or `ISpNotifyCallback`, for example, but these do not inherit from `IUnknown` (and are therefore not considered to be COM interfaces (?)).

Also, what would be the reason for such an callback interface to implement `IUnknown` if you're not supposed to query for it? Just the `AddRef`/`Release` ref-counting functionality?

kennykerr: Yes, lifetime management.

PR #2778, opened 2024-01-06
Truncate standalone test files instead of removing

### LINKED ISSUE #2777: Opening this crate in vscode on Linux leads to files being changed and an endless build loop
### Summary

To reproduce, download this crate and open it in vscode on Linux with rust-analyzer installed. (I did that since I needed to grep for some things the crate does to research Windows-related issues in my code, and I don't have a Windows machine.) This leads to changes showing up in `git diff`, and rust-analyzer hangs in an endless build loop. I reported this with RA at https://github.com/rust-lang/rust-analyzer/issues/16250 but they are saying there isn't much they can do; the reason is that this crate modifies its own source in a build script which causes a reload which causes a build which modifies the source again ad nauseam.

### Crate manifest

_No response_

### Crate code

_No response_

ChrisDenton:
Attempts to fix #2777 by truncating the test files instead of removing them. This may help avoid an endless loop in rust-analyzer (see also https://github.com/rust-lang/rust-analyzer/issues/16250).

cc @RalfJung would you be able to test if this helps?

### DIFF
--- a/crates/tests/standalone/build.rs
+++ b/crates/tests/standalone/build.rs
@@ -167,7 +167,7 @@ fn write_no_inner_attr(output: &str, filter: &[&str]) {
 fn riddle(output: &str, filter: &[&str], config: &[&str]) {
     // Rust-analyzer may re-run build scripts whenever a source file is deleted
     // which causes an endless loop if the file is deleted from a build script.
-    // To workaroud this, we truncate the file instead of deleting it.
+    // To workaround this, we truncate the file instead of deleting it.
     // See https://github.com/microsoft/windows-rs/issues/2777
     _ = std::fs::File::options()
         .truncate(true)
--- a/crates/tests/standalone/build.rs
+++ b/crates/tests/standalone/build.rs
@@ -165,7 +165,14 @@ fn write_no_inner_attr(output: &str, filter: &[&str]) {
 }
 
 fn riddle(output: &str, filter: &[&str], config: &[&str]) {
-    _ = std::fs::remove_file(output);
+    // Rust-analyzer may re-run build scripts whenever a source file is deleted
+    // which causes an endless loop if the file is deleted from a build script.
+    // To workaroud this, we truncate the file instead of deleting it.
+    // See https://github.com/microsoft/windows-rs/issues/2777
+    _ = std::fs::File::options()
+        .truncate(true)
+        .write(true)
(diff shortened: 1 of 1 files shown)


RalfJung:
Yes that does seem to fix the endless loop. :)

kennykerr:
I can't repro that nightly warning, but I fixed it here: https://github.com/microsoft/windows-rs/pull/2775/commits/b5e5000f4b0aa2b029626397d42fe8d91463b997

If you merge master you should be good to go.

kennykerr (dismissed):
Maybe a comment about this being a workaround for rust-analyzer as I'm bound to forget why this is used rather than the simpler `remove_file`. 😊

ChrisDenton:
Thanks for testing! I've added a comment to explain why we're doing this.

crates/tests/standalone/build.rs
```diff
@@ -165,7 +165,14 @@ fn write_no_inner_attr(output: &str, filter: &[&str]) {
 }
 
 fn riddle(output: &str, filter: &[&str], config: &[&str]) {
-    _ = std::fs::remove_file(output);
+    // Rust-analyzer may re-run build scripts whenever a source file is deleted
+    // which causes an endless loop if the file is deleted from a build script.
+    // To workaroud this, we truncate the file instead of deleting it.
+    // See #2777.
```
kennykerr, line 171:
```suggestion
    // See https://github.com/microsoft/windows-rs/issues/2777.
```

Just makes it easier to open the issue in VS Code.

ChrisDenton: Good point!

riverar (dismissed):

crates/tests/standalone/build.rs
```diff
@@ -165,7 +165,14 @@ fn write_no_inner_attr(output: &str, filter: &[&str]) {
 }
 
 fn riddle(output: &str, filter: &[&str], config: &[&str]) {
-    _ = std::fs::remove_file(output);
+    // Rust-analyzer may re-run build scripts whenever a source file is deleted
+    // which causes an endless loop if the file is deleted from a build script.
+    // To workaroud this, we truncate the file instead of deleting it.
```
kennykerr, line 170:
```suggestion
    // To workaround this, we truncate the file instead of deleting it.
```

Typo.

Issue #2764, opened 2024-01-01
For COM interfaces T and U, if T is derived from U, implement From<T> for U


chloekek:
### Suggestion

The `interface_hierarchy` macro and the `CanInto` trait enable upcasting of COM interfaces, but this only appears to be used for arguments to Windows API functions (through the undocumented `IntoParam` trait).

It would be useful to additionally support upcasting of COM interfaces directly. This is always safe and infallible, so `ComInterface::cast` is not suitable. Example code that would compile with this change:

```rust
use windows::Win32::Graphics::Dxgi::{IDXGISwapChain, IDXGISwapChain1};

fn f(a: IDXGISwapChain1) -> IDXGISwapChain
{
    a.into()
}
```

Note that this would be a zero-cost conversion; no clone or call to `QueryInterface` necessary.

I think the implementation would be as simple as having `interface_hierarchy` generate transmuting `From` impls: https://github.com/microsoft/windows-rs/commit/9057d9ea688540399defc7f8b24901fb64643a4a but maybe there is more to it that I am not aware of.

tim-weis:
> maybe there is more to it that I am not aware of

There is, indeed. In COM Theory™, reference counts are strictly per interface. When transmuting from one (owning) representation into a different (owning) representation, `IUnknown`'s [`impl Drop`](https://github.com/microsoft/windows-rs/blob/master/crates/libs/core/src/unknown.rs#L36-L42) is inevitably going to cause havoc.

In practice, this frequently pans out, even if wrong. While I haven't made up my mind whether 2024 is going to be the year when I stop giving a damn, I'll proactively suggest not introducing this.

This isn't to say that ergonomics for upcasting COM interfaces couldn't be improved. It's just that—this time—the simple solution won't do. I'm confident some sharp minds will review this and procure a workable solution.

Happy New Year, regardless!

chloekek:
The move constructor for Microsoft’s own `ComPtr` type (i.e. owning COM interfaces) in C++ seems to suggest otherwise. It implements an upcast simply by copying the pointer: https://github.com/tpn/winsdk-10/blob/9b69fd26ac0c7d0b83d378dba01080e93349c2ed/Include/10.0.14393.0/winrt/wrl/client.h#L272

tim-weis:
Yes, that's the same bug. You'd need to instantiate a COM object that implements a tear-off interface to exercise the error mode.

Raymond Chen published several blog posts on this topic:

[...]

PR #2779, opened 2024-01-08
Implement `From<T>` and `From<&T>` for interface hierarchies

kennykerr:
This can avoid calls to `QueryInterface` or unsafe code to do the same. 

Fixes: #2764

### DIFF
--- a/crates/tests/implement/tests/identity_from.rs
+++ b/crates/tests/implement/tests/identity_from.rs
@@ -1,0 +1,97 @@
+#![allow(non_snake_case)]
+
+use windows::{core::*, Foundation::*};
+
+#[implement(IStringable, IClosable)]
+struct Test;
+
+impl IStringable_Impl for Test {
+    fn ToString(&self) -> Result<HSTRING> {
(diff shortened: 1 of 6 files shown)


crates/tests/implement/tests/identity_from.rs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        let stringable: IStringable = Test.into();
+
+        let inspectable = Into::<&IInspectable>::into(&stringable);
+        assert_eq!(
+            inspectable.GetRuntimeClassName()?,
+            "Windows.Foundation.IStringable"
+        );
+
+        let inspectable = Into::<IInspectable>::into(stringable.clone());
```
MarijnS95, line 51: The canonical way I've seen people call typed `From`/`Into` implementations is via `IInspectable::from()`, and the matching `<&IInspectable>::from(&...)`.

kennykerr: Well then somebody should write a clippy lint. 😊

Issue #539, opened 2021-02-12
Support for VARIANT data type


1Dragoon:
Great, glad you're unblocked! 

Yes, VARIANT support is coming. Much like BSTR, there will likely be a helper struct named Variant that provides friendlier access to the underlying union.

_Originally posted by @kennykerr in https://github.com/microsoft/windows-rs/issues/536#issuecomment-778302738_

kennykerr:
Thanks! This depends on #417, which I'll be wrapping up soon.

kennykerr:
Unions are now supported. VARIANT though has union members that need to be dropped. While the union's memory layout is now correct, it doesn't provide support for dropping, so I'll leave this issue open until I add that.

iainnicol:
Wrapping members in ManuallyDrop, as you mention in the other issue, sounds plausible.

But in the specific case of VARIANT, should we implement Drop manually, to call VariantClear?

dpaoliello:
One more request to finish fleshing out `VARIANT`, please implement the `Default` trait as well - this makes it easier (and consistent with other types) to create one to pass-by-pointer to a function.

mwcampbell:
Also eagerly awaiting this. Will be very helpful for my upcoming implementation of UI Automation provider interfaces in [AccessKit](https://github.com/AccessKit/accesskit).

kennykerr:
I'm getting closer to being able to tackle this. 😉

rgwood:
A related issue I've had in this area: [propvarutil.h](https://docs.microsoft.com/en-us/windows/win32/api/propvarutil/) has a bunch of VARIANT helpers but it seems like they're not all present in `windows-rs`. 

For example, [`InitVariantFromInt32Array()` exists](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/Shell/PropertiesSystem/fn.InitVariantFromInt32Array.html) but it seems like [`InitVariantFromInt32()`](https://docs.microsoft.com/en-us/windows/win32/api/propvarutil/nf-propvarutil-initvariantfromint32) does not. Not sure why.

riverar:
@rgwood See the Remarks on that page and [this FAQ entry](https://github.com/microsoft/windows-rs/blob/master/docs/FAQ.md#why-are-some-equivalents-to-cc-macros-from-the-windows-sdk-missing). 👍
> This is an inline function, with its source code provided in the header.

hcldan:
[...]

PR #2780, opened 2024-01-09
Draft support for `VARIANT` and `PROPVARIANT`

kennykerr:
Hey folks, finally spending some time on `VARIANT` and `PROPVARIANT`. This is not quite done but do give me some feedback on the shape of things. Does this seem useful? Exhaustive coverage of the variant types may not be practical, but would this still be useful with partial support?

Fixes: #539


mwcampbell:
To replace [AccessKit](https://github.com/AccessKit/accesskit)'s current code for constructing `VARIANT`s, this would also need to support `VT_UNKNOWN`. I suppose the next step is to make the generated bindings use the new `VARIANT` wrapper rather than the raw struct, making sure to get ownership transfer right, particularly when a method in a user-implemented interface returns a `VARIANT`, as with `IRawElementProviderSimple::GetPropertyValue` which AccessKit implements.

mwcampbell:
I forgot to say, thanks for this and for all your work on windows-rs. I have no immediate need for this, as I already wrote the ugly code to create `VARIANT`s in AccessKit, but I'm happy to see safe and convenient support for more aspects of the Windows API.

kennykerr:
Thanks @mwcampbell, yes `VT_UNKNOWN` support is on my list. 👍

kennykerr:
Final PR coming soon.

kennykerr:
Final PR is here: #2786

PR #2786, opened 2024-01-10
Add `VARIANT` and `PROPVARIANT` support

kennykerr:
This update adds first-class support for `VARIANT` and `PROPVARIANT` to the `windows-core` crate and by extension the `windows` crate. This does not affect the `windows-sys` crate.

Creating a variant is now very simple:

```rust
let v = VARIANT::from(345u16);
```

The `TryFrom` trait can be used to query for an expected type:

```rust
assert_eq!(u16::try_from(&v)?, 345u16);
```

Most primitive types, including `BSTR`, are supported:

```rust
let v = VARIANT::from(BSTR::from("hello"));
assert_eq!(BSTR::try_from(&v)?, "hello");
```

Natural type coercion is also support:

```rust
let v = VARIANT::from(3.5f64);
assert_eq!(BSTR::try_from(&v)?, "3.5");
```

Standard traits like `Drop`, `Clone`, `PartialEq`, and `Debug` are provided:

```rust
assert_eq!(format!("{v:?}"), "VARIANT { type: 5, value: 3.5 }");
```

Fixes: #539


mwcampbell:
For what it's worth, if this is released in its current state, AccessKit won't be able to upgrade to this version, unless there's some escape hatch I can still use to create a `VARIANT` from an `IUnknown` and return that `VARIANT` in an implemented interface method.

kennykerr:
#2788 adds support for `IUnknown` - is that what you mean?

mwcampbell:
Yes, thanks.

PR #2791, opened 2024-01-12
Harden generic type parameter binding

kennykerr:
Some minor housekeeping around generic type parameter binding.

* The internal `abi` helper wasn't needed as `transmute_copy` works just fine.
* The `into_param` trait function is rightly unsafe.
* Added some more test coverage to cover all paths.


kennykerr:
Weird how 1.62 will warn about unnecessary `unsafe` blocks but nightly doesn't pick it up. 

https://github.com/microsoft/windows-rs/actions/runs/7506895207/job/20439317989

Issue #2771, opened 2024-01-03
Support `[AssociatedEnum]` on `UINT*` pointer (`[Out]`/return) types


MarijnS95:
### Summary

For https://github.com/microsoft/win32metadata/pull/1757

DXGI has some `UINT*` return fields that should be typed/associated with the relevant enumeration type in win32metadata. Overwriting the `UINT*` type with a pointer to an enum type currently generates valid code, but it was recommended to use `AssociatedEnum` (for the same reason as it is used on `UINT` and other types). This however does not currently generate valid code.

Pulling the original report from https://github.com/microsoft/win32metadata/pull/1757#discussion_r1428616836:

> Having `AssociatedEnum` on these (out) pointer types breaks windows-rs completely:
> 
> https://github.com/Traverse-Research/windows-rs/commit/ea1c32772c4da263b828f2822d0177acd8dadf22
> 
> Out values are no longer returned from functions but are passed _by value_ to the function, where _its value_ is cast to a pointer.
> 
> Fwiw:
> 
> ```
> Windows.Win32.Graphics.Dxgi.IDXGIOutput3.CheckOverlaySupport : pFlags : [Out] => [AssociatedEnum(DXGI_OVERLAY_SUPPORT_FLAG),Out]
> Windows.Win32.Graphics.Dxgi.IDXGIOutput4.CheckOverlayColorSpaceSupport : pFlags : [Out] => [AssociatedEnum(DXGI_OVERLAY_COLOR_SPACE_SUPPORT_FLAG),Out]
> Windows.Win32.Graphics.Dxgi.IDXGIOutput6.CheckHardwareCompositionSupport : pFlags : [Out] => [AssociatedEnum(DXGI_HARDWARE_COMPOSITION_SUPPORT_FLAGS),Out]
> Windows.Win32.Graphics.Dxgi.IDXGISwapChain3.CheckColorSpaceSupport : pColorSpaceSupport : [Out] => [AssociatedEnum(DXGI_SWAP_CHAIN_COLOR_SPACE_SUPPORT_FLAG),Out]
> ```

Taking the following metadata of one of these functions:

```c#
unsafe HRESULT CheckColorSpaceSupport([In] DXGI_COLOR_SPACE_TYPE ColorSpace, [Out][AssociatedEnum("DXGI_SWAP_CHAIN_COLOR_SPACE_SUPPORT_FLAG")] uint* pColorSpaceSupport);
```

[...]

PR #2797, opened 2024-01-16
Limit `AssociatedEnum` attribute to input parameters

kennykerr:
The `AssociatedEnum` attribute is only supported on input parameters because it is very problematic in general. This update just ensures that metadata that applies it to output parameters doesn't produce invalid code gen. 

Fixes: #2771


riverar (approved):

riverar:
Side note: We can look at supporting `AssociatedEnum` on output/pointers at the time we ingest newer metadata with those changes.

MarijnS95:
Unfortunately, as reported in https://github.com/microsoft/windows-rs/issues/2771#issuecomment-1894458340 this only seems to make things worse. Raw by-value parameters with an `[In,AssociatedEnun]` annotations are suddenly being treated as pointers, which is the wrong ABI.

PR #2799, opened 2024-01-17
Update `windows-core` docs

kennykerr:
Ensures that the `windows-core` crate docs are generated correctly.  

* The `CanInto` trait is internal and should not be documented. 
* The new `VARIANT` and `PROPVARIANT` types are public and needed a basic description.

### DIFF
--- a/crates/libs/core/src/param.rs
+++ b/crates/libs/core/src/param.rs
@@ -17,6 +17,7 @@ impl<T: Type<T>> Param<T> {
     }
 }
 
+#[doc(hidden)]
 pub trait CanInto<T>: Sized {
     const QUERY: bool = false;
 }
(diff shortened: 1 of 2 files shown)


kennykerr:
Getting a little frustrating that I keep getting some warnings only show up on ubuntu builds and not reproducible on Windows. Anyone else seeing this? @riverar @ChrisDenton

kennykerr:
It might be "user error" in this case - I keep forgetting about the silly `docs` feature...

ChrisDenton:
> It might be "user error" in this case - I keep forgetting about the silly `docs` feature... #2676

Ah. Hopefully we can remove that sometime this year. The Cargo feature to replace it should be in edition 2024.

Issue #2811, opened 2024-01-19
Requesting release to include win32metaddata


LeonarddeR:
### Suggestion

Version [57.0.42](https://github.com/microsoft/win32metadata/releases/tag/v57.0.42-preview) of win32metadata contains [interfaces for SAPI](https://github.com/microsoft/win32metadata/pull/1738) that were missing in the past. I'd like to create a new Rust project based on these interfaces.

ChrisDenton:
If https://github.com/microsoft/win32metadata/issues/1788 is going to find its way into metadata soon then I'd quite like that to be included too. But also I don't want to block a release on that.

kennykerr:
I'm bringing in the latest metadata here #2812. Chris, are the changes you're looking for included?

ChrisDenton:
Oh, I didn't realise that's now closed. I'll take a look...

ChrisDenton:
Yep! That seems to address my issue. Never mind then 🙂

kennykerr:
Sweet! Note that I'm not planning a crates release - #2812 is just for updating the metadata and resulting code gen. Let me know if and when a particular crate release is needed for a production dependency.

mikebattista:
@ChrisDenton I just fixed https://github.com/microsoft/win32metadata/issues/1788 so it's not included in the latest published release.

ChrisDenton:
Oh, hmm... yes you're right. It's in the generated metadata if you generate it manually. But not in the rust docs for #2812. Sorry for getting confused there.

kennykerr:
Happy to suck in another update. Maybe check with @riverar and @ChrisDenton that we have everything we need for the next release.

mikebattista:
I just pushed another update with https://github.com/microsoft/win32metadata/issues/1788.

PR #2812, opened 2024-01-19
Update Win32 metadata

kennykerr:
Fixes: #2811


kennykerr:
@riverar just a heads up that I updated the MSVC libs in https://github.com/microsoft/windows-rs/pull/2812/commits/c3f5af4fa34d7f4779909270063d2000dda26377 locally but since I can't update the other libs locally I grabbed the libs produced by the build and was surprised to see a delta on the msvc lib for arm https://github.com/microsoft/windows-rs/pull/2812/commits/e0e5e09fdc089d23504c0b326cf46a8a5263a748

riverar:
The aarch64 lib in https://github.com/microsoft/windows-rs/commit/c3f5af4fa34d7f4779909270063d2000dda26377 is actually x64, something went awry there.

kennykerr:
Hmm. We need to get GitHub to finally support ARM64.

ydirson:
This PR breaks such code:
```
error[E0277]: the `?` operator can only be applied to values that implement `Try`
 --> src\wmi_extra.rs:5:5
  |
5 |     unsafe { Com::CoInitializeEx(None, Com::COINIT_MULTITHREADED) }?;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the `?` operator cannot be applied to type `HRESULT`
  |
  = help: the trait `Try` is not implemented for `HRESULT`
```

That does not seem intended, or is it?

ydirson:
Oh, and to return a `Result`, `CanReturnMultipleSuccessValues` would need more work, like generating function-specific enums?

PR #2815, opened 2024-01-23
Add JSON validator sample

kennykerr:
This sample follows on from #2814 ([article](https://kennykerr.ca/rust-getting-started/creating-your-first-dll.html)) and is meant to show how you can implement a traditional Win32-style API in Rust.

This Rust implementation maps to the following hypothetical Win32-style API:

```C++
HRESULT __stdcall CreateJsonValidator(char const* schema, size_t schema_len, uintptr_t* handle);
HRESULT __stdcall ValidateJson(uintptr_t handle, char const* value, size_t value_len, char** sanitized_value, size_t* sanitized_value_len);
void __stdcall CloseJsonValidator(uintptr_t handle);
```

Once this sample is done, I'll write the next article for [Getting Started with Rust](https://kennykerr.ca/rust-getting-started).


kennykerr:
I'd appreciate any feedback - need to present a good case for using Rust for systems programming on - and in - Windows.

@ChrisDenton @riverar @tim-weis

riverar:
Just throwing out there, you could wire up to WinRT types in `Windows.Data.Json` for extra pizazz.

kennykerr:
Also, is this a Rust 1.56 bug? https://github.com/microsoft/windows-rs/actions/runs/7632102514/job/20791559920?pr=2815

Why would a dependency in a sample break these unrelated builds?

ChrisDenton:
That error doesn't really make sense to me. The crate version definitely exists and aren't yanked. Maybe going back to old cargo corrupts the cache? Does deleting the `.cargo/registry` help? I think it would be worth opening an issue on the Cargo repo.

ChrisDenton:
Oh, this might be https://github.com/rust-lang/cargo/issues/10623

tl;dr old cargo didn't understand `dep` syntax and this results in an unhelpful error message. Any chance of bumping the MSRV to 1.60?

crates/samples/components/json_validator/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+extern "system" fn CloseJsonValidator(handle: usize) {
+    if handle != 0 {
+        unsafe {
+            _ = Box::from_raw(handle as *mut JSONSchema);
+        }
+    }
+}
+
+// Implementation of the `CreateJsonValidator` function so we can use `Result` for simplicity.
+fn create_validator(schema: *const u8, schema_len: usize, handle: *mut usize) -> Result<()> {
```
ChrisDenton, line 32: This and other functions are `unsafe` because we have to take the caller's word that the pointer is readable for the given length.

ChrisDenton (commented):
Looks good! I think I'd be tempted to use a type alias for the handle type. Or even a transparent new type wrapper if you don't mind explaining how this prevents accidentally mixing handle types on the Rust side of the FFI.

Speaking of handles, I'd be remiss if I didn't mention that technically speaking it should be a pointer like `*mut core::ffi::c_void` or something as it is actually a real pointer to real allocated memory.

crates/samples/components/json_validator/src/lib.rs
kennykerr: The trouble with the pointer vs. handle debate is that the pointer-ness is an implementation detail. In the next installment I'd like to show how to implement the handles in a way that they can be validated, at which point they become merely indexes into a container.

kennykerr:
Looks like bumping MSRV to 1.60 fixed the build. Thanks Chris.

crates/samples/components/json_validator/src/lib.rs
riverar: Opaque handle values can, by chance or purposely, point to _real allocated memory_.

kennykerr: Marked them all `unsafe`.

crates/samples/components/json_validator/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    // The handle is not null so we can safely dereference it here.
+    *handle = Box::into_raw(Box::new(compiled)) as usize;
+
+    Ok(())
+}
+
+// Implementation of the `ValidateJson` function so we can use `Result` for simplicity.
+unsafe fn validate(handle: usize, value: *const u8, value_len: usize) -> Result<()> {
```
jonwis, line 51: Our current helper has a "return a reconstructed json from the parsed DOM" model so that the downstream consumer is only touching "produced by code we own, safely" bits. Is there an example of returning a string here?

jonwis (approved):
This is _much_ easier to read than what we have currently. Thanks @kennykerr !

jonwis:
> Just throwing out there, you could wire up to WinRT types in `Windows.Data.Json` for extra pizazz.

Please don't. :) Use literally anything other than `Windows.Data.*` if you can.

Not everything needed to be in the platform SDK 😞

crates/samples/components/json_validator/src/lib.rs
kennykerr: Sure, `HSTRING` would be simple but it probably makes more sense to return a UTF-8 string that would need to be freed with something like `HeapFree` or `CoTaskMemFree`.

kennykerr: I've updated the PR with such an example. https://github.com/microsoft/windows-rs/pull/2815/commits/864de7961db415b065d3c36c6bb02cdc8d85dc1c

riverar (commented):
Looks good to me; going to skip the explicit approval to avoid chasing green ✅ here while you finish up some tweaks.

kennykerr:
Thanks for feedback all! Here are the recent entries:

* https://kennykerr.ca/rust-getting-started/creating-your-first-dll.html
* https://kennykerr.ca/rust-getting-started/implement-win32-api.html

More to come...

PR #2817, opened 2024-01-25
Remove old Windows 7 support code

kennykerr:
This update finally drops Windows 7 support from the `windows-core` crate [as I did for C++/WinRT](https://github.com/microsoft/cppwinrt/pull/1348) some months ago. Much [like the Rust compiler](https://github.com/rust-lang/compiler-team/issues/651), this "support" was never actually tested as we don't have test infrastructure for Windows 7. As a consequence, some Windows 8 only dependencies had crept in a long time ago.


crates/libs/core/src/imp/bindings.rs
```diff
@@ -1,6 +1,8 @@
 // Bindings generated by `windows-bindgen` 0.52.0
 
 #![allow(non_snake_case, non_upper_case_globals, non_camel_case_types, dead_code, clippy::all)]
+::windows_targets::link!("api-ms-win-core-winrt-error-l1-1-0.dll" "system" fn RoOriginateError(error : HRESULT, message : HSTRING) -> BOOL);
+::windows_targets::link!("api-ms-win-core-winrt-l1-1-0.dll" "system" fn RoGetActivationFactory(activatableclassid : HSTRING, iid : *const GUID, factory : *mut *mut ::core::ffi::c_void) -> HRESULT);
```
AngelicosPhosphoros, line 5: Hi! I have read that this DLLs are not supposed to be used.

> Those DLLs are Windows's implementation detail and are subject to change at anytime. Basically Microsoft started moving around APIs starting from Windows 8. For example, APIs in Windows 8's api-ms-win-core-file-l1-2-0.dll got moved to api-ms-win-core-file-l1-2-1.dll in Windows 8.1. Old software still work because the dlls developer linked to are now just placeholders redirecting calls to the actual implementation. But nobody should link to the implementations directly, as that would be defeating the purpose of having such a redirection. Any software that report those dlls are missing are failing to accommodate the redirection (e.g. reporting a delay-loaded dependency as hard one). 

https://answers.microsoft.com/en-us/windows/forum/all/missing-api-ms-win-core-dlls/d99d1368-0f92-43db-bbdb-7d080f1f96e9

May you fix that, please?

roblabla:
Errf, I still use windows-rs on Windows 7, this move is going to cause me issues.

Starting with Rust 1.76, there is now a dedicated for windows7, called `x86_64-win7-windows-msvc`. Would you accept a PR that `#[cfg]`-outs the call to `RoOriginateError`?

kennykerr:
Just turn it into a nop?

crates/libs/core/src/imp/bindings.rs
riverar: I think you misread that; it's suggesting folks do not link to the **implementations**, i.e. the _target_ of the redirection.

flawrist:
Related:
* https://github.com/stunndard/golangwin7patch/issues/8

PR #2841, opened 2024-02-11
Address new nightly unsafe precondition check panics

riverar:
Passing size=0, ptr=0 into `std::slice::from_raw_parts` is undefined behavior and is no longer safe to rely upon. Also addresses an unaligned pointer used by `Event`.

See also:
https://github.com/rust-lang/rust/issues/120902
https://github.com/rust-lang/rust/pull/120594

### DIFF
--- a/crates/libs/core/src/event.rs
+++ b/crates/libs/core/src/event.rs
@@ -198,6 +198,7 @@ impl<T: Interface> Drop for Array<T> {
 
 /// A reference-counted buffer.
 #[repr(C)]
+#[repr(align(8))]
 struct Buffer<T>(crate::imp::RefCount, std::marker::PhantomData<T>);
 
 impl<T: Interface> Buffer<T> {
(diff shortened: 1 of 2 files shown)


ChrisDenton:
Thank you for looking into this. It seems there's still one more failure:

```
thread 'add_remove' panicked at library\core\src\panicking.rs:155:5:
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
thread caused non-unwinding panic. aborting.
error: test failed, to rerun pass `-p test_event --test tests`
```

Which indirectly points to https://github.com/microsoft/windows-rs/blob/40f671bc3e31f501dd327c4dd679a552f1389682/crates/libs/core/src/event.rs#L161-L167

But that looks totally fine on its own. The empty case is checked and I very strongly doubt the size exceeds `isize::MAX`. Which just leaves alignment of the buffer.

So looking at `Buffer::new`: https://github.com/microsoft/windows-rs/blob/40f671bc3e31f501dd327c4dd679a552f1389682/crates/libs/core/src/event.rs#L205-L216

It uses HeapAlloc, which aligns to 16 (or 8 on 32-bit) but then it writes a header which, is essentially an `AtomicI32`. Meaning the data itself is only 4 byte aligned.

Quick fix here would be to add an `#[repr(align(8))]` to `Buffer` to add padding to the header.

kennykerr:
Thanks all. `#[repr(align(8))]` should do the trick for `Event`.

kennykerr (approved):
Thanks!

kennykerr:
I'll probably publish an update to crates.io soon to avoid dependents having issues with nightly.

kennykerr:
But we should probably get the feature doc search feature integrated before then. Let me know if there's anything else top of mind.

PR #2848, opened 2024-02-16
Add `windows-registry` crate

kennykerr:
This update adds the new `windows-registry` crate in the spirit of `windows-result` (#2847) and `windows-version` (#2702) to provide simple, safe, and efficient access to the Windows registry. The `windows-registry` crate takes advantage of the `windows-result` crate for error handling and works naturally with other `windows-rs` crates but does not depend on the larger `windows` or `windows-sys` crates.

Here's a simple example of reading registry key values:

```rust
use windows_registry::*;

fn main() -> Result<()> {
    let key = LOCAL_MACHINE.open(r#"SOFTWARE\Microsoft\Windows NT\CurrentVersion"#)?;
    println!("{}", key.get_string("ProductName")?);
    Ok(())
}
```

Here's an example of writing and reading values:

```rust
use windows_registry::*;

fn main() -> Result<()> {
    let key = CURRENT_USER.create(r#"software\\windows-rs"#)?;

    key.set_u32("number", 123)?;
    key.set_string("name", "Rust")?;

    println!("{}", key.get_u32("number")?);
    println!("{}", key.get_string("name")?);

    Ok(())
}
```

Here's an example of how you might recursively print keys and values:

```rust
use windows_registry::*;

fn main() -> Result<()> {
    let key = LOCAL_MACHINE.open(r#"SOFTWARE\Microsoft\Windows NT\CurrentVersion"#)?;
    print_tree(&key)?;
    Ok(())
}

fn print_tree(key: &Key) -> Result<()> {
    for subkey in key.keys()? {
        print_tree(&key.open(subkey)?)?;
    }

    for (name, value) in key.values()? {
        println!("{name}: {value:?}");
    }

    Ok(())
}
```


crates/libs/registry/src/key.rs
```diff
@@ -0,0 +1,327 @@
...
+        let value = pcwstr(value);
+
+        let result = unsafe {
+            RegSetValueExW(
+                self.0,
+                pcwstr(name).as_ptr(),
+                0,
+                REG_SZ,
+                value.as_ptr() as _,
+                value.len() as u32 * 2,
```
ChrisDenton, line 105: While the length overflowing a `u32` may be exceedingly rare (and probably indicates something else has gone badly wrong) I do think it'd be good to catch that here rather than silently truncating. Maybe it can return an error? Ditto the other similar functions.

ChrisDenton (commented):
Nice! I think a brief top level example would be nice to have just to give people an idea of how to get started exploring the API.

crates/libs/registry/src/key.rs
kennykerr: How about this: [e8a892e](https://github.com/microsoft/windows-rs/pull/2848/commits/e8a892e6d15f56a889842c5c3a8907056f6479a9)

kennykerr:
I also added a readme example - thanks for the feedback Chris!

crates/libs/registry/src/key.rs
ChrisDenton: Sure, that works for me!

ChrisDenton (approved):

PR #2854, opened 2024-02-16
Release 0.53.0

kennykerr:
This update prepares the way to publish new versions of the Windows family of crates. 

* The `windows-sys` crate won't be updated at this time. 
* The `windows-version` crate is unchanged.
* This will be the first release of the `windows-result` and `windows-registry` crates.
* There will be a minor (compatible) update to the `windows-targets` crates.
* Added a simple readme example for the `windows-result` crate.


riverar:
Just a friendly reminder to double-check my changes, to ensure `features.json` and `web/*` don't end up in a crate. Thanks!

PR #2857, opened 2024-02-19
Improve `HRESULT` constant literal generation

kennykerr:
Just a small quality of life improvement for those of us familiar with `HRESULT` values. Constants are now generated as follows:

```rust
pub const E_INVALIDARG: HRESULT = 0x80070057u32 as _;
```

Rather than: 

```rust
pub const E_INVALIDARG: HRESULT = -2147024809i32;
```

This is a bit more in line with how they are traditionally represented in C++ and just makes them easier to read and identify.


kennykerr:
Looks like some new Rust nightly compiler changes are breaking the build.

tim-weis:
Nice change, makes it *much* easier to interpret those `HRESULT` values.

The only suggestion I have is to introduce a separator (`_`) between the value and type suffix[^1]. I find `0x80070005u32` visually harder to parse than `0x80070005_u32`.

[^1]: As supported by [integer literal expressions](https://doc.rust-lang.org/reference/expressions/literal-expr.html#integer-literal-expressions).

kennykerr:
Done, that's indeed more readable.

PR #2860, opened 2024-02-20
Improve `NTSTATUS` constant literal generation

kennykerr:
Much like #2857, `NTSTATUS` uses the sign bit for severity just like `HRESULT` does and are traditionally represented in C++ as hex values. Constants are now generated as follows to improve readability and identification:

```rust
pub const STATUS_NOT_FOUND: NTSTATUS = 0xC0000225_u32 as _;
```

Rather than:

```rust
pub const STATUS_NOT_FOUND: NTSTATUS = -1073741275i32;
```


kennykerr:
Normally, I'd be explicit but these are always directly assigning/initializing an explicit type so it seems overly redundant.

riverar (approved):

PR #2862, opened 2024-02-20
Simpler package layout for natvis files

kennykerr:
I don't think we need the wordier file paths.


tim-weis:
Seems reasonable.

I initially renamed the .natvis files expecting them to carry over into the PDBs and showing up in `.nvlist`. For some reason that didn't happen. (I'll sink some more time into investigating this when I can.)

The rationale behind moving them into a folder was to declutter the crate root. Doing that felt naive and I think leaving them in the crate root and marking them "hidden" (by convention) does a better job at that.

The only concern I have is whether this affects publishing. "Hidden" files (i.e., files starting with a `.`) follow [special rules](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields). Can you [`--dry-run`](https://doc.rust-lang.org/cargo/commands/cargo-publish.html#options) a `cargo publish` to make sure they aren't getting lost? Not that the visualizers are stable enough to be of much help at this time, but things *are* [moving](https://github.com/MicrosoftDocs/visualstudio-docs/issues/3381).

kennykerr:
I just published an update:

https://github.com/microsoft/windows-rs/releases/tag/0.53.0

Please do kick the tires and let me know if you find any issues. Thanks again for the help!

tim-weis:
Running `.nvlist` (WinDbg Preview) on a re-targeted binary displays both *.natvis* files (named `<crate>-<index>.natvis`), dumping the contents produces the expected files, and `dx`-ing the respective expressions delivers the desired results. (VS 2022 Preview [still](https://github.com/microsoft/windows-rs/pull/2853#discussion_r1493376945) fails to expand the expressions.)

LGTM.

Issue #2717, opened 2023-11-30
Documentation no longer lists required features


tim-weis:
### Summary

Starting with version 0.52.0, the documentation published [here](https://microsoft.github.io/windows-docs-rs/doc/windows/) no longer lists all/some required features.

For example:

* [`OutputDebugStringW`](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/Diagnostics/Debug/fn.OutputDebugStringW.html) doesn't list any features
* [`GetMessageW`](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/WindowsAndMessaging/fn.GetMessageW.html) doesn't list the full set of features

It looks like the requirements for the containing module have dropped off.

ChrisDenton:
Probably #2671. I think it no longer adds namespace features to the list of features.

kennykerr:
Yep the docs no longer explicitly include the feature for the enclosing module. This caused a huge amount of redundant doc comments and markup. A bit less intuitive for sure, but a tradeoff I think is worthwhile until I find a long-term replacement for rustdoc...

junglie85:
`OutputDebugStringW` does not show up in [search](https://github.com/search?q=repo%3Amicrosoft%2Fwindows-rs+OutputDebugStringW&type=code) results to be able to find the module to use, it only shows the kernel32.dll.c definition. Maybe an edge case, but that makes it difficult to find which feature to enable.

junglie85:
That's fair - I tend to hunt around the GitHub repo rather than use the published docs.

tim-weis:
The only reason I care about the required features is to copy them into the `features` table of a *Cargo.toml* file. This used to work well when *all* features were listed.

After the change, things are far less convenient. It's not entirely unworkable as long as only the containing module's feature is required. When more features are required the documentation starts to feel very hostile (just go through the number of steps needed to get [`GetMessageW`](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/WindowsAndMessaging/fn.GetMessageW.html)'s features into a *Cargo.toml*).

Someone who knows their way around this crate may merely be inconvenienced. To anyone new, this may be reason enough to look into alternatives.

Reverting this change would be very much appreciated.

ChrisDenton:
[...]

PR #2865, opened 2024-02-21
Add feature search links

kennykerr:
Still need to add support for versioned links, but this is basic idea. 

Fixes: #2717


ChrisDenton:
The functionality works great for me for the types I've tried. Only minor issue is that on mobile the website is just slightly too wide for more phone. Maybe the version selection box could drop below or above the input on narrow screens?

A possible future enhancement would be to have some way to input a list of types and get a unified list of features rather than entering them one at a time. But I think that's one for the suggestion box and not something that needs thinking about now.

kennykerr:
We could scrap the drop down if the target is configurable via the URL.

riverar:
Adjusted the branch drop-down so it wraps on mobile. Still more work to be done here (e.g. perhaps we shift into a two-inner-row per row layout) but something we can address a bit later.

image

tim-weis:
That looks like a workable solution.

The only suggestion I have is to add a "Copy to clipboard" button. That's not something that needs immediate attention, and could be incorporated at a later time.

tim-weis:
Well, maybe I lied. A cup of coffee later I was reminded about a few more suggestions I had in mind:

* Provide a link to the Rust documentation. I don't know how feasible that is, but it would be convenient to have a quick way to navigate to the signature/type information.
* Provide a link to the MSDN reference documentation. That would need a bit of work but it would be very handy to have.
* Improve the relevance priority in searches. As an example, typing `createprocess` into the search field has the `CreateProcessW` API listed way down. Prioritizing exact (prefix) matches might work better.

I don't know how others feel about those suggestions, but they would definitely streamline *my* workflow.

At any rate, thank you for putting this together and reminding me that a browser-based search doesn't *have to* stall the browser window.

riverar:
All great suggestions, will def. work on those.

PR #2868, opened 2024-02-22
Fix for `windows-target` link doc compat

kennykerr:
Just making sure we preserve link macro compatibility in some cases as shown in this test failure:

https://github.com/microsoft/windows-rs/actions/runs/8005925420/job/21866485024

### DIFF
--- a/crates/libs/targets/src/lib.rs
+++ b/crates/libs/targets/src/lib.rs
@@ -8,7 +8,7 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 #[cfg(all(windows_raw_dylib, target_arch = "x86"))]
 #[macro_export]
 macro_rules! link {
-    ($library:literal $abi:literal $($link_name:literal)? fn $($function:tt)*) => (
+    ($library:literal $abi:literal $($link_name:literal)? $(#[$doc:meta])? fn $($function:tt)*) => (
         #[link(name = $library, kind = "raw-dylib", modifiers = "+verbatim", import_name_type = "undecorated")]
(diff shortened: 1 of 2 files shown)


nbigaouette:
I'm seeing this exact error using `0.52.2`. My build now passes if I downgrade to 0.52.0 (`cargo update --package windows-targets --precise 0.52.0`).

kennykerr:
I noticed that Cargo doesn't reliably grab the latest when an older version is yanked that you happen to already have downloaded. My issue was resolved after I forced a download of 0.52.2 and after that it just worked as before. For example, this will force your local cache to get updated:

```toml
[dependencies.windows-targets]
version = "0.52.2"
```

After running Cargo, you can remove this dependency and its cache should be updated. This seems like a Cargo or crates.io bug but I'm not sure.

kennykerr:
Presumably you can also run `cargo update --package windows-targets --precise 0.52.2`

kennykerr:
You may also be hitting this issue: #2870

nbigaouette:
I just checked my logs and CI was using 0.52.1 🤔

I guess an updated version (0.52.3?) will be released with the linking issue fixed (#2870). I'll see if that version helps out.

kennykerr:
Yep 0.52.3 is on its way.

kennykerr:
0.52.3 has now been published - let me know if you have any further issues.

nbigaouette:
I've relaunched my CI job (still without lock file) and it picked up 0.52.3; it now builds without issue.

Thanks a lot!! 👍

PR #2874, opened 2024-02-23
Run web.yml workflow only in original repository

kaivol:
Don't run web.yml workflow in forks, because it fails there without enabling GitHub Pages.


riverar:
Thanks!

PR #2883, opened 2024-02-26
Add natural error translation for RPC

kennykerr:
This update adds support for `RPC_STATUS` so that you can easily turn it into an `HRESULT` and use automatic error propagation with `?` and `Result` for RPC functions. 

Note I also opened this issue with the Win32 metadata: https://github.com/microsoft/win32metadata/issues/1857


kennykerr:
@lhecker @zadjii-msft - this should simplify things for your RPC interop.

Issue #2787, opened 2024-01-11
`LocalFree` and `GlobalFree` functions return an error if the free succeeds, while returning a success if the free fails


Rileran:
### Summary

Functions [LocalFree](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Foundation/fn.LocalFree.html) and [GlobalFree](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Foundation/fn.GlobalFree.html) return `Err` when the free is successful, and `Ok` when the free fails.

The expected outcome would be the opposite.

Looking at the [documentation for LocalFree](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-localfree) tells us that if the function succeeds, the return value is **NULL** (or invalid).

Looking at the generated source code (for crate version 0.52.0), the function returns `Ok` when the return value is **NOT NULL** (or valid).

```rust
pub unsafe fn LocalFree<P0>(hmem: P0) -> ::windows_core::Result<HLOCAL>
where
    P0: ::windows_core::IntoParam<HLOCAL>,
{
    ::windows_targets::link!("kernel32.dll" "system" fn LocalFree(hmem : HLOCAL) -> HLOCAL);
    let result__ = LocalFree(hmem.into_param().abi());
    (!result__.is_invalid()).then(|| result__).ok_or_else(::windows_core::Error::from_win32)
}
```

### Crate manifest

```TOML
[package]
name = "windows-playground"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
windows = { version = "0.52.0", features = ["Win32_Foundation", "Win32_System_Memory"] }
```

### Crate code

```Rust
use windows::Win32::Foundation::LocalFree;
use windows::Win32::System::Memory::{LocalAlloc, LMEM_FIXED, LMEM_ZEROINIT};

fn main() {
    unsafe {
        let data = LocalAlloc(LMEM_FIXED | LMEM_ZEROINIT, 10).unwrap();
        // the next call fails if free is successfull
        // Error { code: HRESULT(0x00000000), message: "The operation completed successfully." }
        LocalFree(data).unwrap();
    }
}
```

Rileran:
Just saw this is already on issue on the win32metadata repository https://github.com/microsoft/win32metadata/issues/1557, closing this one.

PR #2889, opened 2024-02-27
Workaround for confusing `LocalFree` behavior

kennykerr:
I doubt this will put the issue to rest, but hopefully this at least results in less frequent questions. 🥲

Fixes: https://github.com/microsoft/windows-rs/issues?q=LocalFree

### DIFF
--- a/crates/tests/resources/tests/sys.rs
+++ b/crates/tests/resources/tests/sys.rs
@@ -8,6 +8,7 @@ use windows_sys::{
 #[test]
 fn sys() {
     unsafe {
+        SetLastError(0);
         assert_eq!(IDI_APPLICATION as u16, 32512);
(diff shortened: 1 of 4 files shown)


crates/tests/resources/tests/sys.rs
```diff
@@ -8,6 +8,7 @@ use windows_sys::{
 #[test]
 fn sys() {
     unsafe {
+        SetLastError(0);
```
kennykerr, line 11: This is unrelated but covers a spurious test failure.

unlimitedsola:
@kennykerr can we do this for `GlobalFree` as well? It's also suffering from the same issue, which has been reported [here](https://github.com/microsoft/windows-rs/issues/2787).

PR #2890, opened 2024-02-27
Avoid `Result` transformation for `WIN32_ERROR`

kennykerr:
Much like `NTSTATUS` and `RPC_STATUS`, functions that return `WIN32_ERROR` often require the caller to inspect and branch based on the specific error code. Automatically transforming to a `Result` can make this needlessly complicated as the error code needs to be re-translated back from an `HRESULT`. 

Callers that still want a Result or `?` error propagation can simply use the `ok` helper as in `SomeApi().ok()?`.

### DIFF
--- a/crates/libs/bindgen/src/metadata.rs
+++ b/crates/libs/bindgen/src/metadata.rs
@@ -173,7 +173,6 @@ impl Signature {
                     SignatureKind::ResultVoid
                 }
             }
-            Type::TypeDef(def, _) if def.type_name() == TypeName::WIN32_ERROR => SignatureKind::ResultVoid,
             Type::TypeDef(def, _) if def.type_name() == TypeName::BOOL && method_def_last_error(self.def) => SignatureKind::ResultVoid,
             _ if type_is_struct(&self.return_type) => SignatureKind::ReturnStruct,
             _ => SignatureKind::PreserveSig,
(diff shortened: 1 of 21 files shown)


fgimian:
Is it worth marking `WIN32_ERROR` as `#[must_use]` @kennykerr? My only concern after upgrading to 0.54.0 is that it is now quite possible to forget to handle errors and no warnings are given about this (unlike previous versions which forced using `Result` values).

Cheers
Fotis

kennykerr:
Good idea!

fgimian:
Forgive me for a further question, but I notice various functions do still return `Result` types in 0.54.0, is this intentional? e.g. `CloseHandle`

Was this applied selectively to certain functions or should it be consistent throughout?

Thanks again mate, and keep up the great work!
Fotis

kennykerr:
The transformation is (still) applied to functions that return `HRESULT` or `BOOL`, except under certain conditions where the original signature is preserved. In those cases, you're not really losing anything having it transformed into a `Result`.

fgimian:
> The transformation is (still) applied to functions that return `HRESULT` or `BOOL`, except under certain conditions where the original signature is preserved. In those cases, you're not really losing anything having it transformed into a `Result`.

Ah gotcha, that makes sense, thanks for the clarification 😄

PR #2895, opened 2024-02-27
Add `--derives` option to bindgen to allow emitting custom `#[derive]` attributes

dpaoliello:
Issue Details:
While Rust allows trait `impl` blocks in either the crate that defines the struct or the trait, `#[derive]` attributes must be applied directly to the struct definition. This makes deriving implementations of common traits (like `Default` or `Copy`) for generated structs impossible without modifying the generated rs file.

Fix Details:
Add a new option, `--derive <ty=traits>` that allows a list of traits to derive for a given type.


kennykerr:
Closing in favor of #2907

Issue #2900, opened 2024-02-28
Add `#[must_use]` to `WIN32_ERROR` and other return types that represent error codes


kennykerr:
Is it worth marking `WIN32_ERROR` as `#[must_use]` @kennykerr? My only concern after upgrading to 0.54.0 is that it is now quite possible to forget to handle errors and no warnings are given about this (unlike previous versions which forced using `Result` values).

Cheers
Fotis

_Originally posted by @fgimian in https://github.com/microsoft/windows-rs/issues/2890#issuecomment-1967940864_

kennykerr:
Probably worth applying to `WIN32_ERROR`, `BOOL`, `NTSTATUS`, and `RPC_STATUS`.

kennykerr:
This is complicated by the fact that the metadata models some of these as enums and others as structs...

PR #2907, opened 2024-03-01
Add ability to prepend code to type generation

kennykerr:
Worked with @dpaoliello on a low-level feature to inject code for specific types. This is primarily for inserting type attributes that cannot be derived from metadata. Ideally the metadata tooling will improve to the point where this is no longer needed. 

Fixes: #2900


crates/libs/bindgen/src/rust/mod.rs
```diff
@@ -42,6 +42,20 @@ pub fn from_reader(reader: &'static metadata::Reader, mut config: std::collectio
...
+            if let Some(index) = full_name.rfind('.') {
+                let namespace = &full_name[0..index];
+                let name = &full_name[index + 1..];
+                if let Some(type_def) = reader.get_type_def(namespace, name).next() {
+                    writer.prepend.insert(type_def, value.to_string());
+                    return false;
+                }
+            }
+        }
+        true
```
dpaoliello, line 56: I just want to check: this means that an unused `prepend` value will raise an "invalid configuration value" error?

dpaoliello (approved):

crates/libs/bindgen/src/rust/mod.rs
kennykerr: Correct.

PR #2911, opened 2024-03-05
Fix for new Rust nightly warnings related to unnecessary qualifications

kennykerr:
That's a lot of warnings! I dealt with them in `windows-bindgen` so you don't have to. 🥲

A lot of these stem from the early days of windows-rs where all code was generated inside a macro and thus the context in which it was generated was completely unknown and liable to collisions.


kennykerr:
Those nightly changes are not without consequence...

PR #2915, opened 2024-03-06
Release 0.55.0

kennykerr:
As requested, this release will include an update to the following crates:

* `windows-bindgen`
* `windows-metadata`
* `windows-core`


dpaoliello (dismissed):

kennykerr:
The `windows-core` crate also requires an update since `windows-bindgen` can optionally generate code depending on that crate.

dpaoliello (approved):

riverar:
@kennykerr Do we want to edit `web/features/.env` here too? (I recognize it's not any different than the last release.)

PR #2918, opened 2024-03-07
Simplify version declarations in toml and readme files

kennykerr:
This has a few benefits:

* Simpler release maintenance requiring fewer changes and smaller diffs. 
* Avoids dependencies in compatible updates that may end up getting yanked.
* Simplifies samples and readmes.


kennykerr:
Hmm, seems older toolchains still require the version field.

PR #2924, opened 2024-03-10
Hash impl for HSTRING

zardini123:
Adds Hash implementation to HSTRING.

### DIFF
--- a/crates/libs/core/src/strings/hstring.rs
+++ b/crates/libs/core/src/strings/hstring.rs
@@ -191,6 +191,12 @@ impl Ord for HSTRING {
     }
 }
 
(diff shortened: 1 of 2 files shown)


kennykerr:
Thanks for the contribution! Please add a basic test here just to make sure we have a test that validates:

`crates\tests\core\tests\hstring.rs`

zardini123:

zardini123:

zardini123:
Hi @kennykerr!  Thank you for the feedback.  I have created two tests to validate the API of hashing HSTRING and updated my branch.

Let me know if there is any more needed changes or feedback.

PR #2934, opened 2024-03-17
Defer module name lowercase conversion for Windows

kennykerr:
This update just defers lowercasing function module names in `windows-metadata` to the point where they actually need to be lowercase in code generation to ensure the metadata reader faithfully preserves this information.


riverar:
@abbec If you get a moment, would be great to hear if this change affects you (https://github.com/microsoft/windows-rs/issues/638)

PR #2951, opened 2024-03-25
Fix for new Rust nightly warnings related to dead code

kennykerr:
New dead code detection can see through derived traits and detect fields that aren't used. It caught one case in the Rust code generator and there are a few cases in rdl and winmd generation that are still in the works that I've temporarily allowed.


kennykerr:
@riverar now it's just the lib test failing on the gnullvm libs only it would seem.

PR #2957, opened 2024-03-28
Add support for Arm64EC

dpaoliello:
Rust recently added support for Arm64EC (<https://github.com/rust-lang/rust/pull/119199>), but adding support in the Standard Library (<https://github.com/rust-lang/rust/pull/123144>) requires that windows-bindgen emit the correct `cfg` items for Arm64EC.

This adds support for emitting the Arm64EC `cfg` attributes and reuses the existing AArch64 target crate (to match how MSVC links against the arm64 VC and Windows SDK libs).

### DIFF
--- a/crates/libs/bindgen/src/rust/cfg.rs
+++ b/crates/libs/bindgen/src/rust/cfg.rs
@@ -125,6 +125,7 @@ fn cfg_add_attributes<R: AsRow + Into<metadata::HasAttribute>>(cfg: &mut Cfg, ro
                         }
                         if value & 2 == 2 {
                             cfg.arches.insert("x86_64");
+                            cfg.arches.insert("arm64ec");
                         }
                         if value & 4 == 4 {
                             cfg.arches.insert("aarch64");
--- a/crates/libs/sys/src/Windows/Win32/Devices/DeviceAndDriverInstallation/mod.rs
+++ b/crates/libs/sys/src/Windows/Win32/Devices/DeviceAndDriverInstallation/mod.rs
@@ -2406,7 +2406,7 @@ impl Clone for BUSNUMBER_RESOURCE {
     }
 }
 #[repr(C)]
-#[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "aarch64", target_arch = "arm64ec", target_arch = "x86_64"))]
 pub struct CABINET_INFO_A {
     pub CabinetPath: windows_sys::core::PCSTR,
     pub CabinetFile: windows_sys::core::PCSTR,
@@ -2414,9 +2414,9 @@ pub struct CABINET_INFO_A {
     pub SetId: u16,
     pub CabinetNumber: u16,
 }
-#[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "aarch64", target_arch = "arm64ec", target_arch = "x86_64"))]
 impl Copy for CABINET_INFO_A {}
-#[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "aarch64", target_arch = "arm64ec", target_arch = "x86_64"))]
 impl Clone for CABINET_INFO_A {
     fn clone(&self) -> Self {
         *self
@@ -2440,7 +2440,7 @@ impl Clone for CABINET_INFO_A {
     }
 }
 #[repr(C)]
-#[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "aarch64", target_arch = "arm64ec", target_arch = "x86_64"))]
 pub struct CABINET_INFO_W {
     pub CabinetPath: windows_sys::core::PCWSTR,
     pub CabinetFile: windows_sys::core::PCWSTR,
@@ -2448,9 +2448,9 @@ pub struct CABINET_INFO_W {
     pub SetId: u16,
     pub CabinetNumber: u16,
 }
-#[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "aarch64", target_arch = "arm64ec", target_arch = "x86_64"))]
 impl Copy for CABINET_INFO_W {}
-#[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
+#[cfg(any(target_arch = "aarch64", target_arch = "arm64ec", target_arch = "x86_64"))]
 impl Clone for CABINET_INFO_W {
     fn clone(&self) -> Self {
         *self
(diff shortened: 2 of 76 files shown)


crates/tools/msvc/src/main.rs
```diff
@@ -9,6 +9,7 @@ fn main() {
             "x86" => "i686_msvc",
             "x64" => "x86_64_msvc",
             "arm64" => "aarch64_msvc",
+            "arm64ec" => "arm64ec_msvc",
```
riverar, line 12: Hmm have you seen any tooling that sets Platform=arm64ec? The last official MSVC workflow I read involved using arm64 and passing ARM64EC flags into cl/link. So we'll probably need to tweak the arm64 workflow here and plumb through a `/machine:xxx` argument to be passed into lib.exe (on line 42).

crates/tools/msvc/src/main.rs
```diff
@@ -340,6 +341,7 @@ fn make_reproducible(lib: &std::path::Path) {
 
 #[test]
 fn test_make_reproducible() {
+    // Arm64EC's lib format is significantly different, so exclude it from this test.
```
riverar, line 344: I can take a look into this.

crates/tools/msvc/src/main.rs
```diff
@@ -340,6 +341,7 @@ fn make_reproducible(lib: &std::path::Path) {
...
     for (machine, offset) in [("x86", 0), ("x64", 12), ("arm64", 12)] {
```
riverar, line 345: Altered test with offsets pushed out by 228 bytes:

```rust
#[test]
fn test_make_reproducible() {
    for (machine, offset) in [("x86", 228), ("x64", 240), ("arm64", 240), ("arm64ec", 0)] {
        dbg!(machine);
        let mut def = std::fs::File::create("test.def").unwrap();
        def.write_all(
            r#"
LIBRARY long-library-name-for-placement-in-long-names-table.dll
EXPORTS
A=A.#1
B=B.#2
C=C.#3
"#
            .to_string()
            .as_bytes(),
        )
        .unwrap();
        drop(def);

        let mut cmd = std::process::Command::new("lib");
        cmd.arg("/nologo");
        cmd.arg("/Brepro");
        cmd.arg(format!("/machine:{machine}"));
        cmd.arg("/out:test.lib");
        cmd.arg("/def:test.def");
        cmd.output().unwrap();

        let mut archive = std::fs::File::options()
            .read(true)
            .write(true)
            .open("test.lib")
            .unwrap();

        let mut buf = [0; 24];

        make_reproducible(std::path::Path::new("test.lib"));

        // Archive member header timestamp
        archive.seek(SeekFrom::Start(0x18)).unwrap();
        archive.read_exact(&mut buf[..12]).unwrap();
        assert_eq!(&buf[..12], b"-1          ");

        // COFF file header time/date stamp
        archive.seek(SeekFrom::Start(0x406 - offset)).unwrap();
        archive.read_exact(&mut buf[..4]).unwrap();
        assert_eq!(&buf[..4], [0x00, 0x00, 0x00, 0x00]);

        // Compiler version symbol
        archive.seek(SeekFrom::Start(0x565 - offset)).unwrap();
        archive.read_exact(&mut buf[..12]).unwrap();
        assert_eq!(&buf[..12], b"@comp.id\0\0\0\0");

        // Import header time/date stamp
        archive.seek(SeekFrom::Start(0x577 - offset)).unwrap();
        archive.read_exact(&mut buf[..4]).unwrap();
        assert_eq!(&buf[..4], [0x00, 0x00, 0x00, 0x00]);

        archive.seek(SeekFrom::Start(0x4E9 - offset)).unwrap();
        archive.read_exact(&mut buf[..18]).unwrap();
        assert_eq!(&buf[..18], b"Microsoft (R) LINK");

        drop(archive);

        std::fs::remove_file("test.lib").unwrap();
        std::fs::remove_file("test.exp").unwrap();
        std::fs::remove_file("test.def").unwrap();
    }
}

```

crates/tools/msvc/src/main.rs
riverar, line 1: Add an entry for the undocumented `/<ECSYMBOLS>/` member. Looks safe to skip / doesn't have any reproducibility issues. Will follow up with docs team to get this added to https://learn.microsoft.com/windows/win32/debug/pe-format#coff-file-header-object-and-image cc: @Karl-Bridge-Microsoft

```rust
archive.read_exact(&mut buf[..16]).unwrap();
let skip_member = match &buf[..16] {
    b"/               " => true, // Linker member
    b"//              " => true, // Long names table
    b"/<ECSYMBOLS>/   " => true, // Emulation Compatible symbols
    _ => false,
};
```

crates/tools/msvc/src/main.rs
kennykerr: Presumably this works with `tool_msvc` but we need lib.yml to be updated to confirm.

crates/tools/msvc/src/main.rs
ChrisDenton: If you're taking requests 😉... would it also be possible to document `IMPORT_OBJECT_NAME_EXPORTAS`? Reverse engineering it for the object crate wasn't hard but still.

crates/tools/msvc/src/main.rs
dpaoliello: Excellent, thanks!

crates/tools/msvc/src/main.rs
dpaoliello: Actually, it occurs to me that this is all moot since Arm64EC links against the Arm64 import libraries in VC and the Windows SDK, so we can also reuse our existing AArch64 target crate.

kennykerr: That would be even better and avoid another target lib crate.

kennykerr:
This looks better! If only we could test ARM64 on GitHub. 🙄

riverar:
Arm64ec binaries can link against x64 and Arm64ec libs, but not Arm64 libs.

dpaoliello:
> Arm64ec binaries can link against x64 and Arm64ec libs, but not Arm64 libs.

Ah, good point, looks like a bunch of the libs in the Windows SDK arm64 directory are actually ARM64X (x64 + Arm64EC). So, for simplicity, we can link against the x64 target crate.

kennykerr:
Until we can get ARM testing on GitHub, do either of you @dpaoliello / @riverar have hardware that you can manually `cargo test --all` this on?

riverar:
@dpaoliello Ah right. If you'd like a hand, I'd be happy to stand up arm64ec testing here too. Just need to know what to pull down.

kennykerr:
@dpaoliello if you're stuck in a chicken-egg situation trying to update rust-lang via `windows-bindgen` we could do this in two steps. First add the `cfg` support and then separately update the targets, since Standard Library doesn't use `windows-targets`.

dpaoliello:
That's not the issue: your tests uncovered a gap in LLVM's support for Arm64EC, so I'm going to have to fix that first.

kennykerr:
Sweet!

kennykerr (approved):
Thanks - that's good enough for me.

riverar (approved):

PR #2958, opened 2024-03-28
crates/tests/core/tests/hstring.rs: corrected import location of DefaultHasher

gibsonpil:
The `hstring_hashing_different_strings` and `hstring_equality_combinations` tests in `crates/tests/core/tests/hstring.rs` were trying to import DefaultHasher from `std::hash` even though it is actually located in `std::collections::hash_map`, causing the tests to not compile.

This pull request fixes the imports, resolving the issue.

### DIFF
--- a/crates/tests/core/tests/hstring.rs
+++ b/crates/tests/core/tests/hstring.rs
@@ -124,7 +124,8 @@ fn hstring_to_os_string() {
 #[test]
 fn hstring_hashing_equal_strings() {
     // Checks if two strings of identical contents have the same hash
-    use std::hash::{DefaultHasher, Hash, Hasher};
+    use std::collections::hash_map::DefaultHasher;
+    use std::hash::{Hash, Hasher};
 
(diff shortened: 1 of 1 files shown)


riverar:
What failures are you seeing? Not seeing any CI failures.

gibsonpil:
It doesn't fail on the CI since the Rust maintainers recently made `DefaultHasher` accessible via `std::hash` in v1.76.0. However, anyone running a version of Rust prior to that version will be unable to compile `test_core`. This conflicts with the `windows` crate's MSRV of v1.62.0. 

I should've clarified better in my initial comment, sorry about that.

kennykerr:
Thanks for clarifying, but the tests aren't constrained by the MSRV of the crate being tested. We have separate MSRV tests, but the general test suite is run against stable and nightly only.

gibsonpil:
Ah, I supposed I misinterpreted that crate's MSRV policy then.
Sorry about that.

PR #2961, opened 2024-04-01
add i686-pc-windows-gnullvm target

jeremyd2019:
Adds i686-pc-windows-gnullvm target.

Fixes: #2960

PR #2970, opened 2024-04-04
Exclude -gnullvm target from i686_gnu build.rs

jeremyd2019:
This was missed in #2774 even though the similar check was added to x86_64_gnu, presumably because there was no handling for i686-pc-windows-gnullvm at that time.

Follow-up to https://github.com/microsoft/windows-rs/pull/2961#issuecomment-2032607772

### DIFF
--- a/crates/targets/i686_gnu/build.rs
+++ b/crates/targets/i686_gnu/build.rs
@@ -2,7 +2,8 @@ fn main() {
     let family = std::env::var("CARGO_CFG_TARGET_FAMILY").unwrap();
     let arch = std::env::var("CARGO_CFG_TARGET_ARCH").unwrap();
     let env = std::env::var("CARGO_CFG_TARGET_ENV").unwrap();
-    if family != "windows" || arch != "x86" || env != "gnu" {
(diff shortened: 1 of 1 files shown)


riverar:
Before we merge this in, we'd like to fix the current cross.yml to catch this error condition first. Then we can use that failing test to validate this new PR.

jeremyd2019:
I'm not seeing how you would test that, but OK.  Theoretically, if you had a branch where you reverted #2961 (except for the change to cross.yml), and applied the change from this PR to it, the i686-pc-windows-gnullvm check would fail.  Then #2961 would be the "correct" fix for that failure...

kennykerr:
Really appreciate you helping out here and debugging this further. I guess what I'm trying to say is that these checks don't appear to be necessary any longer and #2973 seems to confirm.

jeremyd2019:
Yes, I figured this was an intentional "belt-and-suspenders" type situation, excluding both via the Cargo.toml dependencies *and* via the build.rs.  But it would certainly make life easier if the targets were only specified in one or the other.

kennykerr:
Closing in favor of #2973.

PR #2972, opened 2024-04-04
contrived testing of #2970 and #2961

jeremyd2019:
I want to show that #2970 and #2961 do what they are supposed to do, so I'm making this draft PR demonstrating appling the changes in a different order so that can be seen.

Theoretically, if you had a branch where you reverted https://github.com/microsoft/windows-rs/pull/2961 (except for the change to cross.yml), and applied the change from this PR to it, the i686-pc-windows-gnullvm check would fail. Then https://github.com/microsoft/windows-rs/pull/2961 would be the "correct" fix for that failure...


jeremyd2019:
https://github.com/microsoft/windows-rs/actions/runs/8559092947/job/23455148213?pr=2972 shows the failure I was expecting on commit 4b8da8c.  https://github.com/microsoft/windows-rs/actions/runs/8559386967/job/23456506812?pr=2972 shows 2d14659 succeed.  The diff now is just #2970, plus adding workflow_dispatch to cross.yml (so I could test it easier on my fork rather than waiting for approval to run workflows here)

PR #2973, opened 2024-04-04
Remove redundant target checks

kennykerr:
This is a test to see whether these checks are really necessary given that `windows-targets` should already eliminate invalid target crate download and compilation. 

And the build seems to confirm this. So unless there are counter arguments, this seems like the way to go.


dpaoliello (approved):
The more that I think about this, the more that I like it - initially I was hesitant since anybody who may have accidently depended on the wrong crate will now get a linker error, whereas before it would silently work. That said, someone may be depending on one of these and *thinking* it is doing something and not realizing that it is a no-op.

ChrisDenton (approved):
I'm very much in favour of build scripts being as minimal as possible.

PR #2976, opened 2024-04-08
Add explicit path support to the `interface` macro

kennykerr:
Continuing on my journey to improve the `interface` macro in #2968 and #2969, this update fixes an issue with the `interface` macro where it required the base interface to be an identifier and didn't support type paths.


kennykerr:
Looks like some new Clippy warnings I need to deal with first.

PR #2979, opened 2024-04-09
Remove workflow action

kennykerr:
There's a discussion about whether this action actually has any effect. Testing to see whether the GitHub build environment is sufficient to succeed without it.


kennykerr:
Seems its still required.

Issue #2984, opened 2024-04-11
Support generic bounds on implement macro


roblabla:
### Suggestion

Currently, the `implement` macro does not support generic bounds. For instance, the following:

```rust
trait MyClass {}

#[implement(IClassFactory)]
struct MyClassFactory<T>(PhantomData<T>);

impl<T: MyClass> IClassFactory_Impl for MyClassFactory<T> {
    // ...
}
```

Will fail to compile with the following error:

```rust
error[E0277]: the trait bound `T: MyClass` is not satisfied
    --> src/lib.rs:185:1
     |
185  | #[implement(IClassFactory)]
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `MyClass` is not implemented for `T`, which is required by `MyClassFactory<T>: IClassFactory_Impl`
     |
note: required for `MyClassFactory<T>` to implement `IClassFactory_Impl`
    --> src/lib.rs:188:23
     |
188  | impl<T: MyClass> IClassFactory_Impl for MyClassFactory<T> {
     |         -------  ^^^^^^^^^^^^^^^^^^     ^^^^^^^^^^^^^^^^^
     |         |
     |         unsatisfied trait bound introduced here
note: required by a bound in `IClassFactory_Vtbl::new`
    --> /Users/roblabla/.cargo_isengard/registry/src/index.crates.io-6f17d22bba15001f/windows-0.52.0/src/Windows/Win32/System/Com/impl.rs:1191:81
     |
1191 |     pub const fn new<Identity: ::windows_core::IUnknownImpl<Impl = Impl>, Impl: IClassFactory_Impl, const OFFSET: isize>() -> IClassFactory_Vtbl {
     |                                                                                 ^^^^^^^^^^^^^^^^^^ required by this bound in `IClassFactory_Vtbl::new`
     = note: this error originates in the attribute macro `implement` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider restricting type parameter `T`
     |
186  | struct MyClassFactory<T: MyClassProvider>(PhantomData<T>);
     |                          ++++++++++++++
```

Doing the suggested change (adding the bound on the struct) doesn't help:

```rust
error[E0107]: struct takes 1 generic argument but 0 generic arguments were supplied
   --> src/lib.rs:186:8
    |
186 | struct MyClassFactory<T: MyClass>(PhantomData<T>);
    |        ^^^^^^^^^^^^^^^^ expected 1 generic argument
    |
[...]

PR #2986, opened 2024-04-11
Better generic support in implement

roblabla:
Fixes: #2984 

It turns out windows-rs already supports generics in the `implement` macro, but would handle generic bounds of the form `<T: Bound>` a bit wrong. It would generate code like this:

```rust
impl<T: AmsiProvider> ::core::convert::From<AmsiClassFactory<T: AmsiProvider>>
for ::windows::core::IUnknown {
    fn from(this: AmsiClassFactory<T: AmsiProvider>) -> Self {
        let this = AmsiClassFactory_Impl::<T: AmsiProvider>::new(this);
        let boxed = ::core::mem::ManuallyDrop::new(::std::boxed::Box::new(this));
        unsafe { ::core::mem::transmute(&boxed.identity) }
    }
}
```

This is wrong, first line should read `impl<T: AmsiProvider> ::core::convert::From<AmsiClassFactory<T>>`, dropping the bound when applying the generic on the right hand side of the for.

Syn has a function for exactly this use-case: `Generics::split_for_impl`, which returns three values: An ImplGeneric to apply to the impl and contains the bounds, a `TypeGenerics` to apply on a type and does not contain the bounds,and the WhereClause with the additional bounds.

This MR reworks the implement macro to use this function, thus fixing this corner case.

### DIFF
--- a/crates/libs/implement/src/lib.rs
+++ b/crates/libs/implement/src/lib.rs
@@ -2,7 +2,7 @@
 Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs>
 */
 
-use quote::{quote, ToTokens};
+use quote::quote;
 
 /// Implements one or more COM interfaces.
 ///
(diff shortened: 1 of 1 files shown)


kennykerr:
Please ensure that you've run all of cargo `test`, `clippy`, and `fmt` on the entire workspace before pushing.

kennykerr:
Closing for now. Feel free to reopen if you'd like to revisit in future.

PR #2994, opened 2024-04-15
Workflow improvements

kennykerr:
Working on some workflow improvements to make it easier to add some needed cross-arch testing.


kennykerr:
These failures go back to https://github.com/rust-lang/cargo/issues/9661 and looks like there's no hope for a fix this this millennium. 🥹

kennykerr:
Thanks, will test that.

PR #3000, opened 2024-04-17
Workflow improvements 3

kennykerr:
Following on from #2998, this again attempts to enable Arm64 testing given the Cargo and GitHub limitations discovered on previous attempts.

### DIFF
--- a/.github/workflows/arm64.yml
+++ b/.github/workflows/arm64.yml
@@ -1,0 +1,306 @@
+name: arm64
+
+on:
+pull_request:
+push:
+  paths-ignore:
+    - '.github/ISSUE_TEMPLATE/**'
+  branches:
+    - master
+
+env:
+RUSTFLAGS: -Dwarnings
+
+jobs:
+check:
+  runs-on: windows-2019
+
+  strategy:
+    matrix:
+      include:
+        - version: stable
+        - version: nightly
+
+  steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Update toolchain
+      run: rustup update --no-self-update ${{ matrix.version }} && rustup default ${{ matrix.version }}-x86_64-pc-windows-msvc
+    - name: Add toolchain target
+      run: rustup target add aarch64-pc-windows-msvc
+    - name: Fix environment
+      uses: ./.github/actions/fix-environment
+    - name: Clean
+      run:  cargo clean
+    - name: Test riddle
+      run:  cargo test -p riddle --no-run
+    - name: Test sample_bits
+      run:  cargo test -p sample_bits --no-run
+    - name: Test sample_com_uri
+      run:  cargo test -p sample_com_uri --no-run
+    - name: Test sample_component_hello_world
+      run:  cargo test -p sample_component_hello_world --no-run
+    - name: Test sample_component_json_validator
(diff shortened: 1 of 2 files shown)


.github/actions/fix-environment/action.yml
```diff
@@ -40,6 +40,12 @@ runs:
             ((Resolve-Path "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\*\bin\Hostx64\x64")
               | Sort-Object -Descending | Select -First 1).ToString() >> $env:GITHUB_PATH
           }
+          "aarch64*"
+          {
+            "C:\Program Files (x86)\Windows Kits\10\bin\10.0.22000.0\x64" >> $env:GITHUB_PATH
+            ((Resolve-Path "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\*\bin\Hostx64\x64")
```
dpaoliello, line 46: Should this be `Hostx64\arm64` ?

.github/workflows/arm64.yml
```diff
@@ -0,0 +1,308 @@
...
+      - name: Update toolchain
+        run: rustup update --no-self-update ${{ matrix.version }} && rustup default ${{ matrix.version }}-x86_64-pc-windows-msvc
+      - name: Add toolchain target
+        run: rustup target add ${{ matrix.target }}
+      - name: Fix environment
+        uses: ./.github/actions/fix-environment
+      - name: Clean
+        run:  cargo clean
+      - name: Test riddle
+        run:  cargo test -p riddle --no-run --target ${{ matrix.target }}
```
dpaoliello, line 38: Could you have re-used test.yml and made `--no-run` part of the matrix?

crates/tools/yml/src/main.rs
```diff
@@ -137,3 +138,65 @@ jobs:
 
     std::fs::write(".github/workflows/clippy.yml", yml.as_bytes()).unwrap();
 }
+
+// Ideally this would just be another matrix dimension in test_yml but Cargo limitations are blocking this for the time being.
```
dpaoliello, line 142: I don't understand why the Cargo limitation blocks this

.github/workflows/arm64.yml
kennykerr: That was the goal but that would require `--target` and then it runs afoul of https://github.com/rust-lang/cargo/issues/9661 and none of the DLL "cdylib" tests would succeed.

crates/tools/yml/src/main.rs
kennykerr: As noted above, if you're cross compiling you need to use `--target` and then the default placement of build artifacts is disrupted and the component (DLL client and server) tests would fail. More info in specific comment:

https://github.com/rust-lang/cargo/issues/9661#issuecomment-2059686735

kennykerr: There's hope in "artifact dependencies" but this feature is not stable and doesn't yet work very well. See this comment:

https://github.com/rust-lang/cargo/issues/9661#issuecomment-2059974262

Luracasmus:
Number 3000! 🎉🎉

PR #3001, opened 2024-04-17
 Workflow improvements 4

kennykerr:
Following on from #3000, this further simplifies testing by consolidating `arm64.yml` and `test.yml` by leveraging some tips provided by @jeremyd2019 and @dpaoliello - thanks!


crates/tools/yml/src/main.rs
```diff
@@ -28,19 +27,35 @@ jobs:
...
+            host: x86_64-pc-windows-msvc
             target: i686-pc-windows-msvc
+            etc:
           - version: nightly
+            host: x86_64-pc-windows-gnu
             target: x86_64-pc-windows-gnu
+            etc:
           - version: stable
+            host: x86_64-pc-windows-gnu
             target: i686-pc-windows-gnu
```
jeremyd2019, line 43: Was it intentional to now cross-compile x86_64 -> i686?

kennykerr: It's not required but works for me.

Issue #2999, opened 2024-04-17
Could  windows_registry::Key expose a raw_handle() API to allow Registry change event using RegNotifyChangeKeyValue


shanyp:
### Summary

Hi
I would like to be able to get notified on Registry key changes, using RegNotifyChangeKeyValue , but it requires the HKEY and while the widnows-registry::Key doesn't expose its inner HKEY

Cinimajig:
As temp workaround, you could `transmute` it and then use it. This code works for me:
```rust
let key = windows_registry::CURRENT_USER.create(r#"SOFTWARE\windows-rs"#)?;
        
// Transforms the value to HKEY.
let hkey = std::mem::transmute::<_, HKEY>(key);

// Dereference it, to use with Reg-functions.
RegNotifyChangeKeyValue(
    hkey,
    true,
    REG_NOTIFY_CHANGE_NAME | REG_NOTIFY_CHANGE_ATTRIBUTES,
    None,
    false
).ok()?;

// Manually close the key, since the original is out of scope.
RegCloseKey(hkey).ok()?;
```
But yeah, it would be nice to be able to get the underlying HKEY or just a way to use directly, instead of transmuting back and forth.

kennykerr:
I can add a `pub fn as_raw(&self) -> isize` method that would return the raw key value. I have `as_raw` methods on other wrappers in windows-rs and they work well.

shanyp:
@kennykerr sounds perfect !

kennykerr:
Here you go: #3003

PR #3003, opened 2024-04-18
Add `as_raw` to `windows-registry`

kennykerr:
This returns the underlying registry key handle for interop purposes. 

Fixes: #2999

### DIFF
--- a/crates/libs/registry/src/key.rs
+++ b/crates/libs/registry/src/key.rs
@@ -37,6 +37,11 @@ impl Key {
         win32_error(result).map(|_| Self(handle))
     }
 
+    /// Returns the underlying registry key handle.
+    pub fn as_raw(&self) -> isize {
+        self.0
+    }
+
     /// Removes the registry keys and values of the specified key recursively.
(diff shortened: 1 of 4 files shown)


shanyp (approved):

crates/libs/registry/src/key.rs
```diff
@@ -37,6 +37,11 @@ impl Key {
         win32_error(result).map(|_| Self(handle))
     }
 
+    /// Returns the underlying registry key handle.
+    pub fn as_raw(&self) -> isize {
+        self.0
+    }
```
tim-weis, line 43: This duplicates the key handle but `Key` also implements `Drop` potentially leaving callers with an invalid handle. Should this be `unsafe`?

Alternatively, this could be `pub fn into_inner(self) -> isize` instead making the transfer of ownership obvious.

kennykerr: There is no transfer of ownership here. This is similar to https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.as_ptr where the underlying value is returned without changing ownership and requiring the caller to use it judiciously.

Issue #2996, opened 2024-04-16
PCWSTR has undocumented undefined behavior on non-windows targets


tyilo:
### Summary

When using `PWCSTR` and `PWSTR` on non-windows targets, `wcslen` is incorrectly used to determine the length of a null-terminated string.

The `as_wide` implementations needs to be fixed in one of the following ways:
- It could be disabled on non-windows targets
- It could instantly panic on non-windows targets
- It could be fixed to correctly count the non-zero `u16` characters on non-windows targets

### Crate manifest

```TOML
[package]
name = "windows-test"
version = "0.1.0"
edition = "2021"

[dependencies]
windows-core = "0.56.0"
```

### Crate code

```Rust
use windows_core::PCWSTR;

fn main() {
    let arr = [b'a' as u16, b'b' as u16, 0, 0];
    let s = PCWSTR::from_raw(&arr as *const u16);
    let slice = unsafe { s.as_wide() };
    // Should output [97, 98]
    // Outputs [97] on Linux
    println!("{:?}", slice);

    let arr = [b'a' as u16, 0];
    let s = PCWSTR::from_raw(&arr as *const u16);
    let slice = unsafe { s.as_wide() };
    // Should output [97]
    // UB on Linux
    println!("{:?}", slice);
}
```

kennykerr:
I can probably disable everything on non-Windows targets, but may I ask why you're "using PWCSTR and PWSTR on non-windows targets"?

tyilo:
> I can probably disable everything on non-Windows targets, but may I ask why you're "using PWCSTR and PWSTR on non-windows targets"?

I'm developing on Linux, so being able to run my tests on Linux would be nice. As my code compiled using a linux target, I assumed the code would also work.

I was only testing some part of the code using windows types, but no win32 API calls.

kennykerr:
Got it, well there are a handful of other functions in `windows-core` that are excluded from non-Windows builds, for testing purposes as well, so I can just add this.

kennykerr:
Actually, this has a bit of a ripple effect. 

Is this also a problem if you use a Windows target as in `--target x86_64-pc-windows-gnu`?

kennykerr:
The `windows-sys` crate provides raw bindings so if you just need the definitions of various Windows types with no run time behavior then that may be more appropriate. 

https://kennykerr.ca/rust-getting-started/windows-or-windows-sys.html

kennykerr:
[...]

Issue #1874, opened 2022-07-06
Implement 32-bit widestrings for Linux/WSL


MarijnS95:
With recent work we've finally enabled `windows-rs` to _run_ on Linux, many thanks for opening up :partying_face:!

However, for my specific use-case of [wrapping](https://github.com/Traverse-Research/hassle-rs) [DirectXShaderCompiler](https://github.com/microsoft/DirectXShaderCompiler) in Rust there's _one more_ thing I need. Widestring types are prevalent in the `DirectXShaderCompiler` API, but don't currently work on Linux.

Wide chars and strings are typically 32-bit, [with Windows being the exception rather than the rule at 16-bit](https://en.cppreference.com/w/cpp/language/types#Character%20types:~:text=wchar_t,a%20distinct%20type). That is also how they're [implemented in `DXC` when compiled for non-Windows](https://github.com/microsoft/DirectXShaderCompiler/blob/76fed39c368cf65167e6db00738b212dac024b49/include/dxc/Support/WinAdapter.h#L385-L393), making them incompatible (read: segfaults) with the current UTF-16 "assumption" inside `windows`.

To support this format we can "simply" apply some `cfg()` changes that turn the anyway-custom `BSTR` / `P*WSTR` into `*mut u32`-carrying structures, with an API to match. In my experiment I have simply written a:

```rust
#[cfg(windows)]
type WCHAR = u16;
#[cfg(not(windows))]
type WCHAR = u32;
```

### Problem 1: converting to and from `str`/`String`

In a previous (unaccepted) contribution I opted to use [`widestring`](https://docs.rs/widestring), since it implements everything we need. That was (understandably) frowned upon as the `windows` crate should be the base for many other things, not the other way around (and not have many dependencies in any case).

The `widestring` implementation simply uses `.chars()` to acquire an iterator over [`char`](https://doc.rust-lang.org/std/char/index.html) which is a "[Unicode code point](https://www.unicode.org/glossary/#code_point)" but exactly fits UTF-32 chars too. It works and makes sense, but I'm no Unicode expert and don't know if this is the right way to implement it.

### Problem 2: Allocations and length

[...]

PR #3004, opened 2024-04-19
Add non-windows support for PWSTR and PCWSTR

tyilo:
As described in #2996, `PWSTR` and `PCWSTR` currently has undocumented undefined behaviour in the `as_wide` method.

This fixes this by using a different implementation on non-windows targets.

Fixes: #2996, #1874

### DIFF
--- a/crates/libs/core/src/strings/pcwstr.rs
+++ b/crates/libs/core/src/strings/pcwstr.rs
@@ -32,7 +32,7 @@ impl PCWSTR {
     ///
     /// The `PCWSTR`'s pointer needs to be valid for reads up until and including the next `\0`.
     pub unsafe fn len(&self) -> usize {
-        #[cfg(target_os = "windows")]
+        #[cfg(windows)]
         let len = {
             extern "C" {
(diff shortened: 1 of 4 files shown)


tyilo:
@microsoft-github-policy-service agree company="Princh"

kennykerr:
You can use #3005 for test coverage.

crates/libs/core/src/strings/pcwstr.rs
```diff
@@ -26,14 +26,41 @@ impl PCWSTR {
         self.0.is_null()
     }
 
+    /// String length without the trailing 0
+    ///
+    /// # Safety
+    ///
+    /// The `PCWSTR`'s pointer needs to be valid for reads up until and including the next `\0`.
+    pub unsafe fn len(&self) -> usize {
+        #[cfg(target_os = "windows")]
```
kennykerr, line 35: Other target cfg tests use the more concise `#[cfg(windows)]` (see hstring.rs) so it's probably best just to keep this consistent.

kennykerr (approved):
Thanks!

PR #3005, opened 2024-04-19
Add Linux smoke test

kennykerr:
The existing Linux smoke test was compiling for a Linux target but not actually running any tests. This updates yml to actually test the code and repurposes the simple test component to actually test something interesting: the difference between `wcslen` on Windows and Linux. 

This test should fail because of: #2996


crates/tests/linux/tests/strings.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[test]
+fn test() {
+    unsafe {
+        let s: PCSTR = s!("hello world");
+        assert_eq!(s.to_string().unwrap(), "hello world");
+
+        // TODO: https://github.com/microsoft/windows-rs/pull/3004 should enable the following test.
+        // let w: PCWSTR = w!("wide world");
+        // assert_eq!(w.to_string().unwrap(), "wide world");
```
kennykerr, line 11: @tyilo please uncomment this test in your PR.

PR #3007, opened 2024-04-20
Use prebuilt gnullvm target

mati865:
With https://github.com/rust-lang/rust/pull/121712 merged we can drop `-Z build-std` workaround.

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -40,10 +40,10 @@ jobs:
         run: rustup target add ${{ matrix.target }}
         if: contains(matrix.target, 'gnullvm') == false
 
-      - name: Add nightly toolchain with rust-src
+      - name: Add nightly toolchain with gnullvm target
         run: |
           rustup default ${{ matrix.version }}
-          rustup component add rust-src
+          rustup target add ${{ matrix.target }}
         if: startsWith(matrix.image, 'ubuntu-') && contains(matrix.target, 'gnullvm') && matrix.version == 'nightly'
 
       - name: Install gcc-mingw-w64-x86-64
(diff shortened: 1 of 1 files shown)


mati865:
I think it will work when new nightly is published and CI is retried.

.github/workflows/cross.yml
```diff
@@ -38,7 +38,7 @@ jobs:
 
       - name: Add toolchain target
         run: rustup target add ${{ matrix.target }}
-        if: contains(matrix.target, 'gnullvm') == false
+        if: contains(matrix.target, 'gnullvm') == false || startsWith(matrix.image, 'ubuntu-') && contains(matrix.target, 'gnullvm') && matrix.version == 'nightly'
 
       - name: Add nightly toolchain with rust-src
```
mati865, line 43: Todo: this is no longer necessary

mati865:
Blocked on https://github.com/rust-lang/rust/pull/124229

jeremyd2019:
I was hoping it would reduce the need for special-case steps, like for instance the Test and Test gnullvm steps are now the same.  Note that llvm-mingw does also ship a mac toolchain, so theoretically could install that and get rid of more special if checks for ubuntu...

mati865:
I've experimented with replacing duplicates with more complicated `if` conditions but it's not that easy. I think the best way to handle it is to leave duplicates until 13 June when current nightly will become stable.



> Note that llvm-mingw does also ship a mac toolchain, so theoretically could install that and get rid of more special if checks for ubuntu...

I was wondering whether it would be beneficial to add it but I don't see much value in mac builds (windows -> windows-gnullvm could me more useful tough).

mati865:
Marked as ready to review as the CI should pass with today's nightly.

kennykerr (approved):

Issue #2638, opened 2023-08-30
Implement drop for HANDLE to automatically call CloseHandle


NiiightmareXD:

Nerixyz:
This isn't really in scope for these crates - see https://github.com/microsoft/windows-rs/issues/2222 and https://github.com/microsoft/windows-rs/issues/2468.

kennykerr:
See #3013 for a solution.

Issue #2468, opened 2023-04-21
Clarification: RAII helpers in-scope for windows-rs?


talagrand:
### Motivation

Projects which are Win32-heavy are peppered with either bespoke RAII types or invocations of scopeguard-type constructs. Are there plans to improve this situation in windows-rs itself or is there room for a wil-type library in the ecosystem? If the former, should we start contributing these types here? There's also a scoping question - e.g. an RAII Handle type vs something higher-level like a SetThreadToken/restore RAII type, or even beyond that higher-level bindings like the wil win32 helpers.

### Rationale and alternatives

A key advantage of having these in windows-rs itself is for ecosystem standardization. Additionally, the various helpers and types could key off of the same feature partitions as the underlying APIs.

kennykerr:
I encourage developers to build WIL-style helpers as discrete crates atop the `windows` or `windows-sys` crates and to treat the `windows-rs` family of crates as merely the generated bindings representing the underlying APIs. The line is a little fuzzy as the `windows` crate necessarily provides some programming aids to make APIs usable but for the most part I try to avoid adding that as much as possible to keep the line between generated bindings and hand-crafted libraries clear.

kennykerr:
See #3013 for a solution.

Issue #2222, opened 2022-12-03
RAII type for HANDLES?


damyanp:
Is there an RAII type for HANDLEs that'll call CloseHandle when dropped?  As far as I can tell the current HANDLE requires an explicit call to CloseHandle.

kennykerr:
No. The Win32 metadata, which defines `HANDLE`, doesn't have a reliable way to express ownership semantics. As you know, there are many uses of `HANDLE`, sometimes with different ways to free the handle, sometimes it is merely borrowed, sometimes representing a psuedo handle, etc. Ideally, the metadata would indicate such semantics on functions like `CreateEventW` to indicate what to do with the resulting handle.

damyanp:
Ok, good to know I wasn't missing something.  Has this already been discussed? Is this in scope for v1.0?

kennykerr:
Ideally, the Win32 metadata can get improved support for ownership semantics. @riverar and I have discussed this at some length. He may have some thoughts as well. But I probably have to come up with some way to infer lifetime/ownership. 

This is a general problem, particularly challenging with DirectX APIs where there are structs with COM interface pointer fields and it's not always clear who owns what.

riverar:
Metadata has quite a bit of work ahead of it to make this work, or at least get it into a state where we can start experimentation.

For example, we're still tossing around candidates for how these attributes look https://github.com/microsoft/win32metadata/issues/389. And we haven't nailed down a fix for free functions with multiple parameters https://github.com/microsoft/win32metadata/issues/423. Or how to deal with functions that return either freeable and non-freeable resources based on other inputs https://github.com/microsoft/win32metadata/issues/423#issuecomment-860187551.

damyanp:
It seems that there are several different problems some of affecting legacy APIs while others are affecting more modern ones.  Is the idea to fix all at once, or can these be prioritized at all?

The D3D APIs I think should be pretty trivial - the COM pointers in D3D12 structs are all weak references.

Chaoses-Ib:
Here is a simple OwnedHandle:
```rust
use windows::Win32::Foundation::{HANDLE, CloseHandle};

/// You should only use this wrapper if the handle can be closed by `CloseHandle`.
#[derive(PartialEq, Eq, Clone, Copy, Debug)]
pub struct OwnedHandle(pub HANDLE);

impl From<HANDLE> for OwnedHandle {
    fn from(handle: HANDLE) -> Self {
        Self(handle)
    }
}

[...]

PR #3013, opened 2024-04-30
Add support for freeing handles automatically

kennykerr:
The Win32 metadata includes information about how to free certain kinds of handles. Since ownership is often unknown or contextual, this is not used to implement the standard `Drop` trait. Instead, a new `Free` trait provided by the `windows-core` crate is implemented for all handles with appropriate metadata. The `free` function can be used to free the handle manually, but an `Owned<T>` wrapper is provided to call `free` automatically when the owned handle is dropped. 

This is supported by nearly 150 different handle types - not just `HANDLE` - including various crypto handle types, registry, GDI and USER32 types, and many more.

Workarounds are provided for the following Win32 metadata bugs:

* https://github.com/microsoft/win32metadata/issues/1891
* https://github.com/microsoft/win32metadata/issues/1892

Here's an example. 

Before:

```rust
let event = CreateEventW(None, true, false, None)?;
SetEvent(event)?;
WaitForSingleObject(event, 0);
_ = CloseHandle(event);
```

After:

```rust
let event = Owned::new(CreateEventW(None, true, false, None)?);
SetEvent(*event)?;
WaitForSingleObject(*event, 0);
```

The BCrypt test provides a more comprehensive example: https://github.com/microsoft/windows-rs/pull/3013/commits/e386c51e024b91451e4b817fd4bfd154b12b685a

Previously, the test was just leaking. You can see how `Owned` supports all kinds of BCrypt handle types through type inference.

Related: #2638 #2468 #2222


crates/libs/core/src/handles.rs
```diff
@@ -0,0 +1,39 @@
+/// Custom code to free a handle.
+///
+/// This is similar to the `Drop` trait, and may be used to implement `Drop`, but allows handles
+/// to be dropped depending on context.
+pub trait Free {
+    /// Calls the handle's free function.
+    fn free(&mut self);
```
ChrisDenton, line 7: This might need to be unsafe. An owned type asserts the right to be able to free. Without that, as you say, whether or not it's ok to free depends on the context. Which we can't, only the user can.

crates/libs/bindgen/src/rust/handles.rs
```diff
@@ -63,12 +63,47 @@ pub fn gen_win_handle(writer: &Writer, def: metadata::TypeDef) -> TokenStream {
...
+        } else {
+            let name = to_ident(function.name());
+            let signature = metadata::method_def_signature(def.namespace(), function, &[]);
+
+            // BCryptCloseAlgorithmProvider has an unused trailing parameter.
+            let tail = if signature.params.len() > 1 {
+                quote! { , 0 }
+            } else {
+                quote! {}
+            };
```
riverar, line 79: Just heads up, I just added metadata for AMSI context/session freeing and this will likely get tangled up here; `AmsiCloseSession` needs two valid inputs 🙃

Perhaps we should mark `BCryptCloseAlgorithmProvider::dwFlags` as `Reserved` then you can key off that and decide whether to abort?

crates/libs/core/src/handles.rs
kennykerr: Good call: https://github.com/microsoft/windows-rs/pull/3013/commits/5986214536be1b760f1378629c2ca7ecc51b869c

crates/libs/bindgen/src/rust/handles.rs
riverar: I can limit my additions to those that fit this form for now. Might be better to then have that discussion on the metadata side of the fence.

crates/libs/core/src/handles.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// Calls the handle's free function.
+    ///
+    /// # Safety
+    /// The handle must be owned by the caller and safe to free.    
+    unsafe fn free(&mut self);
+}
+
+/// A wrapper to provide ownership for handles to automatically drop via the handle's `Free` trait.
+#[repr(transparent)]
+#[derive(Clone, PartialEq, Eq, Default, Debug)]
```
MarijnS95, line 15: I strongly doubt `Owned` should implement `Clone`, at least not safely, because this way you could safely drop the same handle twice because `unsafe fn free()` is hidden away in `drop()`.

MarijnS95: Wow, that was really poor timing with the merge.

crates/libs/core/src/handles.rs
```diff
@@ -0,0 +1,45 @@
+/// Custom code to free a handle.
+///
+/// This is similar to the `Drop` trait, and may be used to implement `Drop`, but allows handles
```
MarijnS95, line 3: Note that intradoc links are nice :)

```suggestion
/// This is similar to the [`Drop`] trait, and may be used to implement [`Drop`], but allows handles
```

crates/libs/core/src/handles.rs
```diff
@@ -0,0 +1,45 @@
...
+/// to be dropped depending on context.
+pub trait Free {
+    /// Calls the handle's free function.
+    ///
+    /// # Safety
+    /// The handle must be owned by the caller and safe to free.    
+    unsafe fn free(&mut self);
+}
+
+/// A wrapper to provide ownership for handles to automatically drop via the handle's `Free` trait.
```
MarijnS95, line 13:
```suggestion
/// A wrapper to provide ownership for handles to automatically drop via the handle's [`Free`] trait.
```

MarijnS95:
This looks like really nice stuff! Is the metadata enriched to know when functions like `CreateEventW()` return an owned handle so that they could possibly return `Owned<>` straight away in the future?

kennykerr:
> This looks like really nice stuff! Is the metadata enriched to know when functions like `CreateEventW()` return an owned handle so that they could possibly return `Owned<>` straight away in the future, forcing the user to handle automatic lifetime management?

This isn't very clear today. Hopefully in future...

crates/libs/core/src/handles.rs
kennykerr: See #3014

PR #3015, opened 2024-05-02
Derive `Copy`, `Clone`, and `Debug` for handles

kennykerr:
This just reduces code gen and improves readability.

### DIFF
--- a/crates/libs/bindgen/src/rust/handles.rs
+++ b/crates/libs/bindgen/src/rust/handles.rs
@@ -100,24 +100,13 @@ pub fn gen_win_handle(writer: &Writer, def: metadata::TypeDef) -> TokenStream {
 
     let mut tokens = quote! {
         #[repr(transparent)]
-        #[derive(PartialEq, Eq)]
+        #[derive(Clone, Copy, Debug, PartialEq, Eq)]
         pub struct #ident(pub #signature);
         #is_invalid
         #free
         impl Default for #ident {
             fn default() -> Self {
                 unsafe { core::mem::zeroed() }
-            }
-        }
-        impl Clone for #ident {
-            fn clone(&self) -> Self {
-                *self
-            }
-        }
-        impl Copy for #ident {}
-        impl core::fmt::Debug for #ident {
-            fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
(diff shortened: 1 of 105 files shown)


kennykerr:
Both #3015 and #3016 are failing with this GNU dlltool error:

https://github.com/microsoft/windows-rs/actions/runs/8924805385/job/24511973178?pr=3015

Unrelated to this change I believe. Any idea what this might be about? Perhaps a change to the GitHub runners?

@riverar @jeremyd2019 @mati865

mati865:
Looks like GHA have changed somehow:
```
C:\msys64\mingw32\bin\dlltool.exe ... -m i386:x86-64
```
32-bit dlltool cannot create x86_64 import libraries.

riverar:
Looping in @mati865 for awareness, as they have some dlltool work in the pipe https://github.com/rust-lang/rust/pull/124138

riverar:
`dlltool.exe` is normally located [by explicitly scouring `PATH`](https://github.com/jfgoog/rust/blob/master/compiler/rustc_codegen_llvm/src/back/archive.rs#L443), which probably got re-ordered in a GHA runner update.

So I suspect adding an explicit dlltool configuration should help here. Let's give it a shot.

mati865:
On the second thought, failing test runs on x86_64 host but targets i686. But it's a proc macro so it should still build for the host.

It could really be a regression then.

riverar:
We manually prepend `C:\msys64\mingw32\bin` onto the path when targeting i686-pc-windows-gnu, not taking proc macros into consideration. So I think the build failure makes more sense to me than the previous successes.

mati865:
Ah, makes sense then. Previous success doesn't make sense though 😄 

It'd probably work if you set PATH to `/mingw64/bin:/mingw32/bin:$PATH` for cross-compilation builds since 64-bit dlltool can generate both kinds of libraries and the linker is called with a target triple prefix.

tim-weis:
Should handle types also `#[derive(PartialOrd, Ord)]` so they can be stored as keys in `BTreeSet`s/`BTreeMap`s? Likewise, `#[derive(Hash)]` would make them eligible as the key in `HashSet`s/`HashMap`s.

kennykerr:
Note that this PR isn't adding these traits - it's merely deriving them rather than generating them directly.

PR #3022, opened 2024-05-06
Handle new `unexpected_cfgs` warnings

kennykerr:
Much like https://github.com/tokio-rs/tokio/pull/6538 I need to silence the new warnings introduced by https://github.com/rust-lang/cargo/pull/13571 about unexpected `cfg` condition names because the alternative is to use a build script which is undesirable and impractical as that introduces a warning about a build script syntax that isn't compatible with the MSRV. 

I was able to keep this for most of the crates and further refine code gen to elide code that depended on excluded features.


epage:
> introduces a warning about a build script syntax that isn't compatible with the MSRV.

You can use the `:` syntax that does work with your MSRV

kennykerr:
Thanks Ed, I'll give that a try.

kennykerr:
Weirdly this causes an entirely different warning. No idea why this is coming from the build script.

```
warning: missing documentation for the crate
 --> crates\libs\targets\build.rs:1:1
  |
1 | / fn main() {
2 | |     println!("cargo:rustc-check-cfg=cfg(windows_raw_dylib)");
3 | | }
  | |_^
  |
  = note: requested on the command line with `-W missing-docs`
```

kennykerr:
So I was able to keep this for most of the crates and further refine code gen to elide code that depended on excluded features, which is great. 🎉

I still have to allow the warning for the `windows-targets` crate due to the unrelated https://github.com/microsoft/windows-rs/pull/3022#issuecomment-2096430603 warning that I've not yet found a workaround for.

kennykerr:
Looks like the error is coming from the `missing_docs` workspace lint which is erroneously thinking the `windows-targets` crate lacks documentation when a build script is added. 🤷

PR #3025, opened 2024-05-09
Add `Ref` and `OutRef` to enhance COM authoring support

kennykerr:
This update introduces a number of improvements to the `implement` and `interface` macros as well as the `windows-core` crate that work together to enhance support for COM interface declarations and implementations as well as general FFI/ABI support. 

Consider the following C++ function signature:

```C++
HRESULT __stdcall DllGetActivationFactory(HSTRING, IActivationFactory**)
```

This function must be exported by DLLs for WinRT (COM) support. The ABI requirements are such that the `HSTRING` and `IActivationFactory` types are "borrowed" in the sense that neither the caller nor the callee must free or drop those values as part of the call itself. This leads to Rust developers having to write something like this:

```Rust
unsafe extern "system" fn DllGetActivationFactory(
    name: std::mem::ManuallyDrop<HSTRING>,
    factory: *mut std::mem::ManuallyDrop<Option<IActivationFactory>>,
) -> HRESULT
```

The implementation is then forced to carefully consider and manage lifetimes and transfer ownership and is easy to get wrong. 

The `Ref` and `OutRef` types represent borrowed types, providing the same memory layout but also offer lifetime management and conveniences to make dealing with in and out parameters safe and convenient:

```Rust
unsafe extern "system" fn DllGetActivationFactory(
    name: Ref<HSTRING>,
) -> HRESULT
```

`Ref` implements the `Deref` trait that returns a reference to the borrowed type while `OutRef` provides the `write` method that will overwrite the memory location with the given value. The `write` method can only be called once and checks for null. It also provides the `is_null` method if you need to conditionally support optional out parameters. 

The `interface` macro will also generate bindings for `Ref` and `OutRef` that behave like the bindings generated by the `windows-bindgen` crate, making it easier to call such interface methods naturally from Rust. This leverages the new `ParamMut` trait that compliments the existing `Param` trait for handling out parameters. It takes care of dealing with uninitialized memory in out parameters as well as dropping `&mut` values in Rust to avoid leaks. 

In addition, the `interface` macro now supports `Result` return types for convenience. The underlying virtual function will still return an `HRESULT` but the `interface` and `implement` macros will automatically translate that for you:

```Rust
#[interface("09428a59-5b40-4e4c-9175-e7a78514316d")]
unsafe trait ITest: IUnknown {
    unsafe fn Void(&self);
    unsafe fn Code(&self) -> HRESULT;
    unsafe fn Result(&self) -> Result<()>;
}
```


crates/libs/core/src/ref.rs
```diff
@@ -0,0 +1,17 @@
+use super::*;
+
+/// A borrowed type with the same memory layout as the type itself that can be used to construct ABI-compatible function signatures.
+#[repr(transparent)]
+pub struct Ref<'a, T: Type<T>>(T::Abi, std::marker::PhantomData<&'a T>);
+
+impl<'a, T: Type<T>> Ref<'a, T> {
+    /// Reads the borrowed value.
+    pub fn read(&self) -> &T::Default {
```
sivadeilra, line 9: Would it make sense for `read` to just be converted to the `Deref` impl?  For COM interfaces, that gives you the ability to call methods directly on it, and you get nice automatic conversions of `&Ref<IFoo>` to `&IFoo`.

crates/libs/core/src/ref_mut.rs
```diff
@@ -0,0 +1,25 @@
+use super::*;
+
+/// A borrowed type with the same memory layout as the type itself that can be used to construct ABI-compatible function signatures.
+///
+/// This is a mutable version of [Ref] meant to support out parameters.
```
sivadeilra, line 5: What would you think of renaming `RefMut` to `OutRef`?

crates/libs/core/src/ref.rs
kennykerr: done!

crates/libs/core/src/ref_mut.rs
kennykerr: I'm trying to borrow 😏 language from Rust where practical. There's `Param` and `ParamMut` and now `Ref` and `RefMut` which mirrors similar naming in the Rust Standard Library. `RefMut` and `ParamMut` take care of all the out parameter conventions, lifetimes, and stack packing rules required by stdcall, Win32, COM, but behaves like `&mut` in Rust.

sivadeilra: Ref and RefMut in the standard library do something different, though.  For example, std::cell::Ref and std::cell::RefMut are both lifetime-tracked pointers a an element in a cell.  The level of the pointer is the same, and they only differ based on whether you've acquired shared or exclusive access to the thing inside the cell.

But in this case, Ref is a pointer to an object, but RefMut is a pointer-to-a-pointer to an object.  And they serve quite different purposes; one is intended as an input parameter, one as an output parameter.

kennykerr: That's fair. I don't disagree. Naming is hard... 🥹

In that case I would probably go with `Param` and `OutParam` for the traits and `Ref` and `OutRef` for the reference wrappers as you suggest.

kennykerr: done!

PR #3041, opened 2024-05-15
Derive standard traits for structs

kennykerr:
Continuing on from #3015, this further derives the standard traits for structs where possible to reduce lines of code and improve readability.


kennykerr:
Looks like deriving `Debug` for function pointers is new in Rust 1.70 so that'll be the new MSRV for the `windows` crate.

crates/libs/bindgen/src/rust/structs.rs
```diff
@@ -135,110 +135,48 @@ fn gen_windows_traits(writer: &Writer, def: metadata::TypeDef, name: &TokenStrea
...
-            }
-        });
+        let features = writer.cfg_features(cfg);
 
         quote! {
             #features
             impl Clone for #name {
                 fn clone(&self) -> Self {
-                    Self { #(#fields),* }
+                    unsafe { core::mem::transmute_copy(self) }
```
dpaoliello, line 179: I'd rather if you left the previous implementation, this feels incredibly dangerous....

dpaoliello (changes requested):

crates/libs/bindgen/src/rust/structs.rs
kennykerr: This is the previous implementation! 🥲 I've just moved it around some, and this is the last bit that's not derived. I have some more work to do to eliminate this but yes that's the goal. This PR didn't actually add any new manual implementations. This is just the last implementation that remains, for now.

dpaoliello: Ah, I see now, GitHub's diffing was confusing me :)

dpaoliello (approved):

PR #3043, opened 2024-05-15
Provide an explicit `ComObject<T>` type that represents a heap-allocated COM object

sivadeilra:
This adds a new type, `ComObject<T>`, which represents a type that was annotated with `#[implement]` and which has been placed in the heap.  `ComObject` is conceptually similar to the ATL/MFC `CComObject` type.

The `ComObject` type and the `#[implement]` macro work together.  They provide efficient (zero-cost) ways to cast to the interfaces that `T` implements. There are no runtime checks (and no addref/release), since we know at compile time that `T` implements those interfaces.

`ComObject` is a counted reference to the heap allocation that contains `T`. Its `Clone` implementation increments the reference count and its Drop implementation decrements the reference count.

Various other improvements;

* The `#[implement]` and `#[interface]` macros can now be used in crates that are `#![no_std]`.  The `windows_core` crate is not yet `#![no_std]`, but I did convert a lot of `std::foo` to `core::foo`.  I don't plan on getting all the way there in this PR, but this helps.
* Added `#[inline(always)]` to many functions that are on important call paths. This is partly for performance but mainly for improving the debugging experience for stepping through interface dispatch.  Right now, if you step into a call to a COM interface, you spend a lot of time stepping through 1-line functions, long before getting to the actual v-call.
* Added unit tests for all new features in this PR.
* `ComObject` has delegating implementations of many core Rust traits, such as `Default`, `Hash`, `PartialEq`, etc.


crates/libs/core/src/imp/mod.rs
```diff
@@ -98,3 +99,12 @@ macro_rules! define_interface {
...
 pub use define_interface;
+
+#[doc(hidden)]
+pub use std::boxed::Box;
+
+#[doc(hidden)]
+pub const E_POINTER: crate::HRESULT = crate::HRESULT(0x80004003u32 as i32);
+
+#[doc(hidden)]
+pub const E_NOINTERFACE: crate::HRESULT = crate::HRESULT(0x80004002u32 as i32);
```
kennykerr, line 110: These are already imported by `pub use com_bindings::*` above.

crates/libs/core/src/interface.rs
```diff
@@ -98,17 +119,93 @@ pub unsafe trait Interface: Sized + Clone {
...
+pub unsafe fn from_raw_borrowed<T: Interface>(raw: &*mut c_void) -> Option<&T> {
     T::from_raw_borrowed(raw)
 }
+
+/// This has the same memory representation as `IFoo`, but represents a borrowed interface pointer.
+///
+/// This type has no `Drop` impl; it does not AddRef/Release the given interface. However, because
+/// it has a lifetime parameter, it always represents a non-null pointer to an interface.
+#[repr(transparent)]
+pub struct InterfaceRef<'a, I>(NonNull<c_void>, PhantomData<&'a I>);
```
kennykerr, line 149: This is very similar to `Ref` from #3025 - be nice to harmonize them somehow.

sivadeilra: They're similar, but I wanted a type that had non-null semantics.

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,274 @@
...
+        T::Outer: ComObjectInterface<I>,
+    {
+        self.as_interface::<I>().to_owned()
+    }
+
+    /// Converts this `ComObject` into an interface that it implements.
+    ///
+    /// This does not need to adjust reference counts because `self` is consumed.
+    #[inline(always)]
+    pub fn into_interface<I: Interface>(self) -> I
```
kennykerr, line 154: This is similar conceptually to `Interface::cast` - may be helpful to use similar/same naming.

sivadeilra: Yeah, that's true.  cast() is dynamic, though, and these are static.  How about we keep cast() for dynamic (so that it's the same as IUnknown::cast()) and think of good names for the as_interface / etc. methods?  Any ideas?

crates/libs/core/src/imp/mod.rs
sivadeilra: Thanks; I've removed these.

crates/libs/core/src/imp/mod.rs
```diff
@@ -10,6 +10,7 @@ mod sha1;
 mod waiter;
 mod weak_ref_count;
 
+pub use crate::com_object::{ComImpl, ComObjectInterface};
```
dpaoliello, line 13: Nit: none of the other `use` declarations prefix with `crate::`

crates/libs/core/src/interface.rs
```diff
@@ -98,17 +119,93 @@ pub unsafe trait Interface: Sized + Clone {
...
+
+impl<'a, I> Copy for InterfaceRef<'a, I> {}
```
dpaoliello, line 151: Derive these traits?

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,274 @@
...
+/// Identifies types that can be placed in `ComObject`.
+///
+/// This trait links types that can be placed in `ComObject` with the types generated by the
+/// `#[implement]` macro. The `#[implement]` macro generates implementations of this trait.
+/// The generated types contain the vtable layouts and refcount-related fields for the COM
+/// object implementation.
+///
+/// This trait is an implementation detail of the Windows crates.
+/// User code should not deal directly with this trait.
+pub trait ComImpl {
```
dpaoliello, line 14: `#[doc(hidden)]` for this and `ComObjectInterface`?

dpaoliello (changes requested):

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,274 @@
...
+    // Note that we _do not_ provide a way to get a mutable reference to the outer box.
+    // It's ok to return &mut T, but not &mut T::Outer. That would allow someone to replace the
+    // contents of the entire object (box and reference count), which could lead to UB.
+    // This could maybe be solved by returning Pin<&mut T::Outer>, but that requires some
+    // additional thinking.
+
+    /// Gets a mutable reference to the object stored in the box, if the reference count
+    /// is exactly 1. If there are multiple references to this object then this returns `None`.
+    #[inline(always)]
+    pub fn get_mut(&mut self) -> Option<&mut T> {
```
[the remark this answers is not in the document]
sivadeilra: I was aiming for consistency with `std::rc::Rc::get_mut`, which has the same name and signature.

Maybe I should just rename `try_take` to `take`?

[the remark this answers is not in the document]
dpaoliello: If we're matching `Rc`, then we should change `try_take` to `into_inner`

crates/libs/core/src/interface.rs
sivadeilra: Can't.  `#[derive]` does the wrong thing with generic types.  It requires that `T` be `Copy`, which is wrong.

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,274 @@
...
+
+impl<T: ComImpl + Ord> Ord for ComObject<T> {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        let inner_self: &T = self.get();
+        let other_self: &T = other.get();
+        <T as Ord>::cmp(inner_self, other_self)
+    }
+}
+
+impl<T: ComImpl + core::fmt::Debug> core::fmt::Debug for ComObject<T> {
```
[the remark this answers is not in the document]
sivadeilra: The `Rc` forwarders do the same thing.  The only thing interesting or relevant inside `ComObject` is the ref count, and even that is rarely interesting.

crates/libs/core/src/com_object.rs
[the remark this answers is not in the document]
sivadeilra: Renamed `try_take` to just `take`.  Keeping `get_mut` for parity with `Rc::get_mut` and for general brevity.  I think `try_*` prefixes make more sense where there's another unconditional method.

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,274 @@
+use crate::{AsImpl, IUnknown, IUnknownImpl, Interface, InterfaceRef};
+use core::ptr::NonNull;
+use std::borrow::Borrow;
+
+/// Identifies types that can be placed in `ComObject`.
```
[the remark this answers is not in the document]
sivadeilra: Linkified everything that could be linkified

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,274 @@
...
+            Ok(outer_box.into_inner())
+        } else {
+            Err(self)
+        }
+    }
+
+    /// Casts to a given interface type.
+    ///
+    /// This always performs a `QueryInterface`, even if `T` is known to implement `I`.
+    /// If you know that `T` implements `I`, then use `as_interface` or `to_interface` because
```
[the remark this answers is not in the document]
sivadeilra: done

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,274 @@
...
+/// This type exists so that you can place an object into the heap and query for COM interfaces,
+/// without losing the safe reference to the implementation object.
+///
+/// Because the pointer inside this type is known to be non-null, `Option<ComObject<T>>` should
+/// always have the same size as a single pointer.
+///
+/// # Safety
+///
+/// The contained `ptr` field is an owned, reference-counted pointer to a _pinned_ Pin<Box<T::Outer>>.
+/// Although this code does not currently use `Pin<T>`,
```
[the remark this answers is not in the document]
sivadeilra: Fixed

crates/libs/core/src/com_object.rs
sivadeilra: In some cases user code needs to import traits, and I think concealing too much of the machinery makes it hard to use the Windows crate.  There are a handful of methods that are meaningful to user code, like `IUnknownImpl::from_inner_ref`, `get_impl`, etc.  I think it's best not to conceal these from docs.

crates/libs/core/src/imp/mod.rs
sivadeilra: Removed and used crate root.

dpaoliello (approved):

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,281 @@
...
+/// `ComObject<T>] instead of `ComObject<T_Impl>`.
+pub trait ComObjectInner {
+    /// The generated `<foo>_Impl` type (aka the "boxed" type or "outer" type).
+    type Outer: IUnknownImpl<Impl = Self>;
+}
+
+/// Describes the COM interfaces implemented by a specific COM object.
+///
+/// The `#[implement]` macro generates implementations of this trait. Implementations are attached
+/// to the "outer" types generated by `#[implemented]`, e.g. the `MyApp_Impl` type. Each
```
kennykerr, line 26:
```suggestion
/// to the "outer" types generated by `#[implement]`, e.g. the `MyApp_Impl` type. Each
```

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,281 @@
...
+    pub fn as_interface<I: Interface>(&self) -> InterfaceRef<'_, I>
+    where
+        T::Outer: ComObjectInterface<I>,
+    {
+        self.get_box().as_interface_ref()
+    }
+
+    /// Gets an owned (counted) reference to an interface that is implemented by this [`ComObject`].
+    #[inline(always)]
+    pub fn to_interface<I: Interface>(&self) -> I
```
kennykerr, line 150: How useful is this in practice? Seems like it wouldn't be much harder just writing `as_interface().to_owned()`.

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+    }
+
+    /// Gets a reference to the shared object stored in the box.
+    ///
+    /// [`ComObject`] also implements [`Deref`], so you can often deref directly into the object.
+    /// For those situations where using the [`Deref`] impl is inconvenient, you can use
+    /// this method to explicitly get a reference to the contents.
+    #[inline(always)]
+    pub fn get(&self) -> &T {
```
kennykerr, line 71: Taking a step back, how useful is `ComObject` outside the context of object construction? Since `get_mut` necessarily can't be used without exclusive ownership, it seems as if `ComObject` would only ever be used for object construction. If that is the case, I'm wondering whether anything can be simplified about this design. It is not overly complicated as it is, just something to consider.

kennykerr: You can of course still use `get` to retrieve a synchronization primitive like `RwLock` stored within the implementation and go from there. So this could well be used during the life cycle of the object. Is that how you see it being used?

sivadeilra: It's useful outside of just construction.  A common pattern in DWriteCore is that we create a bunch of COM objects, and hand out interfaces to the client process, but we also need to keep strongly-typed references to the contents.  Right now I'm doing that with the equivalent that is in the old `com` crate, and having `ComObject` in `windows_core` gets me that much closer to converting DWriteCore to use the Windows crates.

get_mut() itself is mainly useful during construction.  You create a ComObject, then get access to its guts while you're still the only reference holder.  You could get the same effect with RefCell, but since we already have the reference count in ComObject, why not use it?  It's basically free.

crates/libs/core/src/com_object.rs
sivadeilra: It's just a convenience.  I think we'll need to use this stuff a bunch to see what's the most useful.

crates/libs/core/src/com_object.rs
sivadeilra: fixed

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,281 @@
...
+        self.get_box().get_impl()
+    }
+
+    /// Gets a reference to the shared object's heap box.
+    #[inline(always)]
+    pub fn get_box(&self) -> &T::Outer {
```
kennykerr, line 77: Seems like an implementation detail - does it need to be public?

crates/libs/core/src/com_object.rs
```diff
@@ -0,0 +1,281 @@
...
+            // allow (incorrectly) modifying the reference count.
+            unsafe { Some(self.ptr.as_mut().get_impl_mut()) }
+        } else {
+            None
+        }
+    }
+
+    /// Returns `true` if this reference is the only reference to the `ComObject`.
+    #[inline(always)]
+    pub fn is_exclusive_reference(&self) -> bool {
```
kennykerr, line 103: Seems like an implementation detail - does it need to be public?

kennykerr (dismissed):
Looks good - thanks!

PR #3044, opened 2024-05-16
Upgrade win32metadata to 60.0.34

MarijnS95:
https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/60.0.34-preview

For me this most notably includes D3D12 Agility SDK 1.613.1 and annotations that leverage #2771.

PR #3045, opened 2024-05-16
bindgen/default: Improve formatting of the readme

MarijnS95:
Excerpt from #3044

Place package names and versions between backticks to format them as monospaced inline code making it easier to spot and copy, and wrap links between angle brackets to satisfy various Markdown linters.

Finally, apply shell formatting to the `mdmerge` oneliner script.


kennykerr (approved):
Thanks!

PR #3046, opened 2024-05-17
Add simple C++ test 

kennykerr:
I'd like to add some C++ interop validation without overly complicating the build. Here's a simple example to try out a technique. May need to limit this to MSVC targets on Windows. We'll see what the CI build has to say.


kennykerr:
This depends on https://github.com/rust-lang/cargo/issues/7825 not becoming an error.

kennykerr:
Any GNU experts know why this is failing?

The [log](https://github.com/microsoft/windows-rs/actions/runs/9130354092/job/25107078838?pr=3046) includes...

```
"-L" "D:\\a\\windows-rs\\windows-rs\\crates\\targets\\i686_gnu\\lib"
"-lwindows.0.52.0"
``` 

... so the lib should be present and yet its failing to find the import:

```
undefined reference to `_imp__CoTaskMemFree@4'
```

Works fine with MSVC.

kennykerr:
Oh well, at least it works with MSVC. 🤷

PR #3047, opened 2024-05-17
Allow windows, windows_core (and several other crates) to be compiled and used in no_std environments.

sivadeilra:
This allows many of the crates in this repo to be compiled and used in `#![no_std]` environments. This is a requirement for some low-level components, especially within the implementation of Windows itself.

These crates are updated: `windows`, `windows-core`, `windows-registry`, `windows-result`.  All of these crates now define a new feature, `std`, which is enabled by default.  If you _disable_ the `std` feature, then the crate will be compiled with `#![no_std]`. Because the `std` features are enabled by default, existing users of these crates should not see any difference in behavior.

Nearly all of the changes are extremely simple and do not change actual codegen. These just convert `std::ptr::null_mut` with `core::ptr::null_mut`, etc.

For those cases where a crate used dynamic memory containers (Box, String, Vec), I added an explicit dependency on the `alloc` crate and imported the needed types.  These are the same types that are re-exported from `std`, so this also has no effect on codegen.

In a few cases, code was using things from `std` that are currently only available in `std`, such as `OsString` or `Mutex`. In those cases, I placed the code using those items under `#[cfg(feature = "std")]`.  Since the `std` feature is enabled by default, this will not break any existing users of those crates.

I also noticed a bug in passing in `BasicString`. It can construct a `&[u16]` over a null pointer, which is UB.  I've fixed that bug in this PR.


crates/libs/result/src/strings.rs
```diff
@@ -35,7 +35,12 @@ impl BasicString {
     }
 
     pub fn as_wide(&self) -> &[u16] {
-        unsafe { std::slice::from_raw_parts(self.as_ptr(), self.len()) }
```
sivadeilra, line 38: This is UB, in safe code, if the string is a null pointer.

crates/libs/registry/Cargo.toml
```diff
@@ -11,6 +11,10 @@ readme = "readme.md"
 categories = ["os::windows-apis"]
 exclude = ["tests"]
 
+[features]
+default = []
```
dpaoliello, line 15:
```suggestion
default = ["std"]
```

crates/libs/registry/Cargo.toml
```diff
@@ -11,6 +11,10 @@ readme = "readme.md"
...
+std = []
```
dpaoliello, line 16:
```suggestion
std = ["windows-result/std"]
```

dpaoliello (changes requested):

crates/libs/registry/Cargo.toml
sivadeilra: Actually, the more I think about feature injection, I think we don't actually need it unless A-->B where A actually uses something from B's std-dependent exterior API.

crates/libs/registry/Cargo.toml
sivadeilra: thanks; fixed

crates/libs/registry/Cargo.toml
dpaoliello: In which case, we should `check` each crate with no default features to make sure we don't miss dependencies

crates/libs/registry/src/value_iterator.rs
```diff
@@ -79,7 +80,8 @@ impl<'a> Iterator for ValueIterator<'a> {
                             Value::String(String::new())
                         } else {
                             let value = unsafe {
-                                std::slice::from_raw_parts(
+                                // TODO: Vec<u8> does not guarantee alignment for u16
```
kennykerr, line 83: Let's create an issue to avoid forgetting about this TODO.

kennykerr: #3048

crates/libs/registry/Cargo.toml
kennykerr: * I think `windows-registry` could be entirely `no_std`.
* I could separately add a generated yml workflow for testing no default feature builds.

crates/libs/registry/src/value_iterator.rs
sivadeilra: Thanks!

crates/libs/registry/Cargo.toml
sivadeilra: Yeah, I see what you mean.  I'm a bit concerned at the combinatorial explosion in `A{no_std?} --> B{no_std?}` variations, so I'm not sure how to implement this suggestion without things blowing up.  But it's definitely a valid concern...

kennykerr: So I think the `windows-registry` crate doesn't need features of its own, depends on the `windows-result` crate with `default-features = false`, and can be `#![no_std]`.

sivadeilra: Agreed.  I've removed the feature from `windows-registry`; it is now always `no_std` (except in `test` builds).

kennykerr (approved):

crates/libs/registry/Cargo.toml
kennykerr: Looks like this has been addressed with some follow-up work. Will merge now to get this batch of changes in.

PR #3050, opened 2024-05-21
Re-enable `unexpected_cfgs`

ChrisDenton:
This is an attempt to re-enable `unexpected_cfgs` using the new `unexpected_cfgs` config in `Cargo.toml`. This avoids the need for a build script while still checking for typos in global cfgs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,4 @@ exclude = [
 rust_2018_idioms = "warn"
 unused_qualifications = "warn"
 missing_docs = "warn"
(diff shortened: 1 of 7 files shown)


ChrisDenton:
Hm, looks like it isn't enough to just add it to the workspace lint table.

kennykerr:
Well that's less bad. 🥲 Thanks Chris!

kennykerr (approved):

ChrisDenton:
Oh. Hm. Looks like this is expected on stable. A fix to silence the warnings was backported to beta but not current stable. So it will be an issue until the next release.

PR #3051, opened 2024-05-21
Fix soundness hole in ComObject

sivadeilra:
This fixes a soundness hole in `ComObject`. The soundness hole is that we should never allow safe Rust code to hold an _owned_ instance of `MyApp_Impl` objects (implementations of COM objects) because those objects contain reference counts and provide safely-callable methods that adjust those reference counts.  If Rust code holds an owned instance of such a type, then we don't control its lifetime; it may be placed on the stack, in a `static`, etc. and we have no control over that.

This PR closes the soundness hole by providing only one way to get access to any `MyApp_Impl` type -- these types are immediately placed into a heap allocation and the only way to access them is through a `ComObject` reference.

A few other minor improvements:

* `ComObject::as_reference()` and friends now work with `IInspectable`.
* Switched vtable pointers to use `&'static` instead of `*const`.
* Move vtable initializers directly into the function which uses them, so symbol names don't "leak" outward.
* Converted some existing `From` implementations to use safe `ComObject` code.

### DIFF
--- a/crates/libs/core/src/com_object.rs
+++ b/crates/libs/core/src/com_object.rs
@@ -15,10 +15,30 @@ use core::ptr::NonNull;
 /// User code should not deal directly with this trait.
 ///
 /// This trait is sort of the reverse of [`IUnknownImpl`]. This trait allows user code to use
-/// `ComObject<T>] instead of `ComObject<T_Impl>`.
-pub trait ComObjectInner {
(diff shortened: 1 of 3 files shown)


dpaoliello (approved):

kennykerr (approved):
Thanks!

PR #3053, opened 2024-05-23
Add `no-default-features` yml workflow

kennykerr:
Following up on https://github.com/microsoft/windows-rs/pull/3047#discussion_r1605448388 where @dpaoliello suggested we test crates with no default features.

### DIFF
--- a/.github/workflows/no-default-features.yml
+++ b/.github/workflows/no-default-features.yml
@@ -1,0 +1,54 @@
+name: no-default-features
+
+on:
+  pull_request:
+  push:
+    paths-ignore:
+      - '.github/ISSUE_TEMPLATE/**'
+    branches:
+      - master
+
+env:
+  RUSTFLAGS: -Dwarnings
+
+jobs:
+  check:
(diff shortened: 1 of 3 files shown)


kennykerr:
@ChrisDenton here's some more fallout from the `unexpected_cfgs` thing. 

https://github.com/microsoft/windows-rs/actions/runs/9207820449/job/25328694260?pr=3053

kennykerr:
Actually looking closer it might be an unrelated nightly change.

ChrisDenton:
Yeah, I think it's basically saying that `unused_qualifications` is redundant because it's already being set in `rust_2018_idioms`.

ChrisDenton:
Aside: lint changes have been exhausting recently.

kennykerr:
Yes, really losing the value for me.

ChrisDenton:
Oh, looks like you were right the first time. It seems that any group has to be priority -1 even if it doesn't include the lint.

kennykerr:
Yep, that's super confusing.

PR #3056, opened 2024-05-24
Switch to `windows-2022` hosted runners

kennykerr:
In an attempt to fix #3054 build issues, I'm testing whether we can switch to the latest GitHub runners so that we can build with VS 2022 rather than 2019. 

I expect action.yml will also need to be "fixed" as it has some hardcoded paths.

### DIFF
--- a/.github/actions/fix-environment/action.yml
+++ b/.github/actions/fix-environment/action.yml
@@ -54,7 +54,7 @@ runs:
     # 1. Old mingw-w64-* binutils that prevent deterministic builds.
     # 2. Missing llvm-dlltool in the native Windows LLVM package.
     - name: Update packages
-      shell: pwsh
+      shell: bash
       run: |
         pacman -S --noconfirm mingw-w64-x86_64-binutils
         pacman -S --noconfirm mingw-w64-x86_64-llvm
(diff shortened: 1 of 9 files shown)


kennykerr:
@riverar `lib.yml` uses `vswhere` - I assume we can do the same in `action.yml` to avoid these hardcoded paths?

https://github.com/microsoft/windows-rs/blob/40d35fabf553a63c577969e70b4e0ff70e4e11c6/.github/workflows/lib.yml#L63

riverar:
We could, yup. I'm not really a fan of vswhere honestly but that's me.

kennykerr:
I don't really care - just noticed that's what lib.yml does - I don't have it installed locally for some reason so I assume it doens't even come with Visual Studio.

kennykerr:
It would just be nice not to have to hardcode these paths.

ChrisDenton:
vswhere should always be installed with visual studio. And always in the same path: `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe` source: https://github.com/Microsoft/vswhere?tab=readme-ov-file#visual-studio-locator

riverar (approved):

ChrisDenton (approved):
lgtm!

riverar:
Thanks!

PR #3060, opened 2024-05-26
Fix bug in COM interface chain support

sivadeilra:
The definition of a COM interface may inherit from another interface. These are known as "interface chains". The `#[implement]` macro allows designers to specify only the minimal set of interface chains that are needed for a given COM object implementation. The `#[implement]` macro (and the `#[interface]` macro) work together to pull in the implementations of all interfaces along the chain.

Unfortunately there is a bug in the implementation of `QueryInterface` for interface chains. The current `QueryInterface` implementation will only check the IIDs of the interfaces at the root of the chian, i.e. the "most-derived" interface. `QueryInterface` will not search the IIDs of interfaces that are in the inheritance chain.

This bug is demonstrated (detected) by the new unit tests in `crates/tests/implement_core/src/com_chain.rs`. This PR fixes the bug by generating an `fn matches()` method that checks the current IID and then checks the parent interface (if any) by calling its `match()` method. This fixes the unit test.

### DIFF
--- a/crates/libs/interface/src/lib.rs
+++ b/crates/libs/interface/src/lib.rs
@@ -211,6 +211,17 @@ impl Interface {
         let parent_vtable_generics = if self.parent_is_iunknown() { quote!(Identity, OFFSET) } else { quote!(Identity, Impl, OFFSET) };
         let parent_vtable = self.parent_vtable();
 
+        // or_parent_matches will be `|| parent::matches(iid)` if this interface inherits from another
+        // interface (except for IUnknown) or will be empty if this is not applicable. This is what allows
+        // QueryInterface to work correctly for all interfaces in an inheritance chain, e.g.
+        // IFoo3 derives from IFoo2 derives from IFoo.
+        //
+        // We avoid matching IUnknown because object identity depends on the uniqueness of the IUnknown pointer.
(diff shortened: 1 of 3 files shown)


kennykerr:
'matches' definitely needs to check parent IIDs.

dpaoliello (approved):

arlosi (approved):

wravery (approved):

crates/libs/interface/src/lib.rs
```diff
@@ -211,6 +211,17 @@ impl Interface {
         let parent_vtable_generics = if self.parent_is_iunknown() { quote!(Identity, OFFSET) } else { quote!(Identity, Impl, OFFSET) };
         let parent_vtable = self.parent_vtable();
 
+        // or_parent_matches will be `|| parent::matches(iid)` if this interface inherits from another
+        // interface (except for IUnknown) or will be empty if this is not applicable. This is what allows
+        // QueryInterface to work correctly for all interfaces in an inheritance chain, e.g.
+        // IFoo3 derives from IFoo2 derives from IFoo.
+        //
+        // We avoid matching IUnknown because object identity depends on the uniqueness of the IUnknown pointer.
```
kennykerr, line 219: May be worth adding a test for this.

PR #3065, opened 2024-06-03
Basic implementation of impl-on-outer

sivadeilra:
This PR allows COM objects to implement COM interfaces on either the "outer" (`MyApp_Impl`) or "inner" (`MyApp`) types.  This is important because many COM objects need access to the full COM object, not just the contents of the wrapped "inner" type.

For example, COM objects may need to cast themselves to a particular `IFoo` that they implement and then return that `IFoo` pointer as an output parameter of a COM method call.  This is not possible in the current design because COM objects receive method calls that are typed `&self : T` where `T` is the inner type, so they do not have access to the vtables or reference count of the current COM object.

This PR fixes this by allowing COM servers to declare that certain interfaces (interface chains, actually) are implemented on the outer type, not the inner type.  See the changes in `crates/tests/implement_core/src/impl_on_outer.rs` for an example of how to use this.  Briefly, you enable it like this:

```
#[implement(ImplLocation = "Outer", IFoo)]
struct MyApp { ... }

impl IFoo_Impl for MyApp_Impl { ... }
```

This PR is in "draft" state because I do not yet want to update the generated code in `crates/libs/windows/src/Windows` because it is overwhelming. Assuming we move forward with this PR, I will update it with another commit that contains the updated generated code.


crates/tests/implement_core/src/impl_on_outer.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[interface("cccccccc-0000-0000-0000-000000000004")]
+unsafe trait IBar: IUnknown {
+    fn goodbye(&self);
+}
+
+// This tests that we can compile a COM object that has some COM interfaces implemented on the
+// outer object and some on the inner object.
+#[implement(ImplLocation = Outer, IFoo3, ImplLocation = Inner, IBar)]
```
dpaoliello, line 25: Not a fan of this syntax: I'd prefer if it were obvious which interface it applied to (e.g., via parens), if it were only permitted once per attribute.

crates/libs/implement/src/lib.rs
```diff
@@ -505,17 +526,27 @@ impl syn::parse::Parse for UseTree2 {
...
+                if ident == "TrustLevel" {
+                    input.parse::<syn::Token![=]>()?;
+                    let span = input.span();
+                    let value = input.call(syn::Ident::parse_any)?;
+                    match value.to_string().as_str() {
+                        "Partial" => Ok(UseTree2::TrustLevel(1)),
+                        "Full" => Ok(UseTree2::TrustLevel(2)),
+                        _ => Err(syn::parse::Error::new(span, "`TrustLevel` must be `Partial` or `Full`")),
+                    }
+                } else if ident == "ImplLocation" {
```
dpaoliello, line 538: Trying to think of the best way to phrase this to describe to the end-user what it will do.

Currently I'm thinking:
```
ImplInterfaceOn = OuterImpl
ImplInterfaceOn = InnerType
```

crates/tests/implement_core/src/impl_on_outer.rs
```diff
@@ -0,0 +1,63 @@
...
+struct MyApp {}
+
+impl IFoo_Impl for MyApp_Impl {
+    unsafe fn hello(&self) {
```
dpaoliello, line 29: Would like to see these tests verify that this is the correct `self` type

dpaoliello (changes requested):

crates/tests/implement_core/src/impl_on_outer.rs
sivadeilra: This test does verify that.  If you get it wrong, it won't compile.

crates/tests/implement_core/src/impl_on_outer.rs
sivadeilra: Yeah, I'm not thrilled about it.  I went with this since there was already a key/value syntax being used for trust levels.  But those apply to the entire `#[implement]`, not to a specific interface (chain), so it's not really a direct correspondence.

I'm open to better syntax.

riverar: Perhaps only one instance of this attribute should be supported, forcing developers to use separate implement lines? Does some sort of flag make more sense here? Just thinking out loud.

Example:
```rust
#[implement(TrustLevel = Partial)]
#[implement(IFoo, IFoo2, Inner)]
#[implement(IFoo3)]
struct co_class;
```

sivadeilra: Unfortunately you cannot apply `#[implement]` to a given type more than once.  The effects do not merge.

dpaoliello: I was thinking something like:
```rust
#[implement(IFoo3(ImplLocation = Outer), IBar(ImplLocation = Inner))]
```

sivadeilra: I looked at a bunch of options, and landed on this syntax:

```
#[implement(IFoo3 @ Outer, IBar)]
struct ...
```

It's brief, easy to parse, pretty obvious to a reader, and won't break any existing code.

riverar: What about adding a revision/version to the `implement` macro (or wire up an attribute?), then just change the default implementation in "v2"? Should devs _want_ to continue using the inner target when the outer target is available? If not, ...

Example:
```rust
#[implement(version=2, IFoo, IFoo2)]
struct co_class;
```

sivadeilra: I think that's worth thinking about, long-term, and maybe as a redesign of the API.  But I don't think calling it "v2" is the right approach.  It does something different.

For many apps, the "inner" behavior is nice and simple, since the developer only sees a `MyApp` type and never has to know about or think about the generated `MyApp_Impl` or the difference between the two.  I think using the outer type (for interface implementations) will be the exception, and not the rule.

Maybe it would be best to push everything to a single model, where the interfaces are always defined on `MyApp_Impl`.  But we can't make that transition in a single PR without breaking all existing users of the Windows crate.  If we want to do that, we'll need a way to soften the transition, and doing that is outside of what could be done in a single PR.

dpaoliello (dismissed):

kennykerr:
More fundamentally, the problem with the existing implementation is that certain functions - like casting/querying for an interface implemented by the implementation - assume that the type that the `implement` macro is attached to is already boxed and there's no safe/sound way to panic if and when that's not the case. This PR aims to solve that nicely and it should just be "the way" to implement COM objects in Rust. So,

1. I don't think its sufficient to offer an optional alternative.
2. There should just be one way to do this and that should be simplest and most streamlined default rather than something you need to opt in to. 
3. Even if both approaches were sound, having two non-complimentary ways to do essentially the same thing seems very confusing.
4. I don't like the ergonomics of the sample in the PR description even if you remove the `@ Outer` thing. 

```rust
#[implement(IFoo)]
struct MyApp { ... }

impl IFoo_Impl for MyApp_Impl { ... }
``` 

There's no hint that `IFoo_Impl` is even a name to be reckoned with. That's more a problem with the interface definition, but `MyApp_Impl` is likewise something you'd have to "know" a priori. Even `impl IFoo::Impl for MyApp::Impl` would be less mysterious and name-mangily. Maybe something else entirely. I'll play around with it over the next few days and see if I can come up with something a bit more natural. If we are to make a change here, I want to be confident that we are on a good path before we disrupt existing code.

sivadeilra:
The more I think about it, I think we should just flip to `MyApp_Impl` for COM interface implementations and be done with it.  Yes, it's a breaking change, but `windows-rs` has made breaking changes before, and the major version number is still 0.

sivadeilra:
Ok, I implemented support for placing interface impls on `ComObject<T>` instead of on `T` or `T_Impl`.  It looks nice, especially from the POV of the developer implementing a COM object, because `ComObject<MyApp>` gives developers an "go to docs" thing, and it gives them an easy way to discover the methods that are in-scope for `self`.

I created a draft PR, #3071 , just for discussion.  That PR still has the "inner vs. outer" option on `#[implement]`, and I'm leaning toward "this needs to just be a breaking change".  So either 1) we go with the opt-in, or 2) we go with the breaking change.  In either case, I'll update _this_ PR after the discussion, to reflect the final design, rather than #3071.

kennykerr:
I like this!

> 
> ```rust
> impl IFoo_Impl for ComObject<MyApp> {
>    ...
> }
> ```

sivadeilra:
> I like this!

Me too! ...  Unfortunately, I tried taking that all the way to completion, and ran into an unsolvable problem with generics and trait implementations.  It's not going to work.  :(

crates/libs/bindgen/src/rust/extensions/impl/Foundation/Collections/Iterable.rs
```diff
@@ -27,52 +28,54 @@ where
...
 
     fn MoveNext(&self) -> windows_core::Result<bool> {
-        let owner: &StockIterable<T> = unsafe { windows_core::AsImpl::as_impl(&self.owner) };
+        let owner: &StockIterable<T> = &self.owner;
         let current = self.current.load(std::sync::atomic::Ordering::Relaxed);
 
         if current < owner.values.len() {
-            self.current.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+            self.current
+                .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
```
kennykerr, line 64: Is this just formatting?

crates/libs/bindgen/src/rust/extensions/impl/Foundation/Collections/Iterable.rs
```diff
@@ -7,17 +7,18 @@ where
...
-            // TODO: ideally we can do an AddRef rather than a QI here (via cast)...
-            // and then we can get rid of the unsafe as well.
-            Ok(StockIterator { owner: self.cast()?, current: 0.into() }.into())
-        }
+        use windows_core::IUnknownImpl;
+        Ok(windows_core::ComObject::new(StockIterator {
+            owner: self.to_object(),
+            current: 0.into(),
+        })
+        .into_interface())
```
kennykerr, line 21: I wonder, can we implement `Into`/`From` here as a shorthand for the simple case of returning the interface as before - then it would be similar to before: `Ok(StockIterator {...}.into())` ?

crates/libs/bindgen/src/rust/extensions/impl/Foundation/Collections/Iterable.rs
```diff
@@ -27,52 +28,54 @@ where
...
 
-impl<T> IIterator_Impl<T> for StockIterator<T>
+impl<T> IIterator_Impl<T> for StockIterator_Impl<T>
 where
     T: windows_core::RuntimeType,
     T::Default: Clone,
 {
     fn Current(&self) -> windows_core::Result<T> {
-        let owner: &StockIterable<T> = unsafe { windows_core::AsImpl::as_impl(&self.owner) };
+        let owner: &StockIterable<T> = &self.owner;
```
kennykerr, line 41: Much nicer.

crates/libs/bindgen/src/rust/extensions/impl/Foundation/Collections/Iterable.rs
```diff
@@ -85,6 +88,6 @@ where
     type Error = windows_core::Error;
     fn try_from(values: Vec<T::Default>) -> windows_core::Result<Self> {
         // TODO: should provide a fallible try_into or more explicit allocator
-        Ok(StockIterable { values }.into())
+        Ok(windows_core::ComObject::new(StockIterable { values }).into_interface())
```
kennykerr, line 91: As above, be nice if this simple case had a shorthand.

crates/libs/bindgen/src/rust/extensions/impl/Foundation/Collections/Iterable.rs
sivadeilra: Yes.  These files weren't being run through rustfmt before.

crates/libs/bindgen/src/rust/extensions/impl/Foundation/Collections/Iterable.rs
sivadeilra: The `into()` thing still totally works.  I just like the `ComObject::new(...).into_interface()` more because it's more explicit.

crates/libs/bindgen/src/rust/extensions/impl/Foundation/Collections/Iterable.rs
sivadeilra: It still works; just being explicit.  I can change this back, but I like the explicitness.

kennykerr (approved):

kennykerr:
Looks like a new Rust nightly update is failing the merge build workflow.

PR #3066, opened 2024-06-03
Fix Debug impls for COM interfaces

sivadeilra:
Currently, the Debug impl for COM interfaces shows the recursive interface chain, like so:

```
IDWriteFontFace5(IDWriteFontFace4(IDWriteFontFace3(IDWriteFontFace2(IDWriteFontFace(0xNNN)))))
```

That's not very useful. This PR trims it down to just the current interface:

```
IDWriteFontFace5(0xNNN)
```

### DIFF
--- a/crates/tests/implement_core/src/com_object.rs
+++ b/crates/tests/implement_core/src/com_object.rs
@@ -416,6 +416,15 @@ fn common_method_name() {
     assert_eq!(unsafe { ibar2.common() }, std::f32::consts::PI);
 }
 
+#[test]
+fn debug_fmt() {
+    let app = MyApp::new(42);
+    let iunknown: IUnknown = app.to_interface();
+    println!("IUnknown = {iunknown:?}");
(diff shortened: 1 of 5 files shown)


dpaoliello (approved):

crates/tests/implement_core/src/com_object.rs
```diff
@@ -416,6 +416,15 @@ fn common_method_name() {
     assert_eq!(unsafe { ibar2.common() }, std::f32::consts::PI);
 }
 
+#[test]
+fn debug_fmt() {
+    let app = MyApp::new(42);
+    let iunknown: IUnknown = app.to_interface();
+    println!("IUnknown = {iunknown:?}");
+    let ifoo: IFoo = app.to_interface();
+    println!("IFoo = {ifoo:?}");
```
kennykerr, line 425: These are not tests. 😉

Issue #2991, opened 2024-04-12
Add `noexcept` support for WinRT methods/properties


kennykerr:
### Suggestion

I originally introduced the `NoExceptionAttribute` attribute to optimize C++ calling patterns in C++/WinRT, but it applies equally well to Rust.

PR #3070, opened 2024-06-05
Add WinRT `noexcept` support 

kennykerr:
A few years ago, I introduced an optimization for https://github.com/microsoft/cppwinrt that allowed methods and properties to be attributed as `noexcept` to improve call site code generation. While all WinRT methods must return an `HRESULT` on the ABI, `noexcept` methods always return `S_OK` and are thus infallible. This does not change the way C++ code "looks and feels" but C++/WinRT would mark such methods [noexcept](https://en.cppreference.com/w/cpp/language/noexcept_spec) in the C++ sense which allows the C++ compiler to avoid a bunch of exception handling overhead at call sites, reducing code size. 

In Rust, the `noexcept` can further be used to avoid having to `unwrap` method calls that are known to be infallible. This also reduces error handling overhead but additionally simplifies implementations and call sites as method return values may be used directly without unwrapping. 

The test included with this update illustrates the impact of this change. Consider the following IDL definition of an interface:

```C#
namespace Test
{
    interface ITest
    {
        void MethodString(String test);
        void MethodInt32(Int32 test);
        void MethodTest(ITest test);
        String String;
        Int32 Int32;
        ITest Test;

        [noexcept] void MethodStringN(String test);
        [noexcept] void MethodInt32N(Int32 test);
        [noexcept] void MethodTestN(ITest test);
        [noexcept] String StringN;
        [noexcept] Int32 Int32N;
        [noexcept] ITest TestN;
    }
}
```

In Rust, the first batch of methods all return a `Result` while the second do not:

```rust
#[test]
fn test_except() -> Result<()> {
    let test: ITest = Test::default().into();

    test.MethodString(h!("abc"))?;
    assert_eq!(test.String()?, "abc");

    test.MethodInt32(123)?;
    assert_eq!(test.Int32()?, 123);

    test.MethodTest(&test)?;
    assert_eq!(&test.Test()?, &test);

    Ok(())
}

#[test]
fn test_noexcept() {
    let test: ITest = Test::default().into();

    test.MethodStringN(h!("abc"));
    assert_eq!(test.StringN(), "abc");

    test.MethodInt32N(123);
    assert_eq!(test.Int32N(), 123);

    test.MethodTestN(&test);
    assert_eq!(test.TestN().as_ref(), Some(&test));
}
```

Fixes: #2991


crates/libs/bindgen/src/metadata.rs
```diff
@@ -475,6 +475,10 @@ fn method_def_last_error(row: MethodDef) -> bool {
     }
 }
 
+pub fn method_def_is_noexcept(method: MethodDef) -> bool {
+    method.has_attribute("NoExceptionAttribute")
```
kennykerr, line 479: The special "remove" methods for event handlers are also `noexcept` in C++/WinRT. I'll add that as a separate update since that introduces a lot more (generated) code changes in the `windows` crate.

crates/libs/windows/src/Windows/UI/UIAutomation/Core/impl.rs
```diff
@@ -1,5 +1,5 @@
 pub trait ICoreAutomationConnectionBoundObjectProvider_Impl: Sized {
-    fn IsComThreadingRequired(&self) -> windows_core::Result<bool>;
+    fn IsComThreadingRequired(&self) -> bool;
```
kennykerr, line 2: There are a handful of system APIs that adopted the `noexcept` attribute. It really should be applied more liberally and can be done retroactively to improve code size for callers.

kennykerr: @jonwis 👋

eholk: Is adding the `noexcept` attribute meant to be a semver-breaking change? With this PR, it seems like it would become one, but I'm not sure if that's desired/acceptable.

riverar: The crate is pre-release/0.x so there is no _semver-breaking_ change here right?

eholk: Technically, but because of the Rust community's reluctance to ever release 1.0 crates, cargo treats the minor version of a 0.x crate more like the major version of a 1.x crate.

I'm more worried about the long term. I interpreted Kenny's comment to mean we should update the winmd source files to add `noexcept` where we can, and the part about doing it retroactively means if we start doing that after this crate gets to 1.0 each of those would be a breaking change.

I guess this comes down to what the winmd files consider to be breaking changes. Obviously a breaking change there would imply breaking changes for any downstream projections.

crates/libs/bindgen/src/rust/winrt_methods.rs
```diff
@@ -47,38 +60,64 @@ pub fn writer(
...
 
+    let vcall = quote! { (windows_core::Interface::vtable(this).#vname)(windows_core::Interface::as_raw(this), #args #return_arg) };
+
     let vcall = match &signature.return_type {
         metadata::Type::Void => {
-            quote! {
-                (windows_core::Interface::vtable(this).#vname)(windows_core::Interface::as_raw(this), #args).ok()
+            if noexcept {
+                quote! {
+                    _ = #vcall;
```
eholk, line 69: is it worth generating a `debug_assert` here to make sure the call actually returned `S_OK` like its contract says it should?

eholk (approved):

crates/libs/windows/src/Windows/UI/UIAutomation/Core/impl.rs
kennykerr: Yes, that's a good point. It doesn't break the ABI by adding it, but it would be as if the API changed for Rust callers.

jonwis: @oldnewthing as we just had an internal conversation about noexcept at the _SDK ABI_ layer. Projections, no problem with them adding noexcept decorations to enforce that. Generally the Win32 ABI is already implicitly noexcept since we can't throw you a C++ exception from within combase.dll to your caller and have it make any sense.

Adding the annotation is a similar conversation as nullable from a while ago. Knowing when to apply it automatically is hard but you can probably figure out some basic rules.

But, per our API design guidelines (man I should publish those someday) property fetches & writes are not supposed to raise exceptions; only method calls.

crates/libs/bindgen/src/rust/winrt_methods.rs
kennykerr: done!

crates/libs/bindgen/src/rust/winrt_methods.rs
```diff
@@ -47,38 +60,68 @@ pub fn writer(
...
+    let vcall = quote! { (windows_core::Interface::vtable(this).#vname)(windows_core::Interface::as_raw(this), #args #return_arg) };
+
     let vcall = match &signature.return_type {
         metadata::Type::Void => {
-            quote! {
-                (windows_core::Interface::vtable(this).#vname)(windows_core::Interface::as_raw(this), #args).ok()
+            if noexcept {
+                quote! {
+                    let hresult__ = #vcall;
+                    debug_assert!(hresult__.0 == 0);
```
sivadeilra, line 70: Suggest:

```
debug_assert_eq!(hresult__.0, 0, "Method marked `noexcept` returned a failure code");
```

crates/libs/bindgen/src/rust/winrt_methods.rs
```diff
@@ -47,38 +60,68 @@ pub fn writer(
...
         _ if signature.return_type.is_winrt_array() => {
-            quote! {
-                let mut result__ = core::mem::MaybeUninit::zeroed();
-                (windows_core::Interface::vtable(this).#vname)(windows_core::Interface::as_raw(this), #args #return_arg)
-                    .map(|| result__.assume_init())
+            if noexcept {
+                quote! {
+                    let mut result__ = core::mem::MaybeUninit::zeroed();
+                    let hresult__ = #vcall;
+                    debug_assert!(hresult__.0 == 0);
```
sivadeilra, line 83: As above, suggest

```
debug_assert_eq!(hresult__.0, 0, "Method marked `noexcept` returned a failure code");
```

crates/libs/bindgen/src/rust/winrt_methods.rs
```diff
@@ -189,31 +232,53 @@ pub fn gen_upcall(
...
-                        let (ok_data__, ok_data_len__) = ok__.into_abi();
-                        // use `core::ptr::write` since `result` could be uninitialized
-                        core::ptr::write(result__, ok_data__);
-                        core::ptr::write(result_size__, ok_data_len__);
-                        windows_core::HRESULT(0)
+            if noexcept {
+                quote! {
+                    let ok__ = #inner(#this #(#invoke_args,)*);
+                    let (ok_data__, ok_data_len__) = ok__.into_abi();
+                    core::ptr::write(result__, ok_data__);
```
sivadeilra, line 266: This can also be `result__.write(ok_data__);`, for brevity.

crates/libs/bindgen/src/rust/writer.rs
```diff
@@ -1284,12 +1284,14 @@ impl Writer {
...
+            let noexcept = metadata::method_def_is_noexcept(signature.def);
+
             let params = signature
                 .params
                 .iter()
                 .map(|p| self.winrt_produce_type(p, !is_delegate));
 
-            let return_type = match &signature.return_type {
+            let return_type_tokens = match &signature.return_type {
                 metadata::Type::Void => quote! { () },
```
sivadeilra, line 1295: Clippy will usually complain about this and suggest converting to

```
let return_type_tokens = if let metadata::Type::Void = &signature.return_type { ... } else { ... };
```

crates/libs/bindgen/src/rust/winrt_methods.rs
kennykerr: I'm reluctant to add this since it can be repeated oodles of times in generated code. Let's wait and see whether this assertion raises any questions and reconsider. As it is, both sides of the contract will almost always be generated by tools - `cppwinrt` or `windows-bindgen` - and it would be unusual for somebody to write this ABI code by hand.

kennykerr:
https://github.com/microsoft/windows-rs/pull/3070/commits/7dfe20bae4210df4577e82df731d48d92a41741f adds interop testing against C++/WinRT.

crates/libs/bindgen/src/rust/winrt_methods.rs
kennykerr: Will update this separately as it will produce a ton of code gen changes.

PR #3073, opened 2024-06-06
Optimize `IUnknown` identity checks

### LINKED ISSUE #3069: Add test for `IUnknown` identity queries
May be worth adding a test to verify QI works predictably for `IUnknown` identity queries.

_Originally posted by @kennykerr in https://github.com/microsoft/windows-rs/pull/3060#discussion_r1626319409_

sivadeilra:
This fixes #3069 by adding more coverage for IUnknown identity tests.

This also adds a fast-path for comparing interface identity.  If two COM interface pointers have the same pointer view, then they refer to the same object and we can avoid the QueryInterface calls.

### DIFF
--- a/crates/libs/core/src/unknown.rs
+++ b/crates/libs/core/src/unknown.rs
@@ -47,11 +47,17 @@ impl Drop for IUnknown {
 
 impl PartialEq for IUnknown {
     fn eq(&self, other: &Self) -> bool {
+        // First we test for ordinary pointer equality. If two COM interface pointers have the
+        // same pointer value, then they are the same object. This can save us a lot of time,
+        // since calling QueryInterface is much more expensive than a single pointer comparison.
+        //
+        // However, interface pointers may have different values and yet point to the same object.
         // Since COM objects may implement multiple interfaces, COM identity can only
         // be determined by querying for `IUnknown` explicitly and then comparing the
         // pointer values. This works since `QueryInterface` is required to return
         // the same pointer value for queries for `IUnknown`.
-        self.cast::<IUnknown>().unwrap().0 == other.cast::<IUnknown>().unwrap().0
+        self.as_raw() == other.as_raw()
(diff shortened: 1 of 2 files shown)


crates/libs/core/src/unknown.rs
```diff
@@ -47,11 +47,17 @@ impl Drop for IUnknown {
...
+        // same pointer value, then they are the same object. This can save us a lot of time,
+        // since calling QueryInterface is much more expensive than a single pointer comparison.
+        //
+        // However, interface pointers may have different values and yet point to the same object.
         // Since COM objects may implement multiple interfaces, COM identity can only
         // be determined by querying for `IUnknown` explicitly and then comparing the
         // pointer values. This works since `QueryInterface` is required to return
         // the same pointer value for queries for `IUnknown`.
-        self.cast::<IUnknown>().unwrap().0 == other.cast::<IUnknown>().unwrap().0
+        self.as_raw() == other.as_raw()
```
kennykerr, line 59: Neat!

crates/tests/implement_core/src/com_object.rs
```diff
@@ -417,12 +420,80 @@ fn common_method_name() {
...
+// Test that we always get the same IUnknown pointer for an object, regardless of which
+// interface we use to query for it.
+#[test]
+fn iunknown_identity() {
+    let app = MyApp::new(0);
+
+    println!("identity = {:?}", &app.identity as *const _);
+    println!("vtables.0 = {:?}", &app.vtables.0 as *const _);
+    println!("vtables.1 = {:?}", &app.vtables.1 as *const _);
+    println!("vtables.2 = {:?}", &app.vtables.2 as *const _);
```
kennykerr, line 444: Intentional? Are we asserting them elsewhere?

crates/libs/core/src/unknown.rs
kennykerr: I forgot I added this optimization to cppwinrt years ago... 😳

kennykerr (approved):

PR #3074, opened 2024-06-06
Fix warnings from Rust 1.80.0 nightly

sivadeilra:
CI jobs are failing because Rust nightly tools have added new warnings. This PR fixes all warnings as of nightly 2024-06-04.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,6 @@ exclude = [
 
 [workspace.lints.rust]
 rust_2018_idioms = { level = "warn", priority = -1 }
-unused_qualifications = "warn"
+unused_qualifications = "allow" # See https://github.com/microsoft/windows-rs/issues/3076
 missing_docs = "warn"
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(windows_raw_dylib, windows_debugger_visualizer)'] }
-
-# See https://github.com/microsoft/windows-rs/issues/3076
-unused-qualifications = { level = "allow" }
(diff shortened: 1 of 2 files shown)


ChrisDenton:
The change here was `size_of` and `align_of` have now been added to the prelude https://github.com/rust-lang/rust/pull/123168

ChrisDenton:
I think the correct fix would be to also add `use std::mem::{align_of, size_of};` to the top of the module until the library's MSRV is 1.80 or greater.

kennykerr:
This is pretty unpleasant.

I also wonder if when the MSRV is finally caught up whether it will be smart enough to tell us that we don't need the `use` imports any longer.

sivadeilra:
Yeah, I agree.  I think the combination of `align_of` and `size_of` being moved into the prelude *and* the new qualified name warning are making things worse, not better.

It might be best to just add `-Aunused-qualifications` to the clippy action, for now.

I really want to get other PRs unblocked, though, so this is kind of urgent.

ChrisDenton:
Looking at the lint's implementation, it seems there's a special carve out for explicitly global use, e.g. `::std::mem::size_of` (that is prefixed by `::`). So that should work.

Or, yes, allowing the lint.

kennykerr:
I recently had to remove all the global `::` due to another lint. #2911

kennykerr:
Let's just allow the lint and keep changes to a minimum please.

ChrisDenton:
It should be as simple as removing it from here: https://github.com/microsoft/windows-rs/blob/0940e1f6352363cf2c3501bfa2cd182414688a47/Cargo.toml#L18

dpaoliello (approved):

kennykerr (approved):
Thanks Arlie!

PR #3079, opened 2024-06-07
Optimize `write_to_file` in `windows-bindgen` to avoid unnecessary file writes

kennykerr:
Alternative to #3072 to avoid unnecessary file writes that trigger side effects in build systems and IDEs. 

I ended up doing the same thing for cppwinrt years ago as msbuild and razzle had a real problem for similar reasons.

### DIFF
--- a/crates/libs/bindgen/src/lib.rs
+++ b/crates/libs/bindgen/src/lib.rs
@@ -257,7 +257,14 @@ fn write_to_file<C: AsRef<[u8]>>(path: &str, contents: C) -> Result<()> {
             .map_err(|_| Error::new("failed to create directory").with_path(path))?;
     }
 
-    std::fs::write(path, contents).map_err(|_| Error::new("failed to write file").with_path(path))
+    let existing = std::fs::read(path).unwrap_or_default();
+
+    if contents.as_ref() != existing {
+        std::fs::write(path, contents)
+            .map_err(|_| Error::new("failed to write file").with_path(path))?;
+    }
+
+    Ok(())
 }
 
 fn canonicalize(value: &str) -> Result<String> {
--- a/crates/tests/standalone/build.rs
+++ b/crates/tests/standalone/build.rs
@@ -202,15 +202,6 @@ fn write_vtbl(output: &str, filter: &[&str]) {
 }
 
 fn riddle(output: &str, filter: &[&str], config: &[&str]) {
-    // Rust-analyzer may re-run build scripts whenever a source file is deleted
-    // which causes an endless loop if the file is deleted from a build script.
-    // To workaround this, we truncate the file instead of deleting it.
-    // See https://github.com/microsoft/windows-rs/issues/2777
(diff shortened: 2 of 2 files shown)


crates/libs/bindgen/src/lib.rs
```diff
@@ -257,7 +257,14 @@ fn write_to_file<C: AsRef<[u8]>>(path: &str, contents: C) -> Result<()> {
             .map_err(|_| Error::new("failed to create directory").with_path(path))?;
     }
 
-    std::fs::write(path, contents).map_err(|_| Error::new("failed to write file").with_path(path))
+    let existing = std::fs::read(path).unwrap_or_default();
+
+    if contents.as_ref() != existing {
```
kennykerr, line 262: This could be further optimized by first comparing file size before reading existing contents, but this seems good enough for now.

sivadeilra:
Unfortunately, I think there's a compelling reason not to do this unconditionally.  Many build systems expect that the timestamps on the output files of a task will be newer than the timestamps of the inputs of a task, after running a task.  If we don't update the output timestamps, then build systems will re-run bindgen every time.  So this breaks incremental builds in environments that use timestamps.

Maybe this could be a configuration option?

kennykerr:
Interesting, the github workflow doesn't seem to like it either.

sivadeilra:
Maybe it could be a boolean parameter to bindgen: Always update output?  If the output reads the same, then maybe we just "touch" the output to update its last-modified time.  I don't see a way to do that directly in `std::fs`, so maybe we just call `std::fs::File::open` with the "open existing for write" parameters, then do nothing, then close the file.  I'm not sure if that's guaranteed to update the mod time though.

ChrisDenton:
> I'm not sure if that's guaranteed to update the mod time though.

[`File::set_times`](https://doc.rust-lang.org/std/fs/struct.File.html#method.set_times) was stabilized a few versions ago.

sivadeilra:
Ah, good.  I wasn't aware of `File::set_times`.  That's perfect.

Still, I would like this behavior to be controllable.  The caller should be able to choose between "do a single unconditional write" vs. this diffing behavior.  I say this because my team manages build systems (the Windows build lab), and our buidl systems do a _lot_ of introspection into the file access patterns of tools that they execute.  This "read and verify" pattern will be flagged as a very problematic behavior by several major build systems because it looks like a false dependency on a previous state, when it's really not.

kennykerr:
For what it's worth, cppwinrt has done this for years without any trouble. 

https://github.com/microsoft/cppwinrt/blob/master/cppwinrt/text_writer.h#L178

Anyway if this is not desirable, I'd rather just leave the current behavior alone. If developers have problems with rust-analyzer they can temporarily disable it while working on the windows-rs repo. I never use it for this reason. 

We can also separately reduce the overhead and optimize things, like using `windows-bindgen` rather than `riddle` in the standalone tests.

PR #3082, opened 2024-06-08
Allow `windows-result` to work on non-Windows platforms

sivadeilra:
This updates the `windows-result` crate so that it can compile and work on non-Windows platforms.  Right now, there are dependencies on things like `GetErrorInfo`, which are obviously not present on other platforms.

This modifies `Error` so that the `info` field is only present on Windows. Everything that reads/writes that field is also made conditional.

I also manually imported definitions for some HRESULT constants, so that we did not need to rely on bindgen for them.  This was necessary because none of the bindings should be used on non-Windows platforms.  When we convert from `std::io::ErrorKind` to `HRESULT`, we do our best to map things to error codes that make sense, although some of them are a bit of a stretch.

This is one part of fixing #3083.

### DIFF
--- a/crates/libs/result/src/error.rs
+++ b/crates/libs/result/src/error.rs
@@ -5,6 +5,7 @@ use core::ffi::c_void;
 #[derive(Clone, PartialEq, Eq)]
 pub struct Error {
     code: HRESULT,
+    #[cfg(windows)]
     info: Option<ComPtr>,
 }
 
@@ -13,6 +14,7 @@ impl Error {
     pub const fn empty() -> Self {
         Self {
             code: HRESULT(0),
+            #[cfg(windows)]
             info: None,
         }
     }
@@ -20,28 +22,46 @@ impl Error {
     /// Creates a new error object, capturing the stack and other information about the
     /// point of failure.
     pub fn new<T: AsRef<str>>(code: HRESULT, message: T) -> Self {
-        let message: Vec<_> = message.as_ref().encode_utf16().collect();
-
-        if message.is_empty() {
-            Self::from_hresult(code)
-        } else {
-            unsafe {
-                RoOriginateErrorW(code.0, message.len() as u32, message.as_ptr());
+        #[cfg(windows)]
+        {
+            let message: Vec<_> = message.as_ref().encode_utf16().collect();
+            if message.is_empty() {
+                Self::from_hresult(code)
+            } else {
+                unsafe {
+                    RoOriginateErrorW(code.0, message.len() as u32, message.as_ptr());
+                }
+                code.into()
             }
-            code.into()
+        }
+        #[cfg(not(windows))]
+        {
+            let _ = message;
+            Self::from_hresult(code)
         }
     }
 
     /// Creates a new error object with an error code, but without additional error information.
     pub fn from_hresult(code: HRESULT) -> Self {
-        Self { code, info: None }
+        Self {
+            code,
+            #[cfg(windows)]
+            info: None,
+        }
     }
 
     /// Creates a new `Error` from the Win32 error code returned by `GetLastError()`.
     pub fn from_win32() -> Self {
-        Self {
-            code: HRESULT::from_win32(unsafe { GetLastError() }),
-            info: None,
+        #[cfg(windows)]
+        {
+            Self {
+                code: HRESULT::from_win32(unsafe { GetLastError() }),
+                info: None,
+            }
+        }
+        #[cfg(not(windows))]
+        {
+            unimplemented!()
         }
     }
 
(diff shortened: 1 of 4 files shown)


riverar:
What's the supporting rationale for this?

sivadeilra:
> What's the supporting rationale for this?

I've opened an issue which describes the problem: #3083.  Briefly, I need cross-platform COM to work.  The `windows-rs` crates are expected to work on other platforms, even if their functionality is greatly reduced.

riverar:
We previously explored similar changes to support DirectX, its pseudo-COM semantics, and non-Windows targeting, but it showed signs of becoming a tangled web of related issues (e.g. [mismatching wide strings](https://github.com/microsoft/windows-rs/issues/1874)).

> I need cross-platform COM to work [...]

There is no cross-platform COM though, right?

> The windows-rs crates are expected to work on other platforms [...]

I'm not certain if this reflects a strong opinion or a shift in Microsoft's direction for this crate. I believe our current stance is to focus squarely on Windows targets only. @kennykerr thoughts?

sivadeilra:
> There is no cross-platform COM though, right?

There is definitely cross-platform COM in the sense that code uses COM interfaces on different platforms.  DWriteCore is a shipping Microsoft product that has a COM-based API, and ships on Android, iOS, Mac OS, and Linux desktop.

DWriteCore is implemented mostly in Rust.  We are currently using `com-rs` for COM, but `com-rs` is the old deprecated crate for COM support.  The `windows-rs` crates already compile for non-Windows platforms and the repo has GitHub actions which test that it compiles for Linux and Mac.

I understand that many parts of the Windows crates will naturally not work on other platforms. But for cross-platform development and for implementing systems that interoperate with Windows, such as systems that use COM or HRESULTs and such, this is important and is easily achievable.

kennykerr:
> I'm not certain if this reflects a strong opinion or a shift in Microsoft's direction for this crate. I believe our current stance is to focus squarely on Windows targets only. @kennykerr thoughts?

There are certainly many aspects of cross-platform support that are very problematic, such as is the case with strings, but I think we can provide some minimal support for things like the essentials of COM - boiling it down to `IUnknown` - in such a way that it does not negatively impact our core support for the Windows platform. 

I do however think that this PR is a little premature and we need to be a bit more deliberate about the scope of such work and that should be discussed in an issue like #3083 before we get too deep into an implementation such as this PR.

crates/libs/result/src/error.rs
```diff
@@ -1,10 +1,58 @@
 use super::*;
+use crate::HRESULT;
 use core::ffi::c_void;
 
+#[allow(dead_code)] // for cases that are not yet stabilized; see below
+mod winerror {
+    use crate::HRESULT;
+
+    // We define these here for non-Windows platforms.
+    // It's just as easy to use them on Windows, too.
```
kennykerr, line 10: This seems to favor non-Windows. Why not just keep the bindings and use them "just as easily" on non-Windows platforms? Much of the value of the generating bindings is not having to maintain these manual definitions.

crates/libs/result/src/error.rs
```diff
@@ -1,10 +1,58 @@
...
+    pub const WSAEWOULDBLOCK: u32 = 10035;
+    pub const WSAEADDRINUSE: u32 = 10048;
+    pub const WSAEADDRNOTAVAIL: u32 = 10049;
+    pub const WSAENETDOWN: u32 = 10050;
+    pub const WSAENETUNREACH: u32 = 10051;
+    pub const WSAECONNABORTED: u32 = 10053;
+    pub const WSAECONNRESET: u32 = 10054;
+    pub const WSAENOTCONN: u32 = 10057;
+    pub const WSAECONNREFUSED: u32 = 10061;
+    pub const WSAEHOSTUNREACH: u32 = 10065;
```
kennykerr, line 45: This interop with `ErrorKind` seems like it could very well be a separate discussion and potential feature. I'm not convinced it's really worth baking in to the `windows-result` crate but I am confident it is not essential to non-Windows support for COM.

crates/libs/result/src/error.rs
sivadeilra: Yeah, that makes sense.  The problem really arises from how easy it is to accidentally use something that is not available on other platforms.  I can reduce this to focus on the DLL imports that are not available, such as GetErrorInfo.

crates/libs/result/src/error.rs
sivadeilra: Agreed.  The existing conversion is broken, though.

crates/libs/result/src/error.rs
kennykerr: Fortunately, it's easy enough to test - just make sure a linux yml workflow is exercising `windows-core` sufficiently. That would be preferable to manually defining bindings.

crates/libs/result/src/error.rs
```diff
@@ -13,35 +14,54 @@ impl Error {
...
-        Self { code, info: None }
+        Self {
+            code,
+            #[cfg(windows)]
+            info: None,
+        }
     }
 
     /// Creates a new `Error` from the Win32 error code returned by `GetLastError()`.
     pub fn from_win32() -> Self {
```
kennykerr, line 54: Shouldn't this whole method be conditional, same as the `as_ptr` method below?

crates/libs/result/src/error.rs
```diff
@@ -13,35 +14,54 @@ impl Error {
...
         Self {
             code: HRESULT(0),
+            #[cfg(windows)]
             info: None,
         }
     }
 
     /// Creates a new error object, capturing the stack and other information about the
     /// point of failure.
     pub fn new<T: AsRef<str>>(code: HRESULT, message: T) -> Self {
```
kennykerr, line 24: What about just having a different `new` method for non-Windows that lacks the `message` parameter?

crates/libs/result/src/error.rs
sivadeilra: Reverted.  I might circle back to the broken conversion later.

crates/libs/result/src/error.rs
sivadeilra: I tried that, but there are a lot of calls to this in other crates.  I think ideally this becomes `#[cfg(windows)]` in the future, and possibly gets renamed to `fn last()` or something.  `from_xxx` sounds like a conversion, but this reads from ambient thread state.

kennykerr: Sounds good, we can tweak this separately as needed.

crates/libs/result/src/error.rs
kennykerr: Probably similar to reasoning for `from_win32`

kennykerr:
Maybe some basic tests for `Error` and `HRESULT` can be added here:

https://github.com/microsoft/windows-rs/tree/master/crates/tests/linux/tests

That way the `linux.yml` workflow will be sure to exercise it as part of PR validation.

sivadeilra:
Added test coverage for functions that are different on non-Windows platforms.

sivadeilra:
The Mac job runners appear to be failing, but without providing any error logs.  :[

kennykerr (approved):

kennykerr:
I'll just override - those runners are not too reliable.

Issue #3083, opened 2024-06-09
COM can't be used on non-Windows platforms


sivadeilra:
### Summary

I'm working on several cross-platform codebases, such as DWriteCore, which are based heavily on COM.  Currently I'm using the old `com-rs` crate, which is no longer supported.  I've converted everything in my codebase to use `windows-rs`, but currently compilation (linking, specifically) fails due to platform dependencies for Windows.

The first blocking issue I've hit is that the code generated for calling COM methods calls `GetErrorInfo`, which is only defined on Windows.

kennykerr:
Note that non-Windows support has never been in scope. While I have no intention to block such support, this would definitely be considered a new feature, not a bug. Existing test coverage exists merely to support cross-compilation and to ease cross-platform development. 

My general philosophy here continues to be that I am ok with enabling cross-platform support so long as it doesn't substantially impact the complexity and primacy of Windows support.

MRahilly:
Now, android-rs would be something.

MarijnS95:
@sivadeilra this is awesome! It may one day help us migrate [`hassle-rs`](https://github.com/Traverse-Research/hassle-rs) off of Microsoft's deprecated and unmaintained `com-rs` and onto `windows-rs`, while getting autogenerated/auto-maintained bindings _for free_!

Note that you'll face a lot of pushback, see for example https://github.com/microsoft/windows-rs/issues/1874 which @riverar already linked to you in #3082. Though a lot of folks will be very thankful if you do manage to sit this through to the end :crossed_fingers: 

---

For the specific linker-error case, I did however manage to get at least https://github.com/microsoft/windows-rs/pull/1863 merged. This allows you to define symbols like `GetErrorInfo` in your own codebase, to provide a fallback when the symbol isn't defined by "the OS". For example: https://github.com/MarijnS95/windows-rs/commit/e595cda41cc8ca8163de6f15890da1f1e7150d2b#diff-1be65c4aa0a07017c9575e418540179c2b2f7638f022c867c0ba42d39e435e9bR63-R67
(But it's hacky and I hope you can come up with something better)

sivadeilra:
To be clear, I do not want this to become something like WINE.  I don't want additional functionality that tries to bridge the gap between Windows and other platforms.  I just want the subset of `windows-core` and `windows` crates that can sensibly used on other platforms.

[...]

PR #3095, opened 2024-06-12
Use malloc on non-Windows platforms

sivadeilra:
This is part of fixing #3083 .  This allows non-Windows code to use more of the COM support from `windows-core`.

This changes the `heap_alloc` function (and related) to use `malloc` on non-Windows platforms, instead of `HeapAlloc`.

With this change, I can now build DWriteCore using `windows-rs` crates.

### DIFF
--- a/crates/libs/core/src/imp/heap.rs
+++ b/crates/libs/core/src/imp/heap.rs
@@ -8,7 +8,17 @@ use core::ffi::c_void;
 /// This function will fail in OOM situations, if the heap is otherwise corrupt,
 /// or if getting a handle to the process heap fails.
 pub fn heap_alloc(bytes: usize) -> crate::Result<*mut c_void> {
-    let ptr = unsafe { HeapAlloc(GetProcessHeap(), 0, bytes) };
+    #[cfg(windows)]
+    let ptr: *mut c_void = unsafe { HeapAlloc(GetProcessHeap(), 0, bytes) };
+
+    #[cfg(not(windows))]
+    let ptr: *mut c_void = unsafe {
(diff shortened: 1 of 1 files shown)


crates/libs/core/src/imp/heap.rs
```diff
@@ -32,5 +42,19 @@ pub fn heap_alloc(bytes: usize) -> crate::Result<*mut c_void> {
...
+
+    #[cfg(not(windows))]
+    {
+        extern "C" {
+            fn free(ptr: *mut c_void);
+        }
+
+        if !ptr.is_null() {
+            free(ptr);
+        }
```
riverar, line 58: Optional: `free` already handles the null case so could skip the extra check here. I could also think up a niche case where `free` is hooked and is expecting the pass through.

sivadeilra: Sure, that's fine.  I'll make the change.

riverar (approved):

riverar:
Cool!

sivadeilra:
Oh wait, I was wrong.  There's still a bunch of other symbols that I'll need to deal with, for DWriteCore.  Like all of the VARIANT conversions that call into external functions.

kennykerr (approved):

riverar:
@sivadeilra I think we could eliminate those VariantToProp/PropToVariant conversion functions altogether and just do the conversions ourselves (being careful to copy strings, convert filetimes, handle clsids, etc.). Or maybe a smaller step would be to just cfg! them away and see if anyone even needs them.

Issue #3096, opened 2024-06-13
Rust nightly warning about dead code


kennykerr:
The latest Rust nightly compiler raises a new warning about dead code. 

```cmd
D:\git\windows-rs> cargo check -p windows --features Foundation      
warning: struct `IReference` is never constructed
    --> crates\libs\windows\src\Windows\Foundation\mod.rs:1024:12
     |
1024 | pub struct IReference<T>(windows_core::IUnknown, core::marker::PhantomData<T>)
     |            ^^^^^^^^^^
     |
     = note: `#[warn(dead_code)]` on by default

warning: struct `IReferenceArray` is never constructed
    --> crates\libs\windows\src\Windows\Foundation\mod.rs:1281:12
     |
1281 | pub struct IReferenceArray<T>(windows_core::IUnknown, core::marker::PhantomData<T>)
     |            ^^^^^^^^^^^^^^^

warning: `windows` (lib) generated 2 warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.14s
```

This seems like a Rust nightly bug. I don't see what's wrong with this code. 

Originally noticed here: https://github.com/microsoft/windows-rs/actions/runs/9501067497/job/26185702630

kennykerr:
You can for example use `IReference` as follows (from another crate):

```rust
use windows::{core::*, Foundation::*};

fn main() -> Result<()> {
    let r: IReference<i32> = PropertyValue::CreateInt32(123)?.cast::<IReference<i32>>()?;
    assert_eq!(r.Value()?, 123);
    Ok(())
}
```

I may just allow `dead_code` to unblock other work until this is resolved in the Rust compiler.

kennykerr:
This seems to be similar to https://github.com/rust-lang/rust/issues/126169

PR #3097, opened 2024-06-13
The `Debug` derive macro does not need to be qualified

kennykerr:
The `windows-bindgen` crate was inconsistently generating this derive macro due to some previous refactoring. This just switches to consistently using the unqualified form in all cases.


kennykerr:
Build validation blocked by #3096

PR #3098, opened 2024-06-13
Workaround for false dead code warning

kennykerr:
This is a workaround for #3096


ChrisDenton (approved):
As a very temporary workaround to unblock CI this seems ok but ooph that's really unfortunate.

kennykerr:
Yep, I'll keep the #3096 issue open as I really don't like this.

PR #3103, opened 2024-06-15
Allow `unused` to deal with new warning about "unused" private fields in structs 

kennykerr:
This is preferable to allowing `dead_code`. 

Fixes: #3096

### DIFF
--- a/crates/libs/windows/src/lib.rs
+++ b/crates/libs/windows/src/lib.rs
@@ -7,10 +7,8 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 */
 
 #![doc(html_no_source)]
-#![allow(non_snake_case, clashing_extern_declarations, non_upper_case_globals, non_camel_case_types, missing_docs, clippy::all)]
+#![allow(non_snake_case, clashing_extern_declarations, non_upper_case_globals, non_camel_case_types, missing_docs, unused, clippy::all)]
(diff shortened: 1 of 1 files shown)


crates/libs/windows/src/lib.rs
```diff
@@ -7,10 +7,8 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 */
 
 #![doc(html_no_source)]
-#![allow(non_snake_case, clashing_extern_declarations, non_upper_case_globals, non_camel_case_types, missing_docs, clippy::all)]
+#![allow(non_snake_case, clashing_extern_declarations, non_upper_case_globals, non_camel_case_types, missing_docs, unused, clippy::all)]
 #![cfg_attr(not(feature = "docs"), doc(hidden))]
-// TODO: https://github.com/microsoft/windows-rs/issues/3096
```
ChrisDenton, line 12: Should we keep a comment for the time being? There may yet be a more specific lint, if not a revert for the new behaviour.

kennykerr: Added comment.

ChrisDenton (approved):
Great! The narrower the allow the better.

riverar (approved):

crates/libs/windows/src/lib.rs
kennykerr: Ideally this just gets reverted.

PR #3104, opened 2024-06-15
Test cross-compilation with stable gnullvm targets

mati865:
Continuation of https://github.com/microsoft/windows-rs/pull/3007 but this time with a stable toolchain.

### DIFF
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -36,6 +36,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Update toolchain
+        run: rustup update --no-self-update stable && rustup default stable
+
       - name: Add toolchain target
         run: rustup target add ${{ matrix.target }}
         if: contains(matrix.target, 'gnullvm') == false
--- a/.github/workflows/cross.yml
+++ b/.github/workflows/cross.yml
@@ -28,7 +28,6 @@ jobs:
         # https://sourceforge.net/p/mingw-w64/wiki2/Exception%20Handling
         #
         image: [macos-latest, ubuntu-latest]
-        version: [stable, nightly]
         target: [x86_64-pc-windows-gnu, aarch64-pc-windows-gnullvm, x86_64-pc-windows-gnullvm, i686-pc-windows-gnullvm]
     runs-on: ${{ matrix.image }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/cross.yml
```diff
@@ -28,7 +28,6 @@ jobs:
         # https://sourceforge.net/p/mingw-w64/wiki2/Exception%20Handling
         #
         image: [macos-latest, ubuntu-latest]
-        version: [stable, nightly]
```
mati865, line 31: It was used only in `Add nightly toolchain with gnullvm target` step so `x86_64-pc-windows-gnu` actually ran twice on stable despite `nightly` word in the name of one of the jobs.

mati865:
It looks like a preinstalled stable toolchain was not yet updated to the latest version.

kennykerr:
When do you expect that to be updated?

mati865:
Judging by the history of https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md, I'd expect it to happen today, but I don't have the details on how GitHub updates their images.

riverar:
Rather than wait for GitHub, which might be busy/frozen due to events this week, can we just update the tools at start of the workflow (like we already do)?

mati865:
If this fails again, I'll look into incorporating changes from other workflows to make sure Rust is up to date.

kennykerr:
I don't know if this is related but GNU builds have started failing: 

https://github.com/microsoft/windows-rs/actions/runs/9567088234/job/26373942886

ChrisDenton:
It looks like a Windows Server 2022 update is currently in progress so it shouldn't be too long now.

mati865:
Even though GHA docs were [updated yesterday](https://github.com/actions/runner-images/pull/10082) so the compiler should be up-to-date now I added the step to update the toolchain like in the other jobs.

> It looks like a Windows Server 2022 update is currently in progress so it shouldn't be too long now.

That's good, I guess?
Don't know how it's related, though.



> I don't know if this is related but GNU builds have started failing:
> 
> [microsoft/windows-rs/actions/runs/9567088234/job/26373942886](https://github.com/microsoft/windows-rs/actions/runs/9567088234/job/26373942886)

It uses GHA image `20240610.1.0` so it's unlikely to be related. Probably something in nightly or rustup. Maybe it's related to https://github.com/rust-lang/rust/issues/125809?

mati865:
Ah, forgot about Mac.

Added `if: startsWith(matrix.image, 'ubuntu-') && contains(matrix.target, 'gnullvm') || endsWith(matrix.target, 'gnu')` to `Test` step so the tests for gnullvm won't run there, but it's still not ideal since there are 3 macos gnullvm jobs that basically do checkout exit with success.
Still, it's an improvement compared to the master branch, so good to go?

riverar (approved):

kennykerr:
Hmm, just noticed those unused mac tests. We can clean those up separately.

PR #3110, opened 2024-06-17
Simplify how extension code for `windows` crate works

sivadeilra:
This reduces the complexity between `windows-bindgen` and `windows`.  This makes it much easier to write "extension" code for the `windows` crate, by reducing the number of places in the code you need to touch.

This also removes the need to run `cargo run -p tool_bindings` whenever you touch the extension code for the `windows` crate. You don't need to update the bindings if you're just working on the extensions. Instead, you just edit extension code, like normal Rust code.

The extension code has moved from `.../bindgen/src/rust/extensions` to `windows/src/extensions`. This puts the code much closer to the code that it extends.  Also, the extension code is now just a normal set of Rust modules.  This makes the language service (Rust Analyzer) work more naturally; refactoring and go-to-definition and such simply work, rather than pointing to source files that were generated by `bindgen`.

Since the extension code is now just normal Rust modules, I had to add a handful of `use crate::Foo::Bar::BAZ;` imports to each extension source file.  This seems like a good thing; more signs that this is just normal code.

I've staged the changes in a set of commits.  You can see how each piece of extension logic was moved, or you can view all the changes as a single set.  The last commit applies `rustfmt` to the extension code.  This wasn't happening before, since those files were not part of any module tree, so `cargo fmt` overlooked them.


dpaoliello (approved):
Love it! Nice to see a ton of duplicated code and hardcoded lists disappear.

crates/libs/bindgen/src/rust/extensions/mod.rs
```diff
@@ -1,55 +1,38 @@
...
-        }
-        _ => "",
-    }
-    .into()
+/// Generates extension code for a specific namespace
+///
+/// This function allows us to inject code into modules generated by bindgen.
+/// This feature is no longer needed but is preserved in case it is needed
+/// again in the future.
+pub fn gen_mod(_writer: &Writer, _namespace: &str) -> TokenStream {
```
kennykerr: Just get rid of them - we can always add them later if that really makes sense.

kennykerr: Looks like the latest commit still uses this.

kennykerr (approved):
Neat!

PR #3112, opened 2024-06-18
Don't compile `windows` and `windows-sys` in unit test mode

sivadeilra:
This PR improves cycle time for running `cargo test --all`.

The `windows` and `windows-sys` crates do not contain unit tests or doc tests.  Cargo doesn't know that, so it compiles these two very large crates in the `#[cfg(test)]` mode and wastes time scanning for doc tests.

This PR simply disables unit tests and doc tests in both of these crates.  This saves a significant amount of time in running tests.

### DIFF
--- a/crates/libs/sys/Cargo.toml
+++ b/crates/libs/sys/Cargo.toml
@@ -15,9 +15,6 @@ exclude = ["features.json"]
 # This crate does not contain tests. All tests are in separate crates.
 test = false
 doctest = false
-# This crate does not generate docs because it is too large (rustdoc has problems with very large
-# crates) and because the Windows API is documented elsewhere.
-doc = false
 
 [lints]
 workspace = true
--- a/crates/libs/sys/Cargo.toml
+++ b/crates/libs/sys/Cargo.toml
@@ -11,6 +11,14 @@ readme = "readme.md"
 categories = ["os::windows-apis"]
 exclude = ["features.json"]
 
+[lib]
+# This crate does not contain tests. All tests are in separate crates.
+test = false
+doctest = false
+# This crate does not generate docs because it is too large (rustdoc has problems with very large
+# crates) and because the Windows API is documented elsewhere.
+doc = false
+
 [lints]
 workspace = true
 
--- a/crates/libs/windows/Cargo.toml
+++ b/crates/libs/windows/Cargo.toml
@@ -13,6 +13,14 @@ readme = "readme.md"
 categories = ["os::windows-apis"]
 exclude = ["features.json"]
 
+[lib]
+# This crate does not contain tests. All tests are in separate crates.
+test = false
+doctest = false
+# This crate does not generate docs because it is too large (rustdoc has problems with very large
+# crates) and because the Windows API is documented elsewhere.
+doc = false
+
 [lints]
 workspace = true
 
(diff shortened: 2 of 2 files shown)


sivadeilra:
> Should we be seeing a reduction here? https://github.com/microsoft/windows-rs/actions/workflows/test.yml

I certainly see a reduction locally.

kennykerr:
Can you illustrate? I just don't see `cargo test` doing much for the `windows` crate unless you explicitly add `--all-features`.

sivadeilra:
> Can you illustrate? I just don't see `cargo test` doing much for the `windows` crate unless you explicitly add `--all-features`.

What I see happening is that we spend a bunch of time compiling the `windows` crate in test mode, even though it doesn't contain any tests.  Then we "run" it and of course that takes no time, but we spent a long time compiling a useless executable.

I see this happening when I run `cargo test` from the repo root.  When you run `cargo test` directly on the `windows` crate you won't see it because no features are activated.

kennykerr:
Ah, strangely this affects 'cargo test' but not 'cargo test --all' which is what I've trained myself to do. 🤷‍♂️

kennykerr (approved):

crates/libs/sys/Cargo.toml
```diff
@@ -11,6 +11,14 @@ readme = "readme.md"
 categories = ["os::windows-apis"]
 exclude = ["features.json"]
 
+[lib]
+# This crate does not contain tests. All tests are in separate crates.
+test = false
+doctest = false
+# This crate does not generate docs because it is too large (rustdoc has problems with very large
+# crates) and because the Windows API is documented elsewhere.
+doc = false
```
tim-weis, line 20: That doesn't sound right. This crate *does* generate documentation, presumably via the standard publishing process through crates.io (see [here]()https://docs.rs/windows-sys/0.52.0/windows_sys/). Will this still be supported after this change?

crates/libs/windows/Cargo.toml
```diff
@@ -13,6 +13,14 @@ readme = "readme.md"
 categories = ["os::windows-apis"]
 exclude = ["features.json"]
 
+[lib]
+# This crate does not contain tests. All tests are in separate crates.
+test = false
+doctest = false
+# This crate does not generate docs because it is too large (rustdoc has problems with very large
+# crates) and because the Windows API is documented elsewhere.
+doc = false
```
tim-weis, line 22: Similar to the above this crate generates documentation. This is a manual process (as I understand). Unlike the `windows-sys` crate, the documentation here is vital given the numerous transformations the `windows` crate applies to functions. Is documentation generation still supported after this change?

Related: Documentation generation is disabled by default via PR #2676.

crates/libs/sys/Cargo.toml
kennykerr: Tim is of course right. Both crates are used to generated Rust docs.

kennykerr (changes requested):
There aren't any tests but there are docs.

crates/libs/sys/Cargo.toml
sivadeilra: Updated.

kennykerr (approved):

PR #3116, opened 2024-06-20
Build `cdylib` dependencies

kennykerr:
These `cdylib` dependencies weren't building too reliably locally at least. This seems to avoid the local failures. 

Note that the warnings about a lack of linkage target seem to have gone away but the issue has no mention of it so it may be accidental... https://github.com/rust-lang/cargo/issues/7825


kennykerr:
Oh this is probably why I had it as a dependency rather than a build dependency:

https://github.com/microsoft/windows-rs/actions/runs/9597443395/job/26466515022?pr=3116

Otherwise, the DLL is placed in the wrong location. 🙄

kennykerr:
This goes back to not having a reliable way to find the target directory from a dependent crate... https://github.com/rust-lang/cargo/issues/9661

Issue #3119, opened 2024-06-21
Should `windows_registry::Key::(get|set)_string` use `OsStr` instead of `str`?


rami3l:
Hi, and thanks for making this project!

## Background

Coming from https://github.com/rust-lang/rustup/pull/3896, we (the Rustup team) are currently evaluating the possibility of replacing `winreg` with `windows-registry` in Rustup. This is mainly because we use the registry APIs to inspect and change the `PATH` variable, which is a necessary step in Rustup's installation process.

However, the following function signatures look a bit concerning:

https://github.com/microsoft/windows-rs/blob/2ec4e06d366c46ba9634316959f680ea3dc42a5a/crates/libs/registry/src/key.rs#L245
https://github.com/microsoft/windows-rs/blob/2ec4e06d366c46ba9634316959f680ea3dc42a5a/crates/libs/registry/src/key.rs#L95

... which is because being valid UTF8 is [a core invariant](https://doc.rust-lang.org/std/primitive.str.html#invariant) of the `str` type. OTOH, it seems to me that there's no guarantee that we should always get/set valid UTF16LE from/to the registry, and we'd like to make sure that Rustup is still working correctly even if there are non-Unicode bytes in the registry value in question (which could be added by a third-party app?).

For example, we now have the following smoke test that inserts into the registry a sequence of `u16` that would be considered malformed UTF16LE, and it's expected to work:

```rs
                // Set up a non unicode PATH
                let reg_value = RegValue {
                    bytes: vec![
                        0x00, 0xD8, // leading surrogate
                        0x01, 0x01, // bogus trailing surrogate
                        0x00, 0x00, // null
                    ],
                    vtype: RegType::REG_EXPAND_SZ,
                };
                RegKey::predef(HKEY_CURRENT_USER)
                    .open_subkey_with_flags("Environment", KEY_READ | KEY_WRITE)
                    .unwrap()
                    .set_raw_value("PATH", &reg_value)
                    .unwrap();
```
_https://github.com/rust-lang/rustup/blob/4c3ff9ce638e8402fbc2e8755d9e8a2cd148f6be/tests/suite/cli_paths.rs#L427-L440_

## Suggestion

For the exact reason explained above, I believe [`OsStr`](https://doc.rust-lang.org/std/ffi/struct.OsStr.html)/[`OsString`](https://doc.rust-lang.org/std/ffi/struct.OsString.html) is a better alternative to `str`/`String` in APIs like this.

[...]

PR #3120, opened 2024-06-21
Add flexible registry type and byte query support

kennykerr:
The `windows-registry` crate is currently pretty strict when it comes to mapping registry values and types. You cannot for example read the bytes from a registry value unless the type of that value is `REG_BINARY`. This makes it a little difficult to gracefully deal with invalid registry values.

This update adds the `get_type` methods and the resulting `Type` enum so that the actual type of a value can be queried directly. It also updates the `get_bytes` method to support returning the raw byte value even if the type isn't known to be bytes.

Related: #3119

### DIFF
--- a/crates/libs/registry/src/key.rs
+++ b/crates/libs/registry/src/key.rs
@@ -115,6 +115,34 @@ impl Key {
         unsafe { self.set_value(name, REG_BINARY, value.as_ptr() as _, value.len()) }
     }
 
+    /// Gets the type for the name in the registry key.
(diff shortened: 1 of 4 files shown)


ChrisDenton (approved):
Seems good to me. Arguably we shouldn't allow setting non-unicode registry values, even if we can read them, but it is convenient to be able to use `OsStr` and I don't feel there's a strong reason to implement checks.

kennykerr:
Thanks, I'll work on `OsStr` support separately.

PR #3121, opened 2024-06-24
Add `OsString` support to `windows-registry` crate

kennykerr:
Fixes: #3119

### DIFF
--- a/crates/libs/core/src/lib.rs
+++ b/crates/libs/core/src/lib.rs
@@ -8,7 +8,7 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
     windows_debugger_visualizer,
     debugger_visualizer(natvis_file = "../.natvis")
 )]
-#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
+#![cfg_attr(all(not(feature = "std")), no_std)]
 
 extern crate self as windows_core;
 
--- a/crates/libs/registry/Cargo.toml
+++ b/crates/libs/registry/Cargo.toml
@@ -9,7 +9,10 @@ description = "Windows registry"
 repository = "https://github.com/microsoft/windows-rs"
 readme = "readme.md"
 categories = ["os::windows-apis"]
-exclude = ["tests"]
+
+[features]
+default = ["std"]
+std = []
 
 [lints]
 workspace = true
--- a/crates/libs/registry/src/key.rs
+++ b/crates/libs/registry/src/key.rs
@@ -20,7 +20,7 @@ impl Key {
         let result = unsafe {
             RegCreateKeyExW(
                 self.0,
-                pcwstr(path).as_ptr(),
+                encode_utf16(path).as_ptr(),
                 0,
                 null(),
                 REG_OPTION_NON_VOLATILE,
@@ -38,8 +38,15 @@ impl Key {
     pub fn open<T: AsRef<str>>(&self, path: T) -> Result<Self> {
         let mut handle = core::ptr::null_mut();
 
-        let result =
-            unsafe { RegOpenKeyExW(self.0, pcwstr(path).as_ptr(), 0, KEY_READ, &mut handle) };
+        let result = unsafe {
+            RegOpenKeyExW(
+                self.0,
+                encode_utf16(path).as_ptr(),
+                0,
+                KEY_READ,
+                &mut handle,
+            )
+        };
 
         win32_error(result).map(|_| Self(handle))
     }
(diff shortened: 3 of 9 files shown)


crates/libs/core/src/lib.rs
```diff
@@ -8,7 +8,7 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
     windows_debugger_visualizer,
     debugger_visualizer(natvis_file = "../.natvis")
 )]
-#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
+#![cfg_attr(all(not(feature = "std")), no_std)]
```
kennykerr, line 11: #3102 removed the embedded tests used to generate bindings.

crates/libs/registry/Cargo.toml
```diff
@@ -9,7 +9,10 @@ description = "Windows registry"
 repository = "https://github.com/microsoft/windows-rs"
 readme = "readme.md"
 categories = ["os::windows-apis"]
-exclude = ["tests"]
+
+[features]
+default = ["std"]
+std = []
```
kennykerr, line 15: Provides a way to compile with `no_std`.

crates/libs/registry/src/key.rs
```diff
@@ -20,7 +20,7 @@ impl Key {
         let result = unsafe {
             RegCreateKeyExW(
                 self.0,
-                pcwstr(path).as_ptr(),
+                encode_utf16(path).as_ptr(),
```
kennykerr, line 23: I just renamed `pcwstr` to `encode_utf16` to distinguish between UTF-16 encoding and the "wide" encoding provided by `OsString`.

crates/libs/registry/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs>
 */
 
-#![cfg_attr(not(test), no_std)]
+#![cfg_attr(all(not(feature = "std")), no_std)]
```
kennykerr, line 5: #3102 removed the embedded tests used to generate bindings, but the new `std` feature still necessitates using the Rust Standard Library for `OsString` support.

crates/libs/result/src/lib.rs
```diff
@@ -6,7 +6,7 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
     windows_debugger_visualizer,
     debugger_visualizer(natvis_file = "../.natvis")
 )]
-#![cfg_attr(all(not(test), not(feature = "std")), no_std)]
+#![cfg_attr(all(not(feature = "std")), no_std)]
```
kennykerr, line 9: #3102 removed the embedded tests used to generate bindings.

kennykerr:
Wow, that's quite the bizarre nightly behavior. 🤪

image

kennykerr:
Looks like that's a rustfmt bug:

https://github.com/rust-lang/rustfmt/issues/6210
https://github.com/rust-lang/rustfmt/issues/6213

crates/libs/registry/src/key.rs
```diff
@@ -320,14 +327,63 @@ impl Key {
...
+        value: V,
+    ) -> Result<()> {
+        let value = encode_wide(value);
+
+        unsafe { self.set_value(name, REG_SZ, value.as_ptr() as _, value.len() * 2) }
+    }
+
+    /// Gets the value for the name in the registry key.
+    #[cfg(feature = "std")]
+    pub fn get_os_string<T: AsRef<str>>(&self, name: T) -> Result<std::ffi::OsString> {
```
kennykerr, line 344: This would be a lot more efficient if `OsString` were an `HSTRING` but I'm guessing its too late for that now. 🥲

ChrisDenton: Sadly, that ship set sail in 2015...

I do have a PR I need to get back to that will (hopefully) allow passing wide strings to filesystem APIs. Return types are stuck with `OsString` though.

ChrisDenton (approved):
Looks good to me. Seems like the rustfmt bug will fix itself in the next nightly.

ChrisDenton:
Oh it just occurred to me that maybe we could just return a `HSTRING` or a `Vec<u16>` and let users turn that into a `OsString` or whatever they want. It's less ergonomic for the `OsString` case but more flexible in general.

kennykerr:
Yes, I thought about using `HSTRING` instead. It's way better and provides convertibility with `OsString`. I just need to pull `HSTRING` and friends into a new `windows-strings` crate before I can add a dependency from `windows-registry`. I need to do that anyway, so perhaps I'll wait on this PR and we can just use `HSTRING` here.

ChrisDenton:
To be honest, that does sound a lot more appealing to me now that I've considered it some more.

PR #3125, opened 2024-06-25
Add `windows-strings` crate

kennykerr:
This update pulls out the Windows string types from the `windows-core` crate and makes them available directly without a dependency on the `windows` or `windows-core` crates. This will help to make these string types more accessible to library crates such as `windows-registry` (#3121). 

These types are still included in the `windows-core` and `windows` crates as a dependency so it should not affect existing code too much.

### DIFF
--- a/crates/libs/strings/src/heap.rs
+++ b/crates/libs/strings/src/heap.rs
@@ -1,0 +1,40 @@
+use super::*;
+use core::ffi::c_void;
+
+/// Allocate memory of size `bytes` using `HeapAlloc`.
+pub fn heap_alloc(bytes: usize) -> crate::Result<*mut c_void> {
+    #[cfg(windows)]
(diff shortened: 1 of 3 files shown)


kennykerr:
Looks like the readme tests were accidentally disabled - https://github.com/microsoft/windows-rs/pull/3125/commits/cf9b35bb35566244fbef6bd786fa7e4f3a0df75f turns them back on.

kennykerr:
This whole non-Windows appeasement thing is getting old. It's making it too difficult to support Windows, which is the whole point of this project. 🤪 Anyway, the latest version retains basic Linux compatibility for `windows-core` only to support COM. For now, I needed to keep `windows-strings` at least building on non-Windows but I plan to limit non-Windows support to COM and nothing more.

ChrisDenton (approved):
I think this is a good approach, not just for `windows-` crates but also for the windows ecosystem in general.

PR #3126, opened 2024-06-25
Allow `Error` and `Result<()>` to be the same size as `HRESULT`

sivadeilra:
Many codebases use `HRESULT` (whether they use COM or not) and yet never use `IErrorInfo`. For those codebases, the size of the `Result<()>` object can be a substantial cost.  Currently on 64-bit platforms, `size_of<Result<()>>` is 16 bytes, when ideally it would be basically a smartly-typed version of `HRESULT`.  For example, Direct3D, DirectWrite, etc. never use `IErrorInfo` (they don't set any error info objects on failures), so all of the code that uses these codebases will suffer in performance.

This PR allows you to statically disable the error info within `Error`, by enabling the `slim-errors` feature in the `windows-result` crate.  (The `windows` and `windows-core` crates will also propagate the feature to `windows-result`, so that you don't need to add an extra crate dependency.)

The performance costs come from the larger code and larger stack area needed.  There is also a significant penalty because `IErrorInfo` requires that the `Error` object have a `Drop` implementation. This requires the compiler to precisely track the lifetime of `Error` values and even of the `Result<()>` value.  This adds a lot of extra code that almost never runs, to many function bodies.

This PR also adds a new `NonZeroHRESULT`, which is equivalent to `HRESULT` except that it contains a `NonZeroI32`, not an `i32`. This means that it cannot represent `S_OK`, which is fine.  This gives the Rust compiler a "niche", which is a byte pattern that cannot be represented by one variant of an enum, which frees up that byte pattern for a different variant or for enum's "tag".  In our case, the compiler will generate type layouts for `Option<NonZeroHRESULT>` and `Result<(), NonZeroHRESULT>` that are the same size and alignment as the underlying `HRESULT`.  This is great because it allows them to be stored in a single register, and in many situations the compiler can totally avoid type layout conversions and can produce much better code.

The `Error` type now uses `NonZeroHRESULT`.  When the `slim-errors` feature is enabled, `size_of::<Error>() == 4` and `size_of::<Result<()>>() == 4`.  Before this change, `size_of::<Error>() == 16` and `size_of::<Result<()>>() == 24`.

This PR improves these types even when not using `slim-errors`.  After this PR, with `slim-errors` disabled, `size_of::<Result<()>>() == 16`, 8 bytes less, because of the niche optimization relating to `NonZeroHRESULT`.

### DIFF
--- a/crates/libs/result/Cargo.toml
+++ b/crates/libs/result/Cargo.toml
@@ -11,9 +11,10 @@ readme = "readme.md"
 categories = ["os::windows-apis"]
 
 [features]
-default = ["std", "error-info"]
+default = ["std"]
 std = []
-error-info = []
+# slim-errors reduces the size of Error (and Result<()>) to that of a single HRESULT
+slim-errors = []
 
 [lints]
 workspace = true
--- a/crates/libs/result/Cargo.toml
+++ b/crates/libs/result/Cargo.toml
@@ -11,8 +11,9 @@ readme = "readme.md"
 categories = ["os::windows-apis"]
 
 [features]
-default = ["std"]
+default = ["std", "error-info"]
 std = []
+error-info = []
 
 [lints]
 workspace = true
@@ -20,6 +21,9 @@ workspace = true
 [package.metadata.docs.rs]
 default-target = "x86_64-pc-windows-msvc"
 targets = []
+
+[dependencies]
+static_assertions = "1.0"
 
 [dependencies.windows-targets]
 version = "0.52.5"
--- a/crates/libs/result/src/error.rs
+++ b/crates/libs/result/src/error.rs
@@ -2,9 +2,17 @@ use super::*;
 #[allow(unused_imports)]
 use core::mem::size_of;
 
-/// An error object consists of both an error code as well as detailed error information for debugging.
+// We define ErrorT and Error separately because it allows type inference to work correctly
+// in some common situations.  If there were no separate type alias, then expressions like
+// `let e = Error::from(...)` and `let e = Error::new(...)` would have a free (unspecified) type
+// parameter, and so are ambiguous and fail to compile.
+//
+// Most code wants to use the default type parameter, so we give the type alias the `Error` name.
+// For the code that wants to deal directly non-default error details, it can use `ErrorT<...>`.
+
+/// An error object consists of both an error code and optional detailed error information for debugging.
 ///
-/// # Extended error info and the `error-info` feature
+/// # Extended error info and the `slim-errors` feature
 ///
 /// The `Error` contains an `HRESULT` value that describes the error, as well as an optional
 /// `IErrorInfo` object. The `IErrorInfo` object is a COM object that can provide detailed information
@@ -16,13 +24,15 @@ use core::mem::size_of;
 /// info within `Error` has no benefits, but has substantial costs because it increases the size of
 /// the `Error` object, which also increases the size of `Result<T>`.
 ///
-/// The `windows-result` crate provides the `error-info` feature, which gives developers control over
-/// the costs of using `Error`.
+/// The `slim-errors` Cargo feature allows you to _statically_ disable error detail in `Error`.
+/// This allows the `Error` and even the `Result<(), Error>` types to have a very small size; they
+/// have the same size and alignment as `HRESULT` itself. Because of this, they can be returned
+/// directly in a machine register, which gives better performance than writing the (relatively
+/// large) `Error` object (with full error detail) to the stack on every function return.
 #[derive(Clone)]
-pub struct ErrorT<Detail = DefaultDetail>
-where
-    Detail: ErrorDetail,
-{
+pub struct ErrorT<Detail = DefaultDetail> {
+    /// The error code. Because this uses `NonZeroHRESULT`, this provides a "niche" to the Rust
+    /// compiler. This allows the compiler to use more compact representations in some sitatutions.
     code: NonZeroHRESULT,
 
     /// Contains details about the error, such as error text.
@@ -35,29 +45,21 @@ where
 
 /// An error object consists of both an error code as well as detailed error information for debugging.
 ///
-/// # Extended error info and the `error-info` feature
-///
-/// The `Error` contains an `HRESULT` value that describes the error, as well as an optional
-/// `IErrorInfo` object. The `IErrorInfo` object is a COM object that can provide detailed information
-/// about an error, such as a text string, a `ProgID` of the originator, etc. If the error object
-/// was originated in an WinRT component, then additional information such as a stack track may be
-/// captured.
-///
-/// However, many systems based on COM do not use `IErrorInfo`. For these systems, the optional error
-/// info within `Error` has no benefits, but has substantial costs because it increases the size of
-/// the `Error` object, which also increases the size of `Result<T>`.
-///
-/// The `windows-result` crate provides the `error-info` feature, which gives developers control over
-/// the costs of using `Error`.
+/// This is a type alias for the [`ErrorT`] type.
 pub type Error = ErrorT<DefaultDetail>;
 
 /// Specifies the default error detail for the `Error` type.
-#[cfg(all(windows, feature = "error-info"))]
+#[cfg(not(feature = "slim-errors"))]
 pub type DefaultDetail = ErrorInfo;
 
 /// Specifies the default error detail for the `Error` type.
-#[cfg(not(all(windows, feature = "error-info")))]
+#[cfg(feature = "slim-errors")]
 pub type DefaultDetail = NoDetail;
+
+/// On non-Windows platforms, `ErrorInfo` is mapped to `NoDetail` because non-Windows platforms
+/// do not support `IErrorInfo`.
+#[cfg(not(windows))]
+pub type ErrorInfo = NoDetail;
 
 /// This type implements `ErrorDetail`, by containing no error information at all. This type allows
 /// for "slim" `Error` objects, which have the same size as `HRESULT`.
@@ -275,17 +277,6 @@ impl<Detail: ErrorDetail> ErrorT<Detail> {
     pub fn detail(&self) -> &Detail {
         &self.detail
     }
-
-    /*
-    /// The error object describing the error.
-    #[cfg(all(windows, feature = "error-info"))]
-    pub fn as_ptr(&self) -> *mut core::ffi::c_void {
-        self.info
-            .ptr
-            .as_ref()
-            .map_or(core::ptr::null_mut(), |info| info.as_raw())
-    }
-    */
 }
 
 #[cfg(feature = "std")]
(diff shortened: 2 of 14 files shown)


riverar:
I'm not very familiar in this area. Is that a guaranteed and/or documented compiler niche optimization? Would be awkward if this changed and resulted in net zero or negative savings.

sivadeilra:
> I'm not very familiar in this area. Is that a guaranteed and/or documented compiler niche optimization? Would be awkward if this changed and resulted in net zero or negative savings.

It's a guaranteed optimization.  It's the reason that the `NonZeroI32` type (and similar) were added to the language.

There are static assertions in the PR that verify this.  If anything changed, we would see a build break.  But these optimizations are guaranteed, so these assertions are more about documentation than anything else.

crates/libs/result/Cargo.toml
```diff
@@ -13,6 +13,8 @@ categories = ["os::windows-apis"]
 [features]
 default = ["std"]
 std = []
+# slim-errors reduces the size of Error (and Result<()>) to that of a single HRESULT
+slim-errors = []
```
kennykerr, line 17: Cargo features are meant to be additive, so I think this should be in the affirmative - e.g. `rich-errors` - and enabled by default. You can then use `no-default-features` to remove it just like you would for the `std` feature.

crates/tests/result/tests/error.rs
```diff
@@ -10,29 +10,30 @@ windows_targets::link!("kernel32.dll" "system" fn SetLastError(code: u32));
 
 #[test]
 fn empty() {
-    const EMPTY: Error = Error::empty();
-    assert_eq!(EMPTY.code(), S_OK);
-    assert!(EMPTY.as_ptr().is_null());
-    assert_eq!(EMPTY.message(), "The operation completed successfully.");
+    let e: Error = Error::empty();
+    assert_eq!(e.code(), E_FAIL);
```
dpaoliello, line 14: This feel like a pretty non-obvious breaking change - I know that we'll be bumping the version number to indicate a breaking change, but still this should be noted somewhere...

dpaoliello (approved):

crates/tests/result/tests/error.rs
riverar: Not sure I agree with this change; `Error::default() / empty()` defaulting to non-zero is an unexpected behavior. If we can't preserve this, then I'd be in favor of abandoning the entire PR.

crates/libs/result/Cargo.toml
sivadeilra: I took that approach to begin with, but it's quite error prone.  You have to make sure that every dependency on `windows-result` uses `no-default-features`, including the dependency from `windows-core`.

If change the `windows-core` --> `windows-result` dependency to be `no-default-features`, then everyone will get slim errors by default, unless they specifically opt-in to `rich-errors`.  I went pretty far down that path, and I think it makes things worse than treating "slim errors" as a feature.

I'm open to changing this, if the result looks good.

crates/tests/result/tests/error.rs
sivadeilra: It's documented on `Error::new` and the other paths that are affected.

I agree it's a breaking change, but a well-deserved ones.  Errors should be...  errors.

crates/libs/result/src/error.rs
```diff
@@ -1,63 +1,265 @@
...
+pub type DefaultDetail = NoDetail;
+
+/// On non-Windows platforms, `ErrorInfo` is mapped to `NoDetail` because non-Windows platforms
+/// do not support `IErrorInfo`.
+#[cfg(not(windows))]
+pub type ErrorInfo = NoDetail;
+
+/// This type implements `ErrorDetail`, by containing no error information at all. This type allows
+/// for "slim" `Error` objects, which have the same size as `HRESULT`.
+#[derive(Default, Clone)]
```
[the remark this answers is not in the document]
sivadeilra: I'm not sure that's a good idea, because then some of the error detail types are `Copy` while others aren't.

crates/libs/result/src/error.rs
```diff
@@ -1,63 +1,265 @@
...
+pub struct NoDetail;
+
+/// Defines the behavior of error detail objects, which are stored within `Error`.
+pub trait ErrorDetail: Sized {
+    /// Returns an empty error detail.
+    fn empty() -> Self;
```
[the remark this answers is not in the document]
sivadeilra: Not much different, but this way everything is defined on a single trait.  Otherwise I'll need to add `+ Default` to lots of places in this file, for not much real benefit.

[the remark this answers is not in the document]
dpaoliello: Could you not use `pub trait ErrorDetail: Sized + Default {` ?

crates/tests/result/tests/error.rs
sivadeilra: Why is there even an `Error::empty()` at all?  What does an "empty error" even mean?  I think `Error::empty()` should be removed entirely.

dpaoliello: > Why is there even an `Error::empty()` at all? What does an "empty error" even mean? I think `Error::empty()` should be removed entirely.

+1 I'd prefer to remove it than to change its meaning - at least that'd cause build breaks.

crates/libs/result/src/error.rs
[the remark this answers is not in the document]
sivadeilra: Yeah, that would work.  I'll switch to that.

crates/tests/result/tests/error.rs
sivadeilra: Switched to `Default`.  Renamed `Error::empty()` to `Error::fail()`.  Changed several usages of `Error::empty()` to `Error::from_hresult(E_POINTER)`, which is much more accurate.

crates/libs/result/src/error.rs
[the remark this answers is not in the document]
sivadeilra: Done

riverar:
Is there any benchmark data that demonstrates this is more than just a theoretical optimization? Not trying to be a blocker, just trying to prevent unnecessary downstream churn.

sivadeilra:
> Is there any benchmark data that demonstrates this is more than just a theoretical optimization? Not trying to be a blocker, just trying to prevent unnecessary downstream churn.

Yes.  This change alone removes 1.2% of the entire binary size for DWriteCore.  The code is smaller, the exception handling tables are smaller.  Text layout benchmarks show meaningful increases in speed as well, which makes sense because these components make extensive use of COM, including purely inside DWriteCore, not just between DWriteCore and external apps.

Keep in mind that DWriteCore is not yet pure Rust; it's about 65% Rust.  So that number will go even higher as I convert more of DWriteCore to Rust.

kennykerr:
The size optimization looks neat. As someone who spent a great deal of effort optimizing code gen for call sites in C++, I can see how this would make a meaningful improvement. Mainly I'm concerned about the following:

* The use of features here just doesn't seem right. Features are meant to [be additive](https://doc.rust-lang.org/cargo/reference/features.html), as I mentioned above, and not fundamentally change behavior. I think this would be better served by a cfg flag like `windows_raw_dylib`. 

* This PR seemingly rewrites all of the `windows-result` crate, greatly increasing complexity and approachability for what boils down to an `HRESULT`. I think we can have the cfg option to trim down the overhead by removing the `IErrorInfo` field and possibly the `HRESULT` optimization without actually changing the public interface to the crate at all. Previously, this crate provided precisely two types; an `HRESULT` and an `Error` struct. Now there's `ErrorInfo`, `ErrorT`, `NoDetail`, `NonZeroHRESULT`, `ErrorDetail`, and `DefaultDetail` added to the mix. Ideally this stuff can be simplified, less generic, and there should be no new public types. In other words, we should be able to achieve the main goal here essentially by simply removing the original `info` field with a cfg attribute.

* The use of `S_OK` to report a failure without error information is critical to the way null interfaces are modeled in `windows-rs`. This change essentially turns them into `E_POINTER` errors. This means that you cannot distinguish between a COM call that returns `E_POINTER` from a call that returns `S_OK` while returning a null interface pointer.

crates/libs/result/src/error.rs
```diff
@@ -1,63 +1,265 @@
...
+// parameter, and so are ambiguous and fail to compile.
+//
+// Most code wants to use the default type parameter, so we give the type alias the `Error` name.
+// For the code that wants to deal directly non-default error details, it can use `ErrorT<...>`.
+
+/// An error object consists of both an error code and optional detailed error information for debugging.
+///
+/// # Extended error info and the `slim-errors` feature
+///
+/// The `Error` contains an `HRESULT` value that describes the error, as well as an optional
```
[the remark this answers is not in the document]
sivadeilra: Updated most of these.  There's a pointer to the MSDN docs for `IErrorInfo`, but I don't want to dup that every time I mention `IErrorInfo`.

crates/libs/result/Cargo.toml
sivadeilra: After a bunch of iterations, we converged on using a `--cfg` option to control this.  It's not perfect, but it works.

crates/libs/windows/Cargo.toml
```diff
@@ -27,7 +27,7 @@ targets = []
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-windows-core = { path = "../core", version = "0.57.0" }
+windows-core = { path = "../core", version = "0.57.0", default-features = false }
```
dpaoliello, line 30: Should this be removed now?

crates/libs/result/src/error.rs
```diff
@@ -226,3 +228,179 @@ impl core::fmt::Display for Error {
...
+
+        pub(crate) fn into_thread(self) {
+            if let Some(ptr) = self.ptr {
+                unsafe {
+                    crate::bindings::SetErrorInfo(0, ptr.as_raw());
+                }
+            }
+        }
+
+        #[cfg(windows)]
```
dpaoliello, line 302: Not needed since the `mod` is already restricted to Windows

dpaoliello (changes requested):
Is the change to `crates/libs/windows/Cargo.toml` intentional?

crates/libs/windows/Cargo.toml
sivadeilra: agreed; will do

riverar:
Something seems wrong with the `Error::code`, is this expected?

```rust
let error_from_win32 = windows_result::Error::from_win32();
```

```
0:000> dx -r1 error_from_win32
error_from_win32                 [Type: windows_result::error::Error]
    [+0x008] code             : 1398755147 [Type: core::num::nonzero::NonZero<i32>]
    [+0x000] info             [Type: windows_result::error::error_info::ErrorInfo]
```

sivadeilra:
> Something seems wrong with the `Error::code`, is this expected?

Hmmm.  I may have to re-introduce the `NonZeroHRESULT` type, so that NatVis has something to activate it...

riverar:
I wouldn't spend too much time on it. I've spent a heavy amount of time with natvis and due to the split in expression engines (Visual Studio / windbg), it's very difficult to get right. Adding VS expression engine tests to start chipping away at this problem is on my TODO.

dpaoliello (approved):

kennykerr (approved):
Sweet - thanks!

PR #3129, opened 2024-06-26
Lock down `windows-core` internals

kennykerr:
With recent refactoring and simplification of the `windows-core` crate, some parts of the internal implementation can be removed or simplified to avoid abuse or inadvertent dependencies.


crates/libs/core/src/com_object.rs
```diff
@@ -165,7 +165,7 @@ impl<T: ComObjectInner> ComObject<T> {
     /// Gets a borrowed reference to an interface that is implemented by `T`.
     ///
     /// The returned reference does not have an additional reference count.
-    /// You can AddRef it by calling [`Self::to_owned`].
+    /// You can AddRef it by calling [`InterfaceRef::to_owned`].
```
kennykerr, line 168: Just a broken doc link.

PR #3132, opened 2024-06-26
Use conventional testing for `windows_slim_errors`

kennykerr:
Jus a minor tidy to the way #3126 is testing the new `windows_slim_errors` config option to bring it in line with how such things are tested in windows-rs. This also avoids introducing an unnecessary new dependency.


.github/workflows/msrv-windows-result.yml
```diff
@@ -29,13 +29,3 @@ jobs:
         run: cargo check -p windows-result --all-features
       - name: Check Default Features
         run: cargo check -p windows-result
-      - name: Check Slim Errors
```
kennykerr, line 32: This yml is for testing MSRV and nothing more. The `windows_slim_errors` testing is provided by a dedicated slim_errors.yml just like how `windows_raw_dylib` is tested.

.github/workflows/slim_errors.yml
```diff
@@ -0,0 +1,41 @@
...
+on:
+  pull_request:
+  push:
+    paths-ignore:
+      - '.github/ISSUE_TEMPLATE/**'
+    branches:
+      - master
+
+env:
+  RUSTFLAGS: -Dwarnings --cfg windows_slim_errors
```
kennykerr, line 12: This sets `windows_slim_errors` for this test run and then tests the `test_result` crate. The generated test.yml will also test `test_result` but without `windows_slim_errors`.

crates/libs/result/Cargo.toml
```diff
@@ -21,9 +21,6 @@ workspace = true
 default-target = "x86_64-pc-windows-msvc"
 targets = []
 
-[dependencies]
-static_assertions = "1.0"
```
kennykerr, line 25: This has just been moved to the `test_result` crate to avoid a production dependency.

crates/tests/result/tests/hresult.rs
```diff
@@ -80,7 +80,12 @@ fn from_result() {
     let result: Result<()> = Err(Error::new(E_INVALIDARG, "test message"));
     let err = HRESULT::from(result).ok().unwrap_err();
     assert_eq!(err.code(), E_INVALIDARG);
-    assert_eq!(err.message(), "test message");
+
+    if cfg!(windows_slim_errors) {
+        assert_eq!(err.message(), "The parameter is incorrect.");
+    } else {
+        assert_eq!(err.message(), "test message");
+    }
```
kennykerr, line 88: We should probably add a bit of test coverage for how this behaves on non-Windows builds.

kennykerr: Oh, `test_linux` does provide basic coverage for this.

sivadeilra (approved):

PR #3133, opened 2024-06-27
Add `HSTRING` builder and registry support 

kennykerr:
This update adds two features to improve string support and efficiency. 

The `HSTRING` type now includes an `HStringBuilder` type for pre-allocating and constructing an `HSTRING` without requiring an additional copy. This is similar to the `WindowsPreallocateStringBuffer` function but implemented directly in Rust, as I did for C++/WinRT back in the day. 

```rust
const HELLO: [u16; 5] = [0x48, 0x65, 0x6C, 0x6C, 0x6F];
let mut b = HStringBuilder::new(5)?;
b.copy_from_slice(&HELLO);
let h: HSTRING = b.into();
assert_eq!(&h, "Hello");
```

The `windows-registry` crate now also provide `get_hstring` and `set_hstring` for reading and writing registry values using `HSTRING`. This is far more efficient than either `String` or `OsString` and is lossless, while still providing convertibility with `String` and `OsString`.

```rust
key.set_hstring("key", h!("value"))?;
let value: HSTRING = key.get_hstring("key")?;
key.set_hstring("key", &value)?;
```

Fixes: #3119


kennykerr:
Thanks to @riverar for helping me fix that natvis... again. 😂

crates/libs/registry/src/key.rs
```diff
@@ -278,6 +287,40 @@ impl Key {
...
+        let mut ty = 0;
+        let mut len = 0;
+
+        let result = unsafe {
+            RegQueryValueExW(self.0, name.as_ptr(), null(), &mut ty, null_mut(), &mut len)
+        };
+
+        win32_error(result)?;
+
+        if !matches!(ty, REG_SZ | REG_EXPAND_SZ) {
```
riverar, line 302: No change needed now but just wanted to be a nerd and point out that the registry has a much larger set of supported values than this crate currently supports. Would be great to relax this check one day and let developers, for example, set a value of type 0x1234 with hstring data.

crates/libs/strings/src/hstring.rs
```diff
@@ -1,27 +1,27 @@
 use super::*;
 
-/// A WinRT string ([HSTRING](https://docs.microsoft.com/en-us/windows/win32/winrt/hstring))
-/// is reference-counted and immutable.
+/// An ([HSTRING](https://docs.microsoft.com/en-us/windows/win32/winrt/hstring))
+/// is a reference-counted and immutable UTF-16 string type.
 #[repr(transparent)]
-pub struct HSTRING(Option<core::ptr::NonNull<Header>>);
+pub struct HSTRING(pub(crate) *mut HStringHeader);
```
kennykerr, line 6: Using `*mut T` is so much simpler.

crates/libs/strings/src/hstring.rs
```diff
@@ -407,54 +402,3 @@ impl From<HSTRING> for std::ffi::OsString {
         Self::from(&hstring)
     }
 }
-
-const REFERENCE_FLAG: u32 = 1;
-
-#[repr(C)]
-struct Header {
```
kennykerr, line 414: This was just moved to `hstring_header.rs` to share it with the builder.

riverar (approved):

crates/libs/registry/src/key.rs
kennykerr: Why not call `get_bytes` instead of `get_hstring` since it doesn't have this check?

kennykerr: Unless you're saying this check adds no value?

riverar: The non standard registry typed value could still hold a string. So this check would prevent me from retrieving that string if the type is 0x1234. (Not a big deal now, we can add support for alien types in a new PR.)

crates/libs/registry/src/key.rs
```diff
@@ -98,6 +98,15 @@ impl Key {
...
     }
 
+    /// Sets the name and value in the registry key.
+    pub fn set_hstring<T: AsRef<str>>(
+        &self,
+        name: T,
+        value: &windows_strings::HSTRING,
+    ) -> Result<()> {
+        unsafe { self.set_value(name, REG_SZ, value.as_ptr() as _, value.len() * 2) }
+    }
```
ChrisDenton, line 108: I am fine with just providing this method but it does feel like there's a version of this that could be more generic, in a similar way to `AsRef<str>`. We don't need an `&HSTRING` per-se just something that provides an `&[u16]` that's guaranteed to be null-terminated.

My only hesitation with that approach is it does increase complexity. It'd probably need a new type or `unsafe` trait to express the validity requirement.

kennykerr: I was thinking the same thing. In this case it doesn't have to be null terminated, but in general yes that's the problem and why I didn't want to use `PCWSTR` and thus make the method `unsafe`.

kennykerr: The `HSTRING` just seems like a reasonable compromise because it does guarantee the right memory layout and is safe and nicely convertible. Perhaps this is where we can pick up @riverar's suggestion and provide a `set_typed_value` kind of method that takes a slice and provides the added flexibility and micro-optimization if really needed.

ChrisDenton: Ah right, so it could just be `AsRef<[u16]>`? Though that would imply needing an `AsRef` implementation for `HSTRING`.

ChrisDenton: Oh sorry, missed your second comment. That does convince me it's not really needed.

ChrisDenton (approved):

crates/libs/strings/src/hstring_builder.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+}
+
+impl core::ops::Deref for HStringBuilder {
+    type Target = [u16];
+
+    fn deref(&self) -> &[u16] {
+        if let Some(header) = self.as_header() {
+            unsafe { core::slice::from_raw_parts(header.data, header.len as usize) }
```
goffrie, line 60: Doesn't this expose uninitialized memory? i.e. `*HStringBuilder::new(123).unwrap()` is a `[u16]` slice whose contents are uninitialized.

PR #3135, opened 2024-06-28
Clarify support for non-Windows targets

kennykerr:
The windows-rs project is focused on bringing the best of Windows to Rust developers. Various efforts to support non-Windows targets continue to make it too difficult to support Windows, which is the whole point of this project. 

The position we're taking is to provide basic non-Windows compatibility for `windows-core` only to support COM and nothing more. Even this adds regrettable complexity to the `windows-core` crate, but this update at least attempts to make that complexity manageable by providing a simple way for us to partition what parts of `windows-core` is specific to Windows and what is cross-platform. Subsequent updates will limit more of the code that isn't required for COM support.

The [test_linux](https://github.com/microsoft/windows-rs/tree/master/crates/tests/linux) crate covers what little support we offer for non-Windows targets. This is limited to `windows-core` and `windows-result` and is tested in [linux.yml](https://github.com/microsoft/windows-rs/blob/master/.github/workflows/linux.yml) as part of each build. 

Fixes: #3101
Fixes: #3083
Related: #3125

### DIFF
--- a/crates/libs/core/src/imp/mod.rs
+++ b/crates/libs/core/src/imp/mod.rs
@@ -1,21 +1,16 @@
-mod bindings;
+#[cfg(windows)]
+include!("windows.rs");
+
 mod can_into;
 mod com_bindings;
-mod factory_cache;
-mod generic_factory;
 mod ref_count;
 mod sha1;
-mod waiter;
 mod weak_ref_count;
 
-pub use bindings::*;
 pub use can_into::*;
 pub use com_bindings::*;
-pub use factory_cache::*;
-pub use generic_factory::*;
 pub use ref_count::*;
 pub use sha1::*;
-pub use waiter::*;
 pub use weak_ref_count::*;
 
 #[doc(hidden)]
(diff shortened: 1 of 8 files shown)


riverar (approved):

crates/libs/core/src/imp/mod.rs
```diff
@@ -1,21 +1,16 @@
-mod bindings;
+#[cfg(windows)]
+include!("windows.rs");
```
ChrisDenton, line 2: I admit I don't love things like this, Which I guess illustrates the point on the impact of supporting non-Windows platforms.

ChrisDenton (approved):
This sounds like a good balance. Having support tightly scoped to only the essentials on non-Windows platforms still allows other parties to build on top of that if they wish. Just in a way that doesn't negatively impact windows-rs support for Windows.

crates/libs/core/src/imp/mod.rs
kennykerr: Exactly.

sivadeilra (approved):

PR #3143, opened 2024-07-03
Make new `windows-strings` crate Windows-only

kennykerr:
This is a new crate, and I don't want the headache of non-Windows compatibility right out of the gate. 

The only crates that continue to support limited non-Windows builds are:

* `windows-bindgen` and `windows-metadata` for code generation on non-Windows platforms.
* `windows-core` and `windows-result` for COM support on non-Windows platforms.

### DIFF
--- a/crates/libs/windows/src/lib.rs
+++ b/crates/libs/windows/src/lib.rs
@@ -6,6 +6,7 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 [Feature search](https://microsoft.github.io/windows-rs/features/#/0.57.0)
 */
 
+#![cfg(windows)]
 #![doc(html_no_source)]
 #![allow(non_snake_case, clashing_extern_declarations, non_upper_case_globals, non_camel_case_types, missing_docs, clippy::all)]
 #![cfg_attr(not(feature = "docs"), doc(hidden))]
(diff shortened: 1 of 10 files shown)


crates/libs/windows/src/lib.rs
```diff
@@ -6,6 +6,7 @@ Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs
 [Feature search](https://microsoft.github.io/windows-rs/features/#/0.57.0)
 */
 
+#![cfg(windows)]
```
IpFruion, line 9: ❓ Doesn't this make it so that all downstream modules are now only windows available?

I am running into `windows::core` not found running with rust-analyzer in linux. I am good with the `crates/libs/strings/src/lib.rs` change to support windows only.

But I would recommend the 
```
#[cfg(windows])
pub use windows_strings::*;
```
in the `crates/libs/core/src/windows.rs` file

kennykerr: https://github.com/microsoft/windows-rs/actions/workflows/linux.yml validates what non-Windows support is provided by windows-rs and this includes the `windows-core` and `windows-result` crates:

https://github.com/microsoft/windows-rs/blob/master/crates/tests/linux/Cargo.toml

If there's a gap in the testing please create a new issue with a repro. Thanks!

IpFruion: Sounds good will do!

In that test file, it does look like it targets `windows_core` and `windows_result` directly instead of going through `windows-rs` crate (which I think is under `../../libs/windows`) which re-exports `windows::core` and `windows::result`. 

But I will create that issue with reproduction

IpFruion: https://github.com/microsoft/windows-rs/issues/3155

PR #3144, opened 2024-07-03
Implement static COM objects

sivadeilra:
This allows you to define "static COM objects".  Static COM objects are COM objects that are stored in static fields.  This avoids the need for a separate heap allocation for stateless COM objects.  It also allows `const fn` initialization of static COM objects.

A static COM object is _not_ the same as a `ComObject<T>` stored in a static field.  `ComObject<T>` is a pointer, which owns 1 reference count, to an allocated COM object.  `StaticComObject<T>` _directly_ contains the COM object's data, its reference count, and its vtable pointers.

Just like `ComObject<T>`, `StaticComObject<T>` allows you to cast it to any interface chain or interface pointer, and even to create a `ComObject<T>` that points to the `StaticComObject<T>`. This is safe because creating the `ComObject<T>` increases the reference count of the static object; dropping the `ComObject<T>` decreases the reference count, but it will never reach zero so it will never be "freed" (which would be wrong).

This feature is intended to support and simplify apps that need to define factory objects and other stateless objects. There is no need to heap-allocate these objects. This also simplifies their initialization because we simply never need to run any initialization logic at all, so we eliminate "order of initialization" bugs.

### DIFF
--- a/crates/tests/implement_core/src/static_com_object.rs
+++ b/crates/tests/implement_core/src/static_com_object.rs
@@ -1,0 +1,77 @@
+//! Unit tests for `windows_core::StaticComObject`
+
+use std::sync::atomic::{AtomicU32, Ordering::SeqCst};
+use windows_core::{
+    implement, interface, ComObject, IUnknown, IUnknownImpl, IUnknown_Vtbl, InterfaceRef,
+    StaticComObject,
+};
+
+#[interface("818f2fd1-d479-4398-b286-a93c4c7904d1")]
+unsafe trait INumberFactory: IUnknown {
+    fn next(&self) -> u32;
+
+    fn add(&self, x: u32, y: u32) -> u32;
+}
+
+#[implement(INumberFactory)]
+struct MyFactory {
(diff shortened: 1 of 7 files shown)


crates/tests/implement_core/src/static_com_object.rs
```diff
@@ -0,0 +1,77 @@
...
+    unsafe fn add(&self, x: u32, y: u32) -> u32 {
+        x + y
+    }
+}
+
+static NUMBER_FACTORY_INSTANCE: StaticComObject<MyFactory> = MyFactory {
+    x: AtomicU32::new(100),
+}
+.into_static();
+
```
kennykerr, line 35: Ideally there would be consistency here with `ComObject` where you'd use `new`. Just makes it a little harder to learn I have to use `new` for ComObject but `into_static` for `StaticComObject`.

kennykerr: Something we can hopefully improve in future.

kennykerr (approved):

PR #3146, opened 2024-07-04
More variant types

### LINKED ISSUE #2983: Adding more Variant types
### Suggestion

Hey,

Using `windows-rs 0.52.0` I had my own `Variant` helper.
In `windows-rs 0.54.0` I see that `windows::core::Variant` was added, which is nice.
With that change, all `Windows::Win32::System::Variant::VARIANT*` where removed (which completely breaks my `Variant` implementation).

I would be happy to migrate to `windows::core::Variant` but it is missing types support that I used:
- `fn null() -> Variant` to construct a `VT_NULL`  variant
- `TryFrom<&[String]> for Variant` to construct `VT_ARRAY_BSTR`  variant
- `TryFrom<&[u8]> for Variant` to construct `VT_ARRAY_UI1`  variant

If I can, I'm going to have a try at it.

PaulDance:
Dear maintainers,

This is an attempt to close #2983. It is effectively an upstream of remainder of the equivalent higher-level wrapper we have around `VARIANT`.

It contains:
 * A constructor for the `VT_NULL` variant. `is_null` was also added as a bonus for symmetry with `is_empty`.
 * `From<String>` mapping to `VT_BSTR` since `BSTR` also implements it.
 * `From<&[&str]>` mapping to `VT_ARRAY | VT_BSTR` since `BSTR` implements `From<&str>`.
 * `From<&[String]>` mapping to `VT_ARRAY | VT_BSTR` since `BSTR` also implements `From<&String>`.
 * `From<&[u8]>` mapping to `VT_ARRAY | VT_UI1`.
 * Tests for everything. I can always add more if some scenarios were overseen. I actually haven't executed them locally yet in the hope that everything works the first time, so expect them to fail.

The array variants require allocation that is hopefully handled correctly. The `Drop` implementations were updated accordingly.

Also, `is_empty` was `const`, so I thought it would be interesting as a sort of bonus to mark the other methods as such: see first commit. I can remove it though, if it is judged not acceptable.

Cheers,
Paul.

### DIFF
--- a/crates/libs/core/src/imp/bindings.rs
+++ b/crates/libs/core/src/imp/bindings.rs
@@ -661,6 +661,7 @@ pub const VT_I1: VARENUM = 16u16;
 pub const VT_I2: VARENUM = 2u16;
 pub const VT_I4: VARENUM = 3u16;
 pub const VT_I8: VARENUM = 20u16;
+pub const VT_NULL: VARENUM = 1u16;
 pub const VT_R4: VARENUM = 4u16;
 pub const VT_R8: VARENUM = 5u16;
 pub const VT_UI1: VARENUM = 17u16;
--- a/crates/libs/core/src/imp/bindings.rs
+++ b/crates/libs/core/src/imp/bindings.rs
@@ -21,6 +21,9 @@ windows_targets::link!("ole32.dll" "system" fn PropVariantClear(pvar : *mut PROP
 windows_targets::link!("ole32.dll" "system" fn PropVariantCopy(pvardest : *mut PROPVARIANT, pvarsrc : *const PROPVARIANT) -> HRESULT);
 windows_targets::link!("oleaut32.dll" "system" fn VariantClear(pvarg : *mut VARIANT) -> HRESULT);
 windows_targets::link!("oleaut32.dll" "system" fn VariantCopy(pvargdest : *mut VARIANT, pvargsrc : *const VARIANT) -> HRESULT);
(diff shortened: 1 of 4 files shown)


PaulDance:
@microsoft-github-policy-service agree company="HarfangLab"

kennykerr:
Thanks, I'd like to reduce rather than expand `VARIANT` support in `windows-core`, likely by moving this to a dedicated crate like `windows-variant`. I also think that array support should be pushed to a `SAFEARRAY` wrapper rather than being baked into `VARIANT`. This is closely tied to `VARIANT` so I imagine I'd probably combine these in the same crate. At any rate, if you can give me some time to figure out that restructuring it should be easier to get more `VARIANT` type support up and running.

PaulDance:
Alright, ping me when you do :slightly_smiling_face:

PaulDance:
?

kennykerr:
We'll keep the discussion alive on the issue. 😊

PaulDance:
Actually, I just thought about this again: why block this PR's review, iteration and potential merge while waiting for a refactor you want to achieve?

The added APIs would still remain after the refactor, whether they are added before or after, right? Also, in any case, moving the types to a separate crate could potentially constitute an API break, but that is independent of the things proposed here, right?

This and that can be done in parallel, no?

kennykerr:
Sorry I've not had any time to return to this yet. I did have a quick look at your old PR and I'm not sure that's the right sustainable approach in general. We probably need a generic `SAFEARRAY` and then provide convertibility between `VARIANT` and `SAFEARRAY` but I've not given it much more thought yet.

PR #3162, opened 2024-07-16
From trait for u128

ssrlive:
From trait for u128 and function `GUID::to_u128`


kennykerr:
Please open an issue before creating a PR. That way we can discuss the problem you're trying to solve and possible solutions, other constraints and considerations, before you begin investing in code.

Issue #3163, opened 2024-07-16
Windows-core incorrectly propagates the std feature to subcrates


Intege-rs:
### Summary

[Windows-core](https://github.com/microsoft/windows-rs/blob/master/crates/libs/core/Cargo.toml) always enables the std feature on [Windows-strings](https://github.com/microsoft/windows-rs/blob/master/crates/libs/strings/Cargo.toml) and [Windows-results](https://github.com/microsoft/windows-rs/blob/master/crates/libs/result/Cargo.toml)

### Crate manifest

```TOML
[package]
name = "scratch"
version = "0.1.0"
edition = "2021"

[dependencies.windows]
version = "0.58.0"
default-features = false
features = []
```

### Crate code

```Rust
# build.rs
fn main() {
    println!("cargo:rustc-link-arg=/SubSystem:Console");
    println!("cargo:rustc-link-arg=/ENTRY:main");
}

# main.rs
#![no_std] #![no_main]

#[no_mangle]
extern "C" fn main() {
    // result incorrectly re-imports std into a no-std project
    let _ = windows::core::HRESULT(0);
}

// found duplicate lang item `panic_impl`
#[cfg_attr(not(test), panic_handler)]
fn _panic(_: &core::panic::PanicInfo) -> ! {
    loop {}
}
```

kennykerr:
I think I figured out how to test this - PR coming soon.

kennykerr:
Thanks for reporting this issue! Testing revealed that there was a bit more involved in fixing this problem. #3180 provides the fix as well as the test repro and verification.

PR #3164, opened 2024-07-16
Fix std propagation for windows-core

Intege-rs:
Fixes windows-core always enabling std on the result and strings crates

Fixes: #3163⬅️

### DIFF
--- a/crates/libs/core/Cargo.toml
+++ b/crates/libs/core/Cargo.toml
@@ -22,10 +22,12 @@ version = "0.52.6"
 path = "../targets"
 
 [dependencies.windows-result]
+default-features = false
 version = "0.2.0"
 path = "../result"
 
 [dependencies.windows-strings]
+default-features = false
 version = "0.1.0"
 path = "../strings"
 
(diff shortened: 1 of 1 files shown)


kennykerr:
Thanks! The fix looks reasonable to me. What's missing is any way to test this to avoid a regression. I have the `no-default-features.yml` but it would be great to have some enduring validation that this actually fails/works before and after this change. 

https://github.com/microsoft/windows-rs/blob/master/.github/workflows/no-default-features.yml

MarijnS95:
Nice! I just typed up an identical fix for `wgpu` only to see that you beat me by a week. Looks good!

kennykerr:
Still need a test. Will come up with a test and open a new PR.

tim-weis:
@MarijnS95 The man is asking for a test, even offers to come up with one, and your sole response is

> :-1:

What is wrong with you?

MarijnS95:
@tim-weis clickable emoji replies don't exactly allow more elaboration, and going out of my way to write a comment for it defeats the purpose of replying with an emoji (except if it were positive, as you wouldn't be asking for my clarification, most likely).

Either way, to answer your question, nothing wrong with me; but there is something (a lot) wrong with how `windows-rs` is managed. If `windows-rs` wasn't as good as it is and solving real gaps in the Rust ecosystem, I'd stay very far away from engaging in contributions and conversations. I'll leave it as an exercise to the reader to figure out feels unheard - sometimes even disrespected - around here.

Issue #3166, opened 2024-07-16
COM Send example needed


aconverse:
### Suggestion

First of all I love windows-rs and have been using it successfully for a while. But I find myself needing to send a COM object to another thread and am a bit puzzled. I apologize if I'm missing something obvious. 

The ability to safely send COM objects across thread or even process boundaries seems to be a heavily advertised feature of COM going way back to before rust even existed. 

If I just try to send the `IUri` from the COM example I get an error:
```
error[E0277]: `NonNull<c_void>` cannot be sent between threads safely
    --> src/main.rs:39:41
```

I vaguely recall something about marshalling COM objects, so I marshal the object (see the crate code from the form), but I get the same error. Now it's just an `IStream` that contains the `IUnknown`. Should `IStream` be `Send`? It seems like Marshaled COM object `IStream`s are safe to send but arbitrary `IStream`s are not.

Is there an example of sending a COM object to another thread in rust somewhere? I wound up wrapping marshal / unmarshal in my own abstraction here and declaring that as Send. Is that the way to go?

Main.rs below:
```
use windows::core::*;
use windows::Win32::System::Com;
use windows::Win32::System::Com::IStream;
use windows_result::Result;

struct MarshalledComPtr<T: windows::core::imp::CanInto<IUnknown> + Interface> {
    stream: IStream,
    phantom: std::marker::PhantomData<T>,
}

impl<T: windows::core::imp::CanInto<IUnknown> + Interface> MarshalledComPtr<T> {
    fn new(iface: &T) -> Result<Self> {
        unsafe {
            Ok(Self {
                stream: Com::Marshal::CoMarshalInterThreadInterfaceInStream(&T::IID, iface)?,
                phantom: std::marker::PhantomData,
            })
        }
    }
    fn take(self) -> Result<T> {
        unsafe { Com::Marshal::CoUnmarshalInterface(&self.stream) }
    }
}

unsafe impl<T: windows::core::imp::CanInto<IUnknown> + Interface> Send for MarshalledComPtr<T> {}

fn main() -> windows::core::Result<()> {
    unsafe {
        Com::CoInitializeEx(None, Com::COINIT_MULTITHREADED).unwrap();

        let uri = Com::CreateUri(
            w!("http://kennykerr.ca"),
            Com::URI_CREATE_FLAGS::default(),
            0,
        )?;
        // error: uri is not Send due to IUnknown
        //let joiner = std::thread::spawn(|| my_thread(uri));

[...]

PR #3167, opened 2024-07-17
Update main.rs with `CoInitializeEx`

ibeckermayer:
This is part PR, part question. My understanding is that one needs to call `CoInitialize`/`CoInitializeEx` to initialize COM for a given process/thread before making use of any COM interfaces. I see this happening in some examples, however not this one targeted directly at COM.

Can anybody clarify this topic for me? This example seems to run fine for me on my machine, however all documentation I can find seems to suggest that such a call is _required_. So how are we getting away without it here?

### DIFF
--- a/crates/samples/windows/com_uri/src/main.rs
+++ b/crates/samples/windows/com_uri/src/main.rs
@@ -3,6 +3,8 @@ use windows::Win32::System::Com::*;
 
 fn main() -> windows::core::Result<()> {
     unsafe {
+        CoInitializeEx(None, COINIT_MULTITHREADED).ok()?;
+        
         let uri = CreateUri(w!("http://kennykerr.ca"), URI_CREATE_FLAGS::default(), 0)?;
 
         let domain = uri.GetDomain()?;


riverar:
Don't think this is needed as it's not a COM API. Probably a mistake; I suspect the intention here was to showcase use of the `IUri::CreateUri` instead.

kennykerr:
`CoInitializeEx` creates or joins an apartment that certain APIs may require from the calling thread. This does not apply to all COM interfaces. A COM interface is just a vtable, a virtual function table, following certain conventions. Apartments are a feature of COM that many APIs don't require.

kennykerr:
And if you really do need an apartment, it's almost always better to use `CoIncrementMTAUsage`. Here's an example:

https://github.com/microsoft/windows-rs/issues/3166#issuecomment-2231756285

kennykerr:
Yes, you can safely call `CoIncrementMTAUsage` and let the MTA live until the process terminates. This is actually safer since COM lifetime is usually unpredictable.

Issue #3148, opened 2024-07-06
Add support for setting string values with the `REG_EXPAND_SZ` type when modifying the registry


InfyniteHeap:
### Suggestion

Coming from https://github.com/rust-lang/rustup/pull/3896/files#r1666173552. As what we discussed in this review, we should set string values with `REG_EXPAND_SZ` type when modifying `PATH`. So, we need a way to set the string type when calling `set_string()` and `set_hstring()`.

Appendix: We also need a way to verify whether `String` or `HSTRING` has `REG_SZ` or `REG_EXPAND_SZ`. See https://github.com/rust-lang/rustup/blob/044083c9cc58b0ae23eeb1a390ca17ddc6f859d2/src/cli/self_update/windows.rs#L950.

kennykerr:
Here you go: #3184

kennykerr:
#3181 is now merged - can you please test the rustup port against github? I don't plan to publish another crate release until we're confident this is in good shape for the foreseeable future. @InfyniteHeap @rami3l

InfyniteHeap:
> #3184 is now merged - can you please test the rustup port against github? I don't plan to publish another crate release until we're confident this is in good shape for the foreseeable future. @InfyniteHeap @rami3l

I think we need a `Value` constructor, because there are some cases where directly using `Value` are required in rustup.

rami3l:
@kennykerr https://github.com/rust-lang/rustup/pull/3896 looks great with the Cargo patch that points to the latest dev version of the lib!

djc:
I'd like to check some more things before the release, we're left with some unsafe code in rustup that might be nice to migrate into this crate.

kennykerr:
@rami3l - glad to hear things are going well! 

@djc - sounds good - please open a new issue for any bug/feature requests.

InfyniteHeap:
> I'd like to check some more things before the release, we're left with some unsafe code in rustup that might be nice to migrate into this crate.

The reason why I add unsafe code is that, since I cannot manually construct a `Value` to compare to what I got from `get_value()`, I have to manually seperately convert the got `Value` into `Type` and `&[u8]` to fit the requirements. See my latest commit and [this issue](https://github.com/microsoft/windows-rs/issues/3189).

kennykerr:
That issue isn't clear on what it's asking for. By the way, #3190 provides a few small improvements for convertibility.

InfyniteHeap:
[...]

PR #3175, opened 2024-07-23
Make users can set string values with `REG_EXPAND_SZ` type

InfyniteHeap:
What's this all about?

Fixes: #3148.

As sometimes users need to set string values with `REG_EXPAND_SZ` (e.g. when setting `PATH` values, see https://github.com/rust-lang/rustup/pull/3896 for more details), I think it is necessary to add relevant support for this.

In addition, users can now identify whether a string value has a `REG_EXPAND_SZ` type.

### DIFF
--- a/crates/libs/registry/src/key.rs
+++ b/crates/libs/registry/src/key.rs
@@ -91,20 +91,36 @@ impl Key {
         unsafe { self.set_value(name, REG_QWORD, &value as *const _ as _, 8) }
     }
 
-    /// Sets the name and value in the registry key.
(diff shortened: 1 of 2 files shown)


InfyniteHeap:

kennykerr:
Thanks, been away for a few days. I've got an update coming that incudes explicit support for expand string types as well as allocation-free updates.

InfyniteHeap:
> Thanks, been away for a few days. I've got an update coming that incudes explicit support for expand string types as well as allocation-free updates.

Looks great! I've waited for such this update for a long time!😊

Issue #342, opened 2020-10-18
WinRT Futures should update their inner Waker


adumbidiot:
Opening this issue as a continuation of #322. While it is now possible to call `poll` on winrt futures more than once, consecutive `poll` calls will not update the `Waker`. As a result, these futures will hang forever if a future is transferred between tasks after being polled. A simple example of this behavior:
```rust
mod bindings {
    winrt::include_bindings!();
}

use bindings::*;
use crate::test_component::TestRunner;
use futures::FutureExt;

fn main() {
    let mut tokio_rt = tokio::runtime::Builder::new()
        .threaded_scheduler()
        .build()
        .unwrap();
    
    tokio_rt.block_on(async {
        let fut_1 = TestRunner::create_async_action(1000).unwrap();
        let fut_2 = TestRunner::create_async_action(2000).unwrap();
        
        let mut fut_1_fuse = fut_1.fuse();
        let mut fut_2_fuse = fut_2.fuse();
        
        let incomplete_future = futures::select! {
            res_1 = fut_1_fuse => fut_2_fuse,
            res_2 = fut_2_fuse => fut_1_fuse,            
        };
        
        println!("1 future complete, finishing other on different task...");
        
        // Work-around !Send future, this example still works with Send winrt futures and tokio::spawn.
        let local = tokio::task::LocalSet::new();
        
        local.spawn_local(async move {
            incomplete_future.await.unwrap();
        });
        
        local.await;
        
        println!("Both futures complete!");
    });
    
    println!("Done!");
}
```
This example will hang indefinitely and never complete. Unfortunately, I don't this its possible to fix this while maintaining the `Future` impl on `IAsyncxxx` as some extra memory is needed to store the shared reference to the `Waker` to update if needed.

In my opinion, [IntoFuture](https://doc.rust-lang.org/std/future/trait.IntoFuture.html) is probably the best way forward, however it is currently unstable. Implementing this trait will allow `await`-ing on that value by implicity calling `into_future`, just as the `IntoIterator` trait allows iterating over a value that doesn't explicitly implement `Iterator`. In the meantime, maybe we could remove the `Future` impl from `IAsyncxxx`, replacing it with a method to get a wrapper struct implementing `Future`?

kennykerr:
[...]

PR #3177, opened 2024-07-24
Implement `IntoFuture` for WinRT async types

goffrie:
This reverts #3142 while also addressing #342.

I implement `IntoFuture` by wrapping each `IAsyncOperation`/etc into a `windows_core::FutureWrapper` that includes a mutable Waker slot (i.e. `Arc<Mutex<Waker>>`, so that we only need to call `SetCompleted` once. To deduplicate the implementation for the various interfaces that exist, I've added a trait `windows_core::AsyncOperation` that abstracts over the SetCompleted/etc methods.

I believe the `Arc` is necessary as there is no way to cancel a `SetCompleted` callback once it is registered. The `Mutex<Waker>` could be an `AtomicWaker` instead, but that's a fairly complex piece of code that I don't want to pull in.

Additionally, the wrapper future automatically calls `Cancel()` when dropped, as is conventional for Rust futures.

### DIFF
--- a/crates/libs/core/src/future.rs
+++ b/crates/libs/core/src/future.rs
@@ -22,26 +22,26 @@ pub trait AsyncOperation {
     /// Get the result value from a completed operation.
     /// Wraps `self.GetResults()`.
     fn get_results(&self) -> crate::Result<Self::Output>;
+    /// Attempts to cancel the operation. Any error is ignored.
+    /// Wraps `self.Cancel()`.
+    fn cancel(&self);
 }
 
 /// A wrapper around an `AsyncOperation` that implements `std::future::Future`.
 /// This is used by generated `IntoFuture` impls. It shouldn't be necessary to use this type manually.
-pub struct FutureWrapper<T> {
+pub struct FutureWrapper<T: AsyncOperation> {
     inner: T,
     waker: Option<Arc<Mutex<Waker>>>,
 }
 
-impl<T> FutureWrapper<T> {
+impl<T: AsyncOperation> FutureWrapper<T> {
     /// Creates a `FutureWrapper`, which implements `std::future::Future`.
     pub fn new(inner: T) -> Self {
-        Self {
-            inner,
-            waker: None,
-        }
+        Self { inner, waker: None }
     }
 }
 
-impl<T> Unpin for FutureWrapper<T> {}
+impl<T: AsyncOperation> Unpin for FutureWrapper<T> {}
 
 impl<T: AsyncOperation> Future for FutureWrapper<T> {
     type Output = crate::Result<T::Output>;
@@ -64,5 +64,11 @@ impl<T: AsyncOperation> Future for FutureWrapper<T> {
             }
             Poll::Pending
         }
+    }
+}
+
+impl<T: AsyncOperation> Drop for FutureWrapper<T> {
+    fn drop(&mut self) {
+        self.inner.cancel();
     }
 }
--- a/crates/libs/core/src/future.rs
+++ b/crates/libs/core/src/future.rs
@@ -1,0 +1,68 @@
+#![cfg(feature = "std")]
+
+use std::{
+    future::Future,
+    pin::Pin,
+    sync::{Arc, Mutex},
+    task::{Poll, Waker},
+};
+
+/// Wraps an `IAsyncOperation`, `IAsyncOperationWithProgress`, `IAsyncAction`, or `IAsyncActionWithProgress`.
+/// Impls for this trait are generated automatically by windows-bindgen.
+pub trait AsyncOperation {
+    /// The type produced when the operation finishes.
+    type Output;
+    /// Returns whether the operation is finished, in which case `self.get_results()` can be used to get the returned data.
+    /// Wraps `self.Status() != AsyncStatus::Started`.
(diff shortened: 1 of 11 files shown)


goffrie:

seritools:
Yes please, we definitely use this, so we're stuck on an old windows-rs version for now 💀

kennykerr:
Adding an explanation to the readme about how and why using `IntoFuture` is necessary to support WinRT async in Rust would be helpful.

kennykerr:
The clippy build failure has already been addressed - you likely just need to merge master.

crates/libs/core/src/future.rs
```diff
@@ -0,0 +1,74 @@
...
+                })?;
+            }
+            Poll::Pending
+        }
+    }
+}
+
+impl<T: AsyncOperation> Drop for FutureWrapper<T> {
+    fn drop(&mut self) {
+        self.inner.cancel();
```
kennykerr, line 72: WinRT's cancellation support isn't very widely supported and requires a fair amount of interlocking over and above the virtual function dispatch cost. I don't think we should call that here just because it is "conventional in Rust".

crates/libs/core/src/future.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Attempts to cancel the operation. Any error is ignored.
+    /// Wraps `self.Cancel()`.
+    fn cancel(&self);
+}
+
+/// A wrapper around an `AsyncOperation` that implements `std::future::Future`.
+/// This is used by generated `IntoFuture` impls. It shouldn't be necessary to use this type manually.
+pub struct FutureWrapper<T: AsyncOperation> {
+    inner: T,
+    waker: Option<Arc<Mutex<Waker>>>,
```
kennykerr, line 34: This seems like an overly expensive combination for synchronization. I'm not sure how to simplify but I'll play around with it a bit.

kennykerr: Haven't found a simpler approach. 🤷

crates/libs/core/src/future.rs
```diff
@@ -0,0 +1,74 @@
...
+    type Output = crate::Result<T::Output>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Self::Output> {
+        if self.inner.is_complete()? {
+            Poll::Ready(self.inner.get_results())
+        } else {
+            if let Some(saved_waker) = &self.waker {
+                // Update the saved waker, in case the future has been transferred to a different executor.
+                // (e.g. if using `select`.)
+                let mut saved_waker = saved_waker.lock().unwrap();
```
kennykerr, line 56: I think we're just lacking a formal test for this.

crates/libs/core/src/future.rs
seritools: How does this compare to the old implementation? Is this incurring an extra allocation for each Future call? :o

kennykerr: Yes, previously (#3142) I didn't store the `Waker` at all. Now the `Waker` is stored in an `Arc`, which adds a heap allocation, and requires a bunch of interlocking to retrieve the `Waker`.

kennykerr: This is why I'm not excited about this in general - the WinRT async model and the Rust futures library don't seem to have a natural fit.

goffrie: I don't feel like an extra heap allocation is a huge problem since eg. calling SetCompleted also needs to allocate already.
I think it is possible to combine those into one allocation though at the cost of some unsafe code. Would you prefer that?
Also, would you prefer to replace the Mutex with lighter weight synchronization using hand-rolled atomics?

crates/libs/core/src/future.rs
goffrie: The added test `test_async_updates_waker` covers this (it asserts that a future works after it has been transferred between executors).

crates/libs/core/src/future.rs
kennykerr: I'd prefer to keep it as simple as possible. We can always optimize later as needed.

kennykerr: Perhaps at least a comment here somewhere that the `Completed` can only be set once and thus the need for a shared waker.

kennykerr:
Are you planning on clearing up the build issues?

goffrie:
Yeah, I'll have time to poke at this again in a few days.

kennykerr:
I'd rather not keep PRs open indefinitely. Feel free to revisit when you've had time to work through the feedback and are ready for review.

crates/libs/core/src/future.rs
```diff
@@ -0,0 +1,74 @@
...
+                saved_waker.clone_from(cx.waker());
+            } else {
+                let saved_waker = Arc::new(Mutex::new(cx.waker().clone()));
+                self.waker = Some(saved_waker.clone());
+                self.inner.set_completed(move || {
+                    saved_waker.lock().unwrap().wake_by_ref();
```
kennykerr, line 62: For future reference, nothing appears to prevent this from racing ahead and running against a previous waker while the saved waker is being updated above.

kennykerr:
I added a sketch of what I think is needed here: #342

PR #3180, opened 2024-07-29
Fix support for `no_std`

kennykerr:
Rust's `no_std` feature doesn't actually verify that dependencies don't silently add a `std` dependency. It just prevents the current package from taking a direct dependency. This isn't particularly easy to test. 

This PR includes a sequence of commits that will verify success/failure before completion. 

Fixes: #3163 
Fixes: #3164

### DIFF
--- a/crates/tests/no_std/src/lib.rs
+++ b/crates/tests/no_std/src/lib.rs
@@ -1,21 +1,21 @@
 //! Test for `#![no_std]` crates.
 //!
 //! Compiling this crate verifies that the Windows crates can be compiled with their "std"
-//! feature disabled.
+//! feature disabled. The tricky part is that `cargo test` depends on `std` so testing
+//! `no_std` requires using `cargo check` instead. That's what the `no_std.yml` is for.
 
 #![no_std]
 
-#[cfg(test)]
-mod tests {
-    use windows::core::{implement, ComObject};
+#[windows::core::implement]
+struct App;
 
-    #[implement]
-    struct App {}
-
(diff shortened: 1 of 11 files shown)


kennykerr:
OK, we have a green build. The next commit adds a duplicate panic handler. Duplicate because an indirect `std` dependency will bring in a panic handler and conflict.

kennykerr:
And finally, https://github.com/microsoft/windows-rs/pull/3180/commits/bc4f31374864fad276aee18ede002e14efd20c09 avoids the indirect `std` dependencies and provides the actual fix to the problem. This ended up requiring a bit more work than was originally suggested by #3164.

crates/tests/no_std/src/lib.rs
```diff
@@ -1,21 +1,28 @@
...
+}
 
-    // Compilation is sufficient to test.
-    #[test]
-    fn basic() {
-        let object = ComObject::new(App {});
-        drop(object);
-    }
+#[cfg_attr(not(test), panic_handler)]
+fn _panic(_: &core::panic::PanicInfo<'_>) -> ! {
```
dpaoliello, line 26: Can you please add a comment to explain what this is for?

dpaoliello (dismissed):

dpaoliello (approved):

ChrisDenton (approved):
Looks like a good solution. An alternative would be to fiddle with `harness = false` in the `Cargo.toml` for the test (to disable the libtest harness) but getting that wired up right might be more effort than it's worth.

PR #3190, opened 2024-08-05
Add registry `Value` to/from `HSTRING` conversion

kennykerr:
Building on #3184, this update just makes it easier to convert safely between the generic `Value` representation and `HSTRING` for accurate wide string (`[u16]`) representation. It's generally more efficient to query for `HSTRING` directly but at least if you need to deal with `Value` you can now safely and correctly convert to `HSTRING` without first converting to `String` or bytes. 

I also added some missing test coverage.

### DIFF
--- a/crates/libs/registry/src/key.rs
+++ b/crates/libs/registry/src/key.rs
@@ -101,7 +101,7 @@ impl Key {
         name: T,
         value: &windows_strings::HSTRING,
     ) -> Result<()> {
-        self.set_bytes(name, Type::String, value.as_bytes())
+        self.set_bytes(name, Type::String, as_bytes(value))
     }
 
     /// Sets the name and value in the registry key.
@@ -115,7 +115,7 @@ impl Key {
         name: T,
         value: &windows_strings::HSTRING,
     ) -> Result<()> {
-        self.set_bytes(name, Type::ExpandString, value.as_bytes())
+        self.set_bytes(name, Type::ExpandString, as_bytes(value))
     }
 
     /// Sets the name and value in the registry key.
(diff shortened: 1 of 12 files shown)


crates/libs/registry/src/key.rs
```diff
@@ -193,12 +193,24 @@ impl Key {
...
+            // RegSetValueExW expects string data to be null terminated.
+            if matches!(ty, Type::String | Type::ExpandString | Type::MultiString) {
+                debug_assert!(
+                    value.get(value.len() - 2) == Some(&0),
+                    "`value` isn't null-terminated"
+                );
+                debug_assert!(value.last() == Some(&0), "`value` isn't null-terminated");
+            }
+        }
+
```
riverar, line 213: Nit: This doesn't quite cover the `REG_MULTI_SZ` scenario, in which each item is terminated with a `\0`. So someone may pass in an non-terminated sequence and pass these checks.

kennykerr: Sure, this is just a basic smoke test for basic API conformance of strings for internal validation. Multi-strings are validated elsewhere.

kennykerr: In other words, this is still a raw `unsafe` function but at least now it provides the most basic string debug assertion.

crates/libs/registry/src/value.rs
```diff
@@ -71,7 +76,9 @@ impl TryFrom<Value> for String {
     type Error = Error;
     fn try_from(from: Value) -> Result<Self> {
         match from.ty {
-            Type::String | Type::ExpandString => Ok(Self::from_utf16(from.data.as_wide())?),
+            Type::String | Type::ExpandString => {
+                Ok(Self::from_utf16(trim(trim(from.data.as_wide())))?)
```
riverar, line 80: Is this double trim expected?

crates/libs/registry/src/key.rs
riverar: Understood. Could also just throw in something like

```rust
if matches!(ty, Type::MultiString) {
    debug_assert!(
        value.get(value.len() - 2..value.len()) == Some(&[0, 0]),
        "`value` sequence isn't null-terminated"
    );
}
```

Not required, will mark as resolved.

PR #3191, opened 2024-08-06
Improve sample binary build dependencies

kennykerr:
Binary crate dependencies continue to be a weak point in Cargo. This just attempts to make the binary component samples build a little bit more reliably. In these two cases, the dependent (client) crates need the component (server) crates built ahead of their build scripts as it relies on the component to produce the .winmd files that the build scripts use to generate bindings.


kennykerr:
But of course, that doesn't work since the yml tests are executed with an explicit target which means that Cargo doesn't use the default target location for binaries, a further reminder of how poorly binaries are supported in Cargo. 😢 https://github.com/rust-lang/cargo/issues/9661

Issue #1833, opened 2022-06-20
`Event` should not be mutable


kennykerr:
The `Event` implementation is thread-safe but because its methods are mutable it cannot be safely used concurrently. 

Related to #1829

PR #3198, opened 2024-08-12
Immutable Event implementation

lifers:
Implement the Event struct with no need of mutating methods using RwLock. I'd think it's safe from deadlock because:

- RwLock can block when reader and writer threads are contesting access
- When swap lock is acquired, we either have:
  - an exclusive write access from one of add, remove, or clear
  - a shared read access from call and one of add, remove, or clear
- Otherwise,
  - when a change lock is acquired, we have a single read access from one of add, remove, or clear
  - when no lock is acquired, there is no access
- Therefore delegates shouldn't deadlock

Please let me know if there is other considerations for this feature. Thank you!

Fixes: #1833


lifers:

kennykerr:
Additional synchronization should not be required above the swap and change locks. For reference, I wrote the Rust implementation based on the implementation I wrote for C++/WinRT here:

https://github.com/microsoft/cppwinrt/blob/master/strings/base_events.h

That in turn is loosely based on the WRL implementation.

kennykerr:
Basically all we need to do here is make `add`, `remove`, and `clear` take a `&self`. The only tricky bit is convincing Rust that `Array::swap` is thread safe.

lifers:
Since we have exclusive access to the delegate array when we hold the `swap` lock, would you rather have the swap done like this?

kennykerr:
Looks good - is it necessary to use a pointer rather than a mutable reference?

lifers:
I figured out a way to use a mutable reference using UnsafeCell.

Once the class design is complete, I guess we should impl Send and Sync right 🤔

kennykerr:
Yes, it can be `Send` and `Sync`.

kennykerr:
Would appreciate some more eyes on this. The change here is just to make the `Event` type - part of the supporting type system for implementing WinRT events - to not require mutable references to `self` since the implementation is otherwise entirely thread safe and is meant to be used in the context of a multi-threaded COM object that can receive concurrent requests for adding/removing/calling event handlers. 

@ChrisDenton @dpaoliello @sivadeilra

sivadeilra:
Reading.

crates/libs/core/src/event.rs
```diff
@@ -25,47 +29,52 @@ impl<T: Interface> Event<T> {
...
 
     /// Registers a delegate with the event object.
-    pub fn add(&mut self, delegate: &T) -> Result<i64> {
+    pub fn add(&self, delegate: &T) -> Result<i64> {
         let mut _lock_free_drop = Array::new();
         Ok({
             let _change_lock = self.change.lock().unwrap();
-            let mut new_delegates = Array::with_capacity(self.delegates.len() + 1)?;
-            for delegate in self.delegates.as_slice() {
+            // Safety: there is no mutable alias to self.delegates at this point
```
sivadeilra, line 43: This comment says a condition is true, but I don't see how that is guaranteed.  I see calls to modify `delegates` in the `add` and `remove` paths, and I see a call that reads `delegates` in the `call` path.  But I don't see how this guarantees the safety invariant implied by this comment.

How are concurrent calls to `add` and `call` guaranteed to be safe?

ChrisDenton: From my understanding exclusive (mutable) access is granted only by acquiring both locks. Though this is by convention rather then enforced. So acquiring one lock (which all functions do) prevents mutable access.

The pattern for `add`, `remove` and `clear` is to acquire a readonly `change` lock and then later upgrade it to an exclusive lock (by acquiring the `swap` lock too). The upgrade is conditional for `remove` and `clear`.

`call` simply acquires the `swap` lock, which blocks the other functions from upgrading their locks.

sivadeilra (changes requested):
I don't see how this code is safe (sound).  It accesses the `delegates` list while acquiring different locks in different `pub` methods.

Can you explain why this is safe?  If `delegates` is associated with exactly one lock, then why is it not contained within one of the `Mutex` instances?  Since this code acquires different locks in different paths, I don't understand how it is sound.

crates/libs/core/src/event.rs
sivadeilra: Ok, thanks.  I'm going to study it a bit more and see if I agree.

lifers: I see what you mean. My intention is when taking an immutable reference and reading from `add` (e.g. lines 43-48), we can still take an immutable reference at `call` as long as `add` haven't acquired the `self.swap` lock. I noticed that the `current_delegates` reference is still valid at line 55 (which is illegal) so I should've dropped it at line 49. This way `current_delegates` doesn't alias with line 55. Does this make sense?

lifers:
From my understanding:
- We don't want concurrent `add`/`remove`/`clear`, so we have the `self.change` Mutex to prevent it from happening.
- When `add`/`remove`/`clear` is reading `delegates` and hasn't mutated anything, we don't want to block `call` since it only needs an immutable reference.
- However, when `add`/`remove`/`clear` is writing to `delegates`, we do need an exclusive mutable reference so we have another lock `self.swap` for it.
- The only time `call` can't read `delegates` is when `delegates` are being written, or in other words `self.swap` is being locked. So we simply make `call` acquire `self.swap` in order to read it.

sivadeilra (changes requested):
Ok, I re-read this with Chris Denton's comments in mind.

I hate to say it, but I think this code makes the wrong trade-off of safety vs. performance.  First, this code may currently be sound, but it is dangerously close to the edge.  And there are safe abstractions that already provide the semantics that are wanted.

From my reading of this code, `RwLock` would provide exactly the same level of synchronization, and would do so safely.  The only scenario where `RwLock` would have different behavior would be that writers would briefly block reads (`call` paths), but I consider that acceptable for this design because modifying the list should be extremely infrequent, compared to reading the list.

So I think this PR needlessly uses `unsafe` code; there is no advantage here to using `unsafe` code.  Even worse, this PR mixes the implementation of a R/W lock with a particular usage of that R/W lock, so it mixes two things together.  That makes verification harder and makes it more likely that a future PR that looks innocent will actually break the safety invariants.  And there's just no advantage over using `RwLock`.

I'm also troubled by the performance implication of cloning the delegate list every time `call` runs.  That's a lot of allocator calls.

I think this would be improved by replacing all of this with `RwLock<Arc<Array<T>>>`.  The writer paths (`add`, `remove`, etc.) would acquire it in exclusive mode and do the obvious thing.  The `call` path would acquire the `RwLock` in shared mode, call `Arc::clone` (which is a single interlocked increment), then drop the `RwLock` guard.  This would eliminate all of the allocator calls in `call`.

This implementation would be 100% safe code, and obviously correct from a borrow-checker and concurrency point of view.

kennykerr:
Thanks for the feedback. As I mentioned in https://github.com/microsoft/windows-rs/pull/3198#issuecomment-2284051874, this implementation is copied from C++/WinRT which is based on WRL. The choice around performance tradeoffs was done before my time but it was stressed that C++/WinRT must provide the same performance tradeoffs as WRL as they were calculated at some point to be preferable for the OS. Having said that, I don't mind making a different set of performance tradeoffs for Rust if it means the code is simpler and thus easier to reason about. I will say that this implementation (at least the C++ implementations) has stood the test of time and very heavy usage.

However, there are some constraints that must be preserved. I suspect `RwLock<Arc<Array<T>>>` could preserve them but I thought I'd just be explicit about that here so that we're on the same page. That will also help folks understand the existing implementation. The constraints are essentially:

* No locks can be held while invoking delegates (event handlers). This is to avoid COM reentrancy issues and deadlocks.
* No locks can be held _indefinitely_ that would prevent `add` or `remove` calls from acquiring and executing within a lock. This is to avoid DoS for unrelated code. 

If a *simpler* implementation is possible that preserves these constraints, then I'm not opposed to it. 😊

sivadeilra:
Both of those constraints would be met by using `RwLock<Arc<Array<T>>>`.  I understand the need for them.

I think a simpler, provably-safe implementation is the best place to start.  If there is quantitative evidence that this implementation can't meet the same needs as the equivalent C++, then I'm open to adjusting that, but not without some evidence.

kennykerr:
Hey @lifers I realize this is probably more than you bargained for. 😊 Would you like someone else to handle this?

lifers:
Hi, this has gotten more interesting and I'd like to figure this out unless the maintainers need to finish this ASAP with their background knowledge 😄. It takes a while for me to figure out what else is happening inside the initial code. From what I've seen,
- Do we need to keep the fallible allocation that returns `E_OUTOFMEMORY` instead of Rust's default abort?
- Using the structs from Rust will simplify a lot of things but might increase the memory footprint (e.g. using `std::sync::Arc` instead of `imp::RefCount` for reference counting, using `RwLock` to guard the `Array`). Is this something of a concern?

kennykerr:
A few thoughts.

* Consistent error handling via `Result` (and `HRESULT`) is important. 
* The downside to using `RwLock<Arc<Array<T>>>` is that it would result in two allocations. The existing implementations handle this by storing the array and the reference count in a single allocation. This is also how the Rust event's internal `Array` works. 
* Some of the additional reasoning about the existing implementations is persuaded by the understanding that events will almost always only ever have zero or one event handlers so if you're going to optimize that's what you should optimize for. Thus, needing two allocations for storing a single event handler seems wasteful.

I wonder if we can use `RwLock<Array<T>>` instead of `RwLock<Arc<Array<T>>>` given that `Array` already provides a thread-safe reference counting.

sivadeilra:
If `Array` already supports reference-counted lifetimes, then `RwLock<Array<T>>` seems best.  The other approach would be `RwLock<Arc<[T]>>`; that is, `Arc` can directly contain owned slices.  This would also eliminate one of the allocations.

I don't believe that converting allocation failures to `E_OUTOFMEMORY` is feasible nor desirable.  Allocation is already pervasive in the `windows-rs` API.  For example, every COM method that takes a string parameter is generic over that input, and some of the implementations of those generics need to allocate.  They report allocation failure by panicking.  Panicking on memory allocation is the better approach in Rust, for a variety of reasons that I can go into if you're interested.

Does `Array` provide semantics above and beyond simply storing items?  Is there a reason to use `Array` instead of `Vec`, `Box<[T]>`, or `Arc<[T]>`?  In general, I think it's better to align to the Rust standard library, when possible, rather than creating new equivalents.

kennykerr:
The error handling is admittedly a little confusing around allocation failures in `windows-rs` partly because the Rust Standard Library doesn't consistently provide fallible allocation alternatives and partly because some system calls result in `E_OUTOFMEMORY` and so there's an inconsistent experience for Rust developers. In C++ I handled this by translating `E_OUTOFMEMORY` into `std::bad_alloc`. Perhaps the Rust version of `std::bad_alloc` is to panic but that's bigger conversation we can have elsewhere. Let's discuss that offline.

Anyway, `Array<T>` doesn't do anything beyond providing a reference-counted container so if `Arc<[T]>` can do the same in a single allocation then that sounds good to me.

sivadeilra:
So here's a sketch of what I propose:

```
pub struct Event<T> {
    delegates: RwLock<Arc<T>>,
}

impl<T> Event<T> {
    pub fn add(&self, delegate: T) {
        let mut guard = self.delegates.write().unwrap();
        let mut new_list = Vec::with_capacity(guard.len() + 1);
        for old_delegate in guard.iter() {
            new_list.push(old_delegate.clone());
        }
        new_list.push(delegate);
        let new_list_arc: Arc<[T]> = new_list.into(); // <-- converts Vec<T> into Arc<[T]>
        *guard = new_list_arc;
    }

    pub fn clear(&self) {
        let mut guard = self.delegates.write().unwrap();
        guard.clear();
    }

    pub fn remove(&self, delegate: &T) {
        let mut guard = self.delegates.write().unwrap();
        if let Some(i) = guard.iter().find(|old_delegate| old_delegate == delegate) {
            let mut new_list = Vec::with_capacity(guard.len() - 1);
            new_list.extend(guard[..i].iter().cloned());
            new_list.extend(guard[i + 1..].iter().cloned());
            let new_list_arc: Arc<[T]> = new_list.into(); // <-- converts Vec<T> into Arc<[T]>
            *guard = new_list_arc;
        } else {
            // We never found it, so don't modify the list.
        }
    }

    pub fn call(&self) {
        let delegates: Arc<[T]> = {
            let guard = self.delegates.read().unwrap();
            guard.clone()
            // <-- lock is released here
        };
        for delegate in delegates.iter() {
            // ...
        }
    }
```

Obviously this is just a sketch, but the idea should be clear.  The paths that modify the list use `RwLock::write`.  The path that reads the list (`call`) uses `RwLock::read`, but only for a very brief time so that they can increase the refcount on the list.

lifers:
If the fallible allocation is still needed, we could use a simplified `Array` just to allocate like:
```Rust
/// A thread-safe reference-counted array of delegates.
struct Array<T: Interface> {
    buffer: *mut Delegate<T>,
    len: usize,
}

impl<T: Interface> Default for Array<T> {
    fn default() -> Self {
        Self::with_capacity(0).unwrap()
    }
}

impl<T: Interface> Array<T> {
    /// Creates a new, empty `Array<T>` with the specified capacity.
    fn with_capacity(capacity: usize) -> Result<Self> {
        if capacity != 0 {
            let alloc_size = capacity * size_of::<Delegate<T>>();
            let header = heap_alloc(alloc_size)? as *mut Delegate<T>;

            Ok(Self {
                buffer: header,
                len: 0,
            })
        } else {
            Ok(Self {
                buffer: null_mut(),
                len: 0,
            })
        }
    }
```
which will be used as
```Rust
pub struct Event<T: Interface> {
    delegates: RwLock<Arc<Array<T>>>,
}

impl<T: Interface> Event<T> {
    pub fn add(&self, delegate: T) -> Result<()> {
        let mut guard = self.delegates.write().unwrap();
        let mut new_list = Array::with_capacity(guard.len() + 1)?;
        for old_delegate in guard.as_slice().iter() {
            new_list.push(old_delegate.clone());
        }
        new_list.push(Delegate::new(&delegate)?);
        // let new_list_arc: Arc<[Delegate<T>]> = new_list.into(); // <-- converts Vec<T> into Arc<[T]>
        *guard = new_list.into();
        Ok(())
    }

    pub fn clear(&self) -> Result<()> {
        let mut guard = self.delegates.write().unwrap();
        *guard = Array::with_capacity(0)?.into();
        Ok(())
    }
...
    pub fn call(&self) {
          let delegates = {
              let guard = self.delegates.read().unwrap();
              guard.clone()
              // <-- lock is released here
          };
          for delegate in delegates.as_slice().iter() {
              // ...
          }
      }
```

kennykerr:
Don't worry about allocation failure.

sivadeilra:
Honestly, I think providing the "fallible" allocation pathway is worse.  Just stick with the standard Rust approach.  If you are implementing a memory allocation container, the appropriate response to OOM is to call `std::alloc::handle_alloc_error`.  This initiates the panic / stack unwind.

crates/libs/core/src/event.rs
```diff
@@ -1,252 +1,89 @@
...
-            let token = delegate.to_token();
-            new_delegates.push(delegate);
+    pub fn add(&self, delegate: &T) -> Result<i64> {
+        let mut guard = self.delegates.write().unwrap();
+        let mut new_list = if let Some(old_delegates) = guard.as_ref() {
+            let mut new_list = Vec::with_capacity(old_delegates.len() + 1);
+            new_list.extend_from_slice(&old_delegates);
+            new_list
+        } else {
+            Vec::with_capacity(1)
```
kennykerr, line 29: It would be ideal if we could create the `Arc` "with capacity" rather than via a temporary `Vec`. 🥲

kennykerr: Note that `Arc` does provide a `FromIterator` implementation so we could `collect` directly into an `Arc` in some cases.

sivadeilra:
Looks a lot simpler -- thank you!

The Clippy suggestion about a `Default` impl is a good one.

crates/libs/core/src/event.rs
```diff
@@ -16,6 +16,9 @@ pub struct Event<T: Interface> {
     delegates: UnsafeCell<Array<T>>,
 }
 
+unsafe impl<T: Interface> Send for Event<T> {}
+unsafe impl<T: Interface> Sync for Event<T> {}
```
kennykerr: The `Send` and `Sync` implementations seem to have disappeared.

crates/libs/core/src/event.rs
lifers: Didn't see that before, thanks for the info 👍

crates/libs/core/src/event.rs
lifers: Yup accidentally removed it, now it's back

lifers:
By the way, now that `Event<T>` can be used concurrently, should we have tests for that?

kennykerr:
Concurrency tests can be challenging - fortunately the code is now a lot simpler so perhaps just a basic smoke test confirming that it is in fact `Send` and `Sync` may be helpful e.g. can be used with `thread::spawn` or similar.

crates/libs/core/src/event.rs
```diff
@@ -25,228 +23,78 @@ impl<T: Interface> Event<T> {
...
+    pub fn add(&self, delegate: &T) -> Result<i64> {
+        let mut guard = self.delegates.write().unwrap();
+        let new_delegate = Delegate::new(delegate)?;
+        let token = new_delegate.to_token();
+        let new_iter = once(new_delegate);
+
+        let new_list = if let Some(old_delegates) = guard.as_ref() {
+            Arc::from_iter(old_delegates.iter().cloned().chain(new_iter))
+        } else {
+            Arc::from_iter(new_iter)
```
kennykerr, line 40: Beautiful 🥲

crates/libs/core/src/event.rs
```diff
@@ -25,228 +23,78 @@ impl<T: Interface> Event<T> {
...
+        };
 
-            let _swap_lock = self.swap.lock().unwrap();
-            _lock_free_drop = self.delegates.swap(new_delegates);
-            token
-        })
+        *guard = Some(new_list);
```
kennykerr, line 43: The only remaining behavioral difference is that the old delegates aren't dropped outside the lock, as far as I can tell. That's what the `_lock_free_drop` does in the existing implementation for add/remove/clear. I'm not sure if there's a simple way to do that explicitly with `RwLock`.

lifers:
<details>
  <summary>Just an idea for a new test...</summary>
  ```Rust
  #[test]
  fn across_threads() -> Result<()> {
      let event = Arc::new(Event::<EventHandler<i32>>::new());
  
      let check = Arc::new(AtomicI32::new(0));
      let other = Arc::new(AtomicI32::new(0));
  
      // Register event handler on a different thread.
      let event1 = event.clone();
      let check1 = check.clone();
      let _ = std::thread::spawn(move || {
          event1.add(&EventHandler::<i32>::new(move |_, args| {
              check1.fetch_add(*args, Ordering::Relaxed);
              Ok(())
          })).unwrap();
      }).join().unwrap();
  
      // Raise event on the main thread.
      assert_eq!(check.load(Ordering::Relaxed), 0);
      event.call(|delegate| delegate.Invoke(None, 1));
      assert_eq!(check.load(Ordering::Relaxed), 1);
  
      let event2 = event.clone();
      let other2 = other.clone();
      let check2 = check.clone();
      std::thread::scope(move |s| {
          // Call the `check` event handler on a another thread.
          let event3 = event2.clone();
          s.spawn(move || {
              assert_eq!(check2.load(Ordering::Relaxed), 1);
              event3.call(|delegate| delegate.Invoke(None, 2));
              assert_eq!(check2.load(Ordering::Relaxed), 3);
          });
  
          // Register and call `other` event on another thread.
          let event3 = event2.clone();
          s.spawn(move || {
              let other3 = other2.clone();
  
              event3.add(&EventHandler::<i32>::new(move |_, args| {
                  other3.fetch_add(*args, Ordering::Relaxed);
                  Ok(())
              })).unwrap();
  
              assert_eq!(other2.load(Ordering::Relaxed), 0);
              event.call(|delegate| delegate.Invoke(None, 4));
              assert_eq!(other2.load(Ordering::Relaxed), 4);
          });
      });
  
      // Check final value after the thread scope ends.
      assert_eq!(check.load(Ordering::Relaxed), 7);
      assert_eq!(other.load(Ordering::Relaxed), 4);
  
      Ok(())
  }
  ```
</details>

crates/libs/core/src/event.rs
lifers: I think this will do
```Rust
/// Registers a delegate with the event object.
    pub fn add(&self, delegate: &T) -> Result<i64> {
        let mut _lock_free_drop = None;
        Ok({
            let mut guard = self.delegates.write().unwrap();
            let new_delegate = Delegate::new(delegate)?;
            let token = new_delegate.to_token();
            let new_iter = once(new_delegate);

            let new_list = if let Some(old_delegates) = guard.as_ref() {
                Arc::from_iter(old_delegates.iter().cloned().chain(new_iter))
            } else {
                Arc::from_iter(new_iter)
            };

            _lock_free_drop = guard.replace(new_list);
            token
        })
    }
```

kennykerr: Looks good.

sivadeilra: I think using an explicit sequence like this:

```
drop(guard);
drop(old_list); // drop the old list _after_ releasing lock
```

would be more clear, and less likely to be accidentally changed during a refactor.

sivadeilra:
I'm not opposed to that test, but from experience in testing concurrent algorithms, I don't think it actually has much value.  I know, more testing is always supposed to be better, but when it comes to concurrency, tests can often be flakey, or even if they aren't flakey, they don't actually exercise meaningful properties and so they just give a false sense of security.

If you think this test is valuable, and it runs in a very short period of time, I'm fine with it.  However, don't drop a thread joiner; that's nearly always a code smell.  If you think you've already synchronized with the thread, then retain the thread joiner and join on it at the end of the test function.  It should be a no-op, but it gives me higher confidence in the correctness of the test.

kennykerr:
Yep, I'd not worry about testing the concurrency itself and just test that the compiler can actually move the references around - the thing that wouldn't work if `Send` and `Sync` weren't implemented - in the simplest way possible.

crates/libs/core/src/event.rs
```diff
@@ -25,228 +23,86 @@ impl<T: Interface> Event<T> {
...
-                new_delegates.push(delegate.clone());
-            }
-            let delegate = Delegate::new(delegate)?;
-            let token = delegate.to_token();
-            new_delegates.push(delegate);
+    pub fn add(&self, delegate: &T) -> Result<i64> {
+        let mut guard = self.delegates.write().unwrap();
+        let new_delegate = Delegate::new(delegate)?;
+        let token = new_delegate.to_token();
+        let new_iter = once(new_delegate);
```
kennykerr, line 35:
```suggestion
        let new_delegate = Delegate::new(delegate)?;
        let token = new_delegate.to_token();
        let new_iter = once(new_delegate);
        let mut guard = self.delegates.write().unwrap();
```

At this point I'm just nitpicking. 😊 The only way this fails if the delegate wrapper cannot be constructed because the delegate cannot be marshaled. So put that first and grab the lock at the last moment.

lifers: Sure thing 😅

kennykerr (approved):
Appreciate all the work on this - looks great!

sivadeilra (approved):

lifers:
Glad to help! 😁 Thanks everyone for following through.

PR #3201, opened 2024-08-14
Remove uneeded feature dependencies

kennykerr:
Following #2735, these are no longer required.


kennykerr:
https://githubstatus.com doesn't seem to know that GitHub Actions are down...

Issue #3187, opened 2024-08-01
Root level documentation for crates is very sparse


Nerixyz:
### Suggestion

Most crates have the following documentation at the root on docs.rs (what people see when they open docs.rs/\<crate-name\>):

> Learn more about Rust for Windows here: <https://github.com/microsoft/windows-rs>

However, these crates have documentation in their README, visible on crates.io. The same documentation should be visible on docs.rs.
Having a refresher on the basic concept of a crate and a few links to more documentation/examples is really helpful.

I made a small concept in https://github.com/Nerixyz/windows-rs/commit/9dd956563dcc296dea28c6a7f4b522b898403e61. If that's good enough, I'll open a PR.

kennykerr:
I tried doing this a while back #1429 but couldn't get it to work reliably. If its testable and produces good doc generation then it seems like a reasonable improvement.

Nerixyz:
> I tried doing this a while back #1429 but couldn't get it to work reliably. If its testable and produces good doc generation then it seems like a reasonable improvement.

The tests would need to specify the required features, as rustdoc can't infer them from the doctests. That's only an issue for windows and windows-sys. I've added a `cfg_attr` for the required features. Realistically, documentation for both crates is only ever built with all features or the `docsrs` cfg for windows. All other crates don't require any features for the examples (or they are ignored, like in windows-implement/interface).

kennykerr:
Realize you probably didn't see my 👍 - did you still want to pursue this?

Nerixyz:
Oh, yea I missed that.

PR #3202, opened 2024-08-14
docs: add root level documentation for all libraries

Nerixyz:
**What's this all about?**

This PR adds crate-/root-level documentation for all library crates. All crates except `windows-implement` and `windows-interface`, import the documentation (with doctests) from their `README.md`.

`windows-sys` and `windows` include examples with some required features. Rustdoc can't infer them, so the inclusion of the README is only done if they're enabled. For `windows` this also checks if `docsrs` isn't set (although docs.rs should build without any enabled features).

Doctests are running in CI with `cargo test --doc -p <crate>` and documentation is generated with `cargo doc --no-deps -p <crate>`.

You can verify the output with

```powershell
cargo doc --no-deps `
  -p windows-bindgen `
  -p windows-core `
  -p cppwinrt `
  -p windows-implement `
  -p windows-interface `
  -p windows-metadata `
  -p windows-registry `
  -p windows-result `
  -p windows-strings `
  -p windows-targets `
  -p windows-version `
  -p windows `
    -F Data_Xml_Dom `
    -F Win32_Security `
    -F Win32_System_Threading `
    -F Win32_UI_WindowsAndMessaging `
  -p windows-sys `
    -F Win32_Security `
    -F Win32_System_Threading `
    -F Win32_UI_WindowsAndMessaging `
  --open
```

Fixes: #3187


kennykerr (approved):
Thanks!

kennykerr:
> windows-sys and windows include examples with some required features. Rustdoc can't infer them, so the inclusion of the README is only done if they're enabled. 

Just playing around with it and it looks good for most crates but this point for the `windows` and `windows-sys` crates seems problematic as I don't see how this would ever be the case in practice so I'm wondering what it's even for. Is it just to test the readme? `test_readme` already does that.

Nerixyz:
> Just playing around with it and it looks good for most crates but this point for the `windows` and `windows-sys` crates seems problematic as I don't see how this would ever be the case in practice so I'm wondering what it's even for. Is it just to test the readme? `test_readme` already does that.

Oh, I didn't see that test. Then the doctests I added to CI can be removed (since there aren't any doctests outside the readmes).

Issue #3204, opened 2024-08-15
Clean up doc testing


kennykerr:
> Just playing around with it and it looks good for most crates but this point for the `windows` and `windows-sys` crates seems problematic as I don't see how this would ever be the case in practice so I'm wondering what it's even for. Is it just to test the readme? `test_readme` already does that.

Oh, I didn't see that test. Then the doctests I added to CI can be removed (since there aren't any doctests outside the readmes).

_Originally posted by @Nerixyz in https://github.com/microsoft/windows-rs/issues/3202#issuecomment-2291634621_

PR #3205, opened 2024-08-15
Cleanup doc testing

Nerixyz:
**What's this all about?**

Removes the duplicate doctests. I kept the `cargo doc` to ensure the documentation always builds.

Fixes: #3204

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -20,13 +20,6 @@ jobs:
         uses: actions/checkout@v4
       - name: Check
         run: cargo doc --no-deps -p windows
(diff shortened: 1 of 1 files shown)


kennykerr:
And presumably we can revert the changes to `lib.rs` for both `windows` and `windows-sys`.

Nerixyz:
> And presumably we can revert the changes to `lib.rs` for both `windows` and `windows-sys`.

We could. My thinking behind the change was that the readmes now have links to the samples and the feature search, both of which have a version number in them.

kennykerr (approved):

PR #3207, opened 2024-08-16
Remove workaround for "unused" private fields

kennykerr:
Looks like newer nightly builds have fixed this problem and the workaround #3103 is no longer needed.

https://github.com/rust-lang/rust/issues/126169 has also been closed.


kennykerr:
@ChrisDenton I'm not sure when/what happened but it appears to no longer need the workaround. 😊

riverar:
Do we need to leave this in for prior versions down to MSRV? I can't recall if this only showed up in unstable channels for a brief amount of time.

kennykerr:
As far as I know this was nightly only. 

Either way, we test stable/nightly/msrv - I don't think we can afford to test everything in between. 😊

riverar (approved):

PR #3210, opened 2024-08-19
Always treat warnings as errors

kennykerr:
Previously, warnings were treated as errors only for PR validation. This just makes it easier to catch such issues during development and avoid unnecessary PR validation failure.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,7 @@
 [build]
 rustflags = [
+    "-D", "warnings",
+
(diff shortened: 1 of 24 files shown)


MarijnS95:
Running into this now, this seems to be hampering the developer experience. Sometimes when experimenting with bindgen code, warnings appear that disallow running `tool_bindings`, forcing the developer to clean up unused variables and imports before proceeding. Isn't the whole idea of a warning to allow developing code while making them aware ("warning"...) that it won't pass upstream CI (as intended) when submitted as-is?

PR #3213, opened 2024-08-21
Overhaul async and future support

kennykerr:
This update does a few notable things to improve async and future support in the `windows` crate.

* Async support in general is now provided by a crate extension rather than code generation. This approach was first introduced in #3110 and is a lot simpler to maintain as its all implemented in one place.

* WinRT classes that represent async execution are now generated simply as type aliases for their respective async interfaces. This greatly simplifies their definitions and makes it a lot easier to understand what's going on. This does mean they lose their identity as runtime classes but in practice this is not very consequential as all of the things you'd want to do with that identity make little sense here. WinRT classes should never have been used to represent async execution. If such a case arises, we can always bring back some limited version. 

* The four async interfaces still provide a blocking `get` method but now also offer `IntoFuture` implementations that may be used in `async` functions or blocks. The resulting shared `Future` implementation supports multiple contexts and addresses the issue first reported in #342. The implementation is somewhat inspired by #342 and the execution test provided by #3177.


crates/tests/futures/tests/tests.rs
```diff
@@ -0,0 +1,85 @@
...
+        assert!(bytes[0] == 1);
+        assert!(bytes[1] == 2);
+        assert!(bytes[2] == 3);
+
+        Ok(())
+    })
+}
+
+#[test]
+fn switch_context() -> Result<()> {
```
kennykerr, line 45: I'd love a simpler way to test switching execution contexts.

crates/libs/windows/src/extensions/Foundation/Async.rs
```diff
@@ -0,0 +1,275 @@
...
+// An `Async` represents a WinRT async execution object or type. There are precisely four such types:
+// - IAsyncAction
+// - IAsyncActionWithProgress
+// - IAsyncOperation
+// - IAsyncOperationWithProgress
+//
+// All four implementations are provided here and there is thus no need to implement this trait.
+// This trait provides an abstraction over the relevant differences so that the `AsyncFuture`
+// implementation below can be reused for all of them.
+pub trait Async: Interface {
```
erikdesjardins, line 18: FWIW, you can use the sealed trait pattern to guarantee that there can be no external implementations of this trait. See https://rust-lang.github.io/api-guidelines/future-proofing.html

kennykerr: Thanks for the tip! There are some other spots where this might be useful.

PR #3224, opened 2024-08-27
Use relative path for extension

glandium:
In the Firefox repository, we avoid vendoring windows-rs because it's big, and instead pull it some other way. Because cargo doesn't support this kind of setup, we have a fake windows-rs crate that uses the modules from where the real windows-rs is. This falls short when the WindowsAndMessaging feature is enabled because CARGO_MANIFEST_DIR doesn't point to the real windows-rs.


kennykerr (approved):
Thanks, I'd like to get rid of `include_ext` entirely but this seems like a reasonable change for the time being.

Issue #3225, opened 2024-08-27
Missing `SetFileTypeChoices` method for `FileSavePicker`


JunkuiZhang:
### Summary

As mentioned in the remarks section of Microsoft's documentation at [here](https://learn.microsoft.com/en-us/uwp/api/windows.storage.pickers.filesavepicker?view=winrt-26100), you must specify one or more file types using the `FileTypeChoices` property before calling the `PickSaveFileAsync` method. Typically, I would use the `SetFileTypeChoices` method to set this property, but I haven't been able to find it. Could you please advise me on how to set this property?

kennykerr:
Looks like the method is called `FileTypeChoices` and requires the "Foundation_Collections" feature. 

https://github.com/microsoft/windows-rs/blob/88a80f2022ce6796e230922a5c2158a05bc7b0ba/crates/libs/windows/src/Windows/Storage/Pickers/mod.rs#L745-L746

JunkuiZhang:
Well, that method retrieves the file types, but I want to set that property. Wait, do I need to add them manually? Something like this?
```rust
let map = dialog.FileTypeChoices()?;
map.add(...);
```

kennykerr:
That is what the sample does in the docs.

JunkuiZhang:
Sorry to bother you, there are `Set***` methods for most properties in windows-rs. So, at the first place I thought it should be something like this:
```rust
let map = dialog.FileTypeChoices()?;
map.add(...);
dialog.SetFileTypeChoices(map)?;
```
So I thought the method was missing, my apologies!

kennykerr:
No prob, some of these APIs can be a little confusing. This "property" cannot be set but instead always returns a reference that must be updated to "set" the property. 🤷

JunkuiZhang:
Really sorry to bother you again, but could you please guide me how to create an `IVector`? 
```rust
let picker = FileSavePicker::new()?;
let map: IMap<HSTRING, IVector<HSTRING>> = picker.FileTypeChoices()?;
```
When I call `map.Insert(K, V)`, the `V` parameter is expected to be `Param<IVector>`. I tried the following:

```rust
let vector_view: IVectorView<HSTRING> = rust_vec.try_into()?;
map.insert(&HSTRING::from("Example"), &vector_view)?;
```

But it didn't work. I also searched for a solution but couldn't find one. Could you please guide me on how to create an `IVector` for `map.Insert(K, V)`?

kennykerr:
[...]

PR #3226, opened 2024-08-27
Add file dialog sample

kennykerr:
You can read more about the [File Dialog APIs here](https://asp-blogs.azurewebsites.net/kennykerr/Windows-Vista-for-Developers-_1320_-Part-6-_1320_-The-New-File-Dialogs). 

Fixes: #3225


riverar (dismissed):
Should we add maybe a comment somewhere pointing to manifest docs if high DPI is needed (which will be likely)?

kennykerr:
Copied the manifest code from one of the other samples.

riverar (approved):

PR #3235, opened 2024-08-30
Add async `spawn` support

kennykerr:
Building on #3221 and #3213, this update concludes support for async implementations. You can now both consume and produce the four async interfaces defined by the Windows Runtime directly in Rust. 
* `IAsyncAction`
* `IAsyncOperation<T>`
* `IAsyncActionWithProgress<P>`
* `IAsyncOperationWithProgress<T, P>`

As illustrated in #3221, if you happen to have a value ready to return, you can simply use the `ready` function as follows:

```rust
fn main() -> Result<()> {
    let ready = IAsyncOperation::ready(Ok(123));

    assert_eq!(ready.get()?, 123);

    Ok(())
}
```

If you need to spin up a background thread to wait for or calculate the value to return, you can use the `spawn` function as follows:

```rust
use windows::{core::*, Foundation::*};

fn main() -> Result<()> {
    let ready = IAsyncOperation::spawn(|| {
        std::thread::sleep(std::time::Duration::from_millis(1000)); // some lengthy operation
        Ok(123)
    });

    assert_eq!(ready.get()?, 123);

    Ok(())
}
```

While the `ready` implementations return immediately, the `spawn` implementations will take the closure and pass it _by value_ to the Windows thread pool. This is far more efficient and scalable than creating a new thread each time `spawn` is called. 

I have also included extensive tests for both `ready` and `spawn` implementations in this update.


kennykerr:
Disabled the lib workflow as its proving unreliable.

Issue #3237, opened 2024-08-31
Fix lib.yml workflow


kennykerr:
The lib.yml workflow has been failing sporadically and has been disabled.

PR #3239, opened 2024-09-01
Zero hint values in .idata$6 sections for GNU libs

riverar:
Fixes: #3237

In Binutils 2.43, a long-standing quirk related to the emission of the [PE hint/name table](https://learn.microsoft.com/windows/win32/debug/pe-format#hintname-table) (.idata$6) [was reverted](https://sourceware.org/git/?p=binutils-gdb.git;a=blobdiff;f=binutils/dlltool.c;h=2c37ad3c6a3477844ec2fdc50ec0dc4c72061ca4;hp=7a3d20f0af9f3e9684a030b97905b97921a25df4;hb=057a2b4c4b57ef495dd32c9087e4af28bda41737;hpb=cbccccfdf19a388abb10579aea42c8b35c25c933), causing the hints to now be emitted with numerically increasing values. This change was detected during our library workflow pass. Since the order of functions in the upstream metadata is not guaranteed and the new hints are not valid, we are now zeroing out the hints in a post-processing step to restore consistency.

### DIFF
--- a/crates/tools/gnu/src/main.rs
+++ b/crates/tools/gnu/src/main.rs
@@ -1,6 +1,7 @@
 use std::collections::{BTreeMap, BTreeSet};
-use std::io::prelude::*;
+use std::io::{prelude::*, SeekFrom};
 use std::process::{Command, Stdio};
+use std::str::FromStr;
 
 fn main() {
     const ALL_PLATFORMS: [&str; 5] = [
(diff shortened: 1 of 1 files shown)


riverar:
Alternatively, we could toss this PR and just re-generate new libs using the current bintools and bank on upstream metadata function order not changing. (I _think_ if we move functions between namespaces in metadata, they will get emitted into the metadata tables in a slightly different order.)

kennykerr:
Happy to take a one time update to gnu libs to bring them in line and avoid a bunch of code churn.

riverar:
Sure, let's give that a shot then.

PR #3249, opened 2024-09-04
Implement `ExactSizeIterator` for registry keys and values

kennykerr:
This allows the caller to query how many keys and values there are without iterating and offers allocator optimization for `Iterator::collect()`.


kennykerr:
@riverar is considering an alternative where the registry iterators are lazy rather than eager in calculating their exact size. Will hold on this PR until that's been considered.

PR #3250, opened 2024-09-05
Update trim function in windows-registry

amone-bit:
Original trim function is iterate from back to front to trim 0x0000, but I found in some registry value is not only have one 0x0000 in the end, but also in the middle. like below:
`HKEY_CURRENT_USER\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\RecentDocs`
image
if use original trim function, I will get the string that is normal string plus unnormal string.
image
And if I use modified trim function, will work as expected.
image

### DIFF
--- a/crates/libs/registry/src/value.rs
+++ b/crates/libs/registry/src/value.rs
@@ -142,9 +142,10 @@ impl<const N: usize> From<[u8; N]> for Value {
 }
 
 fn trim(mut wide: &[u16]) -> &[u16] {
-    while wide.last() == Some(&0) {
-        wide = &wide[..wide.len() - 1];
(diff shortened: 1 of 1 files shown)


riverar:
I don't think we want to support calling `trim()` on an (undocumented) `REG_BINARY`.

amone-bit:
Sorry, I inspect again, it's my fault. And I have a question that the value type is REG_BINARY in the registry editor, actually the value is should a string represented a document name. I dont understand whether is a windows' design error. In my understanding, the value type should is REG_SZ.

kennykerr:
The purpose of `trim` is to remove trailing null characters from strings and nothing more.

Issue #3220, opened 2024-08-25
windows-bindgen: derive Eq incorrectly emitted for struct containing IReference to struct containing float


mattico:
### Summary

See the IDL below. windows-bindgen should avoid deriving `Eq` for `struct { IReference<StructWithF32> }` similar to how it avoids it for `struct { StructWithF32 }`.

### Crate manifest

```TOML
[package]
name = "test-windows-bindgen"
version = "0.1.0"
edition = "2021"

[dependencies]
windows = { version = "0.58.0", features = ["Foundation"] }
windows-core = "0.58.0"

[build-dependencies]
windows-bindgen = "0.58.0"
```

### Crate code

https://github.com/mattico/test-windows-bindgen

`src/test.idl`:
```
namespace Test
{
    struct StructWithF32
    {
        Single Value;
    };

    struct IReferenceStructWrapper
    {
        Windows.Foundation.IReference<StructWithF32> Value;
    };
}
```

`src/bindings.rs`:

```rust
// ...
#[derive(Clone, Debug, Eq, PartialEq)]
pub struct IReferenceStructWrapper {
    pub Value: Option<windows::Foundation::IReference<StructWithF32>>,
}
// ...
```

```
   Compiling test-windows-bindgen v0.1.0 (C:\Users\matt\Code\test-windows-bindgen)
error[E0277]: the trait bound `StructWithF32: std::cmp::Eq` is not satisfied
   --> src\bindings.rs:339:5
    |
337 | #[derive(Clone, Debug, Eq, PartialEq)]
    |                        -- in this derive macro expansion
338 | pub struct IReferenceStructWrapper {
339 |     pub Value: Option<windows::Foundation::IReference<StructWithF32>>,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::cmp::Eq` is not implemented for `StructWithF32`, which is required by `Option<IReference<StructWithF32>>: std::cmp::Eq`
    |
    = help: the trait `std::cmp::Eq` is implemented for `Option<T>`
    = note: required for `IReference<StructWithF32>` to implement `std::cmp::Eq`
    = note: 1 redundant requirement hidden
    = note: required for `Option<IReference<StructWithF32>>` to implement `std::cmp::Eq`
```

kennykerr:
Good catch - I need to fix that.

kennykerr:
[...]

PR #3255, opened 2024-09-05
Avoid deriving `Eq` for structs containing floating point type parameters

kennykerr:
Previously such structs were only detected when the field, or some nested type, contained a floating-point type. This update extends that to generic types. 

Fixes: #3220


mattico:
I tested this and can confirm it fixed the issue on my code, thanks!

PR #3260, opened 2024-09-05
Make registry key/value iterators lazy

riverar:
This change updates the key and value iterators to support "forward-lazy" iteration, where the underlying key may change during the iteration process. Previously, the iterators assumed a static state for the registry key but this could lead to panic if keys were modified externally during iteration. The updated iterators now account for potential changes, at the expense of some additional registry queries and/or heap activity.

### DIFF
--- a/crates/libs/registry/src/key_iterator.rs
+++ b/crates/libs/registry/src/key_iterator.rs
@@ -3,37 +3,25 @@ use super::*;
 /// An iterator of registry key names.
 pub struct KeyIterator<'a> {
     key: &'a Key,
-    range: core::ops::Range<usize>,
+    idx: u32,
     name: Vec<u16>,
 }
 
 impl<'a> KeyIterator<'a> {
     pub(crate) fn new(key: &'a Key) -> Result<Self> {
-        let mut count = 0;
-        let mut max_len = 0;
+        let info = get_key_info(key)?;
 
-        let result = unsafe {
-            RegQueryInfoKeyW(
-                key.0,
-                null_mut(),
-                null_mut(),
-                null_mut(),
-                &mut count,
-                &mut max_len,
-                null_mut(),
-                null_mut(),
-                null_mut(),
-                null_mut(),
-                null_mut(),
-                null_mut(),
-            )
-        };
-
-        win32_error(result).map(|_| Self {
+        Ok(Self {
             key,
-            range: 0..count as usize,
-            name: vec![0; max_len as usize + 1],
+            idx: 0,
+            name: vec![0; (info.subkey_name_max + 1) as usize],
         })
+    }
+
+    fn resize(&mut self) -> Result<()> {
+        let info = get_key_info(self.key)?;
+        self.name.resize((info.subkey_name_max + 1) as usize, 0);
+        Ok(())
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/libs/registry/src/key_iterator.rs
```diff
@@ -3,66 +3,59 @@ use super::*;
 /// An iterator of registry key names.
 pub struct KeyIterator<'a> {
     key: &'a Key,
-    range: core::ops::Range<usize>,
+    idx: u32,
```
kennykerr, line 6:
```suggestion
    index: u32,
```

No need to abbreviate.

crates/libs/registry/src/key_iterator.rs
```diff
@@ -3,66 +3,59 @@ use super::*;
...
-                debug_assert_eq!(result, ERROR_NO_MORE_ITEMS);
-                None
-            } else {
-                Some(String::from_utf16_lossy(&self.name[0..len as usize]))
+        if result == ERROR_MORE_DATA {
+            if self.resize().is_err() {
+                return None;
             }
-        })
+            return self.next();
```
kennykerr, line 50: This is recursive - not necessarily bad in this scenario but worth calling out if it's not obvious.

crates/libs/registry/src/value_iterator.rs
```diff
@@ -3,80 +3,72 @@ use super::*;
...
+            data: Data::new(info.value_data_max as usize),
         })
     }
+
+    fn resize(&mut self) -> Result<()> {
+        let info = get_key_info(self.key)?;
+        self.name.resize((info.value_name_max + 1) as usize, 0);
+        self.data = Data::new(info.value_data_max as usize);
+        Ok(())
+    }
```
kennykerr, line 28: Not a big deal but might be nice to default initialize and then call `resize` in the `new` method to avoid the redundant initialization here and in the other iterator.

crates/tests/misc/registry/tests/keys.rs
```diff
@@ -14,6 +14,17 @@ fn keys() -> Result<()> {
     let names: Vec<String> = key.keys()?.collect();
     assert_eq!(names, ["one", "three", "two"]);
 
+    let mut names = Vec::<String>::new();
+    let iterator = key.keys()?;
+    for name in iterator {
+        if name == "one" {
+            key.remove_tree("three")?;
+            key.create("seventy")?;
```
kennykerr, line 22: Maybe make it obvious that this requires a larger buffer...

crates/tests/misc/registry/tests/values.rs
```diff
@@ -22,5 +22,23 @@ fn values() -> Result<()> {
         ]
     );
 
+    let mut names = Vec::<(String, Value)>::new();
+    let iterator = key.values()?;
+    for (name, value) in iterator {
+        if name == "string" {
+            key.set_string("string-two", "hello world two")?;
```
kennykerr, line 29: Same here.

crates/libs/registry/src/key_iterator.rs
riverar: Changed.

riverar:
After more testing, I've identified a critical issue with key enumeration. Adding a subkey during this process triggers immediate lexical sorting, which can re-order the subkeys mid-enumeration. Since modification of the key isn't supported during enumeration, it's not possible to safely perform the lazy initialization proposed here. Therefore, I'm closing this PR.

PR #3266, opened 2024-09-09
Remove "riddle" and metadata generation

kennykerr:
In an attempt to streamline and focus on what's most valuable, I'm making the difficult decision to drop support for metadata generation in general. While the existing metadata generation tooling is not ideal for Rust developers (e.g. midlrt and win32metadata), I don't realistically have the bandwidth to provide a Rust version any time soon. I would rather spend what effort I can on what the `windows-rs` project does best which is to support generation of Rust bindings and type system support. The benefit in removing this code is that I can then streamline the implementation of `windows-bindgen` and more rapidly improve type system and authoring support without being encumbered by the future possibility of metadata generation. Perhaps in future I'll have the bandwidth to tackle this again, but it doesn't seem likely in the short term. 

Fixes: #1093


ChrisDenton (commented):
It's unfortunate but completely understandable. I'd much rather the focus be on what's most important right now than spread your efforts too thin.

Do we need one last update to the riddle crate that notes that it's no longer being updated?

PR #3279, opened 2024-09-12
Exclude `web` on most workflows

kennykerr:
All those dependabot PRs don't need to run most of the workflows if they're just updating web hosting dependencies.


riverar (approved):

riverar:
Good call, ha!

PR #3281, opened 2024-09-13
fix: types of SetProcessWorkingSetSize's parameters should be isize

csmoe:
See https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-setprocessworkingsetsize the parameters can be `-1`
> If both dwMinimumWorkingSetSize and dwMaximumWorkingSetSize have the value (SIZE_T)–1, the function removes as many pages as possible from the working set of the specified process.


csmoe:

ChrisDenton:
Thank you but these functions are generated from [win32metada](https://github.com/microsoft/win32metadata) so an issue would need to be opened on that repository describing the problem.

riverar:
`SIZE_T` is a `ULONG_PTR`, emphasis on _unsigned_, so is not likely to get changed on the metadata side. No bug here.

PR #3282, opened 2024-09-13
Move `VARIANT` support to the `windows` crate

kennykerr:
#2786 added richer support for `VARIANT` and `PROPVARIANT` and at the time the only practical way to do that was adding it to the `windows-core` crate. But that meant that everyone using the `windows-core` crate carried around `VARIANT` support even though that's really only meant for some very old Windows APIs in the `windows` crate. Now that #3110 added a simpler way to handle type extensions, we can more naturally provide this same functionality directly in the `windows` crate, unburdening the `windows-core` crate and simplifying the `windows-metadata` and `windows-bindgen` crates as well with less "special" type support.


kennykerr:
It's tricky because `VARIANT` and `PROPVARIANT` are in different Win32 metadata namespaces. I might have to break the dependency apart and implement them separately.

ChrisDenton:
This might be a dumb question but do they need to be in a separate namespace? Or is this something that could reasonably be changed?

kennykerr:
That would greatly simplify this. Arguably they should just 
 be together as the two are interdependent.

kennykerr:
It's far from perfect but greatly simplifies `windows-core` and `windows-bindgen` and lets me keep pursuing improvements to code generation so it's a step in the right direction.

Issue #3285, opened 2024-09-19
Should raw-dylibs defined in windows-targets use import_name_type = "undecorated"?


glandium:
### Summary

In a complex multi-language project, one may end up using both windows-target and the system libraries directly (that's the case of Firefox). When enabling windows-targets's support for raw-dylib, linking fails when using LLVM's lld-link (I haven't tried link.exe, it might actually work, but the result is probably suboptimal too). The cause for the link failure is possibly a bug in lld-link, but it is a manifestation of the fact that the fake import lib that rust creates with raw-dylib does not agree with the real system import lib, wrt symbol names: on 32-bits windows, the system import lib use the `__imp__sym@n` form, while the raw-dylib uses the `__imp_sym` form because of `import_name_type = "undecorated"`.

See https://github.com/llvm/llvm-project/issues/107371#issuecomment-2359947863 for more details.

Removing `import_name_type = "undecorated"` solves the problem.

glandium:
I guess the discussion on the LLVM side points to the problem being on the rust side of raw-dylib.

kennykerr:
@dpaoliello thoughts?

kennykerr:
Anyone know how we might add this...

https://github.com/microsoft/windows-rs/blob/ff6f56df6c9d727c348b3c4cd6d45db59f76fa1b/.github/workflows/raw_dylib.yml#L12-L13

...to a yml strategy/matrix so I can more easily run `raw-dylib` tests across the board? Happy to expand test coverage if there's a gap here.

ChrisDenton:
I did check with `1.71` and it has the same behaviour.

kennykerr:
#3287 passed meaning we have the same testing for `raw-dylib` as non-`raw-dylib`.

kennykerr:
Happy to test rust-lld / lld-link if someone can point me to how that is done. 

I see its enabled by default on Linux but not Windows. Anyone know why?

https://blog.rust-lang.org/2024/05/17/enabling-rust-lld-on-linux.html

ChrisDenton:
There have been compatibility issues in the past. I'm not exactly sure where we are at now but it seems https://github.com/rust-lang/rust/issues/111480 is still open (though might just be that nobody has tested it recently).

ChrisDenton:
> Happy to test rust-lld / lld-link if someone can point me to how that is done.

You can put this in `config.toml`:

```rust
[target.x86_64-pc-windows-msvc]
[...]

PR #3287, opened 2024-09-19
Expand `raw-dylib` testing

kennykerr:
In response to #3285 I thought I'd expand `raw-dylib` testing. Initially this was only tested on nightly Rust but it has since been stabilized so we can expand test coverage.


kennykerr:
Tests are failing because C++ doesn't support `raw-dylib`... 🙂

Switching C++ interop tests to use SDK libs.

PR #3291, opened 2024-09-23
Add `Deref` implementation for `HSTRING`

kennykerr:
The `HSTRING` type represents a UTF-16 string that is modeled as a `[u16]` slice in Rust. This update replaces the `as_wide` method to retrieve the `HSTRING` as a slice with a `Deref` implementation. This makes it a lot easier to work with an `HSTRING` in Rust as conversion is implicit in many cases. 

The same is done for `BSTR` for consistency and code reduction. 

Fixes: https://github.com/rust-lang/rustup/pull/3896#issuecomment-2366820680

### DIFF
--- a/crates/libs/result/src/bstr.rs
+++ b/crates/libs/result/src/bstr.rs
@@ -1,36 +1,26 @@
 use super::*;
+use core::ops::Deref;
 
 #[repr(transparent)]
 pub struct BasicString(*const u16);
 
-impl BasicString {
-    pub fn is_empty(&self) -> bool {
-        self.len() == 0
-    }
+impl Deref for BasicString {
+    type Target = [u16];
 
-    pub fn len(&self) -> usize {
-        if self.0.is_null() {
+    fn deref(&self) -> &[u16] {
+        let len = if self.0.is_null() {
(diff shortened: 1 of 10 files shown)


kennykerr:
Build break is due to unrelated Rust nightly update - investigating.

djc:
Nice!

crates/libs/result/src/bstr.rs
```diff
@@ -1,36 +1,26 @@
...
-impl BasicString {
-    pub fn is_empty(&self) -> bool {
-        self.len() == 0
-    }
+impl Deref for BasicString {
+    type Target = [u16];
 
-    pub fn len(&self) -> usize {
-        if self.0.is_null() {
+    fn deref(&self) -> &[u16] {
```
tim-weis, line 10: What's the reason for this signature rather than its more conventional spelling

```rust
fn deref(&self) -> &Self::Target
```

?

I'm just curious if there's a particular reason.

kennykerr: No reason, just seemed more succinct in this case.

Issue #3300, opened 2024-09-28
`/web` changes are triggering all build workflows


kennykerr:
The #3279 change doesn't appear to work as expected as you can see in #3299.

PR #3312, opened 2024-10-05
Update workflows to ignore paths on pull request

riverar:
Updates all our workflows to ignore the same paths across both push and pull request.

Fixes: #3300 ([example](https://github.com/riverar/windows-rs/pull/3))


kennykerr (approved):
Sweet thanks!

Issue #3315, opened 2024-10-09
`no_std` support for `windows-rs`


sagacity:
### Suggestion

I am working in an environment in which I am quite executable size constrained, but I'd still like to use Direct3D11 in order to perform some rendering tasks. Since `windows-sys` has no support for this, I am currently using the `winapi` crate to fill in support for the relevant D3D COM classes I'm using. However, `winapi` has been unmaintained for three years after `windows-rs` ate its lunch :)

I've tried using `windows-bindgen` to generate bindings and patching them up manually, but I couldn't get the generated code to compile properly (after changing things like `std::ptr::null_mut` to `core::ptr::null_mut` and so on). Could this be a more or less valid path to take, though?

What is a good way forward, generally? I wouldn't mind using bindings that are generated by using a manual bindgen step, but ideally I'd just use the `windows` crate since that would also make all other non-COM code simpler and there wouldn't be a manual bindgen step.

(btw, I realize this issue has been brought up before, so feel free to close if this cannot be realized, but I'd like some guidance on how to avoid depending on an unmaintained crate)

kennykerr:
Good is, the `windows` crate can now be compiled with `no_std`. #3047

Just make sure to specify `default-features = false` as this test illustrates:

https://github.com/microsoft/windows-rs/tree/master/crates/tests/misc/no_std

sagacity:
Fair enough! I'd be happy to take a stab at it since most of it is mechanical conversion, but since testing is indeed difficult I'm not sure I'd end up accidentally breaking stuff.

PR #3317, opened 2024-10-09
Fix remaining `std` references in `windows` and `windows-core` crates for `no_std` builds

kennykerr:
I'll think some more about how to test this, but at least for now this should deal with this issue. 

Fixes: #3315

This also changes the Direct3D12 sample to build with `no_std` for validation and illustration purposes.


sagacity:
I can confirm this branch works fine and I can create a Direct3D device, create vertex buffers, etc. Thanks for the quick response!

kennykerr:
Sweet - thanks for checking!

PR #3321, opened 2024-10-11
Fix nested struct sort order 

kennykerr:
While working on the next version of `windows-bindgen` I noticed that the current version incorrectly sorts nested structs such that the mangled names (since Rust lacks support for nested types) are based on an internal sort order rather than the sort order provided by ECMA-335. 

This usually isn't too noticeable, but some structs like `FLT_PARAMETERS` are now far more sensible. 

I'm fixing it here in the current version to simplify diff validation of the next version of `windows-bindgen`.

### DIFF
--- a/crates/libs/sys/src/Windows/Wdk/NetworkManagement/Ndis/mod.rs
+++ b/crates/libs/sys/src/Windows/Wdk/NetworkManagement/Ndis/mod.rs
@@ -2867,14 +2867,9 @@ pub struct NDIS_INTERRUPT_MODERATION_PARAMETERS {
 #[repr(C)]
 #[derive(Clone, Copy)]
 pub struct NDIS_IPSEC_OFFLOAD_V1 {
-    pub Supported: NDIS_IPSEC_OFFLOAD_V1_2,
-    pub IPv4AH: NDIS_IPSEC_OFFLOAD_V1_0,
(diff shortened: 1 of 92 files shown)


crates/libs/sys/src/Windows/Wdk/NetworkManagement/Ndis/mod.rs
```diff
@@ -3258,32 +3258,32 @@ pub struct NDIS_TCP_CONNECTION_OFFLOAD {
...
     pub Encapsulation: u32,
     pub _bitfield: u32,
 }
 #[repr(C)]
 #[derive(Clone, Copy)]
-pub struct NDIS_TCP_IP_CHECKSUM_OFFLOAD_1 {
+pub struct NDIS_TCP_IP_CHECKSUM_OFFLOAD_0 {
     pub Encapsulation: u32,
     pub _bitfield: u32,
 }
```
riverar, line 3277: I'm concerned about the potential compatibility impact of these struct name changes. Since some structs have identical form, developers may not get compile-time errors, making it easy for this change to go unnoticed in existing codebases. Addressing this and not annoying folks may be quite tricky, hmm...

kennykerr: Fortunately, the nested structs that are identical will not cause any problems. In other cases, you get a compiler error which is relatively simple to follow.

PR #3342, opened 2024-11-14
Update Windows metadata

riverar:
Updates Windows.Win32, Windows.Wdk, and Windows metadata.


crates/targets/baseline/icu.dll.c
riverar, line 1: Just a heads up, breaking lib change here as a result of ICU API remapping for down-level support. https://github.com/microsoft/win32metadata/pull/1958.

riverar:
This is ready for your review @kennykerr, no rush.

crates/libs/bindgen/src/metadata.rs
```diff
@@ -240,7 +240,7 @@ impl Signature {
         }
 
         // Then we see if we can infer retval-like behavior more conservatively.
-        self.params.last().map_or(false, |param| param.is_retval())
+        self.params.last().is_some_and(|param| param.is_retval())
```
kennykerr, line 243: Clippy?

crates/targets/baseline/icu.dll.c
kennykerr: Actually, we should see if there are actually any omitted APIs - if they've just moved to a more compatible library then that may not be a breaking change. So long as none have been removed.

crates/libs/bindgen/src/metadata.rs
riverar: Yep, can pull this out if you'd prefer to address in a standalone PR.

PR #3344, opened 2024-11-14
Update web workflow

riverar:
Update web browser metadata and restricts the workflow to web folder changes.


kennykerr (approved):

riverar:
Works around regression introduced in node 23.2, will revisit when a patch is available. https://github.com/nodejs/node/issues/55826.

PR #3354, opened 2024-11-27
build test - ignore

kennykerr:
This is a smoke test of the next major version of the `windows-bindgen` crate - this PR is just for validation and will be closed - look for a subsequent PR with a complete description of new features and changes.


kennykerr:
Hey @ChrisDenton I added some panic tests to the latest version of `windows-bindgen` and I'm noticing that `std::fs::create_dir_all` and `std::fs::write` sometimes fail silently. Might I be missing something?

https://github.com/microsoft/windows-rs/actions/runs/12060811636/job/33631839763?pr=3354

Both tests that fail to fail.. are not failing inside `write_to_file` here:

https://github.com/microsoft/windows-rs/blob/vnext/crates/libs/bindgen/src/io.rs

ChrisDenton:
Hm, I've not looked at it closely yet but could those tests actually succeed in CI? Because it's running with admin permissions it could succeed writing in windir.

kennykerr:
Is it? I also noticed it silently failing locally with the i686 toolchain although sporadically.

kennykerr:
I'll confirm one way or the other.

ChrisDenton:
Quick thought: i686 might be affected by wow64 path redirection.

ChrisDenton:
Quick test

```rust
pub fn write_to_file<C: AsRef<[u8]>>(path: &str, contents: C) {
    if let Some(parent) = std::path::Path::new(path).parent() {
        std::fs::create_dir_all(parent)
            .unwrap_or_else(|_| panic!("failed to create directory `{path}`"));
    }
    std::fs::write(path, contents).unwrap_or_else(|_| panic!("failed to write file `{path}`"));

    // print the canonical path
    let p = std::fs::canonicalize(path).unwrap();
    println!("{}", p.display());
}
```

For i686 it outputs: `\\?\C:\Users\Chris\AppData\Local\VirtualStore\Windows\output.log`

kennykerr:
OK I'll find a more reliable failure path. 🫠

kennykerr:
ok everything is green except for the lib tests 👍

PR #3356, opened 2024-11-28
fix: remove use of std in windows-strings h! macro

vthib:
This crate can be used in no_std, but some use of std were left in the h! macro. They can be trivially converted to core.

On this subject, i've been using the windows crates in no_std setups, and this is the only issue I reached, the rest seems to work great.

However, the windows 0.58 does not have all the no_std fixes, is there a release planned soon to be able to get those changes?


kennykerr (approved):

kennykerr:
Thanks for the fix! The crates are only published on demand.

Issue #3339, opened 2024-11-09
Reg-free activation should only kick in for `REGDB_E_CLASSNOTREG`


kennykerr:
Other error codes should cause activation to fail.

PR #3365, opened 2024-12-09
Harden reg-free class activation

kennykerr:
Reg-free activation should only be attempted if the class is not registered.
It should not be attempted if the class is registered but fails to activate.

Fixes: #3339


riverar:
What about the `E_NOINTERFACE` scenario?

riverar:
After reflecting on it more ☕... I suppose if I ask you to grab the meat from the fridge and you check, only to find it empty, that should conclude the task. I wouldn't want you searching elsewhere or making assumptions about where else it might be. Looks good to me.

kennykerr:
Yes, this ensures that it does not accidentally "fall through" and attempt reg-free when activation should legitimately fail for one reason or another.

Issue #3335, opened 2024-10-31
`windows-bindgen` should generate `no_std` bindings by default


mwcampbell:
Since the windows crate recently gained support for using no_std while still supporting alloc, it would be useful to have a complete example of an executable or DLL built this way. I read in #3047 that some low-level components, including in Windows itself, use the crate this way. I understand if this is an impractical request because all the real examples are too tightly coupled to Windows's build system or other proprietary bits.

kennykerr:
While we can probably add/update a sample, there's not really much to illustrate here beyond adding `#![no_std]` to your `lib.rs` or `main.rs` and taking it for a spin. The compiler should quickly point out what you might be missing.

kennykerr:
Actually, I attempted to port a sample to use `no_std` and ran into a few minor issues so I'll use this issue to track those and then make sure we have test coverage and a working sample.

kennykerr:
Note to self:  `sample_direct2d` is a good test case for this.

mwcampbell:
What I had in mind for a complete no_std sample was things like setting up a panic handler, allocator (if you've actually figured out a way to use no_std with alloc in user space), and any other bits that are required for non-trivial no_std projects on Windows.

kennykerr:
If you mean the `alloc` crate then you can find various examples by searching this repo. It is used for allocation support under `no_std`. 

If you mean a custom allocator then you may want to look at https://github.com/microsoft/windows-drivers-rs for examples.

kennykerr:
Yes, this issue was mostly resolved by #3359 but I'll open again to address some remaining issues and update the sample.

kennykerr:
Thanks for checking! 🙂

kennykerr:
@mwcampbell I decided against updating the sample as it butched the sample undesirably, but I fixed the underlying code gen and added tests to ensure that everything should build with `no_std` if so desired. 

Please take #3366 for a spin and let me know if you run into any problems.

PR #3366, opened 2024-12-09
`windows-bindgen` should generate `no_std` bindings by default

kennykerr:
This handles a few cases where it assumed `std` when it shouldn't. This is now also more fully tested in `no_std` builds. 

Fixes: #3335


crates/tests/misc/no_std/Cargo.toml
```diff
@@ -37,6 +37,10 @@ default-features = false
 [dependencies.windows]
 path = "../../../libs/windows"
 default-features = false
+features = [
+    "Foundation_Collections",
+    "Win32_Graphics_Direct3D",
+]
```
kennykerr, line 43: This ensures that all the updated code gen is validated in a `no_std` build.

crates/libs/windows/src/extensions/Foundation/Numerics/Vector2.rs
```diff
@@ -22,15 +22,18 @@ impl Vector2 {
     pub fn length_squared(&self) -> f32 {
         self.dot(self)
     }
+    #[cfg(feature = "std")]
```
kennykerr, line 25: `sqrt` is only available in `std` library.

crates/libs/windows/src/extensions/Foundation.rs
```diff
@@ -1,7 +1,10 @@
+#[cfg(feature = "std")]
 pub mod Async;
+#[cfg(feature = "std")]
 pub mod AsyncReady;
+#[cfg(feature = "std")]
 pub mod AsyncSpawn;
-#[cfg(feature = "Foundation_Collections")]
+#[cfg(all(feature = "std", feature = "Foundation_Collections"))]
 pub mod Collections;
```
kennykerr, line 8: All of the async and collection helpers required `std`.

PR #3367, opened 2024-12-09
Simpler bindings generation

kennykerr:
Thanks to #3359, bindings for Win32 APIs can mostly ignore the artificial namespaces imposed by the metadata format. This simplifies the maintenance of binding generation scripts. 

All I've done here is (1) remove the explicit namespace for types and (2) sorted alphabetically.


kennykerr:
@ChrisDenton just fyi - no need to do this for rust-lang but its an option if you find its simpler to maintain.

ChrisDenton:
It does look much simpler both for contributors and maintainers. Not least because it won't break if something is moved to another namespace.

kennykerr:
Yep, that's a good point.

Issue #3320, opened 2024-10-11
`windows-bindgen` should avoid `Param` traits as much as possible


kennykerr:
The `Param` trait provide the convertibility necessary for many APIs but the metadata, particularly for Win32, don't make it clear what those APIs are so the bindings tend to overuse the `Param` trait "just in case". This causes a lot of confusion especially for folks new to Rust or Windows. A common mistake is to pass a raw pointer value when a `HANDLE` is expected. The Rust compiler error would be very straightforward if not for the `Param` trait.

```rust
fn main() {
    let handle = std::ptr::null_mut();
    windows::Wdk::Foundation::NtClose(handle);
}
```

In this case, the compiler error message is very unhelpful.

```
D:\git\sample>cargo check -p hello_world
    Checking hello_world v0.1.0 (D:\git\sample\crates\hello_world)
error[E0277]: the trait bound `*mut _: Param<HANDLE, CopyType>` is not satisfied
 --> crates\hello_world\src/main.rs:3:39
  |
3 |     windows::Wdk::Foundation::NtClose(handle);
  |     --------------------------------- ^^^^^^ the trait `CanInto<HANDLE>` is not implemented for `*mut _`, which is required by `*mut _: Param<HANDLE, CopyType>`
  |     |
  |     required by a bound introduced by this call
  |
  = help: the following other types implement trait `CanInto<T>`:
            `HINSTANCE` implements `CanInto<HMODULE>`
            `HMODULE` implements `CanInto<HINSTANCE>`
            `HWND` implements `CanInto<HANDLE>`
            `IAgileObject` implements `CanInto<IUnknown>`
            `IAgileReference` implements `CanInto<IUnknown>`
            `IGenericFactory` implements `CanInto<IInspectable>`
            `IGenericFactory` implements `CanInto<IUnknown>`
            `IInspectable` implements `CanInto<IUnknown>`
          and 2 others
  = note: required for `*mut _` to implement `Param<HANDLE, CopyType>`
note: required by a bound in `NtClose`
 --> C:\Users\kekerr\.cargo\registry\src\index.crates.io-6f17d22bba15001f\windows-0.58.0\src\Windows\Wdk\Foundation\mod.rs:4:9
  |
2 | pub unsafe fn NtClose<P0>(handle: P0) -> super::super::Win32::Foundation::NTSTATUS
  |               ------- required by a bound in this function
3 | where
4 |     P0: windows_core::Param<super::super::Win32::Foundation::HANDLE>,
[...]

PR #3368, opened 2024-12-09
Prefer optional over convertible parameters

kennykerr:
Thanks to #3359, it's now a lot easier to reason about some of these more nuanced type scenarios. The `windows-bindgen` crate has traditionally overused the `Param` trait to provide cover for various weird Windows API-isms but the `Param` trait tends to be quite confusing when you don't get the parameters just right and in particular for APIs that don't expect polymorphism. In some cases its necessary - specifically for COM or WinRT polymorphism - but in particular it was previously used for all Win32 handle types which was almost always unnecessary. 

This update does two things:

* Avoid applying the `Param` trait for handle types
* Detect more optional parameters and use `Option<T>` for copyable parameter types

When combined this limits the fallout from the first and generally makes the API surface more coherent as optional parameters are actually optional and the vast majority of handle types, which aren't convertible, are simple type parameters. 

In some cases where "handle polymorphism" is actually expected by some API you can simply use the `Into` or `From` traits to perform the conversion. 

Fixes: #3320


liqh99em:
The param "lptimerfunc" type "TIMERPROC" of SetTimer is option itself,   I think it is repeated.

Issue #3391, opened 2024-12-13
`windows-bindgen` should support Rust edition 2024


kennykerr:
There appear to be a bunch of new warnings to deal with...

kennykerr:
I really don't want to switch the edition - the last edition update was brutal - but hopefully we can find a way to at least test this and future-proof windows-rs for the inevitable. 🙂

kennykerr:
This is the most vocal warning. 

https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html

I can probably begin validation against this directly using:

```toml
[workspace.lints.rust]
unsafe_op_in_unsafe_fn = "warn"
```

Also `cargo fix --edition` helps but is a little over eager.

kennykerr:
Also, Rust edition 2024 requires Rust 1.85 as the MSRV so we won't be able to actually switch any time soon.

kennykerr:
Testing locally, #3393 appears to unblock adoption of `windows-rs` crates and code gen for use with Rust 2024 without any warnings. I'll leave windows-rs itself on edition 2021 for now but I think this issue can be closed with #3393.

PR #3393, opened 2024-12-13
Update `windows-bindgen` to support `unsafe_op_in_unsafe_fn`

kennykerr:
This gets us a long way toward supporting Rust edition 2024. This does require an MSRV bump though. So I've just bumped all my 1.60 crates to 1.74 with the exception of `windows-sys` which remains on 1.60 for now. 

Fixes: #3391


ChrisDenton:
Alternatively you do [`#![deny(unsafe_op_in_unsafe_fn)]`](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unsafe-op-in-unsafe-fn), which should work for 1.52+. But bumping the MSRV is admittedly simpler 🙂.

kennykerr:
Thanks, good to know. Unless there's a good reason to stick with an older MSRV I'd rather move things along especially since 1.74 is already over a year old.

ChrisDenton:
Very fair!

Issue #3233, opened 2024-08-29
Consolidate `Param`/`Ref`/`InterfaceRef` etc


kennykerr:
A number of conversion types and traits have been developed to solve different problems. Some like `Param` are used by `windows-bindgen` while others are more suitable for hand-authored APIs. It's just a bit confusing and hard to know what the right tool is where there are so many options.

kennykerr:
We should try and use Ref for interface traits and delegate closures. Its way simpler.

kennykerr:
Ideally we could have an `AsRef` that worked something like this:

```rust
pub trait AsRef<T> where T: Type<T> {
    fn as_ref<'a>(&'a self) -> Ref<'a, T>;
}

impl<T: Type<T>> AsRef<T> for T {
    fn as_ref<'a>(&'a self) -> Ref<'a, T> {
        unsafe { transmute_copy(self) }
    }
}

impl<'a, T: Type<T>> AsRef<T> for &'a T {
    fn as_ref(&self) -> Ref<'a, T> {
        unsafe { transmute_copy(self) }
    }
}

impl<'a, T: Type<T>> AsRef<T> for Option<&'a T> {
    fn as_ref(&self) -> Ref<'a, T> {
        unsafe {
            match self {
                Some(value) => transmute_copy(value),
                None => Ref::zeroed(),
            }
        }
    }
}
```

This would then provide more uniform parameter binding across regular and generic parameters and be far simpler in general. 

As noted in #3394 this doesn't support polymorphism the way that `Param` does though.

kennykerr:
The ability to use a hidden lifetime parameter `elided_lifetimes_in_paths` that makes using `Ref<T>` convenient is being deprecated by Rust, which provides further urgency to simplify and streamline this.

kennykerr:
Another tidbit is that `elided-lifetimes-in-paths` is included with the `rust-2018-idioms` lint group but not later groups like `rust-2021-compatibility` and `rust-2024-compatibility` making me question whether its worth supporting. 

https://doc.rust-lang.org/beta/rustc/lints/groups.html

PR #3394, opened 2024-12-13
Make `Ref` work with more than just interface types

kennykerr:
On the way to fixing #3233, much like #3386, this makes `Ref` work with all Windows types. Technically it already worked with all types but you could only `deref` a value. I had hoped to also land an `AsRef` to replace `Param` but that's not quite there yet. Mainly because `Param` supports polymorphism. Still, this is a step in the right direction and probably all I'll get to before the Christmas break.


crates/tests/winrt/reference/src/lib.rs
```diff
@@ -7,7 +7,7 @@ unsafe extern "system" fn DllGetActivationFactory(
     name: Ref<HSTRING>,
     factory: OutRef<IActivationFactory>,
 ) -> HRESULT {
-    if *name == "test_reference.Reference" {
+    if name.unwrap() == "test_reference.Reference" {
```
kennykerr, line 10: `Ref` still provides `Deref` so this isn't strictly necessary but illustrates using `Ref` methods directly.

PR #3402, opened 2024-12-22
Detect unsupported array parameters

### LINKED ISSUE #3384: All *GetIDsOfNames COM methods have incorrect return types/signatures
### Summary

This issue was previously reported in [#2141](https://github.com/microsoft/windows-rs/issues/2141) and fixed in [#2178](https://github.com/microsoft/windows-rs/pull/2178), but it seems to have been reintroduced with the recent big update to windows-bindgen.

### Crate manifest

```TOML

```

### Crate code

```Rust

```

iancormac84:
This pull request fixes an issue where windows-bindgen confuses an unsupported trailing array parameter for a result value and then generates code that dereferences it and tries to return it.

@kennykerr I believe I have tracked down the source of the reintroduction of the bug that had been reported prior. You mentioned providing a minimal repro in #3384. I added a generated test in the code to ensure that it's working as expected. Please let me know if this is what you expected, and what else, if anything, that I need to do to get this pull request accepted. Thanks!

Fixes #3384

### DIFF
--- a/crates/libs/bindgen/src/types/cpp_method.rs
+++ b/crates/libs/bindgen/src/types/cpp_method.rs
@@ -825,7 +825,7 @@ fn is_param_retval(ty: &Type, param: Param, hint: ParamHint) -> bool {
     {
         return false;
     }
-    if hint.is_array() {
+    if ParamHint::from_param(param).is_array() {
         return false;
     }
 
--- a/crates/libs/bindgen/src/types/cpp_method.rs
+++ b/crates/libs/bindgen/src/types/cpp_method.rs
@@ -825,6 +825,8 @@ fn is_param_retval(ty: &Type, param: Param, hint: ParamHint) -> bool {
     {
         return false;
     }
+    // This is reevaluated to detect unsupported array parameters.
+    // https://github.com/microsoft/windows-rs/issues/3384
     if ParamHint::from_param(param).is_array() {
         return false;
     }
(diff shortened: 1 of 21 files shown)


iancormac84:

kennykerr:
Oh, this looks like it impacts parameters that are attributed as array parameters but that are then discarded as invalid by `windows-bindgen`.

crates/libs/bindgen/src/types/cpp_method.rs
```diff
@@ -825,7 +825,7 @@ fn is_param_retval(ty: &Type, param: Param, hint: ParamHint) -> bool {
     {
         return false;
     }
-    if hint.is_array() {
+    if ParamHint::from_param(param).is_array() {
```
kennykerr, line 828:
```suggestion
    // This is reevaluated to detect unsupported array parameters.
    // https://github.com/microsoft/windows-rs/issues/3384
    if ParamHint::from_param(param).is_array() {
```

kennykerr:
Builds are failing with latest Rust toolchain following the holidays - I'll work on that separately and then we can update these PRs.

crates/libs/bindgen/src/types/cpp_method.rs
kennykerr: Ideally we'd have a formal test beyond code gen but at least this will give future maintainers a hint that something is happening here.

kennykerr (approved):

kennykerr (approved):

Issue #3406, opened 2024-12-28
[DirectX12 Sample] Inconsistent rendering behavior between debug and release builds


cx20:
## Overview
This issue is regarding the DirectX12 triangle sample located at:
[`crates/samples/windows/direct3d12/src/main.rs`](https://github.com/microsoft/windows-rs/blob/master/crates/samples/windows/direct3d12/src/main.rs)

## Issue Description
The triangle sample shows different behavior between debug and release builds:
- Debug build: Triangle renders correctly
- Release build: Triangle doesn't render when the rendering logic relies on WM_PAINT messages

## Reproduction Steps
1. Clone the repository
2. Build and run the DirectX12 triangle sample in debug mode

```
cargo run
```

- Triangle renders correctly
3. Build and run in release mode

```
cargo run --release
```

-  The window appears but remains blank (white background) with no triangle visible

## Root Cause
The current implementation relies on WM_PAINT messages for rendering:
```rust
fn sample_wndproc<S: DXSample>(sample: &mut S, message: u32, wparam: WPARAM) -> bool {
 match message {
     WM_PAINT => {
         sample.update();
         sample.render();
         true
     }
     ...
 }
}
```

This approach may not trigger frequently enough in release builds due to optimization, leading to inconsistent rendering behavior.

## Suggested Fix

Implement continuous rendering in the main message loop instead of relying on WM_PAINT:
```rust
loop {
    let mut message = MSG::default();
    if unsafe { PeekMessageA(&mut message, None, 0, 0, PM_REMOVE) }.into() {
        unsafe {
            _ = TranslateMessage(&message);
            DispatchMessageA(&message);
        }
        if message.message == WM_QUIT {
            break;
        }
    } else {
        // Render when no messages are available
        sample.update();
        sample.render();
    }
}
```

## Environment

Rust Version: rustc 1.85.0-nightly (dd84b7d5e 2024-12-27)
Windows Version: Windows 11 Version 24H2 (OS Build 26100.2605)
Graphics Card: NVIDIA GeForce RTX 2060
Driver Version: GeForce Game Ready Driver 555.97

### Crate manifest

```TOML
https://github.com/microsoft/windows-rs/blob/master/crates/samples/windows/direct3d12/Cargo.toml
```

### Crate code

```Rust
https://github.com/microsoft/windows-rs/blob/master/crates/samples/windows/direct3d12/src/main.rs
```

kennykerr:
[...]

PR #3419, opened 2025-01-06
Fix provenance in direct32 sample

ChrisDenton:
Using transmute creates a pointer without [provenance](https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance). Using an `as` cast instead fixes this.

Fixes: #3406

### DIFF
--- a/crates/samples/windows/direct3d12/src/main.rs
+++ b/crates/samples/windows/direct3d12/src/main.rs
@@ -5,8 +5,6 @@ use windows::{
     Win32::UI::WindowsAndMessaging::*,
 };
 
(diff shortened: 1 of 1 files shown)


damyanp (approved):

ChrisDenton:
Note: Ideally we'd use [`ptr::with_exposed_provenance`](https://doc.rust-lang.org/nightly/std/ptr/fn.with_exposed_provenance.html) to make it more explicit but that won't stabilize until 1.84.

kennykerr (approved):
Thanks!

PR #3421, opened 2025-01-07
Avoid `transmute` where possible

kennykerr:
@ChrisDenton's fix in #3419 reminded me that I have some work to do to remove more of the unnecessary `transmute` calls in generated code. This is just some low-hanging fruit but I have some more work to do to avoid the `as` cast entirely in cases where the parameter and ABI types are identical. With the overhaul in #3359 this is quickly getting easier.


ChrisDenton (approved):
Nice! It's also great that you're still finding ways to reduce the code size.

PR #3422, opened 2025-01-07
Improve search UX, add dark mode, and update dependencies

riverar:
Changes:
- Added dark mode with a theme switcher
- Better sorting: match position + suffix length
- Search terms now added to URL for easy sharing
- Branch switching auto-updates search results
- Improved search result view on mobile
- Updated Node.js dependency


riverar:
@tim-weis Heads up Tim, added a few of the items you requested! (e.g., better result sort)

PR #3427, opened 2025-01-08
Simplify internal `windows-bindgen` caching

kennykerr:
The `windows-bindgen` crate uses an internal static winmd reader but passing it around constantly is rather cumbersome. This simplifies this by using a static `AtomicPtr` that can be retrieved via the `reader` free function. This then helps to simplify some upcoming refactoring and bug fixing. 

I ended up doing the same for `Config` which is similarly passed around and describes what and how to generate bindings from metadata.


kennykerr:
Seems like spurious IO failures are sporadically failing tests.

PR #3428, opened 2025-01-09
Use OS temp directory instead of CARGO_TARGET_TMPDIR in bindgen tests

riverar:
In my testing, I found that `CARGO_TARGET_TMPDIR` is not guaranteed to point to an existing directory, contrary to what is stated in the [Cargo documentation](https://doc.rust-lang.org/cargo/reference/environment-variables.html). To address this, and stabilize the bindgen tests, I replaced references to `CARGO_TARGET_TMPDIR` with calls to `std::env::temp_dir()`.

### DIFF
--- a/crates/tests/bindgen/tests/panic.rs
+++ b/crates/tests/bindgen/tests/panic.rs
@@ -138,22 +138,20 @@ fn subset_namespace() {
 #[test]
 #[should_panic(expected = "failed to create directory")]
 fn failed_to_create_directory() {
-    let test_path = format!(
-        "{}\\failed_to_create_directory",
-        env!("CARGO_TARGET_TMPDIR")
-    );
-
+    let test_path = std::env::temp_dir().join("failed_to_create_directory");
     std::fs::write(&test_path, "test").unwrap();
-    let test_path = format!("{}\\out.txt", test_path);
 
+    let test_path = test_path.join("out.txt");
+    let test_path = test_path.to_str().unwrap();
     bindgen(&format!("--out {test_path} --in default --filter POINT",));
 }
 
 #[test]
 #[should_panic(expected = "failed to write file")]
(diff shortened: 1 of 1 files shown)


ChrisDenton:
> I found that `CARGO_TARGET_TMPDIR` is not guaranteed to point to an existing directory

Can you say more? The docs say Cargo will create it so if it doesn't then that's definitely a bug in Cargo. However, it also says that Cargo doesn't manage it which means that anything could delete it. And if that's happening then it's likely a bug in the testing code somewhere.

kennykerr:
Sporadic build failures suddenly started showing up today like this one:

https://github.com/microsoft/windows-rs/actions/runs/12679261715/job/35339673541

ChrisDenton:
That's on stable, huh. Which definitely hasn't changed recently. Strange. Possibly some change in CI uncovered a bug somewhere. But seems weird.

riverar:
@ChrisDenton I noticed on my local machine (and on the runners via SSH) that the directory was missing during the test run. I do see Cargo creates the folder once during a compile pass but not during a test-run--not conclusive, I realize.

In these bindgen tests, if the `failed_to_write_file` test runs first, `std::fs::create_dir_all` is called, inadvertently creating the `tmp` directory and masking the missing state of the `tmp` location. My thinking was to rely on the OS temp directory to address several issues at once.

But given the errors we're now encountering in this PR, maybe CI could use a break tonight. 😅

ChrisDenton:
What's not clear to me is if Cargo creates the folder once then what deletes it?

kennykerr:
It seems this change does not address the problem:

https://github.com/microsoft/windows-rs/actions/runs/12682378553/job/35376701787?pr=3428

kennykerr:
Thanks for the eyeballs on this issue @riverar @ChrisDenton

PR #3448, opened 2025-01-15
Reintroduce `Ref` lifetime type parameter

kennykerr:
Now that C++ delegates use `Ref` and `OutRef` to properly handle lifetime and ownership for callbacks (#3447), it became evident that `Ref` really should have a lifetime type parameter (#3433). Otherwise its impossible to safely construct a `Ref` on the call side. This does however mean that `Ref` can't be used as a type alias inside a trait (#3435) but since this was only being used for WinRT generics, it seems reasonable to favor making `Ref` safer over generic code that very few will encounter.


kennykerr:
Note to self: [GATs](https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html) may be able to give us the best of both worlds.

PR #3450, opened 2025-01-15
Introducing the `windows-link` crate

kennykerr:
This update introduces the `windows-link` crate as a simpler alternative to the `windows-targets` crate. It provides the exact same `link` macro but uses `raw-dylib` unconditionally and thus has no target-specific dependencies to include import libs. This means that the `windows-link` crate requires Rust 1.71 or later as that was the first version to stabilize `raw-dylib` for all Windows targets. 

The `windows-bindgen` crate is also updated to default to use `windows-link` for functions. You can opt-out by using the new `--link` option and specify a different module for the `link` macro. For example, the `windows-sys` crate continues to use `windows-targets` as it has an older MSRV. It does so by using the following option `bindgen` arguments:

`--link windows_targets`

The `windows-targets` crate is unchanged and will continue to be updated for the time being to support `windows-sys`. It continues to support the `windows_raw_dylib` cfg option for opting in to `raw-dylib` whereas the new `windows-link` crate ignores the `windows_raw_dylib` cfg option and uses `raw-dylib` in all cases.

As an example, consider the following build script.

```rust
windows_bindgen::bindgen(
    "--in default --out src/bindings.rs --flat --sys --filter GetTickCount".split_whitespace(),
);
```

This will produce the following output:

```rust
windows_link::link!("kernel32.dll" "system" fn GetTickCount() -> u32);
```

Add a dependency on the `windows-link` crate and you're all set!

You can then use the new `--link` argument only if you need to change this.

```rust
windows_bindgen::bindgen(
    "--in default --out src/bindings.rs --flat --sys --filter GetTickCount --link windows_targets".split_whitespace(),
);
```

This will produce the following altered output to once again use the `windows-targets` crate.

```rust
windows_targets::link!("kernel32.dll" "system" fn GetTickCount() -> u32);
```


ChrisDenton (approved):
Looks great! `--link` is a really useful feature on its own. And the more we can move toward not needing to bundle binary libs the better.

crates/libs/link/src/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+    )
+}
+
+/// Defines an external function to import.
+#[cfg(all(windows, not(target_arch = "x86")))]
+#[macro_export]
+macro_rules! link {
+    ($library:literal $abi:literal $($link_name:literal)? fn $($function:tt)*) => (
+        #[link(name = $library, kind = "raw-dylib", modifiers = "+verbatim")]
+        extern "C" {
```
ChrisDenton, line 23: This is just a comment, I had to look this up to recall why `extern "C"` is used instead of `$abi`. It was https://github.com/rust-lang/rust/issues/110505 which wasn't fixed until 1.77, a fair bit above the current MSRV.

RalfJung: No, that cannot be true. https://github.com/rust-lang/rust/issues/110505 got closed by https://github.com/rust-lang/rust/pull/119587 which just adds an unstable feature that's still unstable.

EDIT: See https://github.com/microsoft/windows-rs/issues/3672 for more context.

Issue #3437, opened 2025-01-14
C++ delegate parameters are already optional


kennykerr:
The param "lptimerfunc" type "TIMERPROC" of SetTimer is option itself,   I think it is repeated.

_Originally posted by @liqh99em in https://github.com/microsoft/windows-rs/issues/3368#issuecomment-2589287078_

kennykerr:
An example from 0.58. 

```rust
use windows::{Win32::Foundation::*, Win32::System::Console::*};

fn main() {
    unsafe {
        SetConsoleCtrlHandler(Some(callback), true).unwrap();
    }
}

unsafe extern "system" fn callback(_ctrltype: u32) -> BOOL {
    BOOL(1)
}
```

With 0.59 the callback must be double wrapped:

```rust
SetConsoleCtrlHandler(Some(Some(callback)), true).unwrap();
```

PR #3452, opened 2025-01-16
Avoid over-wrapping optional callback parameters

kennykerr:
When a parameter is both optional and a callback, it should not be wrapped in an `Option`.

Fixes: #3437

I have been thinking about removing the `Option` from the callback type itself. At that point we'd not need this workaround and we'd be able to distinguish between optional and required callback parameters, but this is a small and targeted fix to the immediate problem for the time being.


riverar (approved):
Looks OK in my testing.

One thing I observed--maybe known/intentional--is that delegate types are emitted with a Option wrapper, despite being not optional in all dependent functions. Should we drop the Optional at that site instead and then keep the existing code?

Edit: Ah you addressed this in your edit. 👍

```csharp
// Metadata
unsafe HRESULT setPfnMiniPDBErrorCallback2(
  [In] void* pvContext,
  [In] PFNMINIPDBERRORCALLBACK2 pfn);
```

```rust
pub type PFNMINIPDBERRORCALLBACK2 = Option<
    unsafe extern "system" fn(
        pvcontext: *mut core::ffi::c_void,
        dwerrorcode: u32,
        szobjorpdb: windows_core::PCWSTR,
        szlib: windows_core::PCWSTR,
    ) -> windows_core::HRESULT,
>;
```

kennykerr:
Right, I'm leaning toward making the C++ delegate type omit `Option` as I think that would make it easier to use those types generally and they would behave more consistently with other non-nullable types in windows-rs like interfaces that you have to explicitly wrap in `Option` as needed. But that's a bigger change I'll work separately.

PR #3455, opened 2025-01-17
Use unique path for `cppwinrt` temp file

kennykerr:
This is not a problem consumers will face but my CI tests occasionally fail on windows-rs due to tests running in parallel as there's IO contention on the temp file. This update just ensures that each invocation gets a unique file name so there's no contention. Ideally there'd be a simpler way to produce a unique path but I can't think of any but this is simple enough. 

I had previously attempted to alleviate the issue in #3289 but there was still a race as two different tests could race to create the file and then subsequently write the file. Here we just avoid the race entirely.


riverar:
Could also defer to Win32 API `GetTempFileName`.

kennykerr:
That would definitely be more difficult and unsafe. 🫢

PR #3460, opened 2025-01-21
Simplify dependency tracking

kennykerr:
Collecting and combing dependencies continues to be an important but challenging task. This update introduces the `Dependencies` trait that simplifies this by allowing dependencies to be retrieved with a single simple method call.

### DIFF
--- a/crates/libs/bindgen/src/types/class.rs
+++ b/crates/libs/bindgen/src/types/class.rs
@@ -81,7 +81,7 @@ impl Class {
             }
         );
 
-        let factories = required_interfaces.                 iter().filter_map(|interface| match interface.kind {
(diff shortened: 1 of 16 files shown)


crates/libs/bindgen/src/types/class.rs
```diff
@@ -83,7 +81,7 @@ impl Class {
             }
         );
 
-        let factories = required_interfaces.iter().filter_map(|interface| match interface.kind {
+        let factories = required_interfaces.                 iter().filter_map(|interface| match interface.kind {
```
kennykerr, line 84: I also noticed that `cargo fmt` seems to give up on some larger closures. Not sure what's going on there...

Issue #3346, opened 2024-11-19
Add new method to windows_registry::Key open_mut()


ThalusA:
### Suggestion

Hi, 

It would be awesome if there was a way to open registry key with write access instead of only read access.
There is some case where I don't want to create it if it doesn't exist but I want to open my registry key with write access

kennykerr:
Thanks for the suggestion - let me think about what that might look like.

petur-keystrike:
How about allowing the caller to specify the desired access? It could look something like 
```
key = CURRENT_USER.open_with_access(path, KEY_READ | KEY_WRITE);
```
This would make it possible to use other flags, for example to add `KEY_WOW64_32KEY` without needing a new method for each combination.

kennykerr:
Sorry for the delay - #3461 provides control over open/create and read/write options.

Issue #3269, opened 2024-09-12
Suggest windows-registry can support transactions.


amone-bit:
### Suggestion

Like the crate named `winreg` support transactions.

gentoo90:
_Side note_: WinAPI functions used in https://github.com/gentoo90/winreg-rs/blob/master/src/transaction.rs are part of [Windows Kernel Transaction Manager API](https://learn.microsoft.com/en-us/windows/win32/ktm/kernel-transaction-manager-portal) which should probably have its own crate, `windows-ktm` or something that should be used as a dependency by `windows-registry` and by other things that need transactions.

kennykerr:
The Kernel Transaction Manager API is primarily intended for the registry and file system APIs but the latter is [strongly discouraged](https://learn.microsoft.com/en-us/windows/win32/fileio/transactional-ntfs-portal) so that just leaves the registry API. As such, there doesn't appear to be much value in generalizing this beyond the `windows-registry` crate.

kennykerr:
Sorry for the delay - #3461 provides transaction support.

PR #3461, opened 2025-01-22
Add `OpenOptions` to `windows-registry`

kennykerr:
Rather than adding more registry creation methods to satisfy #3269, #3346, or any future needs, this update introduces an `OpenOptions` type modeled after `std::fs::OpenOptions`. This new type provides the options and flags that you can set in order to configure how a registry key is opened. The `Key` type continues to provide the `create` and `open` methods as a convenience but internally they now also use `OpenOptions` for construction. Let's consider a few examples.

Today you can use `create` to create or open a registry key for read and write access:

```rust
let path = "software\\windows-rs\\sample";
let key = CURRENT_USER.create(path)?;
```

What if you only want to open the key if it already exists? Well you can use `open` instead but it offers only read access:

```rust
let key = CURRENT_USER.open(path)?;
```

This is where you can use the new `options` method to gain more control:

```rust
let key = CURRENT_USER.options().write(true).open(path)?;

key.set_u32("name", 123)?;
```

This opens the subkey with write access but only if it already exists. The call will fail if the key does not exist. Want to create the key in all cases? Just set the `create` option:

```rust
let key = CURRENT_USER.options().create(true).write(true).open(path)?;
```

You can also optionally create a `Transaction` object for transactional access to the registry:

```rust
let tx = Transaction::new()?;

let key = CURRENT_USER
    .options()
    .read(true)
    .write(true)
    .create(true)
    .transaction(&tx)
    .open(path)?;

key.set_u32("name", 123)?;

tx.commit()?;
```

Failure to commit the transaction will cause all changes to be rolled back when the transaction object is dropped. They are also not observable outside of the transaction until and unless `commit` is called. 

Finally, I removed the `Default` implementation from the `Key` object. This just seemed confusing as you could try (and fail) to use the various registry functions on an invalid registry key handle. If for some reason you need to hold on to an invalid registry key you can still use the unsafe `from_raw` method.

Fixes: #3269 
Fixes: #3346

### DIFF
--- a/crates/libs/registry/src/transaction.rs
+++ b/crates/libs/registry/src/transaction.rs
@@ -1,0 +1,53 @@
+use super::*;
+
+/// A transaction object.
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct Transaction(pub(crate) HANDLE);
+
+impl Transaction {
+    /// Creates a new transaction.
+    pub fn new() -> Result<Self> {
+        let handle = unsafe { CreateTransaction(null_mut(), null_mut(), 0, 0, 0, 0, null()) };
+
+        if handle == INVALID_HANDLE_VALUE {
+            Err(Error::from_win32())
+        } else {
+            Ok(Self(handle))
+        }
+    }
+
+    /// Commits the transaction.
+    ///
+    /// The transaction rolls back if it is dropped before `commit` is called.
+    pub fn commit(self) -> Result<()> {
+        let result = unsafe { CommitTransaction(self.0) };
+
+        if result == 0 {
+            Err(Error::from_win32())
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Constructs a transaction object from an existing handle.
+    ///
+    /// # Safety
+    ///
+    /// This function takes ownership of the handle.
+    /// The handle must be owned by the caller and safe to free with `CloseHandle`.
+    pub unsafe fn from_raw(handle: *mut core::ffi::c_void) -> Self {
+        Self(handle)
+    }
+
+    /// Returns the underlying transaction key handle.
+    pub fn as_raw(&self) -> *mut core::ffi::c_void {
+        self.0
+    }
+}
+
+impl Drop for Transaction {
+    fn drop(&mut self) {
+        unsafe { CloseHandle(self.0) };
+    }
+}
--- a/crates/libs/registry/src/transaction.rs
+++ b/crates/libs/registry/src/transaction.rs
@@ -40,7 +40,7 @@ impl Transaction {
         Self(handle)
     }
 
-    /// Returns the underlying transaction key handle.
+    /// Returns the underlying transaction handle.
     pub fn as_raw(&self) -> *mut core::ffi::c_void {
         self.0
     }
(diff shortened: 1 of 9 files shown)


Nerixyz:
> You can also optionally create a `Transaction` object for transactional access to the registry:

This example would be great in the README (so it shows up in the crate documentation on docs.rs).

kennykerr:
Done - thanks for the suggestion!

ChrisDenton (approved):
This is a really nice API! The extensibility is a nice feature too as it'll make future feature requests much easier to accommodate.

crates/libs/registry/src/transaction.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// The transaction rolls back if it is dropped before `commit` is called.
+    pub fn commit(self) -> Result<()> {
+        let result = unsafe { CommitTransaction(self.0) };
+
+        if result == 0 {
+            Err(Error::from_win32())
+        } else {
+            Ok(())
+        }
+    }
```
riverar, line 31: Thoughts on automatically rolling back the transaction on failure?

crates/libs/registry/src/transaction.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub fn as_raw(&self) -> *mut core::ffi::c_void {
+        self.0
+    }
+}
+
+impl Drop for Transaction {
+    fn drop(&mut self) {
+        unsafe { CloseHandle(self.0) };
+    }
+}
```
riverar, line 53: Should probably have a way for devs to get the state of the current transaction (via `GetTransactionInformation`), esp. if it gets rolled back after failure by us or the system.

crates/libs/registry/src/transaction.rs
kennykerr: Drop the transaction object and it rolls back automatically.

kennykerr:
@riverar - good feedback but I'd rather not add features opportunistically. If and when such needs present themselves, we can now more easily introduce them.

crates/libs/registry/src/transaction.rs
riverar: After `CommitTransaction` fails, the transaction is in a limbo unknown state, and there's no way to retrieve that state. So my thinking was to just explicitly rollback all failed commits to make the state known. (Granted, developers would lose the ability to retry the transaction.)

kennykerr: Recovery is typically impractical thus simply unwind and drop everything, starting again if desired.

crates/libs/registry/src/transaction.rs
```diff
@@ -0,0 +1,53 @@
...
+    ///
+    /// # Safety
+    ///
+    /// This function takes ownership of the handle.
+    /// The handle must be owned by the caller and safe to free with `CloseHandle`.
+    pub unsafe fn from_raw(handle: *mut core::ffi::c_void) -> Self {
+        Self(handle)
+    }
+
+    /// Returns the underlying transaction key handle.
```
riverar, line 43:
```suggestion
    /// Returns the underlying transaction handle.
```

riverar (approved):

Issue #3468, opened 2025-01-31
Bindgen cannot find nested `enum Version` definition


MarijnS95:
### Suggestion

Hi @kennykerr, are you planning on upgrading `win32metadata` any time soon? [The new Agility SDK is out](https://github.com/microsoft/win32metadata/pull/2055) and I'd like to regenerate `windows-rs` with it (even just for local testing) but hit the following snag:

```
thread 'main' panicked at crates\libs\bindgen\src\tables\field.rs:29:18:
type not found: .Version
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/std\src\panicking.rs:665
   1: core::panicking::panic_fmt
             at /rustc/9fc6b43126469e3858e2fe86cafb4f0fd5068869\library/core\src\panicking.rs:76
   2: windows_bindgen::winmd::reader::Reader::unwrap_full_name
             at .\crates\libs\bindgen\src\winmd\reader.rs:178
   3: enum2$<windows_bindgen::types::Type>::from_ref
             at .\crates\libs\bindgen\src\types\mod.rs:230
   4: enum2$<windows_bindgen::types::Type>::from_blob_impl
             at .\crates\libs\bindgen\src\types\mod.rs:285
   5: enum2$<windows_bindgen::types::Type>::from_blob
             at .\crates\libs\bindgen\src\types\mod.rs:258
   6: windows_bindgen::tables::Field::ty
             at .\crates\libs\bindgen\src\tables\field.rs:29
   7: windows_bindgen::types::cpp_const::impl$3::combine
             at .\crates\libs\bindgen\src\types\cpp_const.rs:142
   8: windows_bindgen::types::impl$3::combine
             at .\crates\libs\bindgen\src\types\mod.rs:984
   9: windows_bindgen::type_map::TypeMap::filter
             at .\crates\libs\bindgen\src\type_map.rs:40
  10: windows_bindgen::bindgen<array$<ref$<str$>,2>,ref$<str$> >
             at .\crates\libs\bindgen\src\lib.rs:171
  11: tool_bindings::main
             at .\crates\tools\bindings\src\main.rs:14
  12: core::ops::function::FnOnce::call_once<void (*)(),tuple$<> >
             at C:\Users\Marijn\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ops\function.rs:250
[...]

PR #3471, opened 2025-02-03
Harden nested type discovery

kennykerr:
It turns out that ECMA-335 may include a valid `TypeNamespace` for a nested `TypeDef`. I have never encountered this until #3468 but I can't find anything in the spec to indicate that this is invalid. So I've hardened the reader to check whether a `TypeDef` is nested directly rather than relying on the absence of a `TypeNamespace` value. 

Scanning the `NestedClass` table is slightly slower, but fortunately it is not much slower as this table is sorted and a binary search can be used. I may revisit this further and attempt to index the types in a single pass instead, but for now this solves the issue encountered by #3468 as I can now also reliably exclude nested types that are not structs. Only nested structs are supported by `windows-bindgen`. 

Fixes: #3468

### DIFF
--- a/crates/libs/bindgen/src/tables/type_def.rs
+++ b/crates/libs/bindgen/src/tables/type_def.rs
@@ -56,6 +56,10 @@ impl TypeDef {
         self.file().equal_range(2, self.index() + 1).next()
     }
 
+    pub fn nested(&self) -> Option<NestedClass> {
+        self.file().equal_range(0, self.index() + 1).next()
+    }
+
     pub fn underlying_type(&self) -> Type {
         let field = self.fields().next().unwrap();
         if let Some(constant) = field.constant() {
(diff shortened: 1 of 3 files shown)


riverar:
I couldn't find it explicitly codified either, but the spec does state that nested type names are scoped by their enclosing type. So I don't think the namespace field serves any meaningful purpose at that point.

Can we instead peek at the _ClassType_ and check for the `nested` accessibility attribute?

riverar (approved):

kennykerr:
Good question.

Here's the nested `Version` type I encountered in the metadata referred to in the #3468 issue:

image

Unfortunately, it does not have any of the nested bits set. This is what it should look like:

image

So it may still be that this is bad metadata. I'm not too sure, but that doesn't work reliably.

riverar:
Interesting. I agree, looks like bad metadata.

Would file a bug on win32metadata; it seems only nested structs are emitted with the appropriate nested attributes.
https://github.com/microsoft/win32metadata/blob/main/sources/ClangSharpSourceToWinmd/ClangSharpSourceWinmdGenerator.cs#L840-L843

(Aside: ILSpy appears to mistakenly refer to these as _visibility_ attributes but is otherwise functioning correctly.)

kennykerr:
Here's a small optimization: https://github.com/microsoft/windows-rs/pull/3471/commits/02eb56bec4e96e22a0bdb62d15e48d676e6e5948

That at least avoids the binary search in most cases.

riverar (approved):

Issue #3474, opened 2025-02-06
Add support for `KEY_WOW64_32KEY`/`KEY_WOW64_64KEY` in registry API


justanotheranonymoususer:
### Suggestion

Probably can be added to `OpenOptions`

kennykerr:
Here you go: #3482

PR #3482, opened 2025-02-11
Generalize support for access rights in `windows-registry`

kennykerr:
Building on #3461, this update generalizes support for access rights so that a caller can request more than just read/write access. I decided against adding an enum declaring the various known access rights as that just leads to the inevitable confusion around which ones to use, how they might be combined, and which obscure access rights to include in the library. If you're unsure just use `read()` or `write()` and move on. 😊

I also refined the `OpenOptions` to be additive-only to simplify things. So rather than having to say `options.read(true).write(true)` you can simply say `options.read().write()` and so on. 

Fixes: #3474

### DIFF
--- a/crates/libs/registry/readme.md
+++ b/crates/libs/registry/readme.md
@@ -41,9 +41,9 @@ fn main() -> Result<()> {
 
     let key = CURRENT_USER
         .options()
-        .read(true)
-        .write(true)
-        .create(true)
+        .read()
+        .write()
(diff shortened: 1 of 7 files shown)


justanotheranonymoususer:
Looks great, thanks! I also like that you can now just use `KEY_QUERY_VALUE` instead of `KEY_READ` when that's the only perm you need.

PR #3484, opened 2025-02-12
Support metadata with `MethodDef` constructor parent resolution

kennykerr:
Some, typically unmerged, metadata may present `MethodDef` constructors for attributes type resolution. This is unusual and thus hard to test but I have manually tested this against some internal metadata to confirm that it covers this edge case in ECMA-335 parsing.


robmikh (approved):
Looks great, thanks!

Issue #3458, opened 2025-01-20
Some APIs are missing in the features tool


Guiguiprim:
### Summary

I'm not sure if this is the right place to report this, if not sorry for the noise.

I discovered https://microsoft.github.io/windows-rs/features/#/0.59.0 recently.

While updating some code from `0.58.0` to `0.59.0`, I had a method not found error on an API that is documented, so I tried the tools to check my feature set, but the search came-up empty.

https://microsoft.github.io/windows-rs/features/#/0.59.0/search/InstalledLocation which should have found `ApplicationModel::Package::InstalledLocation`. 
Looking at the code, this method is under `#[cfg(feature = "Storage_Search"]` (which was what I was missing).

kennykerr:
Thanks, we're aware of this issue. The feature tool needs a bit of work - hopefully that should be wrapped up soon.

kennykerr:
Oh, `InstalledLocation` is a method. I'm not sure whether the index will cover methods. The trouble is that this blows up the size of the index.

riverar:
I'm working on wrapping up the next iteration of the feature search tool, thanks for your patience!

To unblock `Windows.ApplicationModel.IPackage::get_InstalledLocation()` requires
```
"Windows.ApplicationModel",
"Windows.Storage.Search"
```

viruscamp:
I searched `OBJECT_ATTRIBUTES`, and got "Wdk_Foundation", but it needs `Win32_Security` also.

`Wdk::Storage::FileSystem::NtOpenFile` needs `Win32_Security` too.

PR #3486, opened 2025-02-13
Restore and extend feature index generation with method support

riverar:
Restores and extends JSON index generation to include method feature information. While this increases the index size from 11.7MB to 12.7MB (~8.5%), the ability to search method features justifies the size increase. Includes space optimization measures to minimize the JSON payload.

The web tool has been updated to support this altered index format (V2) while maintaining backwards compatibility with existing V1 indices.

Fixes: #3458


crates/libs/bindgen/src/lib.rs
```diff
@@ -195,7 +196,8 @@ where
         namespace: "",
     };
 
-    writer.write(tree)
+    writer.write(tree);
+    index::write();
```
kennykerr, line 200: As this is not specific or applicable to most invocations of `windows-bindgen` we should probably just have another static tool that generates the json like tool_yml and the rest.

riverar: It's currently in bindgen to take advantage of the filters. I *think* everything else is externally accessible but I'd have to double check.

crates/libs/bindgen/src/index.rs
```diff
@@ -0,0 +1,137 @@
...
+            .push(index_item);
+    }
+}
+
+const EXCLUDED_NAMESPACES: &[&str] = &["Windows.Foundation", "Windows.Win32.Foundation"];
+
+#[doc(hidden)]
+pub fn write() {
+    let mut feature_index = Index::new();
+    let reader = Reader::new(expand_input(&["default"]));
```
kennykerr, line 74: So this is not specific to the invocation of `bindgen` so it makes more sense to have it as a separate step. We can add a hidden `features` function that returns the string and then `tool_features` can simply call it. Or we just make the reader public.

crates/libs/bindgen/src/index.rs
```diff
@@ -0,0 +1,137 @@
...
+                    .collect();
+
+                let scoped_name = format!("{}.{}", interface_name, method.name());
+                feature_index.add_item(namespace, &scoped_name, method_features);
+            }
+        }
+    }
+
+    write_to_file(
+        "features.json",
```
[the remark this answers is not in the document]
riverar: Done!

kennykerr (approved):
Thanks for all your work on this!

PR #3502, opened 2025-02-21
Release 0.62.0

kennykerr:
This release includes the following brand new crates:

* The [windows-collections](https://crates.io/crates/windows-collections) crate defines the Windows collection types like `IIterable<T>`, `IVector<T>`, `IMap<K, V>`, and so on (#3483). It also includes all of the stock implementations for creating such collections (#2346, #2350, #2353). This allows these collections to be used without requiring a dependency on the larger `windows` crate. This crate also provides an optimized implementation of the standard `Iterator` trait for the Windows `IIterator<T>` interface (#3476).

* The [windows-future](https://crates.io/crates/windows-future) crate defines the Windows async types like `IAsyncAction`, `IAsyncOperation<T>`, and so on (#3490). It also includes all of the stock implementations for creating such async types (#3221, #3235). This allows these async types to be used without requiring a dependency on the larger `windows` crate. 

* The [windows-link](https://crates.io/crates/windows-link) crate provides linker support for Windows (#3450). This is the evolution of the older `windows-targets` crate but is substantially simpler and more versatile thanks to advances in the Rust compiler since the `windows-targets` crate was unveiled. Notably, it does not depend on or insert any import libs and can be used with custom libraries, not only those provided by the Windows operating system. All of the crates, with the exception of `windows-sys`, now depend on the new `windows-link` crate instead of the older `windows-targets` crate. This greatly simplifies compilation and also greatly reduces the size of dependencies as the `windows-link` crate is tiny. The `windows-bindgen` crate defaults to `windows-link` but also adds the `--link` option to override this as needed. You may for example want to use `--link windows_targets` if you need to stick with the `windows-targets` crate if you cannot change your MSRV to Rust 1.71 or later as that was the first version to stabilize `raw-dylib` for all Windows targets. This then lets you continue to use `windows-bindgen` until you are ready to move to a newer version of Rust. 

* The [windows-numerics](https://crates.io/crates/windows-numerics) crate defines the Windows numeric types to support graphics-oriented math APIs and calculations (#3488). It also also includes all of the stock implementations for overloaded operators and other transformations. This allows these numeric types to be used without requiring a dependency on the larger `windows` crate.

In addition to these new crates, this release includes major updates to the following existing crates:

* The [windows-bindgen](https://crates.io/crates/windows-bindgen) crate provides a number of improvements including new diagnostics (#3498), streamlined and more capable reference support (#3497, #3492), hardened method overloading (#3477), far fewer `transmute` calls, as well as many other critical fixes and improvements. 

* The [windows-core](https://crates.io/crates/windows-core) crate is largely unchanged but required some breaking changes to support `windows-bindgen` type system improvements. 

* The [windows-registry](https://crates.io/crates/windows-registry) crate continues to improve with generalized support for access rights (#3482), open options (#3461), and other minor improvements.

* The [windows](https://crates.io/crates/windows) crate now delegates to the `windows-numerics`, `windows-future`, and `windows-collections` crates for those types, as well as a number of critical fixes and improvements to features and `cfg` guards (#3431), and many other small improvements. 

In addition, this release also includes minor updates to the following existing crates:

* The [windows-result](https://crates.io/crates/windows-result) now includes the `BOOL` type (#3441) as a core type. This allows this ubiquitous type to be used without requiring a dependency on the larger `windows` crate. 

* The [windows-strings](https://crates.io/crates/windows-strings) crate now  depends on the new `windows-link` crate instead of the older `windows-targets` crate.

* The [windows-version](https://crates.io/crates/windows-version) crate now  depends on the new `windows-link` crate instead of the older `windows-targets` crate.

* The [cppwinrt](https://crates.io/crates/cppwinrt) crate includes minor improvements to improve build reliability.


ratijas:
Sorry, I can't find v0.62 on crates.io, and [official documentation](https://microsoft.github.io/windows-docs-rs/) shows v0.60 — what is happening?

kennykerr:
See the release notes [for the latest release](https://github.com/microsoft/windows-rs/releases/tag/63). The windows-rs project now encompasses a [wide range of crates](https://github.com/microsoft/windows-rs/blob/master/docs/readme.md) and it doesn't make sense to release a new version of each crate with every release.

ratijas:
I see.

Well, on some level it does make sense to release all crates with the same version -- even if some of them didn't have any commits since the last release -- just to avoid the confusion. For example, KDE does that for their Frameworks and Plasma releases. They have automated tooling to bump versions simultaneously across all their repos. To be honest it kinda helps stay sane with their amount of repos to manage.

PR #3055, opened 2024-05-23
Dynamic casting to implementation

sivadeilra:
This provides a new feature for COM developers using the `windows-rs` crate. It allows for safe dynamic casting from `IUnknown` to an implementation object. It is based on Rust's `Any` trait.

Any type which is marked with `#[implement(IDynamicCast, ...)]` can be used with dynamic casting. Types must opt-in to using dynamic casting by adding the `IDynamicCast` interface to their interface list.

Example:

```rust
#[implement(IDynamicCast, IFoo)]
struct MyApp { ... }

fn main() {
    let my_app = ComObject::new(MyApp { ... });
    let iunknown: IUnknown = my_app.to_interface();
    do_stuff(&iunknown);
}

fn do_stuff(unknown: &IUnknown) -> Result<()> {
    let my_app: ComObject<MyApp> = unknown.cast_impl()?;
    my_app.internal_method();
    Ok(())
}
```

PR #3506, opened 2025-02-21
Remove `cast` from `implement` macro

sivadeilra:
This is a cleanup item for my PR #3055.  That PR (and many others that I submitted around the same time) greatly improved support for creating COM objects.  This PR cleans up a few minor things from that work.

First, this PR deletes the generated `cast` method on types marked with `#[implement]`.  The `cast` method was an older mechanism for casting from an object to one of its interfaces.  It is extremely dangerous because it relies on memory constraints that are not verifiable (which is why it is marked `unsafe`).  My previous PRs provided much better ways to handle the need for this, with the `to_interface`, `as_interface` etc. methods.  So `cast` is dangerous and unneeded; let's just remove it.  This is a breaking change, but it's needed.

Second, the PR removes an incorrect type bound on a method in `ComObject`.  This is harmless.

### DIFF
--- a/crates/libs/core/src/com_object.rs
+++ b/crates/libs/core/src/com_object.rs
@@ -1,5 +1,5 @@
 use crate::imp::Box;
-use crate::{AsImpl, IUnknown, IUnknownImpl, Interface, InterfaceRef};
+use crate::{IUnknown, IUnknownImpl, Interface, InterfaceRef};
 use core::any::Any;
 use core::borrow::Borrow;
 use core::ops::Deref;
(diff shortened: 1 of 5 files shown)


riverar:
This method is currently in the documentation; what's the proposed new API surface here? Are these methods drop-in replacements or?

sivadeilra:
> This method is currently in the documentation; what's the proposed new API surface here? Are these methods drop-in replacements or?

There are already better replacements: `ComObject::as_interface`, `ComObject::to_interface`.  You can see in the existing code where I changed uses of `cast()` to use these functions.  The replacements are safe and do not require runtime type checks (QueryInterface calls).

Which docs reference `cast`?  I can update the docs.

riverar:
Ah, thanks. I also took it for a spin and realized we weren't talking about `Interface::cast`.

kennykerr (approved):
Nice, less scary code is great.

PR #3507, opened 2025-02-23
Improvements to COM object support

sivadeilra:
This improves COM object support.

The main focus is on improving the internal structure of the `implement` proc macro. I've broken the `implement` function into a bunch of small, focused functions. The `implement` proc macro first parses its input, creates an `ImplementInputs` struct which contains all of the information need to generate output.  Then, it calls `gen_all` to generate the output, which calls `gen_*` to generate each specific output item. This makes the code much easier to understand, debug, and edit.

The codegen functions (the `gen_*` functions) now generate strongly-typed `syn` values, such as `syn::Item`. Each `gen_*` generates the tokens and then parses them into the strongly-typed value. This allows us to catch syntax errors far more quickly, and the `parse_quote` macro will panic at the site of any syntax error. This makes development on the `increment` macro a lot more productive.

Speaking of debugging, I've added some unit tests within the `implement` crate. They don't actually _test_ anything, but that's not the point.  The point of these unit tests is that you can very easily see the output of `implement`, without needing `cargo-expand` (which has its own problems).  You can also step through the code in a debugger.

This PR contains some modest improvements to the code that is generated by `implements`:

* Instead of generating a tuple of interface pointers, this code now generates fields for each interface pointer. Each interface pointer field is named after the interface and its index, such as `interface2_ifoo`. This makes it a lot easier to debug COM objects at runtime.
* The `IUnknownImpl` trait is now an `unsafe` trait. There's no way to implement this trait without writing unsafe code.
* The `IUnknown::Release` implementation now puts the "slow" path (refcount reaches zero) into a separate function. This keeps the fast path (refcount is not zero) as small and fast as possible. For objects with significant destructors, this helps.
* Cosmetic: Renamed `GetTrustLevel` and such to use snake casing, where applicable.
* Removed `from_inner_ref` on `IUnknown`. There is no need for this any more, since all COM interface method implementations are on `Foo_Impl` types now.
* Changed the methods on `WeakRefCount` that can change the refcount to have `unsafe` signatures.  Any change to a refcount could affect memory lifetime, so these operations need to be `unsafe`.  This allows us to expose `&WeakRefCount` for `ComObject<T>`, which simplifies other things. The motivation here is to make it difficult to accidentally leave a memory safety hole (in safe code).
* "Inlined" a little code that touches the reference count and simply directly used the refcount field.


kennykerr:
At an initial reading this PR seems to be doing too much - it would be easier to reason about if this were broken up a bit more into more discrete changes that focused on individual things. Philosophical changes to `WeakRefCount` - which I'm not convinced I agree with - seem unrelated to the refactoring of the code generation of the `implement` macro. Then there are changes to `ComObject` and `IUnknownImpl` which looks good but again are not directly related to code generation in the `implement` macro. Also some renaming of methods which may be good for consistency but again unrelated and just leading to the impression of a large complex change. Given that there are around 900 LOC of new and somewhat intricate code involving proc macros and COM - both of which are complex topics, it would be beneficial to just make one change at a time if at all possible.

kennykerr:
> The main focus is on improving the internal structure of the `implement` proc macro. I've broken the `implement` function into a bunch of small, focused functions. The `implement` proc macro first parses its input, creates an `ImplementInputs` struct which contains all of the information need to generate output. Then, it calls `gen_all` to generate the output, which calls `gen_*` to generate each specific output item. This makes the code much easier to understand, debug, and edit.
> 
> The codegen functions (the `gen_*` functions) now generate strongly-typed `syn` values, such as `syn::Item`. Each `gen_*` generates the tokens and then parses them into the strongly-typed value. This allows us to catch syntax errors far more quickly, and the `parse_quote` macro will panic at the site of any syntax error. This makes development on the `increment` macro a lot more productive.

This sounds like a good meaty PR on its own. I like the sound of this.

> Speaking of debugging, I've added some unit tests within the `implement` crate. They don't actually _test_ anything, but that's not the point. The point of these unit tests is that you can very easily see the output of `implement`, without needing `cargo-expand` (which has its own problems). You can also step through the code in a debugger.

> * Code that is generated for a specific interface now uses the `quote_spanned` or `parse_quote_spanned` macros, with a span that points to the interface name. This improves the error messages when the user forgets to provide an implementation of a COM interface, because the error message will point to the interface name.

Macros seem like they have so much potential in Rust but lack so much in terms of debuggability and testability and contextual ignorance. It would be nice, for consistency to move this into the `test_implement` crate, but that probably makes it harder to keep these details private. It would also be nice to have some [panic tests like this](https://github.com/microsoft/windows-rs/blob/master/crates/tests/libs/bindgen/tests/panic.rs) to ensure we maintain the usability. 

> * Instead of generating a tuple of interface pointers, this code now generates fields for each interface pointer. Each interface pointer field is named after the interface and its index, such as `interface2_ifoo`. This makes it a lot easier to debug COM objects at runtime.

There might be an issue here if an implementation needs to implement both `IVector<HSTRING>` and `IVector<IInspectable>` for example. I encountered such edge cases with C++/WinRT. 

> * The `IUnknown::Release` implementation now puts the "slow" path (refcount reaches zero) into a separate function. This keeps the fast path (refcount is not zero) as small and fast as possible. For objects with significant destructors, this helps.

Neat, we did this in C++/WinRT and it made big difference.

> * Cosmetic: Renamed `GetTrustLevel` and such to use snake casing, where applicable.

We may not be super consistent here but it helps to preserve the original API names to avoid confusion and make it easier to search the code base for related code. 

> * Removed `from_inner_ref` on `IUnknown`. There is no need for this any more, since all COM interface method implementations are on `Foo_Impl` types now.

👍

> * Changed the methods on `WeakRefCount` that can change the refcount to have `unsafe` signatures.  Any change to a refcount could affect memory lifetime, so these operations need to be `unsafe`.  This allows us to expose `&WeakRefCount` for `ComObject<T>`, which simplifies other things. The motivation here is to make it difficult to accidentally leave a memory safety hole (in safe code).

I'm not sure I agree with this - the `WeakRefCount` itself is quite safe. This is also not consistent with Rust's convention with methods like `slice::as_ptr` for example which is completely safe even though it can often be used in very unsafe ways but its not `as_ptr` that's unsafe. 

> * "Inlined" a little code that touches the reference count and simply directly used the refcount field.

👍

sivadeilra:
> There might be an issue here if an implementation needs to implement both IVector<HSTRING> and IVector<IInspectable> for example. I encountered such edge cases with C++/WinRT.

Fortunately, that won't cause a collision. Both interface chains will have a different index, and the generated ident contains the index as well as the name suffix.  So this case would generate something like `interface4_ivector` and `interface5_ivector`.  Not terribly useful for distinguishing them, but no worse than the previous approach which used tuples.

Speaking of tuples, we've been relying on a behavior that is not guaranteed.  The compiler is allowed to reorder the fields within a tuple; tuples are `repr(Rust)`, not `repr(C)`. In our case the compiler never does, but nothing guarantees that.  Generating one field per interface chain lets us guarantee the layout, rather than generating a single field whose type is a tuple.

How about this:
* I'll revert the changes to `IUnknownImpl` and `ComObject` and propose those as separate PRs.
* I'll keep the interface chain renaming. It's purely cosmetic and does not need changes in code outside of the macros.
* And keep the `gen_*` split, of course.

kennykerr:
> The compiler is allowed to reorder the fields within a tuple; tuples are repr(Rust), not repr(C). 

Even when the field is part of a `#[repr(C)]` struct? 😯

sivadeilra:
I believe so, because the layout of types is determined by their definition, but not their context.  In this case, we're getting lucky because these definitions are always "a sequence of non-nullable pointer-sized fields".  But if we had a tuple like `(u8, u64, u8)`, then I would expect the compiler to reorder it to `(u64, u8, u8)` to remove some alignment padding.

We're just getting lucky right now.  ;)

crates/libs/implement/src/gen.rs
```diff
@@ -0,0 +1,579 @@
...
+    let identity_query = quote! {
+        if iid == <::windows_core::IUnknown as ::windows_core::Interface>::IID
+        || iid == <::windows_core::IInspectable as ::windows_core::Interface>::IID
+        || iid == <::windows_core::imp::IAgileObject as ::windows_core::Interface>::IID {
+            break 'found &self.identity as *const _ as *const ::core::ffi::c_void;
+        }
+    };
+
+    let tear_off_query = quote! {
+        let tear_off_ptr = self.count.query(&iid, &self.identity as *const _ as *mut _);
```
dpaoliello, line 318: Just checking, we don't need an explicit `AddRef` here?

crates/libs/implement/src/gen.rs
```diff
@@ -0,0 +1,579 @@
...
+        // of the outer object, never by application code.
+        //
+        // The callers of this function (`into_static` and `into_object`) are both responsible
+        // for maintaining one of our invariants: Application code never has an owned instance
+        // of the outer (implementation) type. into_static() maintains this invariant by
+        // returning a wrapped StaticComObject value, which owns its contents but never gives
+        // application code a way to mutably access its contents. This prevents the refcount
+        // shearing problem.
+        //
+        // TODO: Make it impossible for app code to call this function, by placing it in a
```
dpaoliello, line 395: Do we want to emit this TODO into the generated code?

crates/libs/implement/src/gen.rs
```diff
@@ -0,0 +1,579 @@
...
+/// This works because the `IFoo3` type has an `Into` impl for `IFoo2`.
+fn gen_impl_from(inputs: &ImplementInputs) -> Vec<syn::Item> {
+    let mut items = Vec::new();
+
+    let original_ident = &inputs.original_type.ident;
+    let generics = &inputs.generics;
+    let constraints = &inputs.constraints;
+
+    items.push(parse_quote! {
+        impl #generics ::core::convert::From<#original_ident::#generics> for ::windows_core::IUnknown where #constraints {
```
dpaoliello, line 456: `QueryInterface` also supports `IAgileObject`, does that need to be added here as well?

dpaoliello (commented):
I think my comments are on existing code which has been moved...

crates/libs/implement/src/gen.rs
sivadeilra: Probably, but I'll let Kenny speak to that.  It might be the case that we will always generate the vtable methods, but they'll only be wired up to QI when a user asks for it.

If we do change that, let's change it in a separate PR.

crates/libs/implement/src/gen.rs
sivadeilra: Ordinary comments are not emitted into the token stream, so this won't be emitted.  Rustdoc comments are, though.

crates/libs/implement/src/gen.rs
sivadeilra: This is the same as the previous code, just with slightly different control flow. I don't understand the tear-off interface stuff, so I'll defer to Kenny on that.

kennykerr: The weak ref tearoff shares a ref count with the object. This is all handled by `WeakRefCount`.

crates/libs/implement/src/gen.rs
```diff
@@ -0,0 +1,579 @@
...
+            interface: *mut *mut ::core::ffi::c_void,
+        ) -> ::windows_core::HRESULT {
+            unsafe {
+                if iid.is_null() || interface.is_null() {
+                    return ::windows_core::imp::E_POINTER;
+                }
+
+                let iid = *iid;
+
+                let interface_ptr: *const ::core::ffi::c_void = 'found: {
```
kennykerr, line 338: Hmm are using loop labels the simplest way to express this? Just curious what motivated this.

kennykerr: I actually don't see a loop here - may have to look at the generated code to internalize this. 😊

sivadeilra: It eliminates another "is this null?" check.  You can do this purely with control-flow.

crates/libs/implement/src/gen.rs
kennykerr: Feel free to keep the TODO in the code as a reminder.

crates/libs/implement/src/gen.rs
sivadeilra: There is no loop.  Rust has a concept of "named scopes you can break out of".  They're kind of like gotos, but far more constrained (and structured).

crates/libs/implement/src/gen.rs
kennykerr: While QI for `IAgileObject` is handled automatically by the `implement` macro, developers will never use an infallible conversion like `From` for `IAgileObject` since the whole point of `IAgileObject` is to ask the question to make a run time determination.

kennykerr: I'm also wondering whether we really need these `From` implementation at all as its not really the most elegant way to construct a COM object any longer, but that is a discussion for another issue and PR.

crates/libs/implement/src/gen.rs
kennykerr: Yeah I started having cold sweats when I thought about gotos. 🥲

crates/libs/implement/src/tests.rs
```diff
@@ -0,0 +1,130 @@
+//! These tests are just a way to quickly run the `#[implement]` macro and see its output.
+//! They don't check the output in any way.
+//!
+//! This exists because of some difficulties of running `cargo expand` against the `#[implement]`
```
kennykerr, line 4: Maybe just a comment about how to run a test and see the output - for example I run this:

```
 cargo test -p windows-implement --tests simple_type
```

But cargo test does not automatically display the `stdout` for `println`.

kennykerr: e.g. 

```
cargo test -p windows-implement --tests simple_type -- --nocapture
```

crates/libs/implement/src/gen.rs
sivadeilra: I'm fine with removing them.  I think `ComObject::new(foo).into()` is more clear and obvious than just `IFoo::from(foo)`, since it makes it clear that you're constructing a COM object and isn't much longer.

Will do in a separate PR.

crates/libs/implement/src/gen.rs
```diff
@@ -0,0 +1,579 @@
...
+
+/// Generates the `T::into_static` function. This function is part of how we construct a
+/// `StaticComObject<T>` from a `T`.
+fn gen_into_static(inputs: &ImplementInputs) -> syn::ImplItem {
+    assert!(!inputs.is_generic);
+    parse_quote! {
+        /// This converts a partially-constructed COM object (in the sense that it contains
+        /// application state but does not yet have vtable and reference count constructed)
+        /// into a `StaticComObject`. This allows the COM object to be stored in static
+        /// (global) variables.
```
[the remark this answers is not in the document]
kennykerr: How would a developer see this comment since it lives within code generated inside a macro. I'm thinking of rust doc generation but perhaps this is something else, maybe a `rust-analyzer` feature I've not noticed?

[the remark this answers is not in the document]
sivadeilra: Rust Analyzer will show these comments when hovering over calls to `into_static`.  They'll also show up in the output of `cargo doc`.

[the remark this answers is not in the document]
kennykerr: Oh `cargo doc` for a crate that uses the `implement` macro not the crate itself. Thanks.

PR #3513, opened 2025-02-24
Remove `IUnknownImpl::from_inner_ref` as no longer needed

sivadeilra:
This PR cleans up part of #3055, which improved COM object support.

The `IUnknownImpl::from_inner_ref` method predates the new `ComObject` API. It allows unsafe code to cast from `&T` to `&T_Impl`. This is a very dangerous operation because it "widens" the pointer; it requires that `&T` point to a `T` that is embedded within a `T_Impl`.

This was necessary for some situations before the new `ComObject` API, but is is completely unnecessary now.  The new `ComObject` API provides safe ways to do the same thing.  Also, this cast is unnecessary because all COM interface trait implementations are done on `T_Impl` types, not on `T`, so this method is unnecessary.

### DIFF
--- a/crates/libs/core/src/unknown.rs
+++ b/crates/libs/core/src/unknown.rs
@@ -117,16 +117,6 @@ pub trait IUnknownImpl {
     /// Gets the trust level of the current object.
     unsafe fn GetTrustLevel(&self, value: *mut i32) -> HRESULT;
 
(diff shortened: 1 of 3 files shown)


dpaoliello (approved):

kennykerr (approved):
Excellent thanks!

PR #3516, opened 2025-02-25
Simplify `tool_yml` and extend Clippy coverage

kennykerr:
Building on #3504 this update further refactors `tool_yml` to make it easier to author by splitting each yml file generation into a separate module. Some minor tweaks to yml generation are also included:

* Clippy now checks everything except the test crates since they're too noisy and have intentional test code that Clippy doesn't like. #3504 reduced clippy coverage to just libs but I realized that some tools and samples can benefit from Clippy validaiton.
* The `clippy` and `no-default-features` yml workflows are simplified by removing the unnecessary strategy matrix since those only run on `nightly-x86_64-pc-windows-msvc`.


kennykerr:
Removed the samples from clippy coverage for now as they're running into some build dependency issues.

PR #3519, opened 2025-02-26
Derive `Default` for more structs

kennykerr:
There are cases where `Default` cannot be derived for certain Windows types, but there are many more where it can and doing so improves readability and reduces code generation. This further builds on the likes of #2760, #3015, #3041, #3097, #3270 by deriving `Default` for structs that don't use the `--sys` option, although I have had a request to extend it to `--sys` code generation as well so I'll probably do that in a follow-up PR. 

Notably, Rust won't allow `Default` to be derived for structs with explicit layout or structs with pointer or array fields. These continue to be generated by `windows-bindgen` as before.

### DIFF
--- a/crates/libs/bindgen/src/types/cpp_struct.rs
+++ b/crates/libs/bindgen/src/types/cpp_struct.rs
@@ -151,6 +151,9 @@ impl CppStruct {
 
         let default = if writer.config.sys {
             quote! {}
+        } else if self.can_derive_default() {
+            derive.extend(["Default"]);
+            quote! {}
         } else {
             quote! {
                 #cfg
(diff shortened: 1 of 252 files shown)


crates/libs/bindgen/src/types/cpp_struct.rs
```diff
@@ -224,6 +227,16 @@ impl CppStruct {
...
     }
 
+    fn can_derive_default(&self) -> bool {
+        !self.has_explicit_layout()
+            && self.def.fields().all(|field| {
+                !matches!(
+                    field.ty(Some(self)),
+                    Type::ArrayFixed(..) | Type::PtrConst(..) | Type::PtrMut(..)
+                )
+            })
```
dpaoliello, line 237: Not sure if this is easier to read?

```suggestion
            && !self.def.fields().any(|field| {
                matches!(
                    field.ty(Some(self)),
                    Type::ArrayFixed(..) | Type::PtrConst(..) | Type::PtrMut(..)
                )
            })
```

dpaoliello (approved):

crates/libs/bindgen/src/types/cpp_struct.rs
kennykerr: Not sure either. 🙃

PR #3520, opened 2025-02-26
Implement `Default` for `sys` style structs

kennykerr:
The `--sys` style bindings generated by `windows-bindgen` implement fewer standard traits to minimize code generation and build time. A few are required and `Default` is simple enough that it does not appear to impact build time or complexity. So this update adds a `Default` implementation for structs. Deciding when this can be derived is however more complicated as so many more types decay to pointers in `--sys` style bindings.


kennykerr:
It expands the size of the `windows-sys` crate so that's not great, but I'm more concerned about making `windows-bindgen` more broadly useful so perhaps this is worthwhile. Long term I think it makes more sense for developers to use `windows-bindgen` directly. 

Thoughts? @dpaoliello @ChrisDenton

dpaoliello (approved):
I think the increase in the crate size is worth it - it both makes the Rust bindings more useful and makes it easier to port C code.

ChrisDenton (approved):
Reducing the amount of `unsafe` code end users end up writing is a good thing in my book. Especially when it's trivially safe (and technically could be written in safe code but few are going to bother). `Default` is also just so much more convenient.

riverar:
I'd argue this is moving away from the convention of `-sys` crates being minimal thin FFI bindings without additional abstractions or traits. But not going to die on that hill for a crate that will probably go away in the future.

kennykerr:
Ideally there'd be no need for two different modes of code generation but there are some obvious benefits right now so at the very least I'd like to avoid unnecessary differences that make it harder to use one or the other and this seems like one of those.

Issue #2939, opened 2024-03-19
`scale` and `skew` functions of `windows::Foundation::Numerics::Matrix3x2`


wuweiran:
### Suggestion

[Scale](https://learn.microsoft.com/en-us/windows/win32/api/d2d1helper/nf-d2d1helper-matrix3x2f-scale(d2d1_size_f_d2d1_point_2f)) and [Skew](https://learn.microsoft.com/en-us/windows/win32/api/d2d1helper/nf-d2d1helper-matrix3x2f-skew) methods are available in the C++ SDK. It is a nice-to-have for window-rs.

robmikh:
I actually ran into this on my own projects recently. I can contribute more helper functions, but should I wait until after the crate split?

Separately, I'd love to investigate SIMD support...

kennykerr:
Let me tackle the split and then you can more easily add these and SIMD!

robmikh:
No worries! I'm wrapped up in a project myself so I'm keeping myself busy.

kennykerr:
Note to self - the new `windows-bindgen` "--reference" facility can be used to generate references to crates like `windows-numerics` automatically.

kennykerr:
#3488 introduces the dedicated `windows-numerics` crate.

kennykerr:
#3524 remaps a few more common numerics types. 

I was also looking at the D2D helpers for skew, scale, and rotation. They all have an optional center parameter defaulting to zero. 

For Rust perhaps we can have `scale(x,y)` and `scale_around(x,y,center)`, `rotation` and `rotation_around`, and so on, as its rather common to use a zero center for these transformations.

PR #3527, opened 2025-03-03
Add `scale` and `skew` functions to `Matrix3x2`

robmikh:
This change copies over the [`Matrix3x2F::Scale`](https://learn.microsoft.com/en-us/windows/win32/api/d2d1helper/nf-d2d1helper-matrix3x2f-scale(d2d1_size_f_d2d1_point_2f)) and [`Matrix3x2F::Skew`](https://learn.microsoft.com/en-us/windows/win32/api/d2d1helper/nf-d2d1helper-matrix3x2f-skew) helpers from `d2d1helper.h`. Additionally, the existing `rotation` helper was changed to align with the pattern created with `scale` and `scale_around`, utilizing `Vector2` for the center.

Finally, tests and samples were updated for these changes. A test was also added to make sure `skew_around` properly linked to `D2D1MakeSkewMatrix`.

Fixes: #2939

### DIFF
--- a/crates/libs/numerics/src/matrix3x2.rs
+++ b/crates/libs/numerics/src/matrix3x2.rs
@@ -42,6 +42,17 @@ impl Matrix3x2 {
             M32: center.Y - height * center.Y,
         }
     }
+    pub fn skew(angle_x: f32, angle_y: f32) -> Self {
+        Self::skew_around(angle_x, angle_y, Vector2::zero())
+    }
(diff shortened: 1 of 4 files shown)


kennykerr (approved):
Looks great - thanks!

We should get a dedicated test crate up for numerics under tests/libs rather than this random little test under tests/misc but I can clean that up separately.

Issue #753, opened 2021-04-29
Implementations must support `IMarshal`


kennykerr:
Implementations (e.g. WinRT delegates and those created with the `implement` macro) support `IAgileObject` but `IMarshal` is also required for some languages.

riverar:
No action needed.

Just scribbling a note here that the Real Time Stylus object falls in this bucket. Failure to implement this will result in a `E_INVALIDARG` return from `IRealTimeStylus::AddStylusSyncPlugin`.

kennykerr:
The above example isn't quite right but I have since baked in support for marshaling: #3531

PR #3531, opened 2025-03-06
Add marshaling support for COM implementations

kennykerr:
In many cases, an explicit `IMarshal` interface implementation is not required when implementations support `IAgileObject`. But there are cases when this is insufficient and COM demands an implementation of `IMarshal`.

This update adds a universal implementation of `IMarshal` for all COM objects produced by `windows-rs` including the `implement` macro, delegates, and weak references. This implementation is based on the system's free-threaded marshaler and is implemented identically to the way I originally implemented it for C++/WinRT which has stood the test of time. Specifically, `IMarshal` is implemented as a tear-off to avoid any overhead in the common case where marshaling is not required.

Here's a simple client-server example to illustrate what this enables. Here's the server that uses `CoRegisterClassObject` to register and host a class factory:

```rust
use windows::{Foundation::*, Win32::System::Com::*, core::*};

#[implement(IClassFactory)]
struct Factory;

impl IClassFactory_Impl for Factory_Impl {
    fn CreateInstance(
        &self,
        _: Ref<IUnknown>,
        iid: *const GUID,
        interface: *mut *mut core::ffi::c_void,
    ) -> Result<()> {
        unsafe {
            let unknown: IUnknown = Stringable.into();
            unknown.query(iid, interface).ok()
        }
    }

    fn LockServer(&self, _: BOOL) -> Result<()> {
        Ok(())
    }
}

#[implement(IStringable)]
struct Stringable;

impl IStringable_Impl for Stringable_Impl {
    fn ToString(&self) -> Result<HSTRING> {
        Ok("Stringable".into())
    }
}

const CLSID_SAMPLE: GUID = GUID::from_u128(0x58d39529_4975_4440_b7a6_6d71bfb59559);

fn main() -> Result<()> {
    unsafe {
        CoIncrementMTAUsage()?;
        let factory: IClassFactory = Factory.into();

        CoRegisterClassObject(
            &CLSID_SAMPLE,
            &factory,
            CLSCTX_LOCAL_SERVER,
            REGCLS_AGILE | REGCLS_MULTIPLEUSE,
        )?;

        println!("Press key to end server...");
        getchar();
        Ok(())
    }
}

unsafe extern "C" {
    pub fn getchar() -> i32;
}
```

And here's a client that uses `CoCreateInstance` to remotely connect to the server:

```rust
use windows::{Foundation::*, Win32::System::Com::*, core::*};

const CLSID_SAMPLE: GUID = GUID::from_u128(0x58d39529_4975_4440_b7a6_6d71bfb59559);

fn main() -> Result<()> {
    unsafe {
        CoIncrementMTAUsage()?;

        let stringable: IStringable = CoCreateInstance(&CLSID_SAMPLE, None, CLSCTX_LOCAL_SERVER)?;
        println!("client: {}", stringable.ToString()?);

        Ok(())
    }
}
```

Fixes: #753


Progdrasil:
Hi @kennykerr, thank you for this work! When do you think this will be available in a published `windows-rs` version?

PR #3550, opened 2025-03-18
Mark stale issues and pull requests

kennykerr:


tim-weis:
There's no value to be had from stifling feedback (as showcased [here](https://github.com/microsoft/cppwinrt/issues/1329)).

Anything that can reasonably be delegated to automation ought to. Everything else shan't.

This PR attempts to automate what can't. Can this be rolled back, please?

kennykerr:
I ended up disabling it here as it wasn't working reliably, but I don't agree that it stifles feedback. 

All it does is remind subscribers to keep the conversation alive. One option is to leave issues and PRs open indefinitely but that can just give the false impression that somebody is still going to address or respond to an issue that has long since been forgotten. There are many such abandoned repos. I'd just rather be clear and eventually close such issues and this tool does that. Folks can continue to comment on issues and subscribers will continue to receive those notifications.

Issue #3557, opened 2025-03-24
Hash for windows_registry::Type


allan2:
### Suggestion

Would you be willing to implement `Hash` on `windows_registry::Type`? The use case is collecting into a hash map from `Key::values`.

#589 related

kennykerr:
As with #3558, I'm not sure it necessarily makes sense to implement `Hash` here. To hash values coming from the `ValueIterator` you'd also need hash `Value` and as this is a somewhat low-level polymorphic type it may make more sense to push this to a higher level abstraction that can make a more meaningful decision on what to hash.

kennykerr:
That seems reasonable - feel free to contribute or I'll get around to it when possible.

riverar:
It seems a bit inefficient to hash String+Type pairs to ultimately determine if you support a Type or not, which can be done by simply inspecting the Type directly. But I'm not opposed to the trait either. Thanks for the additional info.

PR #3564, opened 2025-04-01
Derive `Hash` for `Type` in `windows_registry`

allan2:
Fixes: #3557


kennykerr:
Please add a test to `crates\tests\libs\registry\tests\hash.rs` with a basic test just to make sure this works and doesn't get yanked accidentally. Something like this will suffice:

```rust
use windows_registry::*;

#[test]
fn test() {
    let mut set = std::collections::HashSet::<Type>::new();
    set.insert(Type::U32);
}
```

PR #3581, opened 2025-04-19
Add `guid!` macro for compile-time creation of `GUID`s from string literals

yotamofek:
Also, add a const method to `GUID` for parsing string literals. And impl `FromStr` for `GUID` (which makes more sense than a `TryFrom<&str>` impl).

Also, adds a [`LowerHex`](https://doc.rust-lang.org/stable/std/fmt/trait.LowerHex.html) impl for outputting the GUID as a lower-cased string (and an `UpperHex` impl which uses the same formatting as the `Debug` impl)

### DIFF
--- a/crates/libs/core/src/guid.rs
+++ b/crates/libs/core/src/guid.rs
@@ -88,6 +88,60 @@ impl GUID {
             ],
         )
     }
+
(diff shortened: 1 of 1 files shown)


kennykerr:
Thanks for the interest! Please open an issue to discuss new features.

yotamofek:
Will close #3582

PR #3588, opened 2025-04-29
Remove `rayon` dependency

kennykerr:
This is only used to speed up package generation, which is mostly just for the generation of the `windows` and `windows-sys` crates, by parallelizing code generation. Unfortunately, it has proven unreliable on Windows. I'll likely provide a simple parallelizing replacement in the near future.


sivadeilra (approved):

ChrisDenton:
I'm all for simplify the dependency tree but was there a particular issue motivating this?

kennykerr:
> was there a particular issue motivating this

Running `cargo run -p tool_bindings --release` hangs my entire machine pretty regularly. At least once or twice a day I'm forced to hard reboot my dev box... This does not happen with my C++ code generators, which hammer the machine at least as hard.

MRahilly:
Possibly like the file system churn that C++/WinRT implementation headers generate on an older PC.

PR #3595, opened 2025-05-06
Introduce the dedicated `windows-threading` crate

kennykerr:
This update introduces the `windows-threading` crate as a simpler and more efficient way to handle threading on Windows. Although the Standard Library provides basic threading support, the `windows-threading` crate offers additional features and optimizations for Windows-specific threading tasks in a manner that is generally more efficient primarily because it leverages the Windows thread pool rather than necessarily creating new threads. This is a `no_std` crate that can be used with very little overhead and only depends on the `windows-link` crate for access to the Windows threading APIs. 

The existing `windows-future` crate now depends on the `windows-threading` crate for spawning WinRT async operations. I chose to keep this functionality separate as there are cases where the added complexity and overhead of the `windows-future` crate may not be desirable. The `windows-bindgen` crate now also depends on `windows-threading` for speeding up package generation that was previously disabled in #3588. 

The `windows-threading` crate is intentionally simple and easy to use but offers tremendous power to scale up tasks with good data parallelism. I may add more capabilities as need arises. 

The `submit` function submits the closure to the default thread pool. It is similar to the Standard Library `spawn` function but does not require a new thread to be created. It is a very efficient way to run some work in the background and is used by the `windows-future` crate for implementing async operations. Waiting or joining is not currently supported but may be added in the future as a separate `Work` object. 

The `for_each` function calls the closure on each element of the iterator in parallel, waiting for all closures to finish. It is a simple way to parallelize a `for` loop or iterator and is used by the `windows-bindgen` crate for speeding up package generation. 

The `for_each` function relies on the `Pool` struct that represents a private thread pool with its own thread limits. Closures can be submitted to private pool objects rather than the default thread pool to manage concurrency or improve code isolation and the `join` method, or `Drop` implementation, will wait for all closures to run their course. 

Here is a [preview of the crate's readme](https://github.com/microsoft/windows-rs/blob/5d2c8898cd0d4b3810d4bd4b79c3a594ae8cde2e/crates/libs/threading/readme.md) with examples.

The Windows thread pool offers many other facilities but these are especially useful and broadly applicable. For more information on the Windows thread pool see my resources here:

[The Windows Thread Pool and Work](http://msdn.microsoft.com/en-ca/magazine/hh335066.aspx)
[The Thread Pool Environment](http://msdn.microsoft.com/en-us/magazine/hh394144.aspx)
[Thread Pool Cancellation and Cleanup](http://msdn.microsoft.com/en-us/magazine/hh456398.aspx)
[Thread Pool Synchronization](http://msdn.microsoft.com/magazine/hh547107.aspx)
[Thread Pool Timers and I/O](http://msdn.microsoft.com/magazine/hh580731.aspx)
[The Evolution of Threads and I/O in Windows](http://msdn.microsoft.com/magazine/jj883951.aspx)
[10 Practical Techniques to Power Your Visual C++ Apps](https://www.pluralsight.com/courses/ten-techniques-power-visual-cplusplus-apps)


crates/libs/threading/readme.md
Nerixyz, line 1: > The Windows thread pool offers many other facilities but these are especially useful and broadly applicable. For more information on the Windows thread pool see my resources here:

Might be helpful to link some of them in the README, since it's included in the top-level crate documentation - or maybe a small example of the basic functionality.

kennykerr: Updated. Here is a [preview of the readme](https://github.com/microsoft/windows-rs/blob/5d2c8898cd0d4b3810d4bd4b79c3a594ae8cde2e/crates/libs/threading/readme.md).

crates/libs/threading/src/pool.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub fn new() -> Self {
+        let mut e = TP_CALLBACK_ENVIRON_V3 {
+            Version: 3,
+            CallbackPriority: TP_CALLBACK_PRIORITY_NORMAL,
+            Size: core::mem::size_of::<TP_CALLBACK_ENVIRON_V3>() as u32,
+            ..Default::default()
+        };
+
+        unsafe {
+            e.Pool = check(CreateThreadpool(core::mem::zeroed()));
```
ChrisDenton, line 20:
```suggestion
            e.Pool = check(CreateThreadpool(core::ptr::null()));
```

crates/libs/threading/src/pool.rs
```diff
@@ -0,0 +1,76 @@
...
+            try_submit(&*self.0, f);
+        }
+    }
+
+    /// Waits for all submissions to finish.
+    ///
+    /// Dropping the `Pool` will also wait for all submissions to finish.
+    pub fn join(&self) {
+        unsafe {
+            CloseThreadpoolCleanupGroupMembers(self.0.CleanupGroup, 0, core::mem::zeroed());
```
ChrisDenton, line 51:
```suggestion
            CloseThreadpoolCleanupGroupMembers(self.0.CleanupGroup, 0, core::ptr::null_mut());
```

ChrisDenton (approved):
This looks fairly straight forward and is clearly written, thanks! The most tricky thing with this is handling lifetimes but using `'static` or binding to `Pool` are safe options.

Just a couple of small nits.

crates/libs/threading/src/bindings.rs
```diff
@@ -0,0 +1,60 @@
...
+pub type PTP_CLEANUP_GROUP = isize;
+pub type PTP_CLEANUP_GROUP_CANCEL_CALLBACK = Option<
+    unsafe extern "system" fn(
+        objectcontext: *mut core::ffi::c_void,
+        cleanupcontext: *mut core::ffi::c_void,
+    ),
+>;
+pub type PTP_POOL = isize;
+pub type PTP_SIMPLE_CALLBACK = Option<
+    unsafe extern "system" fn(instance: PTP_CALLBACK_INSTANCE, context: *mut core::ffi::c_void),
```
MarijnS95, line 22: Hey @kennykerr, I'm once again generating new `windows-rs` crates against new `win32metadata` and noticed that upstream changed this delegate to receive zero arguments in https://github.com/microsoft/win32metadata/commit/6d6617a8eee57d7c4f6c55d8c23e35064ea69b9e, is that intended/correct?

kennykerr: No, it has two parameters as depicted here.

MarijnS95: That is why I am giving you a heads-up, that this should perhaps be reported and corrected upstream. The change says (https://github.com/microsoft/win32metadata/blob/c7ffce1c97bb724580580f3856b94c6d4ad677ed/scripts/ChangesSinceLastRelease.txt#L5517-L5518):

```
Windows.Win32.System.Threading.PTP_SIMPLE_CALLBACK.Invoke() added
Windows.Win32.System.Threading.PTP_SIMPLE_CALLBACK.Invoke(Instance,Context) removed
```

PR #3603, opened 2025-05-15
Release 64

kennykerr:
# New crates in this release

* The [windows-services](https://crates.io/crates/windows-services) crate provides a simple and safe way to implement Windows services in Rust (#3599). 

* The [windows-threading](https://crates.io/crates/windows-threading) crate provides simple, safe, and efficient access to the Windows threading support (#3595).

# Major version updates to existing crates

* The [windows-metadata](https://crates.io/crates/windows-metadata) crate is a low-level metadata library for ECMA-335, providing both a reader and writer for the ECMA-335 metadata format (#3560) used by .NET, WinRT, and more recently the Win32 metadata. It is technically a brand new implementation but reuses the older crate name from my first attempt at a metadata parser in Rust. The emphasis is on low-level. While it is easy to use, the metadata format is intended for tools to use internally. This implementation is very fast and scalable. 

# Minor version updates to existing crates

The [windows-bindgen](https://crates.io/crates/windows-bindgen) crate has been updated to use the new `windows-threading` crate for parallelism to improve reliability and reduce dependencies (#3588, #3595). 

The [windows-future](https://crates.io/crates/windows-future) crate now uses the new `windows-threading` crate for parallelism (#3595).

The [windows-registry](https://crates.io/crates/windows-registry) crate adds support for key rename (#3567).

The [windows-core](https://crates.io/crates/windows-core), [windows-result](https://crates.io/crates/windows-result), and [windows-strings](https://crates.io/crates/windows-strings) crates no longer require the `windows_debugger_visualizer` option to enable natvis support (#3586).


crates/libs/services/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "windows-services"
-version = "0.0.0"
+version = "0.24.0"
```
tim-weis, line 3: Out of curiosity, why `0.24.0` for a new crate?

kennykerr: I originally attempted to generate namespace-specific crates from metadata. That experiment shaped where we are today in terms of having `windows-bindgen` and other discrete crates. I will likely move all crates to 1.0 at some point and then these versions should be less confusing.

kennykerr: To expand, the experiment generated and published crates like "windows-services" for the APIs in the "Windows.Services" namespace, and many others. The main problem with that approach is that Windows APIs lack robust layering and routinely have bidirectional dependencies which is incompatible with Cargo/Rust.

tim-weis: I feel like we're talking past each other (or I'm missing a crucial detail to follow along). To rephrase the question: Why isn't `windows-services` introduced with version `0.1.0` (just like the other new `windows-threading` crate)? Why `0.24.0` specifically?

kennykerr: Because that's impossible for the reason I mentioned. Have a look at the current version: https://crates.io/crates/windows-services

tim-weis: Oh, now *that* makes perfect sense! I wasn't aware of that period of the evolution, that burnt through makeshift `cargo publish` crates that never were.

Thanks for following up, Kenny.

Issue #3606, opened 2025-05-17
`.natvis` file build failure with vendored sources after upgrading from windows-result 0.3.2 to 0.3.3


wez:
### Summary

I did a cargo update for wezterm today and our Windows builds are now failing:

```
error: couldn't read D:\a\wezterm\wezterm\vendor\windows-result-0.3.3\src\../.natvis: The system cannot find the file specified. (os error 2)
 --> D:\a\wezterm\wezterm\vendor\windows-result-0.3.3\src\lib.rs:2:24
  |
2 | #![debugger_visualizer(natvis_file = "../.natvis")]
  |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `windows-result` (lib) due to 1 previous error
```

In CI, we use `cargo vendor` to enable better caching; it sounds like maybe there is a manifest or ignore issue of some kind that prevents vendoring from working in the latest update?

### Crate manifest

It's too big to copy and paste here :-/

### Additional info

here's the CI job output from the wezterm project:
https://github.com/wezterm/wezterm/actions/runs/15086992317/job/42410788920

wez:
Seems to be a regression introduced by 6a6fd1a61f4ef77667eb1fcaaf9a3e5c1d27fcc7 which unconditionally references that file now, where previously that feature would never be enabled

riverar:
Right, the natvis should always be included now (https://github.com/microsoft/windows-rs/pull/3586). Hmm.

christianhoelzl:
This issue also breaks a vendored use of the latest windows-core v0.61.1, as windows-result v0.3.3 is a transitive dependency here.

kennykerr:
At a cursory glance it looks like a `cargo vendor` bug where it fails to retain the natvis file for some reason. Short term I may have to just drop natvis support from those crates until that is fixed.

tim-weis:
> Short term I may have to just drop natvis support from those crates until that is fixed.

If you have to, please do retain them behind a `windows_debugger_visualizer` configuration gate (like before). They are valuable to me. Indeed, I would appreciate it if you didn't make any changes at all, and rather have clients suffering from a `cargo vendor` bug invest into having *that* issue resolved.

wez:
I'm not personally a fan of punting a resolution for this to cargo vendor; for a widely transitively used dep like this one, thats a very high friction stance to take.

[...]

PR #3607, opened 2025-05-19
`.natvis` workaround for `cargo vendor` 

kennykerr:
This is a quick workaround for what appears to be a `cargo vendor` issue that fails to include some files that are otherwise included by `cargo package`. 

Fixes: #3606


ChrisDenton:
> I couldn't find a way to test cargo vendor as part of the build as it does not appear to support archiving path-based dependencies.

Hm, `cargo package` might be another option for testing as it should work similarly. Remove (or move) the `.git` directory and then run, for example, `cargo package -p windows-result`. That should fail on the current master branch and succeed for this PR.

ChrisDenton (approved):

kennykerr:
Thanks Chris, will debug this when I get a moment - in the meantime #3608 should unblock `cargo vendor` support.

kennykerr:
> Since you are renaming the `.natvis` files could you consider using the crate names instead (i.e., `windows-core.natvis`, `windows-result.natvis`, `windows-strings.natvis`). That makes it easier to know which visualizers are present in a PDB, or loaded when debugging.

Done in #3608

kennykerr:
Thanks for reporting that behavior.

tim-weis:
FYI: The behavior is specific to the [`debugger_visualizer`](https://doc.rust-lang.org/reference/attributes/debugger.html#r-attributes.debugger.debugger_visualizer) attribute. Natvis visualizers for core Rust libraries (like `liballoc.natvis`) are processed differently and retain their respective file names.

Issue #3514, opened 2025-02-25
Add most common `windows-bindgen` options to readme


kennykerr:
Originally posted by @kennykerr in [#3510](https://github.com/microsoft/windows-rs/issues/3510#issuecomment-2678570531)_

MarijnS95:
That'd be awesome. I was just going over another crate migration to `windows-bindgen` and noticed that I had to backtrack from the [_library documentation_ on `windows-bindgen`](http://docs.rs/windows-bindgen) all the way to the `0.61` release tag to find https://github.com/microsoft/windows-rs/pull/3359 again which documents all the new options.

Would be awesome if that could be copy-pasted into docs.rs 👍

MarijnS95:
@kennykerr has anything materialized here? Specifically, I'm curious if there's an option to generate bindings that point into `windows-strings` (for `P(C)STR` & co) which currently point to `windows-core`, but not sure how to set that up with `--reference`.

I expected something like `--reference windows_strings,skip-root,Windows.Win32.Foundation.PCSTR` but it can't find `PCSTR` in that namespace (and `ilspy` only shows `PSTR` there...?).

MarijnS95:
That's the PR I had been looking at, but iirc some things changed since (`--no-deps` is new iirc, and not documented there).

I don't necessarily need to get rid of `windows-core`, but a vocal minority at `iana-time-zone` wants it (requiring use of more complicated and less portable APIs while at it).

Perhaps having a mode where we can pick which crate to use would be useful. For example, `--no-deps` + `--sys` generates bindings against `windows-link` with an MSRV of 1.71. If I want to opt out of that I can remove `--no-deps` and get back to `windows-targets` dependencies, but the `PCSTR` typedef is replaced with `windows_sys::core::PCSTR` too :)

kennykerr:
#3450 added the `--link` option that you can use to control whether it uses `windows-link`, `windows-targets`, or something else. 

The strings are a bit more complicated but I'll give that some more thought.

MarijnS95:
Thanks, that's exactly what I need! Looks like it's available in both `windows-bindgen` `0.60` and `0.61` too 👍

kennykerr:
Sorry for the delay: #3617

MarijnS95:
[...]

PR #3617, opened 2025-06-02
Add `bindgen` argument documentation

kennykerr:
This update adds a description of the most common `bindgen` arguments to the generated documentation. 

Fixes: #3514


ChrisDenton:
Would it be worth adding a quick table at the end? Just listing the major options and a very brief summary.

kennykerr:
Here's what it looks like:

image

Issue #3626, opened 2025-06-12
Should use `extern "system"` as much as possible


Berrysoft:
### Suggestion

`extern "system"` is designed for Windows API, which uses "stdcall" on x86 and "C" on other archs. `windows_targets::link` now distinguishes them again. It becomes very annoying because previously I use `unsafe extern "system" fn` function pointers to store some windows APIs but now they become "C" on non-x86. I don't get the point why you distinguish them.

kennykerr:
I'm not really sure what you're alluding to. Is this in reference to #3622 or something else?

Berrysoft:
I think so. #3622 modifies [crates/libs/targets/src/lib.rs](https://github.com/microsoft/windows-rs/pull/3622/files#diff-6fd35ee92167be2217aca197f703a29f6835760ee209be67a60d7e08e879bbc5).

RalfJung:
I think is is related to my question [here](https://github.com/microsoft/windows-rs/pull/3622#issuecomment-2962027462): for all non-variadic functions, windows-sys should be able to use `extern "system"` on *all* targets: Windows or not, x86-32 or not. For variadic functions it should be able to use `extern "C"` everywhere. (I think we should allow variadic "system" functions and make them behave like "C", which is already implemented but still pending stabilization: https://github.com/rust-lang/rust/issues/136946.)

>  now they become "C" on non-x86

https://github.com/microsoft/windows-rs/pull/3622 didn't change this though... already before that PR, it was "system" on i686-windows and "C" on x86_64-windows. That PR changed some "cdecl" to "C", but this does not seem to be your problem. It also changed the ABI on non-Windows from "system" to "C", but I assume you are building for a Windows target here.

ChrisDenton:
Hm, the link macro could be made to convert "cdecl" to "C" and leave other abis as-is. 

That would avoid the future incompatibilities while allowing "system" to pass through unchanged.

ChrisDenton:
But yes, I'm confused by your PR because it changes something that existed before my PR.

Berrysoft:
> [#3622](https://github.com/microsoft/windows-rs/pull/3622) didn't change this though... already before that PR, it was "system" on i686-windows and "C" on x86_64-windows.

Before that PR, **only when raw-dylib enabled**, it was "C" on x86_64-windows. That PR changed the behavior.

Berrysoft:
> But yes, I'm confused by your PR because it changes something that existed before my PR.

[...]

PR #3627, opened 2025-06-12
Use the given ABI in `link!`

Berrysoft:
What's this all about?

Fixes: #3626

Varargs should not be a problem because they are already `extern "C"`.

### DIFF
--- a/crates/libs/link/src/lib.rs
+++ b/crates/libs/link/src/lib.rs
@@ -5,6 +5,10 @@
 #[cfg(all(windows, target_arch = "x86"))]
 #[macro_export]
 macro_rules! link {
+    // Avoid future-compat lint if bindings are still using the old "cdecl" form of the bindings.
+    ($library:literal "cdecl" $($link_name:literal)? fn $($function:tt)*) => (
+        $crate::link!($library "C" $($link_name)? fn $($function)*);
+    );
     ($library:literal $abi:literal $($link_name:literal)? fn $($function:tt)*) => (
         #[link(name = $library, kind = "raw-dylib", modifiers = "+verbatim", import_name_type = "undecorated")]
         extern $abi {
@@ -18,9 +22,13 @@ macro_rules! link {
 #[cfg(all(windows, not(target_arch = "x86")))]
 #[macro_export]
 macro_rules! link {
+    // Avoid future-compat lint if bindings are still using the old "cdecl" form of the bindings.
+    ($library:literal "cdecl" $($link_name:literal)? fn $($function:tt)*) => (
+        $crate::link!($library "C" $($link_name)? fn $($function)*);
+    );
     ($library:literal $abi:literal $($link_name:literal)? fn $($function:tt)*) => (
         #[link(name = $library, kind = "raw-dylib", modifiers = "+verbatim")]
-        extern "C" {
+        extern $abi {
             $(#[link_name=$link_name])?
             pub fn $($function)*;
         }
@@ -31,8 +39,12 @@ macro_rules! link {
 #[cfg(not(windows))]
 #[macro_export]
 macro_rules! link {
+    // Avoid future-compat lint if bindings are still using the old "cdecl" form of the bindings.
+    ($library:literal "cdecl" $($link_name:literal)? fn $($function:tt)*) => (
+        $crate::link!($library "C" $($link_name)? fn $($function)*);
+    );
     ($library:literal $abi:literal $($link_name:literal)? fn $($function:tt)*) => (
-        extern "C" {
+        extern $abi {
             pub fn $($function)*;
         }
     )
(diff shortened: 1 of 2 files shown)


crates/libs/targets/src/lib.rs
```diff
@@ -40,19 +40,6 @@ macro_rules! link {
...
-        extern "C" {
-            $(#[link_name=$link_name])?
-            pub fn $($function)*;
-        }
-    )
-}
-
 /// Defines an external function to import.
 #[cfg(all(not(windows), not(windows_raw_dylib)))]
 #[macro_export]
```
RalfJung, line 57: I feel like at this point you may as well remove the non-Windows fallback here as well, and use the same ABI everywhere.

crates/libs/targets/src/lib.rs
```diff
@@ -28,7 +28,7 @@ macro_rules! link {
 }
```
RalfJung, line 36: And there's another `extern "C"` block here.

crates/libs/targets/src/lib.rs
Berrysoft: I'm not sure about whether "system" works well on non-Windows targets...

crates/libs/targets/src/lib.rs
RalfJung, line 1: This isn't even the same file that https://github.com/microsoft/windows-rs/pull/3622 changed...

Berrysoft: It is.

crates/libs/targets/src/lib.rs
Berrysoft: OK. "system" should be the same as "C" on other targets.

crates/libs/targets/src/lib.rs
RalfJung: It is not. `crates/libs/targets/src/lib.rs` != `crates/libs/link/src/lib.rs`. Or did the file get moved around? It'd be helpful if you could explain a bit more rather than just bluntly asserting that I am wrong.

Berrysoft: It is. Both files are modified in that PR...

RalfJung: Ah, #3622 changed both files, the `target` one just drowned in the other edits. My bad.

crates/libs/targets/src/lib.rs
```diff
@@ -20,19 +20,6 @@ macro_rules! link {
 macro_rules! link {
     ($library:literal $abi:literal $($link_name:literal)? fn $($function:tt)*) => (
         #[link(name = $library, kind = "raw-dylib", modifiers = "+verbatim")]
```
RalfJung, line 30: FWIW, since all versions of the macro use the same ABI now this could now be done with `cfg_attr` rather than having separate macros.

crates/libs/targets/src/lib.rs
RalfJung: Exactly, "system" is defined as 
- "stdcall" on win32
- "C" everywhere else

crates/libs/targets/src/lib.rs
Berrysoft: Well, we put `cfg`s out of macros to prevent them from leaking to the users. We cannot use `cfg_attr` here.

ChrisDenton:
Could you squish your commits? And maybe change the PR title to reflect the current state.

crates/libs/targets/src/lib.rs
RalfJung: Ah, I guess `cfg_attr`  inside the macro would be evaluated too late, it'd be evaluated in the user crate... that's unfortunate.

crates/libs/link/src/lib.rs
```diff
@@ -5,6 +5,9 @@
 #[cfg(all(windows, target_arch = "x86"))]
 #[macro_export]
 macro_rules! link {
+    ($library:literal "cdecl" $($link_name:literal)? fn $($function:tt)*) => (
+        $crate::link!($library "C" $($link_name)? fn $($function)*);
+    );
```
RalfJung, line 11: I still don't understand which problem this solves. I guess it will fix the lint if you update the macro but don't update the bindings? If that's the only point, this warrants a comment.

That doesn't seem worth the extra code here though. People can just update the bindings to fix the lint. :shrug:

Berrysoft: Well, @ChrisDenton suggests that there're still some old code (authored by other users?) uses "cdecl". This is for compatibility because "cdecl" raises a warning and will become an error soon.

RalfJung: Okay so it is about the lint, thanks. That seems worth a comment.
```suggestion
    // Avoid future-compat lint if bindings are still using the old "cdecl" form of the bindings.
    ($library:literal "cdecl" $($link_name:literal)? fn $($function:tt)*) => (
        $crate::link!($library "C" $($link_name)? fn $($function)*);
    );
```

Berrysoft: Comments added.

kennykerr:
Closing in favor of #3628 - thanks for jumping in with a fix!

PR #3628, opened 2025-06-12
Revert ABI changes to the `link` macros

ChrisDenton:
This simply reverts the changes to the `link` macros made in https://github.com/microsoft/windows-rs/pull/3622. A future version of these crates could change how linking works but let's not do that now.

Fixes: #3626


kennykerr (approved):

kennykerr:
Thanks! Once green I'll republish the crates to avoid `windows-sys` and friends from explicitly depending on a yanked crate.

Issue #3636, opened 2025-06-22
Combination of features `Networking_BackgroundTransfer` and `Security_Credentials` does not build without `deprecated`


joshtriplett:
### Summary

Steps to reproduce:

```text
/tmp$ cargo new testcrate
    Creating binary (application) `testcrate` package
note: see more `Cargo.toml` keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
/tmp$ cd testcrate/
/tmp/testcrate$ cargo add windows --features Networking_BackgroundTransfer,Security_Credentials
    Updating crates.io index
      Adding windows v0.61.3 to dependencies
             Features:
             + Foundation
             + Networking
             + Networking_BackgroundTransfer
             + Security
             + Security_Credentials
             + std
             688 deactivated features
    Updating crates.io index
     Locking 15 packages to latest Rust 1.89.0-nightly compatible versions
/tmp/testcrate$ cargo build --target x86_64-pc-windows-gnu -r
   Compiling proc-macro2 v1.0.95
   Compiling unicode-ident v1.0.18
   Compiling windows-link v0.1.3
   Compiling windows-strings v0.4.2
   Compiling windows-result v0.3.4
   Compiling windows-threading v0.1.0
   Compiling quote v1.0.40
   Compiling syn v2.0.104
   Compiling windows-interface v0.59.1
   Compiling windows-implement v0.60.0
   Compiling windows-core v0.61.2
   Compiling windows-future v0.2.1
   Compiling windows-collections v0.2.0
   Compiling windows-numerics v0.2.0
   Compiling windows v0.61.3
error[E0308]: mismatched types
    --> /home/josh/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/windows-0.61.3/src/Windows/Networking/BackgroundTransfer/mod.rs:1489:20
     |
1489 |             Group: Group::<Identity, OFFSET>,
     |                    ^^^^^^^^^^^^^^^^^^^^^^^^^ expected `usize`, found fn item
     |
     = note: expected type `usize`
             found fn item `unsafe extern "system" fn(*mut c_void, *mut *mut c_void) -> HRESULT {Group::<Identity, OFFSET>}`

error[E0308]: mismatched types
    --> /home/josh/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/windows-0.61.3/src/Windows/Networking/BackgroundTransfer/mod.rs:1490:23
     |
[...]

PR #3638, opened 2025-06-23
Remove "deprecated" feature

kennykerr:
The WinRT `DeprecatedAttribute` has always been challenging to support in Rust due to the fact that it can be applied to types and/or type members. As you can imagine, this impacts Rust feature generation recursively. Add the complexity of balancing support for calling such APIs (where only some interface methods may be available) and implementing such APIs (where all methods and dependent types are required for a complete implementation). 

This "deprecated" metadata is only available for a small portion of the overall Windows API surface area, but is disproportionally complicated to support. I mulled over a few different options but ultimately decided that I am just going to remove it entirely as none of the options I've explored over the years work very effectively or reliably. I came to the same conclusion for [C++/WinRT](https://github.com/microsoft/cppwinrt). I already filter out larger APIs that are almost or entirely deprecated from the [windows](https://crates.io/crates/windows) crate. I'll probably add a tool to scan namespaces to find any additional namespaces that can be excluded entirely from the pre-generated `windows` crate.

As always, you can use [windows-bindgen](https://crates.io/crates/windows-bindgen) to generate just the APIs you need. Whether they are included in the `windows` crate, deprecated or otherwise. 

Note that this PR does not remove any APIs - it merely removes the "deprecated" feature. 

Fixes: #3636

### DIFF
--- a/crates/libs/bindgen/src/config/cfg.rs
+++ b/crates/libs/bindgen/src/config/cfg.rs
@@ -34,11 +34,10 @@ pub fn write_arches<R: HasAttributes>(row: R) -> TokenStream {
 #[derive(Default)]
 pub struct Cfg {
     features: BTreeSet<&'static str>,
-    deprecated: bool,
 }
 
 impl Cfg {
-    pub fn new<R: HasAttributes>(row: R, dependencies: &TypeMap, config: &Config<'_>) -> Self {
+    pub fn new(dependencies: &TypeMap, config: &Config<'_>) -> Self {
         let features: BTreeSet<&'static str> = dependencies
             .keys()
             .filter_map(|tn| {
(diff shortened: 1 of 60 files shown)


joshtriplett:
Is it intentional that, in cases where there's code under `#[cfg(not(feature = "deprecated"))]` and code under `#[cfg(feature = "deprecated")]`, this PR removes the *former* code? What was the polarity of the "deprecated" feature, and were the "not deprecated" branches the right ones to remove?

kennykerr:
Yes, those slots are the placeholders for the vtable so that any subsequent virtual function offsets are preserved whether or not whatever features are enabled or disabled. So removing the "deprecated" feature - if it is the only feature withholding that virtual function - will also remove the placeholder.

PR #3646, opened 2025-06-30
Fixes #3641

### LINKED ISSUE #3641: Add support for running a service exe as a console app.
### Suggestion

All of the Windows services that I've ever written in C or C++ could be run as a console application for one reason or another.

The `windows-services` crate could easily allow this functionality, and already detects if the process was started as a console app: https://github.com/microsoft/windows-rs/blob/6437f5e1e02c287fd19ee3cae3d75a482e7ed5e0/crates/libs/services/src/lib.rs#L132

There's more than 1 way to add this feature, but I think that there is 1 way that breaks the least: Add a `.allow_console()` (or choose a better name) to `Service` that when called allows `.run()` to return if the check on the call to `StartServiceCtrlDispatcherW()` (linked above) sees that it was run on the console. If the `.allow_console()` function is never called, then keep the existing functionaility.

The state in `Service` is all private, so adding to it would only affect code in the same file, but a bit in the existing `u32` could be used, if that's acceptable.

Please add this functionality, as I can't use it without this feature.

Type1J:
This fix adds a way to run an executable as a service or from the command line, which was previously not possible.

There is a call that sets a flag to opt in to this behavior, so the existing functionally should not change.

Fixes: #3641

### DIFF
--- a/crates/libs/services/src/lib.rs
+++ b/crates/libs/services/src/lib.rs
@@ -87,25 +87,37 @@ pub fn set_state(state: State) {
 
 /// A service builder, providing control over what commands the service supports before the service begins to run.
 #[derive(Default)]
-pub struct Service(u32);
+pub struct Service {
+    api_flags: u32,
+    can_run_without_scm: bool,
+}
 
 impl Service {
     /// Creates a new `Service` object.
     ///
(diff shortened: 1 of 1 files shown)


Type1J:

Issue #1881, opened 2022-07-08
Feature request: `*-pc-windows-gnullvm` targets


mati865:
### Motivation

`*-pc-windows-gnullvm` is a new family of Tier 3 Rust targets (currently only `x86_64` and `aarch64`) that need `windows-rs` support to be usable. Some discussions happened over https://github.com/microsoft/windows-rs/pull/1846

### Drawbacks

More maintenance work.

### Rationale and alternatives

To match behaviour of current `*-windows-msvc` and `*-windows-gnu` targets new crates would have to be made, solely for `*-windows-gnullvm`.
Alternatively it could be implemented using `raw-dylib` feature once it stabilises (there is no any E.T.A.) putting any work on these targets on hold.

### Additional context

List of remaining work:
- [ ] adding new target crates with tool to generate imports - I can do it soon
- [ ] testing on CI - this is complicated as there are no prebuilt binaries available yet, I plan to provide this target for CLANG64 subsystem at MSYS2 once it becomes usable (is able to build most popular crates - so it requires windows-rs support), until then I can prepare tarballs with those targets for use on the CI

mati865:
Hello, are there plans to release `windows-sys` version with the new targets?

kennykerr:
Yes, I'm just waiting for one more fix from win32 metadata (https://github.com/microsoft/win32metadata/pull/1276) and then I plan to release an update early this week.

kennykerr:
`windows-sys` 0.42.0 has now been published!

PR #3653, opened 2025-07-08
Test *-pc-windows-gnullvm host toolchains

mati865:
A bit overdue but addresses the last checkbox in https://github.com/microsoft/windows-rs/issues/1881


kennykerr:
Thanks, I'm not sure we need to test this in windows-rs as we've mostly moved to `raw-dylib` by default so its less of an issue specifically about windows-rs and more of a Rust toolchain issue to support different targets.

mati865:
Fair enough, I'll close it then.

Issue #3652, opened 2025-07-08
`windows-core::GUID::new` should call UuidCreate directly rather than bounce through CoCreateGuid


Robo210:
### Suggestion

CoCreateGuid is in ole32.dll, and is a simple forwarder to UuidCreate in rpcrt4.dll. Swapping GUID::new to call UuidCreate directly would allow people to use GUIDs for things outside of COM without needing to link against the COM libraries.

ole32.dll is not always available in every Windows edition, but rpcrt4.dll currently is. It would be nice for the GUID class to not statically link to ole32.dll for those of us targeting the smaller Windows editions.

riverar:
Sounds reasonable.

Can confirm on my system that `CoCreateGuid` forwards to `combase.dll` by way of `api-ms-win-core-com-l1-1-0` and the implementation unconditionally calls `rpcrt4!UuidCreate`. (This also means the GUIDs from `GUID::new` are type 4 and not guaranteed to be globally unique.)

kennykerr:
No objection to this targeted fix but generally speaking this will be a bit of a whack-a-mole as other APIs crop up that may not be available on different Windows editions. Still, happy to make the change if it unblocks adoption for you.

Robo210:
The thought of playing whack-a- mole does concern me. I have a hunch though that the number of thin wrappers like CoCreateGuid is quite limited, and the number of exports that purely forward to another dll is larger but easily knowable. And I believe if it went a step further and linked against an API Set rather than a lib, the latter case could be handled easier (aka I know I need a bunch of functions officially exported from advapi32 that are actually hosted elsewhere, and advapi32 is usually first in the chopping block for the smaller Windows editions).

I was surprised that the crate uses named dlls rather than API Set names. There's probably some compat issues with older versions that don't support API Sets, but it also means newer, smaller OS versions need the reverse forwarders, which also aren't always available even when the needed API Set is.

My immediate concern though is that I need to call ETW functions that take a GUID parameter but I never actually call 'new'. As I understand it the linker must be able to resolve all references in the obj (e.g. to ole32.lib) before the optimizer can strip out the new function as unused and drop the ole32 dependency. And if it fails to optimize it away, then my program will fail to launch on any edition without COM support, even though I don't need or want COM support in the first place. All so I can pass a simple GUID struct to a function.

riverar:
[...]

PR #3656, opened 2025-07-09
Replace `CoCreateGuid` with `UuidCreate` to avoid `ole32.dll` dependency

kennykerr:
Fixes: #3652


crates/libs/core/src/guid.rs
```diff
@@ -21,7 +21,14 @@ pub struct GUID {
...
     pub fn new() -> Result<Self> {
-        unsafe { imp::CoCreateGuid() }
+        let mut guid = Self::zeroed();
+        let result = unsafe { imp::UuidCreate(&mut guid as *mut _ as _) };
+
+        if result == 0 {
+            Ok(guid)
+        } else {
+            Err(Error::from_hresult(HRESULT::from_win32(result as u32)))
+        }
```
riverar: `UuidCreate` returns multiple success values (e.g., `RPC_S_UUID_LOCAL_ONLY`). I believe the `CoCreateGuid` wrapper swallows this one specifically and returns `S_OK`. Double-checking now.

Robo210 (approved):

crates/libs/core/src/guid.rs
riverar: On my copy of 25H2 26200.5670, I verified it has a check for `RPC_S_UUID_LOCAL_ONLY || RPC_S_OK` and returns `S_OK`.

See also: `onecore\com\combase\class\cocrguid.cxx`

kennykerr: Yes, I see that. I'm not sure whether that's significant to preserve but I suppose I can do that easily enough. 👍

kennykerr: OK, I've preserved that behavior.

riverar (approved):

PR #3657, opened 2025-07-10
Add simple C# interop sample/test

kennykerr:
I'd like to add some more interesting C# interop samples and tests (much like I already have for C++) but for now this is just a tiny example to prove out that it works in the build without overly complicating windows-rs - let's see. 🙃

There's a new Rust `cdylib` library crate that will produce a DLL with a simple exported function and a second Rust library crate that just acts as a driver to build and run a simple C# console app that then calls the exported function. This lets us test it as follows:

```
cargo test -p csharp_client
```

It will also be validated as part of the CI build automatically. 

This still seems a little cumbersome but its at least manageable. It does depend on "dotnet.exe" and the .NET SDK being installed so we'll see if the default GitHub runners include that...


kennykerr:
I need to figure out how to pass the Rust `target_arch` to `dotnet` to satisfy the build.

riverar (approved):

PR #3660, opened 2025-07-13
windows: Use new `hints.mostly-unused`

joshtriplett:
Most users of the `windows` crate will use a fraction of its API surface
area.

Nightly rustc provides an option `-Zhint-mostly-unused` to tell it to
defer as much compilation as possible, which provides a substantial
performance improvement if most of that compilation doesn't end up
happening. Cargo plumbs this option through using the new `[hints]`
table. This will cause users of the `windows` crate to default to
setting `hint-mostly-unused`. (Top-level crates can override this if
they wish, using a new profile option.)

Note that setting this hint does not increase the MSRV of the Windows
crate, as old versions of Cargo will ignore it. New versions of Cargo
will respect it automatically (and, until we stabilize it, Cargo will do
nothing unless you pass `-Zprofile-hint-mostly-unused` to cargo).

Some sample performance numbers:
| **Dependency Crate** | **Before** | **`hint-mostly-unused`** | **Delta** |
| :- | -: | -: | -: |
| `windows`, all Graphics/UI features | 18.3s | 10.7s | -42% |
| `windows`, all features | 3m 48s | 2m 55s | -23% |

PR #3661, opened 2025-07-14
Hopefully speed up the test workflow via `hint-mostly-unused`

kennykerr:
Based on #3660, I'm trying to see if we can cut down the time it takes to run the nightly test workflows. That should also prove out the worthiness of this optimization. 

Assuming I got the yml syntax right...


kennykerr:
Closing as not beneficial, given the discussion in #3660.

Issue #3610, opened 2025-05-21
Support extended control codes with `windows-services`


edvard-pettersen-elas-ai:
### Suggestion

Currently the `windows-services` crate only supports regular service control codes due to using `RegisterServiceCtrlHandlerW`. Having support for the extended control codes would be needed for certain types of events, which requires the event handler to be registered with `RegisterServiceCtrlHandlerExW`.

Specifically, the support for the following extended control codes are requested:
* `SERVICE_CONTROL_DEVICEEVENT`
* `SERVICE_CONTROL_HARDWAREPROFILECHANGE`
* `SERVICE_CONTROL_POWEREVENT`
* `SERVICE_CONTROL_SESSIONCHANGE`
* `SERVICE_CONTROL_TIMECHANGE`
* `SERVICE_CONTROL_TRIGGEREVENT`
* `SERVICE_CONTROL_USERMODEREBOOT`

kennykerr:
That seems reasonable - the main thing right now is to come up with a way to test `windows-services` as part of the build - I have some ideas but I need to get that going before I can entertain more features. I will try to get to that asap.

kennykerr:
Sorry for the delay, I think I'm closer to a fix now - should have a PR out in the next day or two.

kennykerr:
Please take a look at #3662 and in particular the tests and samples and let me know if you have any comments. Thanks for the suggestion!

kennykerr:
An update has now been published: https://github.com/microsoft/windows-rs/releases/tag/68

PR #3662, opened 2025-07-14
Windows service improvements for hosting, testing, fallback, and extended commands

kennykerr:
This update provides a major overhaul of the [windows-services](https://crates.io/crates/windows-services) crate and addresses a number of issues to support more scenarios and features while retaining the simplicity of the `Service` builder by default. 

* Provides the foundation for supporting extended controls by switching from `RegisterServiceCtrlHandlerW` to `RegisterServiceCtrlHandlerExW` internally.
* Provides a new `Extended` variant for the `Command` enum to allow services to receive extended control information.
* Provides a new `can_accept` builder method to specify other specific commands to accept.
* Provides a new `can_fallback` builder method to allow the service to provide specific handling when the service is not started by the Service Control Manager.
* The `run` method is no longer fatal so that a caller may decide to gracefully handle and detect when the service is not started by the Service Control Manager.
* Provides a new `handle` method to retrieve the handle representing the service. This may be used to call other service-related functions like `RegisterPowerSettingNotification`. 

Additional samples and tests are now provided thanks to the greater flexibility afforded by the new implementation. 

Fixes: #3610 
Fixes: #3641


edvard-pettersen-elas-ai:
@kennykerr, the API looks good, but there are some additional steps that are required to support `SERVICE_CONTROL_POWEREVENT`. After having specified `can_accept(SERVICE_ACCEPT_POWEREVENT)`, the application also needs to specify which power events to register for notification by calling `RegisterPowerSettingNotification`. This requires access to the handle returned by `RegisterServiceCtrlHandlerExW`, which doesn't seem accessible in the current proposal.

kennykerr:
https://github.com/microsoft/windows-rs/pull/3662/commits/37c0d93901eb2d24106de673202cc8df2b80be1a adds a `handle` method that you can use to retrieve the service handle. You might for example use it when responding to `Command::Start` to call `RegisterPowerSettingNotification`. Something like this:

```rust
let mut power_handle = 0;

Service::new()
    .can_stop()
    .can_accept(SERVICE_ACCEPT_POWEREVENT)
    .run(|service, command| {
        if command == Command::Start {
            power_handle = unsafe {
                RegisterPowerSettingNotification(
                    service.handle(),
                    &GUID_GLOBAL_USER_PRESENCE,
                    DEVICE_NOTIFY_SERVICE_HANDLE,
                )
            };
        }

        if let Command::Extended(command) = command {
            if command.control == SERVICE_CONTROL_POWEREVENT {
                debug_assert!(command.ty == PBT_POWERSETTINGCHANGE);
                let data = unsafe { &*(command.data as *const POWERBROADCAST_SETTING) };
                debug_assert!(&data.PowerSetting == &GUID_GLOBAL_USER_PRESENCE);
                debug_assert!(data.DataLength == 4);

                let data = unsafe {
                    std::slice::from_raw_parts(data.Data.as_ptr(), data.DataLength as usize)
                };

                let data = u32::from_le_bytes(data.try_into().unwrap());
                writeln!(log, "Present: {data}").unwrap();
            }
        }
    })
    .unwrap();
```

kennykerr:
Blocked on #3665

Issue #3650, opened 2025-07-03
The `windows` crate's "std" feature should enable dependent "std" features


kennykerr:
> Figured it out: I needed to add an explicit dependency to the `windows-future` crate to enable its `std` feature, since the `IntoFuture` impl is gated behind that feature, and `windows`'s `std` feature does not enable it. 

 _Originally posted by @fenhl in [#3197](https://github.com/microsoft/windows-rs/issues/3197#issuecomment-3031392685)_

PR #3666, opened 2025-07-16
The `windows` crate's "std" feature should enable dependent "std" features

kennykerr:
This adds the missing conditional "std" dependencies and also adds a test to verify that this behavior is observed.

Fixes: #3650


kennykerr:
I take that back - this is blocked on #3668 - not sure why this bugged surfaced only now but there it is.

PR #3668, opened 2025-07-16
Fix use-after-free in json_validator sample

riverar:
`String::from_utf8_lossy` returns a `Cow<'_, str>` that references the original memory when the input is considered valid UTF-8. Calling `.into_owned()` forces allocation of an owned `String`, allowing us to safely free the FFI memory with `CoTaskMemFree`.

### DIFF
--- a/crates/samples/components/json_validator/src/lib.rs
+++ b/crates/samples/components/json_validator/src/lib.rs
@@ -316,8 +316,8 @@ fn sanitized_value() {
         );
         let sanitized = std::slice::from_raw_parts(sanitized_alloc, sanitized_len);
         let sanitized = String::from_utf8_lossy(sanitized);
-        CoTaskMemFree(Some(sanitized_alloc as _));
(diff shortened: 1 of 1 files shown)


crates/samples/components/json_validator/src/lib.rs
```diff
@@ -316,8 +316,8 @@ fn sanitized_value() {
         );
         let sanitized = std::slice::from_raw_parts(sanitized_alloc, sanitized_len);
         let sanitized = String::from_utf8_lossy(sanitized);
```
kennykerr: Maybe a comment like "This does not in fact return a `String`" or just something specific like `.into_owned()` to force the `String` to be allocated. Just want to leave a breadcrumb but I guess the PR does that so its probably fine as is.

kennykerr (approved):
Well spotted!

crates/samples/components/json_validator/src/lib.rs
riverar: Happy to do that, esp. since it's a sample.

PR #3673, opened 2025-07-21
Add MSYS2 package cache

riverar:
Optimizes our composite action with the addition of a weekly MSYS2 package cache, saving 1+ minutes per call.

I tested full MSYS2 environment caching (~1.5GiB) but found that cache restore time cancelled out all savings.


riverar:
Didn't achieve the gains I saw in smaller scale testing. Closing for now.

PR #3675, opened 2025-07-24
Apply clippy workflow to all but tests

kennykerr:
Right now the clippy workflow only validates libraries and tools. This means that a lot of samples and test crates go unchecked. Although all crates are tested, clippy still points out some more pedantic but often valuable suggestions. Clippy cannot be run on tests at this time as there are too many false positives but I think it should be doable to validate everything else.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -28,6 +28,126 @@ jobs:
         uses: ./.github/actions/fix-environment
       - name: Check cppwinrt
         run:  cargo clippy -p cppwinrt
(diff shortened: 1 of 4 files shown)


riverar:
After https://github.com/microsoft/windows-rs/pull/3677 is merged, adjust fix-environment to below and you should be good to go.

```yml
- name: Fix environment
  uses: ./.github/actions/fix-environment
  with:
    target: x86_64-pc-windows-msvc
```

kennykerr:
Thanks for the build help @riverar 👍

PR #3687, opened 2025-07-29
Fix parallel build reliability with `build-dependencies`

kennykerr:
Many of the binary (DLL) component tests depend on sibling crates that don't actually need a crate dependency but need predictable build order to ensure that the component's winmd is generated. Cargo isn't very good at this in general but I did notice that some of my test crates were relying on `dependencies` rather than `build-dependencies` resulting in frequent parallel build failures as the build scripts routinely ran before the crate dependency's build script resulting in frequent "missing winmd" errors.

While not foolproof, this should make it a lot easier to quickly run "cargo check --all" or "cargo clippy --all" and get more of the many many crates in this repo to compile successfully in parallel.

### DIFF
--- a/crates/samples/components/json_validator_winrt_client/Cargo.toml
+++ b/crates/samples/components/json_validator_winrt_client/Cargo.toml
@@ -10,6 +10,4 @@ windows = { workspace = true, features = ["Win32_Foundation"] }
 
 [build-dependencies]
 windows-bindgen = { workspace = true }
-
-[dependencies]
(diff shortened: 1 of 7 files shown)


kennykerr:
Ugh, that build failure reminds me why I didn't do this from the start. A build dependency rather than a crate dependency means the `target` output directory for the DLL is unpredictable hence "Cargo isn't very good at this in general". 🙄

kennykerr:
For reference, the issue where I originally reported this: https://github.com/rust-lang/cargo/issues/9661

PR #3691, opened 2025-07-31
Address most Clippy warnings in test code

kennykerr:
By default, windows-rs does not run Clippy against tests but I have been steadily working on the noisy issues and I think this should be the last batch before I can flip the switch and let the Clippy workflow keep the tests honest. 

This update doesn't change the yml workflow as that will require one test to be disabled or removed. I'll do that separately.


kennykerr:
Remaining test is blocked on https://github.com/rust-lang/rust/issues/137018#issuecomment-3140607760 and https://github.com/microsoft/rust_debugger_test/issues/14

PR #3692, opened 2025-07-31
Turn on Clippy testing of all crates and tests

kennykerr:
Following #3691, this turns on all of the remaining Clippy test coverage. 

I do [expect it to fail](https://github.com/microsoft/windows-rs/pull/3691#issuecomment-3140639900) for the time being, but may highlight other potential issues.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -29,176 +29,360 @@ jobs:
         with:
           target: x86_64-pc-windows-msvc
       - name: Check cppwinrt
-        run:  cargo clippy -p cppwinrt
+        run:  cargo clippy -p cppwinrt --tests
       - name: Check csharp_client
-        run:  cargo clippy -p csharp_client
+        run:  cargo clippy -p csharp_client --tests
       - name: Check csharp_component
-        run:  cargo clippy -p csharp_component
+        run:  cargo clippy -p csharp_component --tests
       - name: Check helpers
-        run:  cargo clippy -p helpers
+        run:  cargo clippy -p helpers --tests
       - name: Check sample_bits
-        run:  cargo clippy -p sample_bits
+        run:  cargo clippy -p sample_bits --tests
       - name: Check sample_com_uri
-        run:  cargo clippy -p sample_com_uri
+        run:  cargo clippy -p sample_com_uri --tests
       - name: Check sample_consent
-        run:  cargo clippy -p sample_consent
+        run:  cargo clippy -p sample_consent --tests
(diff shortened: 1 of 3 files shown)


kennykerr:
Confirmed there's the failure: https://github.com/microsoft/windows-rs/actions/runs/16656986097/job/47144369160?pr=3692

kennykerr:
Closing for now.

kennykerr:
Reopening now that https://github.com/rust-lang/rust/issues/137018#issuecomment-3140607760 is fixed and I can use `allow(unsupported_calling_conventions)` as a workaround.

kennykerr:
So even though it appears that this lint has now been fixed in nightly, it does not work in stable, which is understandable if the fix has not made it into the stable builds yet. The error message is however a little odd:

```
error: lint `unsupported_calling_conventions` has been removed: converted into hard error
 --> crates\tests\misc\debugger_visualizer\tests\test.rs:3:10
  |
3 | #![allow(unsupported_calling_conventions)]
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `-D renamed-and-removed-lints` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(renamed_and_removed_lints)]`
```

Previously, the nightly toolchain stated that [it was an unknown lint](https://github.com/rust-lang/rust/issues/137018#issuecomment-3140607760), but the stable toolchain says it has been removed. 🤷 

fyi @RalfJung

Anyway, I think I may just remove this test outright until and unless a supportable alternative is found.

RalfJung:
https://github.com/rust-lang/rust/issues/137018 hasn't yet propagated to stable, so naturally the lint doesn't work on stable. It'll be in the release coming out this week.

RalfJung:
The message about unknown/removed lints is just a lint, so it's easy to support this across releases: just allow `renamed_and_removed_lints`. The error even tells you that much. You just can't expect lints to behave the same in future rustc versions, that has always been the case. `-D warnings` is *not* a configuration that is stable under upgrades or downgrades.

kennykerr:
Sure, I understand. I think I'm mostly just observing that it sounds a little confusing that a lint that does not exist has been removed or renamed. Or maybe I don't understand. 🙃

RalfJung:
Yeah that's fair. There used to be a lint of that name which got removed and then we re-enabled it for this 2nd round of ABI deprecations. That's why old rustc say "removed".

And then separately there was a bug you noticed where even nightly rustc called this an "unknown" lint.

PR #3696, opened 2025-08-02
Don't implement `Send` for `windows_threading::Pool`

### LINKED ISSUE #3695: windows-threading lifetime soundness issue
### Summary

After looking into #3694, I was confused how the windows_threading::Pool example worked vs (scoped) threads without unsafe. I suspected there may be a lifetime bug in pool.submit and i was able to get a segfault when executing an example.

```
PS C:\Users\prussian\source\example-winthread> cargo run
   Compiling example-winsvc v0.1.0 (C:\Users\prussian\source\example-winthread)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.24s
     Running `target\debug\example-winsvc.exe`
error: process didn't exit successfully: `target\debug\example-winsvc.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

see example code below.

### Crate manifest

```TOML
[package]
name = "example-winthread"
version = "0.1.0"
edition = "2024"

[dependencies]
windows-threading = "0.1"
```

### Crate code

```Rust
fn main() {
    use windows_threading::Pool;
    let pool = Pool::new();
    pool.set_thread_limits(1, 2);
    pool.submit(|| {
        let s = "Hello, world!";
        pool.submit(|| println!("{s}"));
    });
    pool.join();
}
```

ChrisDenton:
This is a quick fix #3695 that avoids changing the API except as necessary but it does restrict calling `submit` to a single thread per pool. Hopefully this will allow most current code to keep working.

In the future `Pool` could have some way to allow submitting `'static` closures from other threads or a way to scope submits to ensure callbacks finish running before the scope ends.


ChrisDenton:
Ok, looking at this more I do think we need a scope API to ensure this is robust so I've changed this PR to implement that,

While investigating this (and part of what drove this conclusion) I found that the service threading sample is not technically sound. Specifically these lines:

https://github.com/microsoft/windows-rs/blob/897ea820d050d43844dff905ecade78462760f1a/crates/samples/services/thread/src/main.rs#L16-L20

`service` is a reference whose lifetime ends when the callback returns. However, here it's being sent to another thread which can potentially outlive the callback.

I think this requires changes in the service API to support either a reference counted `Service` object or else some other way to retrieve the `Service` from other threads.

crates/samples/services/thread/src/main.rs
```diff
@@ -1,44 +1,54 @@
...
+
 fn main() {
-    let pool = Pool::new();
+    let pool: Pool = Pool::new();
     pool.set_thread_limits(1, 1);
 
-    Service::new()
+    let service_original = Arc::new(RwLock::new(Service::new()));
+    let service = Arc::clone(&service_original);
+    service_original
```
kennykerr, line 12: This seems overly complicated. The original `Service::run` terminated the process upon completion so that the `run` method was `-> !` and the lifetime was much simpler. This was changed in #3662 to make it possible to test `Service` but perhaps that was a mistake.

kennykerr:
I definitely found a friction between Rust's lifetime model and the various restrictions in the threadpool APIs. I much prefer the addition of the `scope` API to restricting the pool to being single-threaded since the threadpool API itself is not single-threaded. I'll play around with this a bit but that seems like the right direction.

crates/tests/libs/threading/tests/pool.rs
```diff
@@ -42,14 +43,15 @@ fn multi() {
...
+        for _ in 0..10 {
+            pool.submit(|| {
+                windows_threading::sleep(10);
+                let mut writer = set.write().unwrap();
+                writer.insert(windows_threading::thread_id());
+            })
+        }
+    });
 
     pool.join();
```
kennykerr: I think this `join` is unnecessary. 
```suggestion
```

crates/tests/libs/threading/tests/pool.rs
```diff
@@ -61,23 +63,22 @@ fn single() {
...
+    pool.scope(|pool| {
+        for _ in 0..10 {
+            pool.submit(|| {
+                let mut writer = set.write().unwrap();
+                writer.insert(windows_threading::thread_id());
+            })
+        }
+    });
 
     pool.join();
```
kennykerr: I think this `join` is unnecessary. 
```suggestion
```

crates/tests/libs/threading/tests/pool.rs
```diff
@@ -61,23 +63,22 @@ fn single() {
...
     assert_eq!(set.read().unwrap().len(), 1);
 }
 
 #[test]
-fn send_sync() {
-    fn send<T: Send>(_: T) {}
+fn sync() {
```
kennykerr: Revert this?

crates/samples/services/thread/src/main.rs
kennykerr: Or the `Service::run` needs a similar `scope` pattern?

kennykerr:
I played around with this today - I don't love how this impacts `Service` but that admittedly says more about `Service` than this change, which seems like the right thing to do and the service design can be improved separately.

crates/samples/services/thread/src/main.rs
ChrisDenton: I wonder if it might be better (and simpler) to separate `Service` into two: one that constructs the service and another that accesses it afterward. The latter could then be freely moved, cloned, etc because it's just accessing global state (via appropriate locks).

Using the scoped pattern may be another option. I'm still not sure if that'd be more or less ergonomic though. It does need to be able to call the closure multiple times which complicates things. It might help if `run` consumed `self` rather than taking `&mut self`. I'd want to play about with this a bit more to see.

ChrisDenton:
I've been thinking a lot about the Service API and yeah, I don't love the workaround in this PR. It does need changes to the Service API itself but I'd be happy to leave that to follow up work.

crates/tests/libs/threading/tests/pool.rs
ChrisDenton: Ah, yeah, that was left over from a previous incarnation of this PR :).

kennykerr (approved):

PR #3703, opened 2025-08-06
Make use of more prelude items

kennykerr:
These are a just a few more items from the Rust prelude that can be adopted to make the code less verbose.

https://doc.rust-lang.org/std/prelude/index.html


kennykerr:
A little surprised there isn't a lint for this...

Issue #3705, opened 2025-08-08
WinRT delegates are constructed with an `FnMut` even though they may be called concurrently


kennykerr:
I think this is incorrect for the same reason that the traits for COM interface trait methods expect a `&self` rather than a `&mut self`. 

Related: #1497 #1511 #1655

PR #3706, opened 2025-08-08
Delegate closures should be immutable

kennykerr:
They may not look like it, but WinRT delegates are implemented as COM interfaces and may well be called concurrently from other languages just like any COM interface method. 

Fixes: #3705


ChrisDenton (approved):
Oh, ooph. Yeah, this is necessary.

PR #3708, opened 2025-08-11
Update `windows-future` with `join` and `when` functions

kennykerr:
I originally implemented support for the Windows `IAsyncXxx` interfaces in the [windows-future](https://crates.io/crates/windows-future) crate to match how I had implemented it for [C++/WinRT](https://github.com/microsoft/cppwinrt) and while the async pattern must be implemented in exactly the same way to interop reliably, the language-specific helpers need not look exactly the same. The C++ version included a blocking `get` method to wait for the results. This was based on C++'s `std::future` but I realized that this is not very intuitive for Rust developers who might equate this more naturally with Rust's `JoinHandle`. I might create a thread and wait for its result using the Standard Library as follows:

```rust
let thread = std::thread::spawn(|| 123);

let result = thread.join().unwrap();
println!("{result}");
```

This update renames the existing `get` method to `join` so that it follows the same pattern:

```rust
let future = windows_future::IAsyncOperation::spawn(|| Ok(123));

let result = future.join().unwrap();
println!("{result}");
```

The `join` method works exactly the same and is implemented in the same way as before, although the implementation is now more streamlined to simplify maintenance. 

This update also adds the `when` method. While the `windows-future` crate implements the Standard Library's `IntoFuture` and `Future` traits, this isn't for everyone and you might just want a simple way to execute a closure when the future completes but without blocking the calling thread as `join` does. Wiring this up is syntactically complicated for historical reasons:

```rust
let future = windows_future::IAsyncOperation::spawn(|| Ok(123));

future
    .SetCompleted(&windows_future::AsyncOperationCompletedHandler::new(
        |sender, _| {
            let result = sender.unwrap().GetResults().unwrap();
            println!("{result:?}");
            Ok(())
        },
    ))
    .unwrap();
```

That's what `when` is for. It greatly simplifies this scenario:

```rust
let future = windows_future::IAsyncOperation::spawn(|| Ok(123));

future.when(|result| println!("{result:?}")).unwrap();
```


kennykerr:
Comments on the name `when`? Options:

* `when`
* `then`
* `chain`
* something else

Permik:
I'd just call it `.await_then`
The waiting is happening asynchronously and then the closure is called

_The reasoning feels self evident, but nonetheless:_
```
v--------- Asynchronous
A-wait <-- Wait
```
`then` is used basically everywhere in std to mean a lazily executed block.

kennykerr:
Naming is hard! 🙃

I'm not sure I want to use the word "await" due to its association with Rust futures and I'd rather have something simple like `when` or `then`. While `then` is common in Rust, it typically means something that happens *under a certain condition* but that's not the case here. So I may just stick with "when" as I can't find a direct parallel in the Rust Standard Library that won't potentially mislead the casual reader.

PR #3714, opened 2025-08-13
Adds basic Cargo package validation

kennykerr:
As the number of crates in the repo has grown, it has become a bit more tedious to ensure that all of the package fields are consistent and correct. This update just adds a simple test crate that validates some of the most important package manifest settings for consistency and correctness. Notably, it checks the following:

* There should not be an `authors` field as [that is deprecated](https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field)
* The `edition` should be `2021`
* The `license` should be "MIT OR Apache-2.0"
* The `repository` should be "https://github.com/microsoft/windows-rs"
* The `readme` should be "readme.md" as Cargo assumes "README.md"
* The `categories` should be "os::windows-apis"
* There should be a non-empty description.
* Unpublished crates (samples/tests/tools) should omit most fields including version, license, repository, readme, categories, and description


kennykerr:
Sigh, older versions of Cargo still require the `version` field.

PR #3716, opened 2025-08-13
Prefer `expect` to `allow` for Rust and Clippy lints

kennykerr:
The benefit of `expect` is that it warns if the lint is not actually triggered and thus helps to identify outdated or misapplied lints. 

I continue to use `allow` for macros and generated bindings. 

Unfortunately, the `expect` attribute is new to Rust 1.81 so I have updated the MSRV for most crates as they are due for breaking change updates with the exception of `windows-version` which I will leave as is to avoid bumping the MSRV.


kennykerr:
The docs aren't clean on what version of Rust stabilized `expect` but evidently it is after the MSRV for the `windows` crate...

Looks like it is 1.81

crates/libs/result/src/hresult.rs
```diff
@@ -80,7 +79,7 @@ impl HRESULT {
                     flags |= FORMAT_MESSAGE_FROM_HMODULE;
 
                     module = LoadLibraryExA(
-                        b"ntdll.dll\0".as_ptr(),
+                        c"ntdll.dll".as_ptr() as _,
```
kennykerr, line 82: This is a false positive as Clippy insists I use c"..." but that fails to compile since the result is a signed pointer whereas `LoadLibraryExA` expects an unsigned pointer...

I also don't see how this is better 🙃

Issue #3694, opened 2025-08-01
windows-services, using the service parameter in the closure passed to run with standard library threads.


adedomin:
### Summary

I was experimenting with the latest release (0.25.0) of windows-services, but was struggling on how to use the service parameter passed to the callback. Rust rightfully points out I am leaking that parameter when I store the JoinHandle of the thread running my service code (so I can join it later). I saw the [example](https://github.com/microsoft/windows-rs/blob/master/crates/samples/services/thread/src/main.rs#L6-L20) but now I'm wondering if this is possible using the standard library's threads.

the compile error:

```
Compiling example-winsvc v0.1.0 (C:\Users\prussian\source\example-winsvc)
error[E0521]: borrowed data escapes outside of closure
  --> src\main.rs:10:21
   |
5  |         let mut thread = None;
   |             ---------- `thread` declared here, outside of the closure body
...
8  |             .run(|svc, comm| match comm {
   |                   --- `svc` is a reference that is only valid in the closure body
9  |                 windows_services::Command::Start if thread.is_none() => {
10 |                     thread = Some(s.spawn(|| {
   |                     ^^^^^^ `svc` escapes the closure body here

For more information about this error, try `rustc --explain E0521`.
error: could not compile `example-winsvc` (bin "example-winsvc") due to 1 previous error
```

### Crate manifest

```TOML
[package]
name = "example-winsvc"
version = "0.1.0"
edition = "2024"

[dependencies]
windows-services = "0.25"
```

### Crate code

```Rust
use windows_services::{Service, State};

fn main() {
    std::thread::scope(|s| {
        let mut thread = None;
        if Service::new()
            .can_stop()
            .run(|svc, comm| match comm {
                windows_services::Command::Start if thread.is_none() => {
                    thread = Some(s.spawn(|| {
                        println!("Hello, World!");
                        svc.set_state(State::Stopped);
                    }));
                }
                windows_services::Command::Stop => {
                    if let Some(jh) = thread.take() {
                        jh.join();
                    }
                }
                _ => (),
            })
            .is_err()
        {
            eprintln!("not running as service");
        };
    });
}
```

adedomin:
[...]

PR #3723, opened 2025-08-14
static lifetime service argument.

adedomin:
The Windows Service API seems to want a singleton. This PR splits Service off from a ServiceBuilder. The builder holds the fallback and accept value, then constructs, only one global Service struct which gets stored as a private static variable. `'static` References of that variable are passed when the fallback and run closure are invoked.

Some other warts with this, unit tests have to be separate files, otherwise they share the same process which will have an initialized singleton `Service`.

I haven't fully tested the examples nor all the features which is why I'm leaving it in draft.

Fixes: #3694


adedomin:

adedomin:
My latest commit seems to have broken the examples and I'm not sure why. Apologize for the noise, it was working before my last change so I'll close this out for now.

Issue #3494, opened 2025-02-18
Allow `windows-targets` to use the `raw-dylib` feature on newer rust versions by default


silverstillisntgold:
### Suggestion

Rust 1.71 [fully stabilized](https://github.com/rust-lang/rust/pull/109677/) the `raw-dylib` feature. Currently, `windows-targets` will only use this feature when the `windows_raw_dylib` build flag is set, but I suspect most consumers of `windows-targets` are using downstream crates and are probably never going to go through the docs to see that this flag should be set on systems with newer rust versions. And it's not realistic to expect them to, I only happened to discover this behavior after stumbling across a random closed issue in a crate that uses `windows-targets`. Ideally, linking would always be done through `raw-dylib` when any rustc version supporting it is available.

My understanding is that this should be trivial to do with a simple build script to enable `windows_raw_dylib` when the rustc version is new enough. This would increase the work cargo has to do, but would also remove the need for linking against the provided static libraries, which should be a net win in compile times and binary sizes.

If this were to be done, would there be a need to provide some other config flag for users to disable the use of `raw-dylib`? Or is it generally better to use it if it's available? Maybe there's a benefit to bringing in the libraries that I'm unaware of.

kennykerr:
The `windows-link` crate #3450 has been introduced to address this issue. It should be released soon and all new code generation will use this by default.

unlimitedsola:
@kennel Hi, I saw `windows` already migrated to the `raw-dylib` `windows-link` crate:
https://github.com/microsoft/windows-rs/blob/2030b20046636b74e7c882a0977078e68c00722a/crates/libs/windows/Cargo.toml#L22

However, it seems `windows-sys` is still on the `windows-targets` crate: https://github.com/microsoft/windows-rs/blob/6b11cb57a1a45492785c230103804138067f491c/crates/libs/sys/Cargo.toml#L13

I wonder if this was intentional and is there a plan to also migrate `windows-sys` to use `raw-dylib`?

tim-weis:
> I wonder if this was intentional

`windows-sys` has a MSRV of 1.60, so it cannot rely on features introduced in 1.71.

kennykerr:
Some of the biggest dependents of `windows-sys` have since bumped their own MSRV to 1.70 so its probably time to think about moving `windows-sys` forward...

PR #3724, opened 2025-08-15
Switch `windows-sys` to use `windows-link` and thus `raw-dylib`

kennykerr:
The [windows-sys](https://crates.io/crates/windows-sys) crate is the last library crate provided by [windows-rs](https://github.com/microsoft/windows-rs) to depend on the older [windows-targets](https://crates.io/crates/windows-targets) crate for linking rather than its modern replacement, the [windows-link](https://crates.io/crates/windows-link) crate. Some of the biggest dependents of `windows-sys` have since bumped their own MSRV to 1.70 so its probably time to think about moving `windows-sys` forward.

The `windows-link` crate uses `raw-dylib` exclusively, providing simpler linking, faster and smaller download and compilation, and requiring no dependency on pre-built import libs. `raw-dylib` was stabilized on Windows with Rust 1.71 so that becomes the new MSRV for the next version of the `windows-sys` crate. 

Fixes: #3494

### DIFF
--- a/crates/libs/targets/readme.md
+++ b/crates/libs/targets/readme.md
@@ -1,6 +1,8 @@
 ## Import libs for Windows
 
-The [windows-targets](https://crates.io/crates/windows-targets) crate includes import libs, supports semantic versioning, and optional support for raw-dylib.
+The [windows-targets](https://crates.io/crates/windows-targets) crate includes import libs, supports semantic versioning, and optional support for raw-dylib. 
+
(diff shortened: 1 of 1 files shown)


unlimitedsola:
This is good news and I'm excited for the upcoming release!

A friendly reminder that we should update the MSRV stated [here](https://kennykerr.ca/rust-getting-started/windows-or-windows-sys.html) (`1.82` for `windows` and `1.71` for `windows-sys`) because it is referenced in the `README.md` of this project.

PR #3730, opened 2025-08-27
Simplify construction of `Ref` and `OutRef` directly

kennykerr:
#3025 added the the `Ref` and `OutRef` types that represent borrowed types, providing the same memory layout but also offer lifetime management and conveniences to make dealing with in and out parameters safe and convenient:

```Rust
unsafe extern "system" fn DllGetActivationFactory(
    name: Ref<HSTRING>,
) -> HRESULT
```

This made it much simpler to implement ABIs like `DllGetActivationFactory` but didn't provide much in the way of calling functions with those same ABI types. The thinking was that these would typically be called from some other language like C++, which is typically the case. 

In order to make it easier to define ABI boundaries directly in Rust that might be called from Rust itself, this update adds some conveniences to safely construct `Ref` and `OutRef` types from local Rust types again with the same memory layout and lifetime. This is mainly just adding more convenient `From` implementations for `Ref` and `Default` implementations for both. The tests illustrate how this might be used.

### DIFF
--- a/crates/libs/core/src/out_ref.rs
+++ b/crates/libs/core/src/out_ref.rs
@@ -29,3 +29,9 @@ impl<'a, T: Type<T>> From<&'a mut T::Default> for OutRef<'a, T> {
         unsafe { core::mem::transmute(from) }
     }
 }
+
+impl<T: Type<T>> Default for OutRef<'_, T> {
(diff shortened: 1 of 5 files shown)


kennykerr:
Blocked on #3731

PR #3742, opened 2025-09-04
Add Windows 10 testing

kennykerr:
The GitHub runner known as `windows-2025` are basically Windows 11 VMs but Rust still supports Windows 10 and I'm hoping that we can use the `windows-2019` runner to provide some Windows 10 coverage. Unfortunately, [the repo for the runner images](https://github.com/actions/runner-images) has `windows-2919` marked as "deprecated" so we'll see...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,6 +22,10 @@ jobs:
           - version: stable
             host: x86_64-pc-windows-msvc
             target: x86_64-pc-windows-msvc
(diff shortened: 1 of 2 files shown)


kennykerr:
Dead in the water: https://github.com/microsoft/windows-rs/actions/runs/17465680076/job/49600702484?pr=3742

More info: https://github.com/actions/runner-images/issues/12045

Looks like there's no support for testing Windows 10. 😟

PR #3743, opened 2025-09-04
Remove `ole32.dll` dependency from `windows-core` crate for compatibility

kennykerr:
Some editions of Windows, mostly used internally by Microsoft, don't include `ole32.dll`. This update attempts to remap a handful of functions that are particularly problematic. This PR should validate that this is viable for now on Windows 11 and we just need to manually test on Windows 10 to make sure this can support the Rust toolchain in general which still targets Windows 10.

### DIFF
--- a/crates/libs/bindgen/src/libraries.rs
+++ b/crates/libs/bindgen/src/libraries.rs
@@ -26,8 +26,7 @@ fn combine_libraries(
                 continue;
             };
 
-            // Windows libs are always produced with lower case module names.
-            let library = ty.method.module_name().to_lowercase();
+            let library = ty.method.module_name();
             let impl_map = ty.method.impl_map().unwrap();
             let flags = impl_map.flags();
             let name = impl_map.import_name().to_string();
--- a/crates/libs/bindgen/src/tables/method_def.rs
+++ b/crates/libs/bindgen/src/tables/method_def.rs
@@ -40,8 +40,25 @@ impl MethodDef {
             .next()
     }
 
-    pub fn module_name(&self) -> &'static str {
-        self.impl_map().map_or("", |map| map.scope().name())
+    pub fn module_name(&self) -> String {
+        let module_name = self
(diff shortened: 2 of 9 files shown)


riverar:
Spot checked Windows 10 22H2 and all those functions were present in combase.dll, so looks safe so far.

kennykerr:
Thanks, I also manually checked `test_core` and `test_implement` (which covers these functions) on Windows 10  so I think we should be good.

roblabla:
This broke Windows 7 support, which doesn't have combase.dll.  Would it be possible to reintroduce the `ole32.dll` when building for the `win7` target?

riverar:
Hey @roblabla, we don't see sufficient value in adding additional complexity here to support Windows 7 at this time. (Sorry!)

Perhaps you can include a small forwarder with your target?

lib.rs:
```rust
macro_rules! stub {
    ($name:ident) => {
        #[allow(non_snake_case)]
        #[unsafe(export_name = stringify!($name))]
        pub extern "C" fn $name() {
            unreachable!()
        }
    };
    ($($name:ident),+ $(,)?) => {
        $(stub!($name);)+
    };
}

stub!(
    CoCreateFreeThreadedMarshaler,
    CoIncrementMTAUsage,
    CoTaskMemAlloc,
    CoTaskMemFree,
    RoGetAgileReference,
);
```

build.rs:
```rust
fn main() {
  println!("cargo::rustc-link-arg=/DEF:module.def");
}
```

module.def:
```
EXPORTS
    CoCreateFreeThreadedMarshaler=OLE32.CoCreateFreeThreadedMarshaler @865
    CoIncrementMTAUsage=OLE32.CoIncrementMTAUsage @919
    CoTaskMemAlloc=OLE32.CoTaskMemAlloc @963
    CoTaskMemFree=OLE32.CoTaskMemFree @964
    RoGetAgileReference=OLE32.RoGetAgileReference @1236
```

PR #3753, opened 2025-09-09
The `windows-bindgen` crate should favor `windows-result` in code generation

kennykerr:
When generating bindings for non-COM APIs it can be possible to avoid all of the COM dependencies required by the `windows-core` crate if the bindings just refer to the `windows-result` crate directly. This update favors the more dedicated `windows-result` crate in such cases.


MarijnS95:
Should this also apply to `::BOOL`? That is typically used for result/error handling and seems to [exist inside the `windows-result` crate](https://github.com/microsoft/windows-rs/blob/2b37e78681bb2dc67c79a6a555dc5f7adf5b8dab/crates/libs/result/src/lib.rs#L38) for that reason?

kennykerr:
Yes, there's also `HRESULT` and I need to add a dedicated test crate to validate that it works with only a `windows-result` dependency.

kennykerr:
This is going to need some more work untangling some of the type system support from `windows-core` that is not specific to COM. I'm going to close this for now but keep working on a long-term solution.

kennykerr:
Take two: #3763

PR #3754, opened 2025-09-09
docs: update README to include guidance on using looser semver ranges for dependencies

wmmc88:
This pull request updates the documentation for both the `windows` and `windows-sys` crates to recommend using wider semver dependency ranges in your `Cargo.toml` files. The goal is to improve compatibility across the Rust ecosystem and reduce issues with duplicate crate versions. The documentation now also provides guidance on testing your code against the minimum supported version.

Dependency versioning guidance:

* Added a new "Dependency Version Ranges" section to both `crates/libs/windows/readme.md` and `crates/libs/sys/readme.md`, explaining the benefits of using wider semver ranges for the `windows` and `windows-sys` crates and providing example `Cargo.toml` entries. [[1]](diffhunk://#diff-9bb56b7df2ffea5648b586b558aefa0462b4f0f3ef5f4790b35f34e089550f1eR52-R86) [[2]](diffhunk://#diff-bee298c5e7f8027b48aa32b97c44192396b02661d4bd0668c196ce335c76a3b1R39-R73)
* Updated the initial dependency instructions in both readme files to reference the new guidance section. [[1]](diffhunk://#diff-9bb56b7df2ffea5648b586b558aefa0462b4f0f3ef5f4790b35f34e089550f1eL10-R10) [[2]](diffhunk://#diff-bee298c5e7f8027b48aa32b97c44192396b02661d4bd0668c196ce335c76a3b1L10-R10)

Testing recommendations:

* Included instructions for using `cargo-minimal-versions` and `cargo-hack` to verify your crate works with the minimum supported dependency versions, ensuring better reliability for consumers. [[1]](diffhunk://#diff-9bb56b7df2ffea5648b586b558aefa0462b4f0f3ef5f4790b35f34e089550f1eR52-R86) [[2]](diffhunk://#diff-bee298c5e7f8027b48aa32b97c44192396b02661d4bd0668c196ce335c76a3b1R39-R73)

### DIFF
--- a/crates/libs/sys/readme.md
+++ b/crates/libs/sys/readme.md
@@ -7,7 +7,7 @@ The `windows-sys` crate is a zero-overhead fallback for the most demanding situa
 - [Releases](https://github.com/microsoft/windows-rs/releases)
 - [Feature search](https://microsoft.github.io/windows-rs/features)
 
-Start by adding the following to your Cargo.toml file:
+Start by adding the following to your Cargo.toml file (see [Dependency Version Ranges](#dependency-version-ranges) for guidance on version specification):
 
 ```toml
 [dependencies.windows-sys]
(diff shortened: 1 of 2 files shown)


kennykerr:
I have been mulling over this and I think it just takes up too much prime real-estate on the readme (and thus crates.io) for these crates to focus on a topic that is not specific to these crates and [already documented in Cargo book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html). It is also a lot harder to actually test this in practice with even the provided testing suggestion not covering all of the pitfalls like breaking changes in intermediate versions and other side effects like target-specific breaking changes. 

I think it should suffice to update the readme examples to just use a conservative range of known versions and leave the rest for dependents to figure out on their own. Something like this:

https://github.com/microsoft/windows-rs/compare/range-example?expand=1

What do you think?

riverar:
I appreciate the 'design with versioning in mind' guidance, though I'm curious about the practical impact for greenfield projects that can only start with a single version.

wmmc88:
> I have been mulling over this and I think it just takes up too much prime real-estate on the readme (and thus crates.io) for these crates to focus on a topic that is not specific to these crates and [already documented in Cargo book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html). It is also a lot harder to actually test this in practice with even the provided testing suggestion not covering all of the pitfalls like breaking changes in intermediate versions and other side effects like target-specific breaking changes.
> 
> I think it should suffice to update the readme examples to just use a conservative range of known versions and leave the rest for dependents to figure out on their own. Something like this:
> 
> https://github.com/microsoft/windows-rs/compare/range-example?expand=1
> 
> What do you think?

I wanted to provide some justification/reasoning on why its a range, and why its specifically important for users of these crates. I'm fine with trimming down the lines about testing, or maybe cutting that completely.

> I appreciate the 'design with versioning in mind' guidance, though I'm curious about the practical impact for greenfield projects that can only start with a single version.

If you're an application, then it doesn't really matter if you only put a single version since (hopefully) any lib crates you use (that also take dependency on windows-rs) are specifying a range. If you are designing a greenfield library meant to be used by the masses, then it is in your own interest to have this wider range, or else you'll get people opening issues complaining about conflicting versions.

kennykerr:
Yes, a sentence recommending a version range should suffice. Perhaps linking to https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html for more information.

PR #3763, opened 2025-09-19
Improve `windows-bindgen` dependency targeting consistency and flexibility

kennykerr:
Today it can be a little hard to manage your dependencies when `windows-bindgen` introduces dependencies in generated code. Although `--no-deps` is useful for raw or sys-style bindings to limit or eliminate dependencies, it is not so easy to limit or eliminate dependencies for the default richer bindings generated by `windows-bindgen` and required for more convenient access to COM and WinRT APIs. 

This update does two things:

1. Improve consistency and convenience by default when `windows-core` is required to avoid a growing set of dependencies by using `windows-core` uniformly for most dependencies where feasible.

2. Allow for more granular and specific dependency targeting and avoid the `windows-core` dependency unless required with the new `--specific-deps` option. 

Let's say for example you generate bindings for the `IsCharLowerA` function

```rust
windows_bindgen::bindgen([
    "--out",
    "src/bindings.rs",
    "--filter",
    "IsCharLowerA",
    "--flat",
])
.unwrap();
```

Today the resulting bindings look something like this:

```rust
// before
#[inline]
pub unsafe fn IsCharLowerA(ch: i8) -> windows_core::Result<()> {
    windows_link::link!("user32.dll" "system" fn IsCharLowerA(ch : i8) -> windows_core::BOOL);
    unsafe { IsCharLowerA(ch).ok() }
}
```

Here you'll need to add dependencies for both the `windows-core` and `windows-link` crates. 

Following this change, the generated bindings will use `windows-core` exclusively to avoid the need for the extra explicit dependency:

```rust
// after
#[inline]
pub unsafe fn IsCharLowerA(ch: i8) -> windows_core::Result<()> {
    windows_core::link!("user32.dll" "system" fn IsCharLowerA(ch : i8) -> windows_core::BOOL);
    unsafe { IsCharLowerA(ch).ok() }
}
```

As you can see, only the `windows-core` crate is assumed by the generated bindings. Note that this is merely a convenience to simplify dependency and version management. You will still get a dependency on the `windows-link` crate in both cases. You just don't need to manage this dependency yourself in the latter case. 

On the other hand, you may want to avoid a dependency on the `windows-core` crate entirely because it is non-trivial and only really required for COM support. In such cases, you can opt in to manage your dependencies more directly with the new `--specific-deps` option that will instruct `windows-bindgen` to use the most specific crate dependencies where feasible. 

Take for example the `WindowsStringHasEmbeddedNull` function. As before, this would previously require a dependency on both the `windows-core` and `windows-link` crates:

```rust
// before
#[inline]
pub unsafe fn WindowsStringHasEmbeddedNull(
    string: &windows_core::HSTRING,
) -> windows_core::Result<windows_core::BOOL> {
    windows_link::link!(<bla bla bla>);
    unsafe {
        let mut result__ = core::mem::zeroed();
        WindowsStringHasEmbeddedNull(core::mem::transmute_copy(string), &mut result__)
            .map(|| result__)
    }
}
```

Following this change, it will by default only require the `windows-core` crate:

```rust
// after (default)
#[inline]
pub unsafe fn WindowsStringHasEmbeddedNull(
    string: &windows_core::HSTRING,
) -> windows_core::Result<windows_core::BOOL> {
    windows_core::link!(<bla bla bla>);
    unsafe {
        let mut result__ = core::mem::zeroed();
        WindowsStringHasEmbeddedNull(core::mem::transmute_copy(string), &mut result__)
            .map(|| result__)
    }
}
```

To avoid a dependency on the `windows-core` crate entirely and instead target the specific - and much smaller - crates directly you only need to add the new `--specific-deps` option to produce the following:

```rust
// --specific-deps
#[inline]
pub unsafe fn WindowsStringHasEmbeddedNull(
    string: &windows_strings::HSTRING,
) -> windows_result::Result<windows_result::BOOL> {
    windows_link::link!(<bla bla bla>);
    unsafe {
        let mut result__ = core::mem::zeroed();
        WindowsStringHasEmbeddedNull(core::mem::transmute_copy(string), &mut result__)
            .map(|| result__)
    }
}
```

As you can see, the bindings now directly reference the `windows-strings`, `windows-result`, and `windows-link` crates. This option is not the default because this level of control is not for everyone but if that's you then enjoy!


MarijnS95:
Thanks for the detailed outline! I haven't tried this out yet but will keep a list of crates to test this out on when the next update lands on crates.io. Two things:

1. This PR doesn't update the argument table nor examples in the module-doc of `crates/libs/bindgen/src/lib.rs`, making your elaborate description get lost and this argument harder to find;
2. I've only briefly mentioned it in https://github.com/microsoft/windows-rs/issues/3514#issuecomment-3272322552, but have you considered a similar case or argument for when `windows` is already required ([for references](https://github.com/Traverse-Research/amd-ext-d3d-rs/blob/6a76b30d39519dec02f58cfc0903716e599ff985/bindings.txt#L8-L10)), but [some of the generated bindings still require `windows_core`](https://github.com/Traverse-Research/amd-ext-d3d-rs/blob/6a76b30d39519dec02f58cfc0903716e599ff985/src/bindings.rs#L41) despite possibly being able to reference that via `windows::core::`?

Issue #3768, opened 2025-09-23
COM objects using implement macro cannot opt-out of IInspectable or IAgileObject


wravery:
### Summary

In the [`identity_query`](https://github.com/microsoft/windows-rs/blob/b3547dde8bd81aa671d99456c5c41fe0c94e8c6b/crates/libs/implement/src/gen.rs#L312) fragment, it checks for `IUnkown`, `IInspectable`, and `IAgileObject`. I'm not sure what the implications are for supporting `IInspectable`, but supporting [`IAgileObject`](https://learn.microsoft.com/en-us/windows/win32/api/objidlbase/nn-objidlbase-iagileobject) implies that the object is free threaded. This makes it impossible to implement apartment threaded interfaces, e.g. `IDispatch`, and count on the COM runtime to call you on the same thread where the object was created, or even to guarantee that you're only called on one thread at a time.

### Crate manifest

```TOML
[package]
name = "test-non-agile"
version = "0.1.0"
edition = "2024"

[dependencies]
windows-core = "0.62"
windows-implement = "0.60"
windows-interface = "0.59"

[dependencies.windows]
version = "0.62"
features = [
    "Win32_System_Com"
]

```

### Crate code

```Rust
#![allow(non_snake_case, non_camel_case_types)]

use std::thread::{self, ThreadId};
use windows::{Win32::System::Com::*, core::*};
use windows_core::Interface;
use windows_implement::implement;
use windows_interface::interface;

#[interface("3EB545F5-07A9-469F-8242-EF9A6D9AB9E0")]
unsafe trait IMyInterface: IUnknown {
    fn SayHello(&self) -> HRESULT;
}

#[implement(IMyInterface)]
struct MyInterface {
    thread_id: ThreadId,
}

impl IMyInterface_Impl for MyInterface_Impl {
    unsafe fn SayHello(&self) -> HRESULT {
        if self.thread_id == thread::current().id() {
            println!("Hello from the same thread!");
        } else {
            println!("Hello from a different thread!");
        }
        HRESULT(0)
    }
}

fn main() {
    let obj = MyInterface {
        thread_id: thread::current().id(),
    };
    let interface: IMyInterface = obj.into();
    unsafe {
        interface.SayHello().ok().expect("Failed to call SayHello");
    }
    assert!(interface.cast::<IAgileObject>().is_err(), "Object should not be agile");
}

```

cnaples79:
I've submitted a PR to fix this issue.

[...]

PR #3769, opened 2025-09-24
implement: Add Identity=IUnknown and Agile=false options to #[implement] macro

cnaples79:
Motivation:
- Addresses #3768: allow pure COM implementations that must not expose IInspectable or IAgileObject. Default #[implement] always exposes both; this adds opt-outs.

What’s new:
- Identity = IInspectable | IUnknown (default IInspectable)
  - Choose identity interface; IUnknown removes WinRT identity.
- Agile = true | false (default true)
  - Toggle IAgileObject exposure via QueryInterface.
- Defaults preserve current behavior.

Implementation:
- Identity field and VTABLE_IDENTITY types depend on Identity.
- QueryInterface branches conditional for IInspectable and IAgileObject.
- From<T> for IInspectable and ComObjectInterface<IInspectable> emitted only when Identity=IInspectable.

Tests:
- New: crates/tests/libs/implement/tests/identity_opts.rs
  - #[implement(ITest, Identity = IUnknown, Agile = false)]
  - Asserts cast::<IInspectable>() and cast::<imp::IAgileObject>() fail; IUnknown still valid.

Backwards compatibility:
- No behavior change unless options are specified.

Example:
- #[implement(IMyComInterface, Identity = IUnknown, Agile = false)]

### DIFF
--- a/crates/libs/implement/src/gen.rs
+++ b/crates/libs/implement/src/gen.rs
@@ -68,8 +68,10 @@ fn gen_impl_struct(inputs: &ImplementInputs) -> syn::Item {
     let original_ident = &inputs.original_ident;
     let vis = &inputs.original_type.vis;
 
-    let mut impl_fields = quote! {
-        identity: &'static ::windows_core::IInspectable_Vtbl,
+    let mut impl_fields = if matches!(inputs.identity, IdentityKind::IUnknown) {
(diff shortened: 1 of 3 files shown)


kennykerr:
Thanks for the contribution - I opted to implement just the agility control so I'll close this in favor of #3770.

Issue #3775, opened 2025-09-26
windows-bindgen API should raise a warning if rustfmt is not available


djc:
### Summary

Due to (still unclear) changes in GitHub Actions, it appears `rustfmt` was not available by default in some runs, which caused [some of my workflows to fail](https://github.com/chronotope/chrono/actions/runs/18030619548/job/51306080050) with errors where checked-in bindings differed from freshly generated bindings only in that the checked-in bindings were formatted while the freshly generated ones were not.

My chrono test `unwrap()`s the `Warnings` result from `windows_bindgen::bindgen()`, and I'm a little surprised that there is no explicit error about this. While I understand that it's nice to optimistically format only if `rustfmt` is available, I think it would be better if there was a (potentially optional) warning if rustfmt was not available.

### Crate manifest

```TOML

```

### Crate code

```Rust

```

kennykerr:
I was just bit by this yesterday. https://github.com/microsoft/windows-rs/pull/3772

kennykerr:
This should do the trick: #3778

PR #3778, opened 2025-09-26
Add `windows-bindgen` warning when `rustfmt` fails

kennykerr:
The new warning can be ignored as with other warnings, but this gives you early detection of problems if you prefer to fail the build on any warnings. 

Fixes: #3775


kennykerr:
Blocked on #3779

Issue #3781, opened 2025-09-28
Option to Generate Function Pointer Types for windows-sys Win32 APIs


uint-nate:
### Suggestion

As shown in [the delay load sample here](https://github.com/microsoft/windows-rs/blob/5c4cb21af8e2bb2bc3101fa28e818e9ae2c265a9/crates/samples/windows/delay_load/src/main.rs#L36), windows-rs provides the ability to dynamically load Windows Win32 APIs at runtime with `GetProcAddress`.

However if a developer wants to dynamically load an API, the developer needs to copy-paste the function signature even though windows-sys already knows the signature that needs to be used here (since it used that signature to generate the bindings in the first place).

From the delay_load example, [Kenny had to re-declare ShellMessageBoxW's signature](https://github.com/microsoft/windows-rs/blob/5c4cb21af8e2bb2bc3101fa28e818e9ae2c265a9/crates/samples/windows/delay_load/src/main.rs#L36) even though windows-rs "already knows" that signature since it generates [the function signature already for linking against](https://docs.rs/windows-sys/0.61.1/windows_sys/Win32/UI/Shell/fn.ShellMessageBoxW.html):
```rust
type ShellMessageBoxW = unsafe extern "C" fn(
    happinst: usize,
    hwnd: usize,
    lpctext: PCWSTR,
    lpctitle: PCWSTR,
    fustyle: u32,
    ...
) -> i32;
// vs this generated from windows_sys
windows_link::link!("shlwapi.dll" "C" fn ShellMessageBoxW(happinst : super::super::Foundation:: HINSTANCE, hwnd : super::super::Foundation:: HWND, lpctext : windows_sys::core::PCWSTR, lpctitle : windows_sys::core::PCWSTR, fustyle : super::WindowsAndMessaging:: MESSAGEBOX_STYLE, ...) -> i32);
```

I propose that windows-sys (or the bindings generation code only) provide an option to generate function pointer types along side the `sys` style `windows_link` bindings for the dynamic loading scenarios. This would allow Rust app developers to load Windows functionality at runtime without having to copy-paste the windows-sys generated bindings and hand-edit them from `raw-dylib` bindings into the function pointer types they want to use.

The binding generation code could look something like this in my head:
```rust
[...]

PR #3782, opened 2025-09-28
[DRAFT] Prototyping windows-sys fn ptr types

jonesnl:
Prototype solution for #3781

This is not ready for review, I haven't run any tests or built anything more than a local project, I just want to have something to point to in #3781 that provides a rough idea of what I think is a direction to go in for solving that issue.

For reference, I used code built on this to generate the following bindings: https://github.com/jonesnl/pluginrust/blob/master/bindings/src/bindings.rs

```rust
        pub type EXPERIMENTAL_WebAuthNDecodeGetAssertionRequestFn = unsafe extern "system" fn(cbencoded : u32, pbencoded : *const u8, ppgetassertionrequest : *mut *mut _EXPERIMENTAL_WEBAUTHN_CTAPCBOR_GET_ASSERTION_REQUEST) -> windows_sys::core::HRESULT ;
        windows_link::link!("webauthn.dll" "system" fn EXPERIMENTAL_WebAuthNDecodeGetAssertionRequest(cbencoded : u32, pbencoded : *const u8, ppgetassertionrequest : *mut *mut _EXPERIMENTAL_WEBAUTHN_CTAPCBOR_GET_ASSERTION_REQUEST) -> windows_sys::core::HRESULT);
```

### DIFF
--- a/crates/libs/bindgen/src/types/cpp_fn.rs
+++ b/crates/libs/bindgen/src/types/cpp_fn.rs
@@ -27,6 +27,35 @@ impl CppFn {
         self.type_name().write(config, &[])
     }
 
+    pub fn write_fn_ptr(&self, config: &Config<'_>, underlying_types: bool) -> TokenStream {
+        let ptr_name = self.method.name().to_string() + "Fn";
+        let name = to_ident(&ptr_name);
+        let abi = self.method.calling_convention();
+        let signature = self.method.signature(self.namespace, &[]);
+
+        let params = signature.params.iter().map(|param| {
+            let name = param.write_ident();
+            let ty = if underlying_types {
+                param.underlying_type().write_abi(config)
+            } else {
+                param.write_abi(config)
+            };
+            quote! { #name: #ty }
+        });
+
+        let return_sig = config.write_return_sig(self.method, &signature, underlying_types);
+
+        let vararg = if config.sys && signature.call_flags.contains(MethodCallAttributes::VARARG) {
+            quote! { , ... }
+        } else {
+            quote! {}
+        };
+
+        link_fmt(quote! {
+            pub type #name = unsafe extern #abi fn(#(#params),* #vararg) #return_sig;
+        })
+    }
+
     pub fn write_link(&self, config: &Config<'_>, underlying_types: bool) -> TokenStream {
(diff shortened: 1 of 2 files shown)


uint-nate:

(shouldn't be necessary since I'm a MSFT employee but 🤷)

uint-nate:
@microsoft-github-policy-service agree company="Microsoft"

uint-nate:
Updated so that if you specify `--sys` and `--sys-fn-ptrs`, then you'll get the function pointer types along side the function bindings.

Also attempted and hopefully successfully added a test for it. Not really familiar with the CI process here but I think the system yelled at me enough to get it mostly right?

This is my first Rust code that isn't just me playing around on toy projects, so happy to take any feedback around best practices that I may be ignorant of 😀.

uint-nate:
https://github.com/uint-nate/pluginrust/blob/master/bindings/src/bindings.rs is updated to use this new binding code.

crates/libs/bindgen/src/types/cpp_fn.rs
```diff
@@ -27,6 +27,35 @@ impl CppFn {
...
+
+        let vararg = if config.sys && signature.call_flags.contains(MethodCallAttributes::VARARG) {
+            quote! { , ... }
+        } else {
+            quote! {}
+        };
+
+        link_fmt(quote! {
+            pub type #name = unsafe extern #abi fn(#(#params),* #vararg) #return_sig;
+        })
```
kennykerr:
```suggestion
        quote! {
            pub type #name = unsafe extern #abi fn(#(#params),* #vararg) #return_sig;
        }
```

`link_fmt` is just a workaround for `rustfmt` which lacks the ability to format the code within the `link!` macro.

uint-nate: Removed link_fmt, note it did change the format of the generated function pointers.

uint-nate:
The current failures aren't coming from my code FWIW. If you want me to address them I can look into them though if that's the normal flow of things. Not sure if there was a library update or something.

kennykerr:
That's a common problem as CI tests against Rust nightly builds as an early warning system. This PR is now blocked on #3789 where I address this.

PR #3787, opened 2025-10-06
Add readme existence check to package validation

kennykerr:
In preparation for #3786 I discovered that Cargo doesn't actually validate whether the `readme.md` that Cargo.toml points to even exists until you attempt to publish the crate. So the workaround (#3785) also needs to ensure that all the readme files exist.

This update then adds the missing readme files and also adds a basic check to the package validation to cover this scenario.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,8 +15,8 @@ on:
       - master
 
 jobs:
-  fmt:
-    runs-on: ubuntu-22.04
+  publish:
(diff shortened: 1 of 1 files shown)


kennykerr:
https://blog.rust-lang.org/2025/09/18/Rust-1.90.0/ also adds a new `cargo publish` feature that, while I'm not yet sure can handle this workspace, may be useful for validation so I've added a dry-run validation workflow to this update as well.

PR #3788, opened 2025-10-06
Add dedicated `windows-enclaves` crate

kennykerr:
The Win32 metadata includes some support for Windows virtualization-based security enclaves but these APIs are a little tricky because you need to import the APIs from a different system library depending on whether you are inside or outside of the enclave. This update thus does the following:

* Remaps the general-purpose enclave APIs so that they are imported by default from `kernel32.dll` for crates like `windows` and `windows-sys`. 
* Adds a dedicated `windows-enclaves` crate that provides dedicated bindings for use within an enclave where all of the functions provided by this crate are specifically imported from `vertdll.dll`. 

I'm opening this PR for testing and validation as I am not too familiar with enclaves. Hopefully this will help make using enclaves easier from Rust.


crates/libs/enclaves/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+
+// EnclaveCopyIntoEnclave
+// EnclaveCopyOutOfEnclave
+// EnclaveRestrictContainingProcessAccess
+// InterlockedPushListSList
+// LdrDisableThreadCalloutsForDll
+// LdrResolveDelayLoadedAPI
+// RtlGetLastNtStatus
+// RtlRaiseStatus
+// RtlUnhandledExceptionFilter
```
kennykerr: @joshwatson just need to patch these in by hand for now and then we should be ready to take it for a spin. 😊

crates/libs/bindgen/src/tables/method_def.rs
```diff
@@ -49,12 +49,21 @@ impl MethodDef {
...
         } else {
-            self.impl_map()
+            let mut result = self
+                .impl_map()
                 .map_or("", |map| map.scope().name())
-                .to_lowercase()
+                .to_lowercase();
+
+            // Workaround for https://github.com/microsoft/windows-rs/pull/3788
+            if result == "vertdll.dll" {
```
joshwatson, line 62: Is it possible to completely remove anything that uses [winenclaveapi.h](https://learn.microsoft.com/en-us/windows/win32/api/winenclaveapi/) too? The APIs in that header file are only present in vertdll.dll but this change will attempt to link them to kernel32.dll; I don't know how much of an issue this would be since a normal process can't call the APIs, but I want to make sure there's no confusion. I'll comment on the APIs in their mod.rs as well

kennykerr: Yes, that's why I am inclined to remove them from the `windows` and `windows-sys` crates entirely as they can accidentally get linked even if you don't call them directly if you happen to enable the feature that allows those bindings to compile.

crates/libs/sys/src/Windows/Win32/System/Environment/mod.rs
```diff
@@ -1,13 +1,13 @@
...
-windows_link::link!("vertdll.dll" "system" fn EnclaveSealData(datatoencrypt : *const core::ffi::c_void, datatoencryptsize : u32, identitypolicy : ENCLAVE_SEALING_IDENTITY_POLICY, runtimepolicy : u32, protectedblob : *mut core::ffi::c_void, buffersize : u32, protectedblobsize : *mut u32) -> windows_sys::core::HRESULT);
-windows_link::link!("vertdll.dll" "system" fn EnclaveUnsealData(protectedblob : *const core::ffi::c_void, protectedblobsize : u32, decrypteddata : *mut core::ffi::c_void, buffersize : u32, decrypteddatasize : *mut u32, sealingidentity : *mut ENCLAVE_IDENTITY, unsealingflags : *mut u32) -> windows_sys::core::HRESULT);
-windows_link::link!("vertdll.dll" "system" fn EnclaveVerifyAttestationReport(enclavetype : u32, report : *const core::ffi::c_void, reportsize : u32) -> windows_sys::core::HRESULT);
+windows_link::link!("kernel32.dll" "system" fn EnclaveGetAttestationReport(enclavedata : *const u8, report : *mut core::ffi::c_void, buffersize : u32, outputsize : *mut u32) -> windows_sys::core::HRESULT);
```
joshwatson: These five APIs (EnclaveGetAttestationReport, EnclaveGetEnclaveInformation, EnclaveSealData, EnclaveUnsealData, EnclaveVerifyAttestationReport) are only present in vertdll, they're not in kernel32. Since they can't be called outside of an enclave, it's actually okay to remove these completely from windows-sys and only put them in windows-enclaves

In general, I think anything that starts with "Enclave" is going to be an Enclave-specific API that does not need to be in windows-sys, only in windows-enclaves

crates/libs/sys/src/Windows/Win32/System/Environment/mod.rs
```diff
@@ -34,7 +34,7 @@ windows_link::link!("kernel32.dll" "system" fn SetCurrentDirectoryW(lppathname :
 windows_link::link!("kernel32.dll" "system" fn SetEnvironmentStringsW(newenvironment : windows_sys::core::PCWSTR) -> windows_sys::core::BOOL);
 windows_link::link!("kernel32.dll" "system" fn SetEnvironmentVariableA(lpname : windows_sys::core::PCSTR, lpvalue : windows_sys::core::PCSTR) -> windows_sys::core::BOOL);
 windows_link::link!("kernel32.dll" "system" fn SetEnvironmentVariableW(lpname : windows_sys::core::PCWSTR, lpvalue : windows_sys::core::PCWSTR) -> windows_sys::core::BOOL);
-windows_link::link!("vertdll.dll" "system" fn TerminateEnclave(lpaddress : *const core::ffi::c_void, fwait : windows_sys::core::BOOL) -> windows_sys::core::BOOL);
+windows_link::link!("kernel32.dll" "system" fn TerminateEnclave(lpaddress : *const core::ffi::c_void, fwait : windows_sys::core::BOOL) -> windows_sys::core::BOOL);
```
joshwatson, line 32: Just leaving a comment here that this one is correct, it does live in kernel32 (and isn't something an enclave can call)

crates/libs/windows/src/Windows/Win32/System/Environment/mod.rs
```diff
@@ -25,27 +25,27 @@ pub unsafe fn DestroyEnvironmentBlock(lpenvironment: *const core::ffi::c_void) -
 }
 #[inline]
 pub unsafe fn EnclaveGetAttestationReport(enclavedata: Option<&[u8; 64]>, report: Option<*mut core::ffi::c_void>, buffersize: u32, outputsize: *mut u32) -> windows_core::Result<()> {
-    windows_core::link!("vertdll.dll" "system" fn EnclaveGetAttestationReport(enclavedata : *const u8, report : *mut core::ffi::c_void, buffersize : u32, outputsize : *mut u32) -> windows_core::HRESULT);
+    windows_core::link!("kernel32.dll" "system" fn EnclaveGetAttestationReport(enclavedata : *const u8, report : *mut core::ffi::c_void, buffersize : u32, outputsize : *mut u32) -> windows_core::HRESULT);
```
joshwatson: As above, all of these "Enclave*" APIs are not present in kernel32, so they can just be removed entirely and only present in windows-enclaves if possible. Otherwise, they should be vertdll

kennykerr:
I have excluded all Enclave functions exported by `vertdll.dll` from the `windows` and `windows-sys` crates with the exception of `CallEnclave` and `TerminateEnclave`. The following Enclave functions remain. The metadata for these all indicate some library other than `vertdll.dll`.

`kernel32.dll`:
* `CreateEnclave`
* `InitializeEnclave`
* `IsEnclaveTypeSupported`
* `LoadEnclaveData`

`api-ms-win-core-enclave-l1-1-1.dll`:
* `DeleteEnclave`
* `LoadEnclaveImageA`
* `LoadEnclaveImageW`

It seems these should all just be loaded from `kernel32.dll`.

riverar:
Not all those APIs are implemented in `kernel32` sadly. (Those last three are in `kernelbase` for example.)

crates/tools/bindings/src/enclaves.txt
```diff
@@ -0,0 +1,105 @@
...
+// InterlockedPushListSList
+// LdrDisableThreadCalloutsForDll
+// LdrResolveDelayLoadedAPI
+// RtlGetLastNtStatus
+// RtlRaiseStatus
+// RtlUnhandledExceptionFilter
+
+AcquireSRWLockExclusive
+AcquireSRWLockShared
+CallEnclave
```
riverar, line 23: Need this one to interact with Enclaves.

kennykerr:
Sigh, it looks like various Enclave APIs intended to be called from outside of the Enclave are not exclusively exported from `kernel32.dll` and thus a simple workaround for the faulty metadata isn't really feasible.

joshwatson:
> Sigh, it looks like various Enclave APIs intended to be called from outside of the Enclave are not exclusively exported from `kernel32.dll` and thus a simple workaround for the faulty metadata isn't really feasible.

It should only be that `CallEnclave` needed a workaround; the rest of the host process APIs should already have the correct metadata. The `CallEnclave` metadata should also be correct now as well, I had the authors update it: https://learn.microsoft.com/en-us/windows/win32/api/enclaveapi/nf-enclaveapi-callenclave

Is the metadata not what the microsoft learn pages are generated from? Is there somewhere else I should just get the metadata updated so that the `windows` and `windows-sys` crates generate the right bindings without workarounds?

kennykerr:
> Is the metadata not what the microsoft learn pages are generated from?

Sadly not yet. Hopefully we get to the point of having canonical metadata in future. 

https://github.com/microsoft/windows-rs/tree/master/crates/libs/bindgen/default

https://kennykerr.ca/rust-getting-started/how-are-crates-built.html

crates/libs/enclaves/src/bindings.rs
```diff
@@ -0,0 +1,1146 @@
...
+    target_arch = "arm64ec",
+    target_arch = "x86_64"
+))]
+impl Default for FLOATING_SAVE_AREA {
+    fn default() -> Self {
+        unsafe { core::mem::zeroed() }
+    }
+}
+pub type HANDLE = *mut core::ffi::c_void;
+pub type HEAP_FLAGS = u32;
```
uint-nate, line 554: It'd be nice if the constants for HEAP_FLAGS and other enum/integer + #define style types ended up in here as well, [like how it works here](https://github.com/microsoft/windows-rs/blob/945130accc25ac18a47054115e861ca704a37eb5/crates/libs/sys/src/Windows/Win32/System/Memory/mod.rs#L166)

kennykerr:
Closing for now. Happy to keep chatting about this and help you guys get an Enclaves crate published. It looks like you have a few repos like [VbsEnclaveTooling](https://github.com/microsoft/VbsEnclaveTooling) and [vbs-enclave-rs](https://github.com/microsoft/vbs-enclave-rs) to potentially host this and own it outright, which probably makes the most sense since I don't know the first thing about how to test and support Enclaves.

PR #3791, opened 2025-10-15
Don't link combase.dll on win7 target

roblabla:
Since #3743 was merged, win7 builds have been broken due to linking functions from the combase.dll which doesn't exist on win7 (the functions are instead in ole32.dll).

This PR fixes the issue by checking for target_vendor to chose whether to link against ole32 or combase.

### DIFF
--- a/crates/libs/bindgen/src/tables/method_def.rs
+++ b/crates/libs/bindgen/src/tables/method_def.rs
@@ -41,6 +41,12 @@ impl MethodDef {
     }
 
     pub fn module_name(&self) -> String {
+        self.impl_map()
+            .map_or("", |map| map.scope().name())
+            .to_lowercase()
+    }
+
+    pub fn combase_hack(&self) -> bool {
         const combase_functions: [&str; 5] = [
             "CoCreateFreeThreadedMarshaler",
             "CoIncrementMTAUsage",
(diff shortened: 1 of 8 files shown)


kennykerr:
Thanks for the suggestion, but Windows 7 is not supported by this project. It looks like @riverar offered a workaround in #3743 that may be worth exploring.

roblabla:
This project is absolutely foundational to the Rust ecosystem. I understand not wanting to advertise or provide win7 support yourself, but why refuse to pull a simple patch that fixes the immediate defect with no impact to the normal rust targets? Especially for a crate whose only purpose is providing language bindings, this seems to have minimal maintenance impact.

I saw the suggestion after I did the change, but this requires every app that wants a win7 build to pull this patch, greatly reducing the usefulness of the Rust win7 target.

I hope you will reconsider this position.

Vaider7:
+1

Rust gives so easy way for conditional compilation, and Rust has separate target for win7, there's should be no reason to say "Windows 7 is not supported by this project" on community patches. Such small things never brings maintenance burnout, but gives a broader compatibility

Despite Microsoft deprecated Win 7 a long time ago, many companies will continue to use deprecated things like Win XP, Win 7, Centos 7 until their hardware just die. And if such companies are your customers, you have no choice, you must find a solution for them. So rejecting such patches just add more headache for everyone without having any benefits

Issue #3796, opened 2025-10-21
Need `*ClassLongPtr*` on 32-bit


Berrysoft:
### Summary

There are aliases `*WindowLongPtr*`, but missing the class ptr ones.

### Crate manifest

```TOML

```

### Crate code

```Rust

```

kennykerr:
https://kennykerr.ca/rust-getting-started/where-are-the-macros.html

I'd rather not expand support for these macros and come up with a better way to support such API-specific capabilities, perhaps with a dedicated crate.

Berrysoft:
Well I don't think that legacy Win32 GUI support is on your plans. I just don't see a reason to have `*WindowLongPtr*` without `*ClassLongPtr*`.

kennykerr:
GUI is definitely on my list of things to tackle in some way. 😊

> I just don't see a reason to have *WindowLongPtr* without *ClassLongPtr*.

I agree but the point I am making is that I would be more inclined to remove the former than add the latter.

PR #3797, opened 2025-10-21
Add aliases *ClassLongPtr* for 32-bit

Berrysoft:
What's this all about?

Fixes: #3796 ⬅️ Be sure to refer to an existing issue here!

I don't know how to regenerate the bindings, so here are only the bindgen changes.


kennykerr:
See issue discussion.

Issue #3801, opened 2025-10-23
Option to generate more methods for WinRT interfaces?


Berrysoft:
### Suggestion

I would like a bindgen option (off by default) to generate binding methods for WinRT interfaces. For example, for class `Application`, I would like to generate safe method wrappers for `IApplication` and `IApplicationOverrides`. I will generate the wrappers by myself, so it won't affect the official bindings.

kennykerr:
The `windows-bindgen` crate allows you to generate bindings for whatever interfaces you choose.

Berrysoft:
Thanks, but for example, there are method wrappers for UISettings in `windows`, but no such wrappers for IUISettings. How to generate it?

Berrysoft:
Oh, I know where the problem is. Could you add an option to ignore the `ExclusiveToAttribute`?

kennykerr:
There is already such an option: `--implement`

https://docs.rs/windows-bindgen/latest/windows_bindgen/fn.bindgen.html

Berrysoft:
Oh, yes. It generates wrappers like `IUISettings_Impl::GetColorValue`, but I want to call a method like `IUISettings::GetColorValue`. Now it's only possible to call it though the raw vtable, or `UISettings::GetColorValue`.

https://github.com/microsoft/windows-rs/blob/1b092112731cdc133525b3cc79e8dbd1577abe09/crates/libs/bindgen/src/types/interface.rs#L232-L255

kennykerr:
The `ExclusiveTo` attribute indicates that this interface is exclusive to a given class (implementation). Unless you are implementing that class, you should simply use that class to call the methods of that interface.

Berrysoft:
Do be so impatient please. I need to call the interfaces for COM interop; especially the `*Overrides` interfaces, which would be useful when implementing composable objects. It fine that you don't provide helpers to implement them, because I can do it manually. I just want some convenience.

kennykerr:
Best of luck! It's certainly possible but not perhaps as convenient as it could be. That's what #3767 is about.

Issue #3767, opened 2025-09-22
Consider implementing limited support for COM aggregation


kennykerr:
I removed this support as it was only needed for Xaml and overly complicated for the vast majority of other cases but perhaps there is a simple way to support it in a limited fashion, most likely with a dedicated macro.

> @kennykerr I appreciate your help with #3242 very much. I've mentioned the need for composable inheritance previously, or at least a way to hook into QI so that it was possible to implement it by hand without unrolling the `implement` macro. .NET has [ICustomQueryInterface](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.icustomqueryinterface), something like this would've definitely been helpful even without proper composable support.
> 
> You can see the main perpetrator here:
> 
> * https://github.com/Alovchin91/winui3-rs/blob/25bb415df3676a834c5ce54a4fe46c14c5e7f14a/winui3/src/xaml_app.rs
> 
> If I want to use any XAML at all, even if it's in an "island", -- I must inherit from `(I)Application` because its factory turns it into a singleton reused later on. I don't like it either, but there's no way around it. I've tried implementing it with the "stock" `implement` macro, but it works until it doesn't, see this commit: https://github.com/Alovchin91/winui3-rs/commit/83cb3eb67e34f2984c4c3c9b0a0e808d593aa774
> 
> FYI, I have no plans to publish it, at least as a "winui3" crate, because I don't plan to support the entirety of the API surface (**and WebView2 is not a part of my plans, at least currently**). I'm only looking at a very constrained set of use cases, like a Window with a SwapChainPanel (because Window itself is not a UIElement, unfortunately). 

 _Originally posted by @Alovchin91 in [#3762](https://github.com/microsoft/windows-rs/issues/3762#issuecomment-3315273883)_

Berrysoft:
I have a solution that's a little hacky and doesn't introduce new macros.

https://github.com/compio-rs/winui3-rs/blob/winui3-lite/winui3/src/compose.rs

The hacky part is the trait `ChildClass`.
* It needs the Factory object to create a compose object.
* It replaces the vtable to extend the `QueryInterface` method.

Then it works. The object is created with `Compose::compose` and the inner base interface should be accessed with `Compose::base`. It would be better if the `*Overrides` interfaces also exposes methods wrappers, in order to make the users calling base methods easier.

kennykerr:
Fixed as of #4360

PR #3802, opened 2025-10-26
bindgen: add --overrides

Berrysoft:
Fixes: #3801 

...although the issue has been closed.

Related: #3767

Not sure whether you'll like it, but at least it's better than nothing.

This PR adds `--overrides` flag for bindgen to generate method wrappers for `*Overrides` interfaces. It uses `contains` instead of `ends_with` because WinRT adds new interfaces when extending, e.g., `IVirtualizingLayoutContextOverrides2`.

This small change helps me a lot. I have generated new bindings with it for [`winio-winui3`](https://crates.io/crates/winio-winui3), and an example of using it in production (and with composable classes) is here: [backdrop.rs](https://github.com/compio-rs/winio/blob/e857a9a3d42b8b5a2ac5093ca1cb9403bdbad741/winio-ui-winui/src/ui/backdrop.rs#L100-L102).

### DIFF
--- a/crates/libs/bindgen/src/config/mod.rs
+++ b/crates/libs/bindgen/src/config/mod.rs
@@ -22,6 +22,7 @@ pub struct Config<'a> {
     pub sys: bool,
     pub sys_fn_ptrs: bool,
     pub implement: bool,
+    pub overrides: bool,
     pub specific_deps: bool,
     pub derive: &'a Derive,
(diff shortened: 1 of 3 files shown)


kennykerr:
I'd rather take a more holistic approach to (re) adding support for composable types and COM aggregation.

PR #3825, opened 2025-12-01
Clean up CodeQL workflow by removing languages

kennykerr:
Removed unused languages from CodeQL workflow matrix.

This is to resolve the security alerts which are false positives but a challenge due to some of the non-code artifacts in this repo used for metadata and lib validation.

https://github.com/microsoft/windows-rs/security


kennykerr:
Looks like CodeQL can't keep up with this repo....

PR #3835, opened 2025-12-16
Temporarily pin clippy tests to nightly-2025-12-11 to avoid ICE

riverar:
The nightly Rust toolchain is now pinned to a specific version to temporarily work around a known Clippy issue.


kennykerr:
Time to revert?

PR #3849, opened 2026-01-08
Release 72

kennykerr:
Happy new year folks! This is the first release of 2026 and focused on publishing some needed `windows-bindgen` updates so that Windows can ingest and update its Rust support internally. 

This release only updates the [windows-bindgen](https://crates.io/crates/windows-bindgen) crate (0.66.0) while other crates in the Windows family remain unchanged at this time.

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -23,10 +23,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


MarijnS95:
Happy new year @kennykerr! As with the previous releases, would you be able to explicitly include in https://github.com/microsoft/windows-rs/releases/tag/72 that `windows-bindgen` (and only `windows-bindgen`) received a major version bump to `0.66`? Reading the description I wasn't sure if other crates received a (minor/patch) bump as a consequence.

PR #3867, opened 2026-01-30
 Add union support for `windows-rdl`

riverar:
Building on #3861, this update adds end-to-end support for unions.

An example in .rdl:
```rust
mod Test {
    union Name {
        a1: u8,
        a2: u16,
        a3: f32,
    }
}
```

The reader will encode this interface into a .winmd file as such:
```csharp
.class public explicit ansi sealed Test.Name
{
  .field [0] public uint8 a1
  .field [0] public uint16 a2
  .field [0] public float32 a3
}
```

### DIFF
--- a/crates/libs/rdl/src/writer/struct.rs
+++ b/crates/libs/rdl/src/writer/struct.rs
@@ -5,20 +5,18 @@ pub fn write_struct(item: &metadata::reader::TypeDef) -> TokenStream {
     let name = format_ident!("{}", item.name());
     let fields = item.fields().map(|field| write_field(namespace, &field));
 
-    if item
+    let keyword = if item
         .flags()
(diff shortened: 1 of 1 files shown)


crates/libs/rdl/src/writer/struct.rs
```diff
@@ -5,9 +5,20 @@ pub fn write_struct(item: &metadata::reader::TypeDef) -> TokenStream {
...
+        quote! {
+            union #name {
+                #(#fields)*
+            }
+        }
+    } else {
+        quote! {
+            struct #name {
+                #(#fields)*
+            }
```
kennykerr: Minor but may be easier to read if the switch just toggles the "token" that then gets inserted into a common quote at the end.

riverar: Fixed.

PR #3868, opened 2026-01-31
Add RDL syntax reference documentation

riverar:
Building on #3861, adds documentation covering the RDL syntax thus far. While RDL currently mirrors Rust syntax closely, having dedicated documentation establishes a foundation as the language evolves to express Windows-specific concepts.


kennykerr (approved):
Thanks, that's a good start. We'll need to try and keep it updated as we go along.

PR #3869, opened 2026-02-02
Add `Owned::into_raw` to relinquish ownership of an handle

### LINKED ISSUE #3857: `Owned` should have a way to relinquish ownership
### Suggestion

Consider the following code:
```rust
pub struct Foo {
    handle: ManuallyDrop<Owned<HSOMETHING>>;
};

impl Foo {
    fn new() -> Result<Self> {
        let hsomething = /* ... */;
        
        if !hsomething.is_invalid() {
            let owned = unsafe { Owned::new(hsomething) };
            // some failure prone init here
            Ok(Self { handle: ManuallyDrop::new(owned) })
        } else {
            Err(Error::from_thread())
        }
    }
}

impl Drop for Foo {
    fn drop(&mut self) {
        unsafe { ManuallyDrop::drop(&mut self.handle) };
        // do some cleanup *after* the handle is freed.
    }
}
```

Owned is being used in `new()` to allow the usage of stuff like `?` and to properly release the handle in case of initialization failure. However, because `Owned` has no way to relinquish ownership of the underlying handle, you then have to wrap it with `ManuallyDrop` or `std::mem::forget`, which doesn't feel amazing. If `Owned` had an `into_raw()` like `Box`, then this would allow the struct to directly hold an `HSOMETHING` and to call `free()` where'd they want in `drop()`, which would make the code nicer.

As a sidenote, I am confused why `Owned::new` doesn't consume the handle type, because it allows to construct several `Owned` off the same handle, or to call `free()` after constructing an `Owned`, leading to double-frees. While it requires unsafe, so requires you to ensure soundness, it would still feel more Rust-idiomatic to consume the handle type.

Also, it would be nice if handle types had `is_valid()` to avoid the double negation of `!is_invalid()`.

sylveon:
This API mirrors `Box::into_raw` but operates on Windows handles

Add basic unit tests for `Owned`

Fixes #3857

### DIFF
--- a/crates/libs/rdl/rdl.md
+++ b/crates/libs/rdl/rdl.md
@@ -1,0 +1,223 @@
+# Rust Definition Language (RDL) Documentation
+
+RDL provides a Rust-like syntax for defining Windows types and APIs. It serves as an interface definition language for generating bindings and implementations for use with any language across the Windows ecosystem.
+
+## Quick Start
+
+```rust
+// Define an interface
+interface ISprocket {
+    fn GetStatus(&self) -> SprocketStatus;
+    fn Spin(&self, speed: f32);
+}
+
+// Define a struct
+struct Sprocket {
+    TeethCount: u32,
+    Diameter: f32,
+}
+
+// Define an enum with explicit discriminants
+#[repr(i32)]
+enum SprocketStatus {
+    Idle = 0,
+    Spinning = 1,
+    Locked = 2,
+    Malfunctioning = 3,
+}
+```
+
+## Language Reference
+
+### Basic Syntax
+
+#### Comments
+
+```rust
+// Single-line comments start with //
+
+/* Multi-line comments
+   use C-style syntax */
+```
+
+### Type Definitions
+
+#### Modules (Namespaces)
+
+Modules provide a way to group types and APIs, simulating namespaces.
+
+**Syntax:**
+```rust
+mod ModuleName {
+    /* ... */
+}
+```
+
+Modules may be nested:
+```rust
+mod Contoso {
+    mod Sprockets {
+        struct Sprocket {
+            TeethCount: u32,
+            Diameter: f32,
+        }
+
+        #[repr(i32)]
+        enum SprocketStatus {
+            Idle = 0,
+            Spinning = 1,
+            Malfunctioning = 2,
+        }
+
+        interface ISprocketFactory {
+            fn CreateSprocket(&self, teeth: u32, diameter: f32) -> Sprocket;
+            fn GetStatus(&self, s: Sprocket) -> SprocketStatus;
+        }
+    }
+}
+```
+
+---
+
+#### Enums
+
+Enums define a set of named constants.
+
+**Syntax:**
+```rust
+#[repr(type)]
+enum EnumName {
+    Variant1 = value1,
+    Variant2 = value2,
+}
+```
+
+**Attributes:**
+- `#[repr(type)]` - Specifies the underlying integer type
+  - Supported types: `i8`, `u8`, `i16`, `u16`, `i32`, `u32`, `i64`, `u64`
+
+**Example:**
+```rust
+#[repr(i32)]
+enum SprocketStatus {
+    Idle = 0,
+    Spinning = 1,
+    Locked = 2,
+    Malfunctioning = 3,
+}
+```
+
+---
+
+#### Structs
+
+Structs define composite data types with named fields.
+
+**Syntax:**
+```rust
+struct StructName {
+    FieldName: Type,
+    // ...
+}
+```
+
+**Example:**
+```rust
+struct Sprocket {
+    TeethCount: u32,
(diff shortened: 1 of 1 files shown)


crates/libs/core/src/handles.rs
```diff
@@ -24,6 +24,21 @@ impl<T: Free> Owned<T> {
...
+    /// # Safety
+    /// 
+    /// The caller is now responsible for freeing the returned handle.
+    #[must_use = "losing the handle will leak it"]
+    pub unsafe fn into_raw(o: Self) -> T {
+        let o = core::mem::ManuallyDrop::new(o);
+
+        // we want to move out the inner handle but Rust
+        // does not let us do partial moves if the drop trait
+        // is implemented. this is okay because we no longer drop o
```
riverar: Optional: Reduced word count

If you reject this suggestion, please fix the capitalization at least.

```suggestion
        // Extract the inner handle without running Drop
```

sylveon: The comment exists to justify the use of mem::read, without it it doesn't compile:
image

riverar: Fair enough. Feel free to ignore that part.

crates/libs/core/src/handles.rs
```diff
@@ -24,6 +24,21 @@ impl<T: Free> Owned<T> {
...
         Self(x)
     }
+
+    /// Consumes the `Owned<T>` and relinquishes ownership of the handle.
+    /// 
+    /// # Safety
+    /// 
+    /// The caller is now responsible for freeing the returned handle.
+    #[must_use = "losing the handle will leak it"]
+    pub unsafe fn into_raw(o: Self) -> T {
```
kennykerr: This is not an unsafe operation.

crates/libs/core/src/handles.rs
kennykerr: All of this trickery is a sign that something's wrong. You're trying to treat handles as if they can be duplicated but the `Free` trait does not require `Clone`. The solution is to require `Clone`. E.g.

`
pub trait Free : Clone { ...
`

sylveon: We're not duplicating the handle, since the `drop` is suppressed. This is also how the standard library does it, for example with `ManuallyDrop::take` and `MaybeUninit::assume_init_read` (both of which tell you to ensure that the instance is not used again, but that we enforce here by consuming the `Owned`).

What happens is the following:
- `into_raw` consumes the `Owned`
- Within `into_raw`, the `Owned` is moved into `ManuallyDrop`, which means that the drop of `T` is suppressed, if any.
- We read the `T` and move it as the return value.

In the end, all invariants are preserved. Callers only see one instance of `T`, and `drop` is only called once the return value of `into_raw` is dropped.

crates/libs/core/src/handles.rs
sylveon: This mirrored `Box::into_raw` also being unsafe but I can remove it.

crates/libs/core/src/handles.rs
sylveon: More here: https://internals.rust-lang.org/t/move-out-of-deref-for-manuallydrop/19216 https://github.com/rust-lang/rfcs/pull/3466

As long as `o.0` or `o` is not used after (which it isn't), this `ptr::read` is safe.

kennykerr: Sure, I'm not saying it doesn't achieve the ultimate purpose. I'm saying its unnecessarily complicated. All handles are both `Clone` and `Copy` so requiring that of the `Free` trait is not a problem and avoids the unsafe read.

crates/libs/core/src/handles.rs
kennykerr: [Box::into_raw](https://doc.rust-lang.org/stable/std/boxed/struct.Box.html#method.into_raw) does not appear to be unsafe. Regardless, unsafe has a very specific and narrow definition in Rust. In this case, the unsafe operation is the creation of the `Owned` since the library or compiler cannot verify that the handle is valid, exclusively owned, and safe to free with the associated function. Once created, the assumption is that these conditions have been met and effectively leaking the handle value at this point does not impose a safety concern.

crates/libs/core/src/handles.rs
sylveon: I wanted to avoid changing the API surface and adding Clone as requirement, but since you're okay with that I'll go ahead and do the change.

crates/libs/core/src/handles.rs
sylveon: I could've sworn it was when I F12'd it lol. Will fix.

riverar:
After scratching my head, I took a peek at GitHub Status and looks like Actions is broken right now. https://www.githubstatus.com/ Standby.

kennykerr:
`cargo fmt` should fix the build break.

sylveon:
Formatted

PR #3873, opened 2026-02-06
Add free (Win32) function support for `windows-rdl`

kennykerr:
Building on #3861, this update adds parsing support for Win32-style free functions. These are traditional API functions like `CreateProcess` or `CoInitialize`. Here's an example of a .rdl definition:

```rust
mod Test {
    #[link(name = "ticker.dll", abi = "system")]
    fn GetTickCount() -> u32;

    #[link(name = "version.dll", abi = "C")]
    fn GetVersion() -> u32;
}
```

Under the hood, this is a little more complicated as ECMA-335 doesn't like the idea of methods without an associated type. So such methods are added to a special per-namespace type called "Apis":

```
.class public auto ansi sealed Test.Apis
	extends [mscorlib]System.Object
{
	.method public hidebysig static pinvokeimpl("ticker.dll" nomangle winapi) 
		uint32 GetTickCount () cil managed 
	{
	}
	.method public hidebysig static pinvokeimpl("version.dll" nomangle cdecl) 
		uint32 GetVersion () cil managed 
	{
	}
}
```

The same goes for Win32 style constants like `CO_E_NOTINITIALIZED` but I'll get constants working separately.


riverar:
Want me to update https://github.com/microsoft/windows-rs/blob/master/crates/libs/rdl/rdl.md?

kennykerr:
Thanks, probably with a more realistic example. 🙃

PR #3874, opened 2026-02-06
Document `windows-rdl` function support

riverar:
Adds `windows-rdl` documentation for recently added (#3873) external function signatures.

### DIFF
--- a/crates/libs/rdl/rdl.md
+++ b/crates/libs/rdl/rdl.md
@@ -179,7 +179,7 @@ interface ISprocket {
 
 #### Functions
 
-Functions declare external function signatures provided by the system. The `#[link]` attribute specifies the library name and ABI.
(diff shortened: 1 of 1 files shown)


crates/libs/rdl/rdl.md
```diff
@@ -177,6 +177,35 @@ interface ISprocket {
 
 ---
 
+#### Functions
+
+Functions declare external function signatures provided by the system. The `#[link]` attribute specifies the library name and ABI.
```
kennykerr:
```suggestion
Functions declare external function signatures provided by another library. The `#[link]` attribute specifies the library name and ABI.
```

It doesn't have to be the operating system.

riverar: Good catch, fixed.

PR #3892, opened 2026-02-17
[bindgen] ArrayShape read corrupts the reader

youyuanwu:
I am experimenting scraping c headers into winmd and use windows-bindgen to generate bindings from winmd here: [youyuanwu/bnd](https://github.com/youyuanwu/bnd). There is a discrepancy between metadata and bindgen for treating ArrayShape, which causes reader corruption and panic for reading elsewhere. 

ArrayShape write in metadata crate writes 4 bytes:
https://github.com/microsoft/windows-rs/blob/65c0c07b5b88ac031cf56acddd9f892efee1634a/crates/libs/metadata/src/writer/file/mod.rs#L426

While read in bindgen crate read 3 bytes:
https://github.com/microsoft/windows-rs/blob/65c0c07b5b88ac031cf56acddd9f892efee1634a/crates/libs/bindgen/src/types/mod.rs#L310

This PR fixes bindgen to read the ArrayShape correctly. I tested this fix with structs with anonymous multi-dimensional arrays c header.

### DIFF
--- a/crates/libs/bindgen/src/types/mod.rs
+++ b/crates/libs/bindgen/src/types/mod.rs
@@ -308,11 +308,20 @@ impl Type {
                 .unwrap_or(&Self::Void)
                 .clone(),
             ELEMENT_TYPE_ARRAY => {
+                // ECMA-335 II.23.2.13 ArrayShape:
+                //   Rank NumSizes Size* NumLoBounds LoBound*
                 let kind = Self::from_blob(blob, enclosing, generics);
-                let _rank = blob.read_usize();
-                let _count = blob.read_usize();
-                let bounds = blob.read_usize();
-                Self::ArrayFixed(Box::new(kind), bounds)
+                let _rank = blob.read_usize(); // Rank (unused — always 1 for fixed arrays)
+                let num_sizes = blob.read_usize(); // NumSizes
+                let len = if num_sizes > 0 { blob.read_usize() } else { 0 }; // Size[0]
+                for _ in 1..num_sizes {
+                    blob.read_usize(); // drain remaining sizes (multi-rank not used)
+                }
+                let num_lo_bounds = blob.read_usize(); // NumLoBounds
+                for _ in 0..num_lo_bounds {
+                    blob.read_usize(); // drain lower bounds (always 0 for C arrays)
+                }
+                Self::ArrayFixed(Box::new(kind), len)
             }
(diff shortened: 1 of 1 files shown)


crates/libs/bindgen/src/types/mod.rs
```diff
@@ -308,11 +308,20 @@ impl Type {
...
+                let num_sizes = blob.read_usize(); // NumSizes
+                let len = if num_sizes > 0 { blob.read_usize() } else { 0 }; // Size[0]
+                for _ in 1..num_sizes {
+                    blob.read_usize(); // drain remaining sizes (multi-rank not used)
+                }
+                let num_lo_bounds = blob.read_usize(); // NumLoBounds
+                for _ in 0..num_lo_bounds {
+                    blob.read_usize(); // drain lower bounds (always 0 for C arrays)
+                }
+                Self::ArrayFixed(Box::new(kind), len)
```
kennykerr, line 324: https://github.com/microsoft/windows-rs/blob/65c0c07b5b88ac031cf56acddd9f892efee1634a/crates/libs/metadata/src/reader/blob.rs#L166-L183

For reference, I already deal with this in the `windows-metadata` crate but haven't ported `windows-bindgen` over to my newer metadata reader yet.

youyuanwu: Cool. U can decide whether to take this change or use the new metadata reader. (Depending on ur plan to release/publish the fix.)

kennykerr: It will be a while before I can port `windows-bindgen` over. The focus right now is `windows-rdl` - which is based on `windows-metadata` - and will also make it much easier to test this sort of thing. This PR for example is a little problematic because there's no easy way to test it, but with .rdl we'll easily be able to write isolated unit tests for this sort of thing. So once #3861 is a little further along, I'll begin to use that for self-testing and then port `windows-bindgen` over as well.

youyuanwu: Ok I will close this and open an issue instead.

youyuanwu:
Opened an issue to track this for further fix: #3893

kennykerr:
Thanks, if it's not urgent I can look at upgrading windows-bindgen when I get a moment.

youyuanwu:
It is not urgent. I can use my forked branch for now.

PR #3895, opened 2026-02-18
 Add formatter to `windows-rdl`

riverar:
Adds a custom token-based formatter for Windows RDL using the `logos` crate, replacing the previous formatters (`rustfmt` and `prettyplease`). While those crates are fantastic for standard Rust, they operate on Rust's AST and offer little flexibility for RDL making it difficult to enforce our formatting conventions without fighting the tools. This new formatter gives us full control over spacing, indentation, and layout rules, and provides a foundation for extending formatting behavior as the language rapidly evolves.

### DIFF
--- a/crates/libs/rdl/src/formatter/mod.rs
+++ b/crates/libs/rdl/src/formatter/mod.rs
@@ -71,7 +71,6 @@ pub fn format(input: &str) -> String {
     let mut indent_level = 0;
     let mut paren_depth = 0;
 
-    let newline = if input.contains("\r\n") { "\r\n" } else { "\n" };
     let tokens: Vec<_> = Token::lexer(input).spanned().collect();
     let mut i = 0;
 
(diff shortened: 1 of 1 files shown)


kennykerr:
image

crates/libs/rdl/src/formatter/mod.rs
```diff
@@ -0,0 +1,279 @@
...
+    #[regex(r"[\s\t\r\n]+", logos::skip)]
+    Whitespace,
+}
+
+pub fn format(input: &str) -> String {
+    let mut output = String::new();
+    let mut indent_level = 0;
+    let mut paren_depth = 0;
+
+    let newline = if input.contains("\r\n") { "\r\n" } else { "\n" };
```
kennykerr: Why not just always normalize to `\n`?

riverar: Would potentially conflict with whatever `core.autocrlf` is set to locally. I have it set to `true` on my machine for this repository, for example.

riverar: Made `LF` the canonical format, matching our `.gitattributes`. Resolved.

kennykerr (approved):
Thanks!

PR #3900, opened 2026-02-19
Add nested types (structs) for `windows_rdl`

riverar:
Building on #3861, this update adds support for nested types within structs.

An example in .rdl:
```rust
#[win32]
mod Test {
    struct S {
        a: u32,
        b: struct {
            x: f32,
            y: f32,
        },
    }
}
```

And the matching metadata:
```
.class public sequential ansi sealed Test.S
{
    .class nested public sequential ansi sealed S_0
	{
        .field public float32 x
        .field public float32 y
	}

	.field public uint32 a
	.field public valuetype [Test]Test.S_0 b
}
```

### DIFF
--- a/crates/libs/rdl/src/formatter/mod.rs
+++ b/crates/libs/rdl/src/formatter/mod.rs
@@ -72,10 +72,10 @@ pub fn format(input: &str) -> String {
     let mut paren_depth = 0;
 
     let tokens: Vec<_> = Token::lexer(input).spanned().collect();
-    let mut i = 0;
+    let mut token_idx = 0;
 
-    while i < tokens.len() {
-        let (token, span) = &tokens[i];
+    while token_idx < tokens.len() {
+        let (token, span) = &tokens[token_idx];
 
         let token = match token {
             Ok(token) => token,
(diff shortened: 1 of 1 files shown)


kennykerr:
Hopefully nobody needs to use this for new APIs, but it will be indispensable for some of the thornier existing structs like `CONTEXT`. 🙃

crates/libs/rdl/src/formatter/mod.rs
```diff
@@ -107,7 +107,13 @@ pub fn format(input: &str) -> String {
                 indent_level -= 1;
                 push_indent(&mut output, indent_level);
                 output.push('}');
-                output.push('\n');
+                if matches!(tokens.get(i + 1), Some((Ok(Token::Comma), _))) {
+                    output.push_str(",\n");
+                    i += 2;
```
kennykerr: Maybe `i` needs a better variable name to make it more obvious what's going on here.

riverar: Agreed. Not super proud of the number of parenthesis used here either. Will take another look at this.

PR #3901, opened 2026-02-20
Add Windows UI Xaml Diagnostics API bindings

riverar:
Add bindings for the Windows UI Xaml Diagnostics API. This API provides runtime inspection and live editing capabilities for profilers and visual debuggers of system XAML-based UI elements.


riverar:
Went with bindgen workflow instead!

PR #3909, opened 2026-02-26
Add fixed array support for `windows-rdl`

kennykerr:
While #3904 added support for WinRT arrays, which are only supported as parameters, this update adds support for Win32-style fixed arrays, which are support both as struct fields as well as parameters. Here's an example:

```rust
#[win32]
mod Test {
    interface IFixed {
        fn In(&self, input: [i32; 4]);
        fn Out(&self, output: &mut [i32; 5]);
    }

    struct S {
        a: [u8; 1],
        b: [u16; 2],
        c: [u32; 3],
    }
}
```

The resulting ECMA-335 looks as follows:

```
.namespace Test
{
	.class interface public auto ansi abstract Test.IFixed
	{
		.method public hidebysig newslot abstract virtual 
			instance void In (
				[in] int32[...3] input
			) cil managed 
		{
		}

		.method public hidebysig newslot abstract virtual 
			instance void Out (
				[out] int32[...4]& output
			) cil managed 
		{
		}
	}

	.class public sequential ansi sealed Test.S
	{
		.field public uint8[...0] a
		.field public uint16[...1] b
		.field public uint32[...2] c
	}
}
```

Note that WinRT does not support fixed arrays. 

Builds on #3861

### DIFF
--- a/crates/libs/bindgen/src/types/cpp_method.rs
+++ b/crates/libs/bindgen/src/types/cpp_method.rs
@@ -184,11 +184,7 @@ impl CppMethod {
                         }
                     }
                 }
-                Type::BOOL => {
-                    if last_error {
-                        return_hint = ReturnHint::ResultVoid
-                    }
-                }
+                Type::BOOL if last_error => return_hint = ReturnHint::ResultVoid,
                 Type::GUID => return_hint = ReturnHint::ReturnStruct,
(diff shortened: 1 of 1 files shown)


kennykerr:
Blocked on unrelated new clippy warning: #3910

PR #3914, opened 2026-02-27
Add Guid support for `windows_rdl`

riverar:
Building on #3861, this update adds support for specialized Guid constants.

An example in .rdl:
```rust
#[win32]
mod Test {
    const CLSID_SystemClock: GUID = 0xe436ebb1_524f_11ce_9f53_0020_af0ba770;
}
```

And the matching metadata:
```csharp
.field public static valuetype [mscorlib]System.Guid CLSID_SystemClock
.custom instance void [Windows]Windows.Foundation.Metadata.GuidAttribute::.ctor(uint32, uint16, uint16, uint8, uint8, uint8, uint8, uint8, uint8, uint8, uint8) = (
    01 00 b1 eb 36 e4 4f 52 ce 11 9f 53 00 20 af 0b
    a7 70 00 00
)
```

### DIFF
--- a/crates/libs/rdl/src/writer/mod.rs
+++ b/crates/libs/rdl/src/writer/mod.rs
@@ -219,7 +219,7 @@ fn write_const_guid(_namespace: &str, item: &metadata::reader::Field) -> TokenSt
         value as u64 & 0xffffffffffff,
     );
 
-    let literal = syn::LitInt::new(&value, proc_macro2::Span::call_site());
+    let literal = syn::LitInt::new(&value, Span::call_site());
     quote! { const #name: GUID = #literal; }
 }
 


crates/libs/rdl/tests/guid.rdl
```diff
@@ -0,0 +1,4 @@
+#[win32]
+mod Test {
+    const CLSID_SystemClock: GUID = 0xe436ebb1_524f_11ce_9f53_0020_af0ba770;
```
kennykerr: Curious. `windows-bindgen` uses `0xe436ebb1_524f_11ce_9f53_0020af0ba770` rather than `0xe436ebb1_524f_11ce_9f53_0020_af0ba770`...

crates/libs/rdl/src/reader/const.rs
```diff
@@ -19,3 +36,74 @@ pub fn encode_const(encoder: &mut Encoder, item: &syntax::Const) -> Result<(), E
...
+) -> Result<(), Error> {
+    let syn::Expr::Lit(syn::ExprLit {
+        lit: syn::Lit::Int(lit),
+        ..
+    }) = &item.expr
+    else {
+        panic!("expected integer literal");
+    };
+
+    let value: u128 = lit.base10_parse().expect("invalid guid literal");
```
kennykerr: Might be simpler to reuse `encode_neg_lit_int` from `reader/mod.rs` here.

crates/libs/rdl/src/reader/const.rs
```diff
@@ -19,3 +36,74 @@ pub fn encode_const(encoder: &mut Encoder, item: &syntax::Const) -> Result<(), E
...
+    let field = encoder.output.Field(
+        name,
+        ty,
+        metadata::FieldAttributes::Public | metadata::FieldAttributes::Static,
+    );
+
+    let guid_typeref = encoder
+        .output
+        .TypeRef("Windows.Win32.Foundation.Metadata", "GuidAttribute");
```
kennykerr: Hmm, I wonder if we should just stick with the `Windows.Foundation.Metadata` attributes as much as possible. I get that this is Win32 metadata. Just feels goofy to have two definitions of `GuidAttribute`.

crates/libs/rdl/tests/guid.rdl
riverar: I noticed that too. Want to shift over to that?

kennykerr: Yes, if its not too much trouble - thanks!

PR #3943, opened 2026-03-06
Test `windows-rdl` on Linux

kennykerr:
#3942 highlighted that the tests were failing on Linux.


kennykerr:
This should be fixed on #3942 is merged.

PR #3950, opened 2026-03-08
Add clipboard sample

arelove:
Adds a sample demonstrating clipboard text operations using Win32 API bindings via GlobalAlloc, SetClipboardData, and GetClipboardData.

### DIFF
--- a/crates/samples/windows/clipboard/src/main.rs
+++ b/crates/samples/windows/clipboard/src/main.rs
@@ -1,83 +1,41 @@
 use windows::{
-    core::*,
-    Win32::Foundation::*,
-    Win32::System::DataExchange::*,
-    Win32::System::Memory::*,
+    core::*, Win32::Foundation::*, Win32::System::DataExchange::*, Win32::System::Memory::*,
     Win32::System::Ole::*,
 };
 
 fn main() -> Result<()> {
-    let text = "Hello from windows-rs!";
-
-    set_clipboard_text(text)?;
-    println!("Written to clipboard: {text}");
-
-    let result = get_clipboard_text()?;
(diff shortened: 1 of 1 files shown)


arelove:

kennykerr:
There's something off about this PR - PR validation isn't kicking in... 🤔

crates/samples/windows/clipboard/src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+        if ptr.is_null() {
+            GlobalFree(Some(hmem))?;
+            return Err(Error::from_thread());
+        }
+        std::ptr::copy_nonoverlapping(wide.as_ptr(), ptr, wide.len());
+        let _ = GlobalUnlock(hmem);
+
+        OpenClipboard(None)?;
+
+        let result = (|| {
```
kennykerr: Seems a little contrived for a sample. Can we simplify?

arelove: Thank you for the feedback! You're right, it can be simpler. I'll revise the sample and incorporate the other comments as well. Thanks again!

arelove:
Simplified - moved everything into main with no helper functions. Ran cargo clippy, cargo fmt, and cargo run -p tool_yml - all clean.

kennykerr:
This is partly why there aren't samples for every API. Many APIs are just very difficult to use correctly and samples can then become weighed down with all the potential pitfalls that must be accounted for in production code.

arelove:
Sure, I can see how clipboard gets weighed down with edge cases for what should be a simple sample. I'll close this PR and focus on APIs that are easier to demonstrate cleanly. Thanks for the feedback!

PR #3951, opened 2026-03-08
Add registry sample

arelove:
Add a registry sample that reads version information from HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion using RegOpenKeyExW and RegQueryValueExW.

### DIFF
--- a/crates/samples/windows/registry/Cargo.toml
+++ b/crates/samples/windows/registry/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "sample_registry"
+version = "0.0.0"
+edition = "2021"
+publish = false
+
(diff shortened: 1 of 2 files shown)


tim-weis:
There already is a registry "sample", sort of, in the `version` crate:

https://github.com/microsoft/windows-rs/blob/82b4d4e0385af1410e93b1efbc7eb5dc662bb611/crates/libs/version/src/lib.rs#L71-L94

Granted, this is using the sys-style bindings and avoids the convenient `w!`-macro and `Result` transformations to keep dependencies to a minimum.

Regardless, it's generally a good idea to start with an issue to discuss the merits of adding a sample ahead of creating a pull request.

PR #3977, opened 2026-03-11
Revert "Add .github/copilot-instructions.md"

kennykerr:
Reverts microsoft/windows-rs#3969 - it doesn't work.


walbourn:
In what sense does it not work? I have them in my github projects and github copilot reads them.

That said, my file is a bit more generic: https://github.com/microsoft/DirectXTex/blob/main/.github/copilot-instructions.md

kennykerr:
For context, I've been using Copilot heavily for two weeks but purely on GitHub in the browser. There's a lot I don't know. 🙂 I don't know if that makes a difference. It repeatedly failed to format code before opening a PR or pushing commits and this would break the build validation and waste time. So somebody suggested I add a copilot instruction file to the repo. I made it really simple and even got Copilot to add it. It then proceeded to ignore it entirely. I even asked Copilot to confirm its working - it didn't even know what it was. 

> I have them in my github projects and github copilot reads them.

How do you know it reads them?

walbourn:
I type into my AI context when when opening a project:

`Read the copilot instructions file'

kennykerr:
https://github.com/microsoft/windows-rs/pull/3999#issuecomment-4060606700 🙃

tim-weis:
The documentation situation isn't ideal. [About customizing GitHub Copilot responses](https://docs.github.com/en/copilot/concepts/prompting/response-customization) suggests that:

> Instead of repeatedly adding this contextual detail to your prompts, you can create custom instructions that automatically add this information for you.

The text then explains that adding a file at `.github/copilot-instructions.md` would automatically ingest those instructions. As you've learned, this isn't universally true. I *guess* these `copilot-instructions.md` are honored only when **reviewing** submissions. They (seemingly) aren't active when submitting work to an agent.

[About repository custom instructions](https://docs.github.com/en/copilot/concepts/prompting/response-customization#about-repository-custom-instructions) explicitly lists *"Agent instructions"*"

> \[...\] These are specified in files called `AGENTS.md`, `CLAUDE.md`, or `GEMINI.md`.

Could you try moving the instructions to one of those file names and try again?

PR #3981, opened 2026-03-11
Port sample from IDL to RDL

kennykerr:
[windows-rdl](https://github.com/microsoft/windows-rs/pull/3861) has a way to go, but is now complete enough to port a simple WinRT component from IDL to RDL! 🎉

This serves as an example of what it takes to port a simple Rust-based WinRT component. Obviously the same can be done for other languages. The corresponding C++/WinRT sample client has no trouble with the new .winmd file produced by `windows-rdl`. 

The RDL is a little wordier than the corresponding MIDL3 syntax. That's because MIDL3 has a bunch of syntactic conveniences and inferences and defaults. I'll open some issues to deal with some of these. In particular, it would be nice if RDL can infer the interface GUIDs in the same way.

Many thanks to @riverar for his help getting to this point.

### DIFF
--- a/crates/libs/rdl/src/error.rs
+++ b/crates/libs/rdl/src/error.rs
@@ -1,4 +1,3 @@
-#[derive(Debug)]
 pub struct Error {
     pub message: String,
     pub file_name: String,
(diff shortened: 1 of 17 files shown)


kennykerr:
Looks like the Clippy build failure is unrelated to this PR and as a result of a Rust nightly update.

riverar (approved):
Nice!

Issue #3989, opened 2026-03-12
`windows-bindgen` appears to ignore cdecl calling conventions


kennykerr:
It is a little obscure but there are rare cases where Win32-style callbacks use an alternative calling convention. Today the metadata encodes callbacks as delegates but that has a few drawbacks, like not being able to use the `ImplMap` table where calling convention information is stored in ECMA-335. Instead it uses the `UnmanagedFunctionPointer` attribute. 

I'm thinking we can perhaps represent callbacks like we represent import functions except that they won't have an import library. The RDL syntax is in question. We can go in two directions. 

Currently functions and callbacks are define as follows:

```rust
delegate fn CallbackC();
delegate fn CallbackS();

#[link(name = "test.dll", abi = "C")]
fn CallC(f: CallbackC);

#[link(name = "test.dll", abi = "system")]
fn CallS(f: CallbackS);
```

The `link` attribute is inspired by Rust but of course that attribute is not without its challenges. The "delegate" keyword is a little weird - it makes sense in WinRT but not really in the context of Win32. Anyway, one option is to support an ABI string as follows:

```rust
delegate "C" fn CallbackC();
delegate "system" fn CallbackS();

#[link(name = "test.dll", abi = "C")]
fn CallC(f: CallbackC);

#[link(name = "test.dll", abi = "system")]
fn CallS(f: CallbackS);
```

The other option is to scrap the delegate keyword for callbacks and just make them look more like functions, without an import library:

```rust
#[link(abi = "C")]
fn CallbackC();

#[link(abi = "system")]
fn CallbackS();

#[link(name = "test.dll", abi = "C")]
fn CallC(f: CallbackC);

#[link(name = "test.dll", abi = "system")]
fn CallS(f: CallbackS);
```

At this point, we could go even further and maybe use something a little more Rusty:

```rust
extern "C" fn CallbackC();
extern "system" fn CallbackS();

#[library("test.dll")]
extern "C" fn CallC(f: CallbackC);

#[library("test.dll")]
extern "system" fn CallS(f: CallbackS);
```

kennykerr:
@ChrisDenton / @riverar preferences?

ChrisDenton:
My initial impression is that you've listed them in reverse order of my preference :). The third option is the most appealing to me as it's the one that looks most like "normal" rust.  And, yeah, "delegate" looks weird to me outside of the context of WinRT.

kennykerr:
Just to be clear... this is your favorite?

[...]

PR #3995, opened 2026-03-13
Updates `windows-rdl` function and callback syntax

kennykerr:
As discussed in #3989, this updated the RDL syntax to look more like normal Rust for functions and callbacks.

Fixes: #3989


kennykerr:
Wow copilot actually gave good PR feedback! 🙃

PR #3996, opened 2026-03-14
Update Windows metadata (10.0.26100.7705)

riverar:
This update brings the metadata for the Win32 portion of the Windows SDK up to the latest version published here:

https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/70.0.11-preview

Fixes #3896

PR #4007, opened 2026-03-15
Add support for nested enum types within structs

riverar:
Tweaks bindgen and metadata reader to handle upcoming nested enum types (e.g. `GdiplusStartupInputEx::Version`). These types currently slip through as top-level types without an enclosing context. They are now collected in the `CppStruct` map alongside structs and are similarly emitted with a qualified name to avoid clashes.

Related: #3996


riverar:
We don't need this after all. Bad upstream metadata.

PR #4065, opened 2026-03-22
Align metadata PE headers with loader expectations

riverar:
Adjusts the PE file writing logic for greater correctness, improving compatibility with Microsoft Interface Definition Language (MIDL) tooling (e.g., midlrt.exe). Specifically, this corrects the calculation of the PE optional header `SizeOfImage` (in memory) value to now account for `SectionAlignment`, and pads the raw section data to `SizeOfRawData` (on disk).

A new test was added to verify that generated metadata can be loaded by `LoadLibraryEx` to ensure PE correctness.

Builds on https://github.com/microsoft/windows-rs/pull/3861


crates/tools/rdl_roundtrip/tests/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+use windows_core::*;
+
+use crate::bindings::Windows::Win32::System::LibraryLoader::{
+    LoadLibraryExA, LOAD_LIBRARY_AS_IMAGE_RESOURCE,
+};
+
+#[test]
+fn load_winmd() {
+    unsafe {
+        LoadLibraryExA(
```
kennykerr: This test should probably live in test_metadata and can just use the writer `File` and then a local definition of LoadLibrary to avoid the circular dependency on downstream crates.

riverar: Agreed, moved. Much cleaner.

PR #4173, opened 2026-04-09
Enable the `unused_qualifications` lint

kennykerr:
Manual attempt of #4168 with `cargo fix --lib --allow-dirty --broken-code`


kennykerr:
Need to compensate for `no_std`...

PR #4189, opened 2026-04-16
Add `Clang` conversion to `windows-rdl`

kennykerr:
This update adds the beginnings of Clang-based header-to-rdl conversion. Using RDL for your APIs is prefereable, but if you happen to have an existing API defined in some set of header files, you can use the new `Clang` builder to convert the C/C++ definitions into RDL definitions. For example, here's how you might turn an API defined in some .h file into a .winmd file for generating bindings in any supported language:

```rust
// Convert api.h to api.rdl
windows_rdl::clang()
    .input("api.h")
    .input("Windows.winmd")
    .output("api.rdl")
    .namespace("Sample")
    .write()
    .unwrap();

// Convert api.rdl to api.winmd
windows_rdl::reader()
    .input("api.rdl")
    .input("Windows.winmd")
    .output("api.winmd")
    .write()
    .unwrap();
```

Just a thin slice is currently working. Now we need to flesh this out to bring parity with the RDL reader/writer and generally support most/all of the quirks in the Windows SDK headers. I'll also be removing the header2rdl experimental tooling soon.

The `clang` reader in the example above will convert a simple header like this:

```C++
enum Name {
    One = 1,
    Two = 2,
};
```

Into the following RDL:

```rust
#[win32]
mod Sample {
    #[repr(i32)]
    enum Name {
        One = 1,
        Two = 2,
    }
}
```

The `reader` will then convert the RDL into the following metadata:

```
.namespace Sample
{
	.class public auto ansi sealed Sample.Name
		extends [mscorlib]System.Enum
	{
		.field private specialname rtspecialname int32 value__
		.field public static literal valuetype [enum.winmd]Sample.Name One = int32(1)
		.field public static literal valuetype [enum.winmd]Sample.Name Two = int32(2)

	}
}
```


youyuanwu:
@kennykerr Is scraping headers into different partitions/namespaces supported? Headers can have interdependencies, and the win32-metadata has a lot of knobs to tune which symbol should go to which partition. Sub included header can be configured to go into a sub namespace. Do u plan to support this?

PR #4309, opened 2026-04-25
Append underscores to constants that conflict with their type name

riverar:
Updates the logic for generating constants to avoid naming conflicts when the constant name matches its type name. It does this by appending one or more underscores to the constant name (e.g., `SymTagEnum::SymTagEnum` > `SymTagEnum::SymTagEnum_`).


crates/libs/bindgen/src/types/cpp_const.rs
```diff
@@ -40,13 +40,20 @@ impl CppConst {
             }
         }
 
-        let name = to_ident(self.field.name());
+        let field_ty = self.field.field_type(None, config.reader).to_const_type();
+        let tn = field_ty.type_name().name();
+        let name = if !tn.is_empty() && self.field.name() == tn {
+            let mut renamed = String::from(tn);
+            renamed.push('_');
+            to_ident(&renamed)
```
kennykerr:
```suggestion
            to_ident(&format!("{tn}_"))
```

Something like that would probably work.

riverar: Thanks!

kennykerr (approved):

Issue #4392, opened 2026-05-06
error types need to be manually excluded from `windows-bindgen`


sylveon:
### Summary

To make use of the changes from #3783, it is required to add
```rust
.filter("!Windows.Win32.Foundation.WIN32_ERROR")
.filter("!Windows.Win32.Foundation.NTSTATUS")
.filter("!Windows.Win32.System.Rpc.RPC_STATUS")
```

to our `windows-bindgen` invocation, otherwise it still creates the types and uses them instead of `windows-result`, leading in duplicated types and user confusion. The `windows` crate itself does this. This is inconsistent with `HRESULT`, which is automatically forwarded to `windows-result` without a manual exclusion.

### Crate manifest

```TOML
[package]
name = "ttb_bindings"
edition = "2024"

[dependencies.windows-core]
git = "https://github.com/microsoft/windows-rs"
rev = "f5eb67d1262d15ddade5851328cc484a9ae86219"

[build-dependencies.windows-bindgen]
git = "https://github.com/microsoft/windows-rs"
rev = "f5eb67d1262d15ddade5851328cc484a9ae86219"
```

### Crate code

```Rust
use std::{env, path::Path};
use windows_bindgen;

fn main() {
    let out_dir = env::var_os("OUT_DIR").unwrap();
    let dest_path = Path::new(&out_dir).join("bindings.rs");

    windows_bindgen::builder()
        .output(dest_path.to_string_lossy().as_ref())
        .no_allow()
        .implement()
        .filter("Windows.Win32.Foundation")
        .filter("Windows.Win32.Graphics.Gdi")
        .filter("Windows.Win32.System.Com")
        .filter("Windows.Win32.System.Console")
        .filter("Windows.Win32.System.Diagnostics.Debug")
        .filter("Windows.Win32.System.LibraryLoader")
        .filter("Windows.Win32.System.Ole")
        .filter("Windows.Win32.System.ProcessStatus")
        .filter("Windows.Win32.System.SystemServices")
        .filter("Windows.Win32.System.Threading")
        .filter("Windows.Win32.UI.Accessibility")
        .filter("Windows.Win32.UI.WindowsAndMessaging")
        .filter("Windows.Win32.UI.Xaml.Diagnostics")
        // These drags in a bunch of HTML/XML stuff
        .filter("!Windows.Win32.System.Com.Urlmon")
        .filter("!Windows.Win32.System.Diagnostics.Debug.Extensions")
[...]

PR #4401, opened 2026-05-08
Remove more specific references from type map

sylveon:
Fixes the root cause of #4392 instead of working it around.

Verified output of `tool_bindings` is exactly the same as previously, despite not filtering types that are in other crates.

### DIFF
--- a/crates/libs/bindgen/src/filter.rs
+++ b/crates/libs/bindgen/src/filter.rs
@@ -48,14 +48,14 @@ impl Filter {
         false
     }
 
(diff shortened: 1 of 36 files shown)


kennykerr (approved):
This looks reasonable - I'm a little nervous about messing with this as I suspect it will change quite a bit more once more of the new metadata tooling stabilizes but this looks like a reasonable improvement for now. Thanks!

sylveon:
Since there's a test it should avoid breaking things thankfully.

PR #4482, opened 2026-05-29
Avoid deprecated Node.js action in LLVM build dependency

kennykerr:
Noticed it here:

https://github.com/microsoft/windows-rs/actions/runs/26594266811


kennykerr:
Doesn't seem to solve the problem.

Issue #4500, opened 2026-06-01
windows-reactor: NavigationView: Pane background doesn't respect Backdrop material and memory is not released after closing the pane


iKineticate:
### Suggestion

I’m building a WinUI 3 application using the latest windows-rs crate. While working with NavigationView, I encountered two issues:

1. **Pane background does not support the configured Backdrop**
Even after setting a `Backdrop` (e.g. Mica, Acrylic) on the window, the NavigationView pane background always remains a solid color and does not apply the backdrop material. The rest of the window correctly shows the backdrop effect.

2. **Memory usage increases when opening the pane and is not released after closing**
When the NavigationView pane is opened (either via hamburger button or programmatically), the application’s memory usage jumps significantly. After the pane is closed, the memory usage stays elevated and never returns to the previous level,.

Environment:  Windows 11 25H2 ( Windows on ARM )

<img width="1180" height="932" alt="Image" src="https://github.com/user-attachments/assets/4636f486-dc5b-42f9-80ad-060350759c7e" />

kennykerr:
#4501 fixes a memory leak I discovered - can you verify this fixes your issue?

iKineticate:
Hi, thank you for looking into this. I updated to the build that includes the fix from #4501 (commit cd50817), but unfortunately the memory issue with NavigationView still persists.

After opening and closing the pane, the memory usage increases and never goes back down to the baseline level. The behavior is exactly the same as I originally reported.

Is there any additional information I can provide to help diagnose this? I can also share a minimal repro project if needed.

Environment: Windows 11 25H2 ( **Windows on ARM** )

<img width="1662" height="1052" alt="Image" src="https://github.com/user-attachments/assets/c8993b45-358f-4494-99a3-d3f446cb30cb" />

kennykerr:
Thanks! I fixed the issues locally in windows-reactor - it may be that you're hitting this:

https://github.com/microsoft/microsoft-ui-xaml/issues/5978

kennykerr:
Yes, if you can share a minimal repro that would be great.

iKineticate:
I’ve confirmed that the memory leak is a WinUI 3 NavigationView bug (not related to windows-rs).

One more question: the `NavigationView` pane background doesn’t seem to respect the window’s backdrop material (Mica/Acrylic). Is this by design, or is there a plan to support it?

Thanks!

iKineticate:
[...]

PR #4501, opened 2026-06-01
`windows-reactor` fix leak in sub trees

### LINKED ISSUE #4491: windows-reactor: Is there a memory leak?
### Summary

In the StackPanel, continuously adding items causes memory usage to rise. When I switch to another page and then return to the StackPanel, the content is reset; however, the memory usage does not decrease.

### Crate manifest

```TOML

```

### Crate code

```Rust

```

kennykerr:
- Composable constructor bug → leaked COM objects (big, unbounded native memory growth)
 - Missing map removes → leaked Rust HashMap entries holding closures, event revokers, and Vec<ControlId> (smaller, but still real)

Fixes #4491 and #4500


crates/samples/reactor/winui_leak_repro/src/main.rs
```diff
@@ -0,0 +1,177 @@
...
+use windows_reactor::DispatcherTimer;
+use windows_reactor::app::App;
+use windows_reactor::core::ControlKind;
+use windows_reactor::core::backend::{Backend, ControlId};
+use windows_reactor::winui::backend::WinUIBackend;
+
+fn main() -> windows_reactor::Result<()> {
+    App::new().title("WinUI Leak Repro").run_custom(|_app| {
+        eprintln!("╔══════════════════════════════════════════════════════════════╗");
+        eprintln!("║        WinUI 3 Memory Leak Reproduction                     ║");
```
kennykerr: That was just some test code that has since been removed.

PR #4511, opened 2026-06-03
Move Direct2D sample rendering onto a worker thread

damyanp:
Reworks the existing `direct2d` reactor sample so that all Direct3D/Direct2D/DXGI work happens on a dedicated render thread instead of the UI thread. The swap chain (the only agile object) is handed back to the UI thread for attachment to the `SwapChainPanel`. UI behaviour is unchanged: WinUI buttons still add/remove animated circles.

- **Worker thread owns all rendering state.** `mod render` keeps `D2DState` and the render loop private; the public surface is just `RenderThread` (`new` / `set_circle_count` / `resize`) and an opaque `SendSwap` exposing `swap_chain()`.
- **No thread_local, no manual marshalling.** The swap chain is delivered to the UI thread via `use_async_state` (the setter is `Send` and marshals back onto the UI thread). The panel lives in a UI-thread `use_ref`; the two are joined in a `use_effect` keyed on the swap chain.
- **Reactive count.** Circle count is reactor state; a `use_effect` pushes changes to the worker. The initial count is threaded through `RenderThread::new` so it has a single source of truth.
- **Clean shutdown via Drop.** Dropping `RenderThread` sends `Shutdown` and joins the worker, so teardown happens on any drop path.

Only the `Send` `SendSwap` payload crosses the thread boundary; the unsafe `Send` impl is justified by DXGI swap-chain agility plus single-owner presentation.


kennykerr (approved):
It works great - thanks! 

The Copilot comments are worth considering but it may suffice for a sample.

PR #4523, opened 2026-06-03
Reactor/Canvas type naming issues

kennykerr:
Just a doc summarizing some naming issues to consider in future work.


damyanp:
@kennykerr - winui also has a Canvas control. Maybe `windows-canvas` needs to find a new name entirely?

PR #4535, opened 2026-06-05
Mark owning component dirty on off-thread use_async_state writes

damyanp:
## Problem

An off-thread `use_async_state` setter only requested a root re-render; it never marked its owning component's dirty flag. When such a component was nested under another (e.g. a sample hosted in a `NavigationView`), the reconciler's `can_skip_update` skipped the structurally-identical child because `is_component_state_dirty` (which reads `peek_state_dirty`) returned `false`. The component's `use_effect` therefore never ran until an unrelated sync `use_state` change forced a re-render.

This surfaced in the Direct2D swap-chain sample: it would not begin rendering until the user clicked a button.

## Fix

Add a `Send + Sync` `async_dirty: Arc<AtomicBool>` to `RenderCx`, folded into `peek_state_dirty`/`take_state_dirty` alongside the existing `state_dirty` cell. `AsyncSetState::call` now sets it on a real value change, so off-thread async writes mark the owning component dirty and the reconciler re-renders it even when nested.

## Tests

Added regression tests in `async_state.rs`:
- `async_setter_marks_owning_component_dirty_for_rerender` — an off-thread async write now marks the component dirty. Verified to **fail without** this fix and **pass with** it.
- `async_setter_equal_value_does_not_mark_dirty` — an equal-value write does not mark the component dirty.

Full reactor test suite passes; `cargo clippy -p windows-reactor` is clean.

### DIFF
--- a/crates/libs/reactor/src/core/render_context.rs
+++ b/crates/libs/reactor/src/core/render_context.rs
@@ -2,6 +2,7 @@ use std::any::Any;
 use std::cell::{Cell, Ref, RefCell};
 use std::fmt;
 use std::rc::Rc;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex, OnceLock};
 use std::thread::{self, ThreadId};
 
@@ -138,6 +139,10 @@ impl<A: 'static> From<Dispatch<A>> for super::callback::Callback<A> {
 pub struct AsyncSetState<T: Send + 'static> {
     cell: Arc<Mutex<Box<dyn Any + Send>>>,
     marshaller: UiMarshaller,
+    /// Owning component's async-dirty flag, set on a real change so the
+    /// reconciler re-renders the component even when nested under a parent
+    /// whose element tree is unchanged.
+    dirty: Arc<AtomicBool>,
     type_name: &'static str,
     _marker: std::marker::PhantomData<fn(T)>,
 }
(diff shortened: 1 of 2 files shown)


crates/libs/reactor/src/core/render_context.rs
```diff
@@ -257,6 +267,11 @@ pub struct RenderCx {
     /// checks this to force re-render of the owning component even when the
     /// parent's element tree is structurally identical.
     state_dirty: Rc<Cell<bool>>,
+    /// `Send + Sync` counterpart to `state_dirty`, set by `AsyncSetState`
+    /// when an off-thread write changes a value. Folded into
+    /// [`Self::peek_state_dirty`] so nested components re-render after an
+    /// async state update, not just root components.
+    async_dirty: Arc<AtomicBool>,
```
damyanp: Wondering if this should just be combined with async_dirty?

damyanp:
Apologies, I haven't had a chance to review this carefully, but I figured sharing earlier rather than later might help move a discussion along.  Does this make sense?  This came up while getting #4536 to work.

crates/libs/reactor/src/core/render_context.rs
damyanp: Combining them feels right?

kennykerr (approved):

PR #4541, opened 2026-06-06
Add code generator for `windows-reactor` backend

kennykerr:
This PR replaces the hand-written dispatch code in `windows-reactor` with a metadata-driven code generation tool `tool_reactor`. A single TOML file `winui.toml` declares all 52 WinUI controls - their properties, events, setter patterns, and emit strategies. The tool reads this TOML alongside .winmd metadata files to generate 2,290 lines of type-safe Rust dispatch code across three files: `generated_bindings.rs`, `generated_set_prop.rs`, and `generated_attach_event.rs`. It also generates the binding filter entries `generated.txt` and drives `windows-bindgen` to produce `bindings.rs`, all in a single `cargo run -p tool_reactor` invocation.

The tool infers as much as possible from WinUI metadata: field names, COM method names, value types, copy semantics, enum variants, setter patterns (direct/optional/IReference/TextBlock/enum-map), and interface resolution. The TOML only needs explicit overrides for non-standard cases - custom setters `setter_fn`, non-default event handlers, ergonomic field renames, and emit strategies other than the default optional. 

Generated code is optimized: 132 redundant `.cast()` calls were eliminated from `set_prop` and 20 from `attach_event` by detecting when a `Handle` variant's WinRT class Derefs to its default interface. Match arms with identical bodies are collapsed via wildcard patterns and OR-patterns, removing ~60 duplicate arms. Compound types that can't be expressed in TOML (BrushBinding, ColorArgb, FlyoutDef, ImageSource, ContentDialogResult) coexist as hand-written supplements alongside generated bindings.

CI validation is added via `gen.yml`, which runs `tool_reactor` and fails if generated files are out of date. The filter file split between `base.txt` (hand-written: composition, dispatching, native interop) and `generated.txt` (auto-derived from TOML) ensures that adding a new control or property to the TOML automatically picks up the correct binding filter entries without manual intervention.


docs/reactor-codegen.md
```diff
@@ -0,0 +1,137 @@
...
+| `invoke_string_getter` | String from sender |
+| `invoke_i32_args` / `invoke_i32_getter` | i32 from args or sender |
+
+### Emit modes
+
+- `optional` *(default)* — emitted if `Some` (field: `Option<T>`)
+- `always` — always emitted (field: `T`)
+- `when_true` / `when_false` — emitted if bool matches (field: `bool`)
+- `non_default` — emitted if `!=` default value (field: `T`)
+
```
damyanp, line 107: I spent a little while trying to understand why these conditional modes exist.  This is what copilot & I came up with:

```suggestion

Conditional modes (`optional`, `when_true`, `when_false`, `non_default`) skip the
binding when the field is at its default. This is an optimization, not just style:
an omitted binding *means* "the control's default". The reconciler relies on this —
when a prop present in the old bindings is absent from the new ones, `diff_props`
resets it via `PropValue::Unset`, which `generated_set_prop.rs` maps back to the
control's native default. So "no binding" and "default value" are equivalent on both
mount and update, which is why emitting a default-valued binding is redundant.
```

damyanp (approved):

docs/reactor-codegen.md
kennykerr: I'm trying very hard to get rid of all the overrides but its pretty challenging.

kennykerr: Simplified it further so that many control entries assume defaults derived from metadata:

```toml
["Microsoft.UI.Xaml.Controls.CalendarView"]
IsTodayHighlighted = {}
IsGroupLabelVisible = {}
IsEnabled = {}
SelectedDatesChanged = {}
```

PR #4544, opened 2026-06-08
`windows-canvas` effects and device loss

kennykerr:
Small improvements to `windows-canvas` to handle device loss and creation efficiently for resources. 

The `windows-canvas` clock sample (`canvas_clock`) now avoids creating resources on each frame. The sample now also provides the same shadow effect as my original Pluralsight Direct2D sample (`sample_direct2d`). 

The only missing piece is the animation support from the original sample. I'm not yet sure how to handle that in `windows-reactor` and `windows-canvas` but I have some ideas...


damyanp (approved):
LGTM.  Some of this will likely need to be reworked as full device management and effects support is added (assuming you're aiming to have feature parity with Win2D) but that's probably ok.

I have a PR blocked behind #4535, #4536 and #4543 that is fleshing out what's needed to support shared devices (things don't scale well if you create a device for each widget).

kennykerr:
Thanks, I'll complete this for now and then we can work toward Win2D parity.

PR #4548, opened 2026-06-09
`windows-reactor` fix button style

### LINKED ISSUE #4545: windows-reactor: Button style prop mismatch: Prop::Style sent but backend expects Prop::StyleVariant
## Problem

Calling .accent(), .subtle(), or .text_link() on a button produces:

```
windows-reactor: unhandled prop (Style, I32(2)) for Button #2
```

The style does not take effect because the widget layer and backend disagree on the property name.

## Root cause

In `crates/libs/reactor/src/core/widgets/button.rs` line 65, the button widget emits `Prop::Style`:

```rust
out.push(Binding::Prop(Prop::Style, PropValue::I32(self.style.0)));
```

But the WinUI backend in `crates/libs/reactor/src/winui/backend/mod.rs` only handles `Prop::StyleVariant` for buttons (the match arm that looks up named styles like "AccentButtonStyle", "SubtleButtonStyle", "TextBlockButtonStyle"):

```rust
(Prop::StyleVariant, PropValue::I32(v), Handle::Button(b)) => { ... }
```

`Prop::Style` falls through to `unhandled_prop` in debug builds and is silently ignored in release.

## Fix

Change `Prop::Style` to `Prop::StyleVariant` in `button.rs:65`:

```diff
- out.push(Binding::Prop(Prop::Style, PropValue::I32(self.style.0)));
+ out.push(Binding::Prop(Prop::StyleVariant, PropValue::I32(self.style.0)));
```

This is a one-line fix. The backend already handles `Prop::StyleVariant` correctly.

kennykerr:
Verified with `cargo run -p reactor_minimal --example button` 

Fixes: #4545


kennykerr:
Merging this fix but am also going to apply some regression testing to catch such errors automatically.

PR #4562, opened 2026-06-11
`windows-reactor-setup` replace framework-dependent `resources.pri`

### LINKED ISSUE #4546: windows-reactor-setup: not good to copy `resources.pri`
https://github.com/microsoft/windows-rs/blob/514163716f02870055aa4251c74c0a5f612a0c94/crates/libs/reactor-setup/src/lib.rs#L45-L52

WinUI3 needs a pri file to start after version 1.8. It is because of https://github.com/microsoft/WindowsAppSDK/pull/4965 . It is a good refactor PR, but it violates the [document](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/win32/mrm/nf-mrm-mrmgetfilepathfromname), which says that the method `MrmGetFilePathFromName` will return a non-exist path with `S_OK`. I have created a PR to fix it, but nobody in MS cares: https://github.com/microsoft/WindowsAppSDK/pull/5828

I think the file copying in `windows-reactor-setup` is a patch at the wrong place. It introduces more files to mitigate the bug. As my solution has been ignored for a long time, I hope you maintainers could contact with the real person in charge to solve the root issue.

riverar:
`windows-reactor-setup` now embeds a dummy `resources.pri` instead of staging it from the Microsoft.WindowsAppSDK.InteractiveExperiences NuGet package, reducing our on-disk footprint by ~100KiB and reducing our build time footprint by ~94MiB.

The file must remain in place until https://github.com/microsoft/WindowsAppSDK/pull/5828 is fixed.

Fixes #4546


Alovchin91:
Please take a look at the idea with `ResourceManagerRequested` here: https://github.com/microsoft/windows-rs/issues/4546#issuecomment-4680053279
Could remove the need for an empty `resources.pri` file in the output directory altogether, effectively allowing a Reactor app to be a single executable file (I suppose).

riverar:
That'll put us in the middle of all custom resource manager requests and increase the risk we mishandle legitimate requests right? What if the dev wants to specify their own resource manager for their needs?

(The good news is we won't have to keep this code around for long. All signs are pointing to this getting fixed in the next Windows App SDK patch release.)

Alovchin91:
Does `windows-reactor` currently expose an API to attach a handler to `ResourceManagerRequested` or to get an instance of an `Application`? 🤔 If it does, and since this is hopefully getting fixed soon, maybe there’s no need for any mitigation on the reactor’s side, apart from documenting the issue?

riverar:
@Alovchin91 We don't currently but may in the future. I'm warming up to your idea, because it would potentially reduce hits to the disk which could be very expensive. But at the expense of the handler, per-thread invocations, and additional QI/function calls to even set this up. Thinking...

Alovchin91:
I would expect that if you ever want to expose this capability, you would probably expose it as a reactor-friendly API and not just a direct WinRT event subscription 🤔 That should be possible without too much hassle even with this workaround.

kennykerr (approved):

riverar:
Going to merge as-is for now. Appreciate the suggestion. Dropping `resources.pri` (when not needed) is the right long-term call, but the setup cost isn't worth it given that this workaround only exists because of an upstream bug in the first place. There's also something to be said for keeping the problem visible as leverage when pushing upstream to fix the root cause. We'd rather spend that energy there and eliminate the need for any of this.

If the winds shift, or this drags out beyond a few months, we'll definitely reconsider.

PR #4582, opened 2026-06-15
`windows-reactor` eliminate dead code

kennykerr:
For some reason I've not been able to pin down yet, the `dead_code` and `unused` Rust lints don't seem to kick in for some/most bindings, forcing these recurring maintenance sweeps. 

cc @ChrisDenton hoping you might have a genius idea for me. 🙃


kennykerr:
FYI https://gist.github.com/kennykerr/b6b1e00b029d16fe4047202505327a77 is a minimal repro.

ChrisDenton:
If the type is never meant to escape the crate then using `pub(crate)` on the type would fix.

kennykerr:
Yep, that's the workaround I'm employing and is effective. It does however seem like the Rust lints, or whatever, are being overly cautious here.

ChrisDenton:
It would be worth opening an issue on rust-lang/rust. I'm not sure how easily fixable but it seems like there should at least be an issue (and I'm not finding one yet).

kennykerr:
https://github.com/rust-lang/rust/issues/157961

kennykerr:
https://github.com/microsoft/windows-rs/pull/4585 is the workaround that detects and eliminates thousands of lines of dead code.

PR #4635, opened 2026-06-24
Add minimal feature search

### LINKED ISSUE #4631: `windows-rs` feature search?
### Suggestion

The usual https://microsoft.github.io/windows-rs/features link was removed from the readme (it was first moved to `docs/crates.md` - now deleted), and is now dead.

https://docs.rs/windows/latest/windows/ and https://microsoft.github.io/windows-docs-rs/doc/windows/ both have the same dead feature search link.

This was an incredibly useful feature for working with this family of crates, is there an alternative?

kennykerr:
Fixes: #4631


crates/tools/features/src/main.rs
```diff
@@ -0,0 +1,290 @@
...
+    if (!hit) continue;
+    count++;
+    if (shown >= LIMIT) continue;
+    shown++;
+    const features = [];
+    const own = featureOf(ns);
+    if (own) features.push(own);
+    if (extras) for (const ei of extras) { const f = featureOf(DATA.namespaces[ei]); if (f) features.push(f); }
+    const cell = features.length
+      ? features.map(f => '<code class="feature">' + f + '</code>').join(' + ')
```
tim-weis: Can this be changed to render features enclosed in ASCII double quotes (`"`), concatenated by `, ` instead of ` + `? This would enable copy-pasting the output straight into Cargo.toml, just like the previous feature search did.

kennykerr: Thanks, yes I was planning to fix that but I'm waiting on @riverar to decide if this is sufficient or he wants to shoot for something a little more elaborate. If not, I'll get this updated.

riverar: We can do both. So let's perhaps fix this for short term and then I'll follow up.

kennykerr: Done!

PR #4702, opened 2026-07-16
`metadata`: Add audiopolicy.h

riverar:
Adds audiopolicy interfaces and types. Is compatible with https://github.com/microsoft/windows-rs/pull/4700.


kennykerr (approved):

crates/tools/win32/src/main.rs
```diff
@@ -364,6 +364,7 @@ const HEADERS: &[&str] = &[
     "audioclientactivationparams.h",
     "spatialaudiometadata.h",
     "spatialaudioclient.h",
+    "audiopolicy.h",
```
kennykerr: Prefer appending headers to the end of the list unless there is a good reason to add it in some specific order.

riverar: Fair, though note there's these comment blocks at the end of the array which suggests they should be last. (I'm guessing I can ignore that.)

```rust
...
    // Apiset umbrella fallback (appended last). Modern APIs that no per-DLL host library
    // claims — `CreateFileMapping2`, the `*FromApp` file APIs, the package-dependency APIs,
    // … — are exported only through an `api-ms-win-*` apiset contract (or a host DLL the
    // umbrella records). Because resolution is first-wins and these come last, every
    // classic host-DLL mapping above is untouched; only the previously-empty residue is
    // stamped with the faithful DLL the umbrella records, matching the reference metadata,
    // which likewise leaves no function with an empty library.
    "onecore.lib",
    "onecoreuap.lib",
    // Enclave runtime, resolved dead last. `vertdll.dll` is the VBS-enclave runtime and
    // also exports host-side synchronization/enclave functions (`WaitOnAddress`,
    // `WakeByAddress*`, `CallEnclave`, `TerminateEnclave`); importing those from
    // `vertdll.dll` faults a normal process at load. Placing it after the apiset umbrella
    // lets those functions resolve to their loadable `api-ms-win-*` contract first, leaving
    // `vertdll.lib` to stamp only genuinely enclave-only residue (`EnclaveSealData`, …).
    "vertdll.lib",
];
```

kennykerr: I meant the end of the header list not the lib list. Yes, I'd stick any specific lib at the end but before the umbrella libs.

PR #4712, opened 2026-07-18
`windows-composition`

kennykerr:
Docs for future crate

### DIFF
--- a/docs/crates/windows-composition.md
+++ b/docs/crates/windows-composition.md
@@ -1,0 +1,119 @@
+# windows-composition
+
+> **Status: planned — this crate does not exist yet.** This page is the design
+> plan for a proposed `windows-composition` crate. Nothing here is shipped.
+
+- 📦 Not published
+- 🧩 Proposed sibling of [`windows-canvas`](windows-canvas.md) and
+  [`windows-reactor`](windows-reactor.md)
+
+## Purpose
+
+`windows-composition` would wrap the `Microsoft.UI.Composition` API behind safe
+Rust types — a `Compositor` and its visual tree (`Visual`, `SpriteVisual`,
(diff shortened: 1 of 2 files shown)


docs/crates/windows-composition.md
```diff
@@ -0,0 +1,119 @@
...
+> **Status: planned — this crate does not exist yet.** This page is the design
+> plan for a proposed `windows-composition` crate. Nothing here is shipped.
+
+- 📦 Not published
+- 🧩 Proposed sibling of [`windows-canvas`](windows-canvas.md) and
+  [`windows-reactor`](windows-reactor.md)
+
+## Purpose
+
+`windows-composition` would wrap the `Microsoft.UI.Composition` API behind safe
```
Alovchin91, line 12: @kennykerr Isn't WinUI [moving away](https://www.windowscentral.com/microsoft/windows-11/what-is-windows-k2-everything-you-need-to-know-saving-windows-11#:~:text=One%20of%20the%20ways,sections%20of%20the%20menu) from the lifted compositor back to the system one though? The new limited-access `CompositionEngine` [API](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.composition.compositionengine?view=windows-app-sdk-2.0) in WinAppSDK 2.3 suggests this report might be accurate.

kennykerr: I certainly hope that's true. I'm implementing `windows-composition` to handle both but require the non-system compositor for `windows-reactor` interop. Hopefully in future that can be eliminated.

kennykerr: Here it is: https://github.com/microsoft/windows-rs/pull/4721

PR #4777, opened 2026-08-03
`windows-csharp`

kennykerr:
> **_Note:_** this is a prototype to improve C# Reactor performance and not meant to compete with C#/WinRT. 

`windows-csharp` is an experimental code generator that projects Windows metadata directly into C#. It uses the same metadata model as `windows-rs` but emits direct COM and native ABI calls instead of using the managed-wrapper model employed by C#/WinRT.

The current branch is a credible experimental alternative to CsWinRT for performance-sensitive, metadata-controlled applications. It is not yet a general replacement, but it is broad enough to build real Composition and WinUI applications without `WinRT.Runtime.dll`, `Microsoft.Windows.SDK.NET.dll`, or generated Windows SDK NuGet projections.

[More info here](https://github.com/microsoft/windows-rs/blob/csharp/docs/crates/windows-csharp.md).


nikolayvpavlov:
This sounds like a solid approach, why isn't it considered a full replacement of CsWinRT, esp. if it would improve performance?

mamoreau-devolutions:
> This sounds like a solid approach, why isn't it considered a full replacement of CsWinRT, esp. if it would improve performance?

I'm certainly interested!

kennykerr:
The C# team is aware of this - hopefully they can benefit from some of these experiments. The `windows-rs` project will continue to focus on Rust and support services like metadata.

PR #4779, opened 2026-08-04
`windows-default`

kennykerr:
This is a new metadata support crate that contains the default metadata for publication. This is the same metadata that was previously available through the `windows-bindgen` crate but now necessarily needs to be shared with the other metadata and code gen crates like `windows-rdl` and `windows-clang`. 

For the most part this should not make any difference to existing `windows-bindgen` scenarios. It just means that those other crates don't have to also carry the same large binary dependencies.


riverar:
Thanks! Will use immediately!

PR #4951, opened 2026-09-17
`windows-reactor` update to Windows App SDK 2.5.1

riverar:
Updates the Windows App SDK (and related dependencies) from version 2.4.0 to 2.5.1 (https://github.com/microsoft/WindowsAppSDK/releases/tag/v2.5.1).


kennykerr (approved):

kennykerr:
I'd love a `tool-appsdk` that we can simply rerun and it grabs the latest of everything rather than having to touch all these things manually. Then gen.yml can also validate that everything is up to date like all the other tools.

